PR #313, opened 2016-08-03
allow bind to random port with INADDR_ANY

aliclark:
Alternatively, it's possible the entire 'if' statement could be removed to rely on the _host_isInterfaceAvailable for checks?

### DIFF
--- a/src/host/shd-host.c
+++ b/src/host/shd-host.c
@@ -888,15 +888,26 @@ static in_port_t _host_getRandomFreePort(Host* host, in_addr_t interfaceIP, Desc
     NetworkInterface* interface = host_lookupInterface(host, interfaceIP);
     in_port_t randomNetworkPort = 0;
 
-    if (interfaceIP == htonl(INADDR_ANY) || (interface && networkinterface_hasFreePorts(interface))) {
-        gboolean freePortFound = FALSE;
-        while (!freePortFound) {
-            gdouble randomFraction = random_nextDouble(host->random);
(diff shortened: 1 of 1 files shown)


robgjansen:
I think your solution fails in the case where there are no free ports.

I think we want to replace this line in the beginning of that function from this:

```
NetworkInterface* interface = host_lookupInterface(host, interfaceIP);
```

to this

```
NetworkInterface* interface = NULL;
if(interfaceIP == htonl(INADDR_ANY)) {
    interface = host->defaultInterface;
    Address* myAddress = networkinterface_getAddress(host->defaultInterface);
    interfaceIP = (in_addr_t)address_toNetworkIP(myAddress);
} else {
    interface = host_lookupInterface(host, interfaceIP);
}
```

Can you test if this works for you?

robgjansen:
@aliclark any progress here?

aliclark:
Hi Rob, sorry for the radio silence, just that it's getting closer to my deadline. I'm keen to take a look if you're ok with a slight delay. One thing I wanted to investigate was what should happen if I do bind(127.0.0.1:4567), and then bind(0.0.0.0:0)=>bind(0.0.0.0:4567). I think the second one should be denied, but will double check.

robgjansen:
OK. No worries about the delay; I would rather have you take your time and test it properly after your deadline. Good luck with the deadline!

aliclark:
Hi Rob, how about this? ^ seemed to work in a quick run on minimal-config

I think the other way is problematic because networkinterface_getAddress will return the ipHint so
and so would ignore an existing bind on 127.0.0.1

robgjansen:
@aliclark - I created a new test case in `src/test/bind` and I think I corrected the problem. Please open a new issue if it is not corrected. Thanks!

PR #342, opened 2017-06-12
Add socket errors numbers

hiromipaw:
Add error numbers in tgen logs for socket error. Makes it easier to debug crashes. 1st step towards better error handling.


robgjansen:
I've incorporated something similar into my private branch, along with better handling of partial reads/writes during socks handshakes - this will all be included in the next Shadow release. Thanks for the PR!

Issue #350, opened 2017-10-07
Build issue on Solus with glibc debug symbols


JRud52:
I am getting this error while trying to run the build script.
 
```
jrud@solus_thinkpad ~/documents/school/thesis/shadow $ ./setup build
2017-10-07 13:11:58,469 [INFO] running 'cmake /home/jrud/documents/school/thesis/shadow -DCMAKE_INSTALL_PREFIX=/home/jrud/.shadow -DCMAKE_EXTRA_INCLUDES=/home/jrud/documents/school/thesis/shadow/build;/home/jrud/.shadow/include -DCMAKE_EXTRA_LIBRARIES=/home/jrud/documents/school/thesis/shadow/build;/home/jrud/.shadow/lib' from '/home/jrud/documents/school/thesis/shadow/build'
-- The C compiler identification is Clang 4.0.1
-- Check for working C compiler: /usr/bin/clang
-- Check for working C compiler: /usr/bin/clang -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- 
-- -------------------------------------------------------------------------------
-- Current settings: (change with '$ cmake -D<OPTION>=<ON|OFF>')
-- SHADOW_DEBUG=OFF
-- SHADOW_PROFILE=OFF
-- SHADOW_TEST=OFF
-- SHADOW_EXPORT=OFF
-- -------------------------------------------------------------------------------
-- 
-- CMAKE_BUILD_TYPE Release enabled.
-- Check if the system is big endian
-- Searching 16 bit integer
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for stdint.h
-- Looking for stdint.h - found
-- Looking for stddef.h
-- Looking for stddef.h - found
-- Check size of unsigned short
-- Check size of unsigned short - done
-- Using unsigned short
-- Check if the system is big endian - little endian
-- CMAKE_MODULE_PATH = /home/jrud/documents/school/thesis/shadow/cmake/;/home/jrud/.shadow/include;/home/jrud/documents/school/thesis/shadow/build
-- Decompressing /home/jrud/documents/school/thesis/shadow/resource/topology.graphml.xml.xz
-- Found components for RT
-- RT_INCLUDES = /usr/include
-- RT_LIBRARIES = /usr/lib/librt.so
-- Found components for DL
-- DL_INCLUDES = /usr/include
[...]

PR #356, opened 2017-10-15
Add support for Debian 9.

ahf:
This patch adds support for bulding Shadow on Debian 9 (stretch). On
Debian 9 the debug symbols of ld.so have to be found within the
`/usr/lib/debug/.build-id/...` directory tree. The checksum that is used
to find the debug symbols in the build-id tree can be found using the
`readelf -n file` command.

I have not tested this patch on Debian i386, but found the path to 
the i386 `ld-2.24.so` file by looking at https://packages.debian.org/stretch/i386/libc6/filelist

### DIFF
--- a/src/external/elf-loader/extract-system-config.py
+++ b/src/external/elf-loader/extract-system-config.py
@@ -6,6 +6,22 @@ import getopt
 import os
 
 
+def find_build_id(path):
+    file = os.popen('readelf -n {}'.format(path), 'r')
+    lines = file.readlines()
+    regex = re.compile(r'^    Build ID: (\w+)$')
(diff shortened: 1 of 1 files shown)


ahf:
Notice that I did not test this patch on anything else than Debian 9.

jtracey:
Ahh, so that's what those are. If you don't mind, could you add `find_build_id(/usr/lib64/ld-2.25.so)` as well to fix #350 (i.e., Solus support)?

ahf:
I think Solus support should be added as a follow up patch once this is merged. I have no way of testing Solus support.

Issue #358, opened 2017-11-03
Shadow hangs on loading an incomplete topology


pastly:
I gave shadow an incomplete topology and 5 days to verify it, and it couldn't. 

I'm working on another way to verify a topology's completeness.

PR #359, opened 2017-11-03
Check topology completeness by counting edges at all vertices.

pastly:
Addresses issue #358.

Testing this on a very large complete topology and the same topology after removing one edge, it no longer hangs. It also returns the correct answer. It just takes the usual ~15 minutes to load the darn thing. Verification of completeness is on the order of 30 seconds.

The idea is to count edges at every vertex. I think this works ... as long as you never have more than one edge between a pair of vertices.

### DIFF
--- a/src/main/routing/shd-topology.c
+++ b/src/main/routing/shd-topology.c
@@ -134,6 +134,8 @@ static gboolean _topology_isComplete(Topology* top, gboolean *result) {
     igraph_vit_t vit;
     int ret;
     igraph_integer_t vcount = igraph_vcount(graph);
+    gboolean is_success = FALSE;
+    gboolean is_complete = FALSE;
     /*
      * Determines if a graph is complete by:
      * - knowning how many vertexes there are
@@ -144,12 +146,14 @@ static gboolean _topology_isComplete(Topology* top, gboolean *result) {
     ret = igraph_vs_all(&vs);
     if (ret != IGRAPH_SUCCESS) {
         critical("igraph_vs_all returned non-success code %i", ret);
-        return FALSE;
+        is_success = FALSE;
+        goto done;
     }
     ret = igraph_vit_create(graph, vs, &vit);
     if (ret != IGRAPH_SUCCESS) {
         critical("igraph_vit_create returned non-success code %i", ret);
-        return FALSE;
+        is_success = FALSE;
+        goto done;
     }
     while (!IGRAPH_VIT_END(vit)) {
         igraph_vector_t iedges;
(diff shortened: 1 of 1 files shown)


src/main/routing/shd-topology.c
```diff
@@ -122,6 +122,54 @@ static gboolean _topology_loadGraph(Topology* top, const gchar* graphPath) {
...
+        return FALSE;
+    }
+    ret = igraph_vit_create(graph, vs, &vit);
+    if (ret != IGRAPH_SUCCESS) {
+        critical("igraph_vit_create returned non-success code %i", ret);
+        return FALSE;
+    }
+    while (!IGRAPH_VIT_END(vit)) {
+        igraph_vector_t iedges;
+        igraph_vector_init(&iedges, 0);
```
robgjansen: There is a memory leak here.

If `igraph_vector_init` returns successfully, then the vector `iedges` needs to be deallocated using `igraph_vector_destroy` after the `iedges` is no longer needed, before the while loop loops, and before this function returns. Otherwise the memory that was allocated inside of `igraph_vector_init` will be leaked.

See:
  - http://igraph.org/c/doc/ch07.html#igraph_vector_init
  - http://igraph.org/c/doc/ch07.html#igraph_vector_destroy

src/main/routing/shd-topology.c
```diff
@@ -122,6 +122,54 @@ static gboolean _topology_loadGraph(Topology* top, const gchar* graphPath) {
...
+        if (ret != IGRAPH_SUCCESS) {
+            critical("error computing igraph_incident\n");
+            return FALSE;
+        }
+        long int ecount = igraph_vector_size(&iedges);
+        if (ecount < vcount) {
+            info("Vert id=%li has %li incident edges to %li total verts "
+                "and thus this isn't a complete graph",
+                IGRAPH_VIT_GET(vit), ecount, vcount);
+            *result = FALSE;
```
robgjansen: Please add a check that the `result` pointer is non-null before dereferencing it.

src/main/routing/shd-topology.c
```diff
@@ -122,6 +122,54 @@ static gboolean _topology_loadGraph(Topology* top, const gchar* graphPath) {
...
+            return TRUE;
+        }
+        IGRAPH_VIT_NEXT(vit);
+    }
+    *result = TRUE;
```
robgjansen: Please add a check that the `result` pointer is non-null before dereferencing it.

pastly:
Instead of responded to each of your comments (not sure if that would email spam you), I'll just comment here once saying that I've responded with code to everything.

src/main/routing/shd-topology.c
```diff
@@ -122,6 +122,71 @@ static gboolean _topology_loadGraph(Topology* top, const gchar* graphPath) {
...
+            critical("error computing igraph_incident\n");
+            is_success = FALSE;
+            igraph_vector_destroy(&iedges);
+            goto done;
+        }
+        igraph_integer_t ecount = igraph_vector_size(&iedges);
+        if (ecount < vcount) {
+            info("Vert id=%li has %li incident edges to %li total verts "
+                "and thus this isn't a complete graph",
+                IGRAPH_VIT_GET(vit), ecount, vcount);
```
robgjansen: Since you changed `ecount` and `vcount` types to `igraph_integer_t`, you need to cast them to `long int` here in the `info()` log function. (We can't guarantee that `igraph_integer_t` is defined as a `long int`everywhere, and since you are using the `%li` format string, you need to cast.)

It would be better if igraph provided something like an `IGRAPH_INTEGER_FORMAT` macro that always inserted the correct format string, but I wasn't able to find it.

pastly:
Did the casting.

You can find a squashed version of these commits in [issue358_squahsed](https://github.com/pastly/shadow/tree/issue358_squashed01).

PR #361, opened 2017-11-12
Parallel testing

jtracey:
add a -j option for running tests in parallel, and set the default to 4 threads


robgjansen:
This seems useful, thanks!

PR #370, opened 2017-12-06
Parse packet data too in parse-shadow.py

pastly:
I'd find it useful if packet data from shadow's log would be parsed, so I fixed this bug.

If this is of limited use for people in general, we may want to consider the impact this has on the time it takes `parse-shadow.py` to run and the size of `stats.shadow.json.xz`.


pastly:
Parsing packet data (in addition to bytes) took between 24 and 36 hours to complete on 5 large (20k client) experiments. `stats.shadow.json.xz` are between 430 and 910 MB in size.

Going to parse without packet data to compare.

pastly:
Leaving packet data out:

took 16 hours and `stats.shadow.json.xz` is 318MB

robgjansen:
OK, so I think leaving packet data out by default, and adding an option to include it when parsing would be the best. I'll merge once that is added. Thanks!

robgjansen:
Note that larger stats files will also cause the plotting script to consume more memory. So I think it's best to leave packet data out by default and add an option that will enable parsing it.

pastly:
It took 8.5 hours to plot two simulations that had packet data, and the plot script consumed about 500GB of memory.

Just piling on more specific information arguing for that flag.

PR #393, opened 2018-01-22
Start sims at a time that isn't the 1970s

pastly:
For better or worse, some programs assume the current time is adequately farther
foward than 1 Jan 1970. Before this patch, Shadow simluations would start with
time == 0.

But not anymore! This patch starts simulations at an arbitrary time (16 Feb
1995) that should be far enough away from 1970 for most applications.

A simple test program called `time()`, `clock_gettime()` (with both `REAL` and `MONOTONIC`), and `gettimeofday()` to verify that they all respect the changes.

The user is still able to specify a kill/stop time as a number of seconds since simulation start.

### DIFF
--- a/src/main/core/shd-worker.c
+++ b/src/main/core/shd-worker.c
@@ -240,7 +240,7 @@ void worker_sendPacket(Packet* packet) {
 
 static void _worker_bootHost(Host* host, Worker* worker) {
     worker_setActiveHost(host);
-    worker->clock.now = 0;
+    worker->clock.now = SIMTIME_MIN;
     host_continueExecutionTimer(host);
     host_boot(host);
     host_stopExecutionTimer(host);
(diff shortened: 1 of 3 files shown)


pastly:
Regarding Tor simulations,

the V3BandwidthFile seems to have a timestamp on the first line, and if it is too old (more than 3 days old), the authorities don't use it.

Fix for that is to change the first line from `1` to `792892801`

pastly:
Processes aren't starting at the correct time. Don't merge yet.

pastly:
Ok that's better.

robgjansen:
Your changes cause the shadow log files to be incomprehensible because the simtime starts at `220248:00:00.000000000`.

```
00:00:00.014457 [thread-0] 220248:00:00.000000000 [message] [fileserver~11.0.0.1] [shd-tracker.c:417] [_tracker_logNode] [shadow-heartbeat]
```

This can be avoided by keeping the Shadow time as is, and only *emulating* that it is 1995 (or whatever year) to the application. See commit 50720e887598df3bfd016429d3fface51bf2cbe7.

PR #396, opened 2018-02-09
add support for specifying the plugin start symbol in the config

dhuseby:
I am researching using Shadow to simulate blockchain networks and wanted to try and see if I could get it to run a Rust executable.  It turns out that the Rust compiler doesn't allow the creation of a C callable dynamic linked library with the "main" symbol exported.  To get around this, I modified Shadow to add a "startsymbol" attribute to the plugin tags in the config.  If it isn't specified, it defaults to "main" so the existing behavior is preserved.  I also modified a few tests to verify it works.  I think this may also be useful for other situations also.

### DIFF
--- a/src/main/core/shd-slave.c
+++ b/src/main/core/shd-slave.c
@@ -27,6 +27,9 @@ struct _Slave {
     /* paths to programs (i.e. shadow plug-ins) that are run by virtual processes */
     GHashTable* programPaths;
 
+    /* start symbols for programs (i.e. shadow plug-ins) */
+    GHashTable* programStartSymbols;
+
     GMutex lock;
     GMutex pluginInitLock;
 
@@ -98,6 +101,9 @@ Slave* slave_new(Master* master, Options* options, SimulationTime endTime, guint
     /* we will store the plug-in programs that are loaded */
     slave->programPaths = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
 
+    /* we will store the plug-in program start symbols if specified in the config */
+    slave->programStartSymbols = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
+
     /* the main scheduler may utilize multiple threads */
 
     guint nWorkers = options_getNWorkerThreads(options);
@@ -144,6 +150,7 @@ gint slave_free(Slave* slave) {
     worker_logAndFreeGlobalObjectCounts();
 
     g_hash_table_destroy(slave->programPaths);
+    g_hash_table_destroy(slave->programStartSymbols);
 
     g_mutex_clear(&(slave->lock));
     g_mutex_clear(&(slave->pluginInitLock));
@@ -196,7 +203,7 @@ gdouble slave_nextRandomDouble(Slave* slave) {
     return r;
 }
 
-void slave_addNewProgram(Slave* slave, const gchar* name, const gchar* path) {
+void slave_addNewProgram(Slave* slave, const gchar* name, const gchar* path, const gchar* startSymbol) {
     MAGIC_ASSERT(slave);
 
     /* store the path to the plugin with the given name, so that we can retrieve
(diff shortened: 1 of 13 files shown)


dhuseby:
Here's my example Rust application: [https://github.com/dhuseby/shadow-rust-test](https://github.com/dhuseby/shadow-rust-test)
It uses the tokio async io/net/futures crates to make a simple client and server.  The main start symbol is named "start".  When compiling the binary executable, main just calls start.  When compiling the dylib, the main symbol and code get stripped and start is exported as a C-callable symbol.

Shadow seems to correctly load the plugin dylib but it hangs due to a segfault related to Rust's TLS handling code in its stdlib.  The segfault is triggered by the TLS destructor registering code in the Rust stdlib.  I'm still investigating, I'm not exactly sure what's going on with it yet.  Maybe Rust doesn't follow the exact same convention as dylibs written in C and Shadow isn't getting all of the TLS information from the library?

dhuseby:
There must be a bug in the RTLD_DI_STATIC_TLS_SIZE calculation when using Rust binaries.  I just wrote a [simple C driver](https://github.com/dhuseby/shadow-rust-test/blob/master/src/driver.c) to test running the Rust code by calling the start symbol and it runs without a segfault.  I'll file a bug.

robgjansen:
I've been wondering if we could get compiled rust code running in Shadow. I'll need to take a deeper look at this, but thanks for taking a stab at this and contributing your code and findings!

src/main/core/shd-slave.c
```diff
@@ -27,6 +27,9 @@ struct _Slave {
     /* paths to programs (i.e. shadow plug-ins) that are run by virtual processes */
     GHashTable* programPaths;
 
+    /* start symbols for programs (i.e. shadow plug-ins) */
+    GHashTable* programStartSymbols;
```
robgjansen: Ideally, the start symbols and the paths would be stored in the same object and stored in a single hash table keyed by the name, rather than in two separate hash tables.

src/main/core/shd-slave.c
```diff
@@ -207,6 +214,15 @@ void slave_addNewProgram(Slave* slave, const gchar* name, const gchar* path) {
     } else {
         g_hash_table_replace(slave->programPaths, g_strdup(name), g_strdup(path));
     }
+
+    if(startSymbol == NULL) return;
```
robgjansen: Our usual style would be

```
if(startSymbol == NULL) {
    return;
}
```

Too bad we don't have a style guide!

src/main/host/shd-process.c
```diff
@@ -473,15 +481,15 @@ static void _process_loadPlugin(Process* proc) {
     /* make sure it has the required init function */
     gpointer symbol = NULL;
 
-    symbol = dlsym(proc->plugin.handle, PLUGIN_MAIN_SYMBOL);
+    symbol = dlsym(proc->plugin.handle, _process_getPluginStartSymbol(proc));
```
robgjansen: We could use something like:

```
_process_getPluginStartSymbol(proc) ? _process_getPluginStartSymbol(proc) : PLUGIN_DEFAULT_SYMBOL
```

and then avoid allocating a new string to copy the const string below.

src/main/host/shd-process.c
```diff
@@ -564,6 +572,11 @@ Process* process_new(gpointer host, guint processID,
     utility_assert(pluginName);
     proc->plugin.name = g_string_new(pluginName);
     proc->plugin.path = g_string_new(pluginPath);
+    if(pluginSymbol == NULL) {
+        proc->plugin.startSymbol = g_string_new(PLUGIN_DEFAULT_SYMBOL);
```
robgjansen: Along with the previous comment, we could leave proc->plugin.startSymbol NULL here, and handle NULL strings properly in places that use it. That way we don't allocate the constant PLUGIN_DEFAULT_SYMBOL string.

robgjansen:
@dhuseby I wonder if your [rust test case][https://github.com/dhuseby/shadow-rust-test] could be added as a Shadow test/ under src/test/ without requiring rust as a dependency. Do you know if cmake has a way to check if rust is installed? I'm thinking we could do something like the following in `src/test/CMakeLists.txt`:

```
if(HAVE_RUST)
    add_subdirectory(rust)
endif()
```

and then put your example in `src/test/rust`. Seems like cmake should have the ability to check for rust, but I if it doesn't I guess we could write our own check. Would you be interesting in submitting the cmake code and your rust example as a test case?

dhuseby:
Sorry, I've been MIA the last week for personal business. I'll catch up this week.

PR #398, opened 2018-02-10
Rust support

### LINKED ISSUE #397: STATIC_TLS_SIZE calculation doesn't work with Rust plugins
While researching using Shadow to simulate blockchain networks, I tried writing a [simple client and server in Rust](https://github.com/dhuseby/shadow-rust-test) that also compiled to a dylib to use as a Shadow plugin.  Rust doesn't allow a dylib to export the "main" function as a C callable symbol.  That required me to modify Shadow and submit a pull request to allow the [start symbol to be defined in the config file](https://github.com/shadow/shadow/pull/396).

Once I got around the start symbol problem, I was able to execute a simple two-node simulation and it crashes the thread on startup.  I jumped into GDB and saw that a segfault was coming from the Rust code that sets up destructors for TLS stored objects.  To figure out if it is Shadow or Rust, I wrote a [simple driver in C](https://github.com/dhuseby/shadow-rust-test/blob/master/src/driver.c) to load the Shadow plugin dylib and jump to the start symbol.  That works perfectly, so it doesn't seem to be a problem with the Rust runtime environment setup.

I investigated a little farther and I'm fairly certain the problem is in the elf loader that tries to calculate the correct value for Shadow's custom RTLD_DI_STATIC_TLS_SIZE info type.  I haven't walked through the elf loader code yet to see what it is doing and why it seems to be missing TLS information in the Rust dylib.  That's next.

jtracey:
Along with #396, this gets a [simple client/server setup written in rust](https://github.com/dhuseby/shadow-rust-test) working.

Fixes #397

### DIFF
--- a/src/external/elf-loader/glibc.c
+++ b/src/external/elf-loader/glibc.c
@@ -168,7 +168,7 @@ __attribute__ ((__regparm__ (1))) ___tls_get_addr (struct tls_index *ti)
 #endif
 
 EXPORT void internal_function
-_dl_get_tls_static_info (size_t * sizep, size_t * alignp)
+_dl_get_tls_static_info (size_t *sizep, size_t *alignp)
 {
   // This method is called from __pthread_initialize_minimal_internal (nptl/init.c)
   // It is called from the .init constructors in libpthread.so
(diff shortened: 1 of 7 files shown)


jtracey:
@robgjansen you probably want to take a look at the second commit, since that's not elf-loader code

dhuseby:
this is great, I'll test it out on my end.

robgjansen:
The shadow changes look great! Thanks @jtracey!

Issue #381, opened 2017-12-22
higher performance gdb support


jtracey:
Right now, gdb appears to have some operation it runs on every load of a shared object which is ~linear time in the number of shared objects loaded so far. I haven't looked at the problematic part of gdb's code yet, but I strongly suspect this is because gdb is iterating through the entire linkmap to find the newly loaded object, which will pretty much always be at the end of the linkmap. Since Shadow now functions by dynamically loading large numbers of shared objects, this makes using gdb with Shadow take quadratic time in the number of hosts. Currently, this gives acceptable but not great performance for small experiments, and horrible performance for large ones. I can think of two possible solutions to this problem:
 - fix gdb upstream, by finding the code that is doing this, and patching it in some way that is ABI-compatible with both glibc and elf-loader (e.g., iterating from the end of the linkmap backwards will give constant time performance for most use cases).
 - use elf-loader to trick gdb, by not adding newly loaded objects to the global linkmap until a particular function is called on it (which can be called by the user from the gdb interface). We might even be able to add all of the new objects in a single pass, though I suspect that gdb only supports adding objects one at a time right now.

The first option has the advantage of allowing Shadow to continue being debugged without having to teach the user any weird hacks and workarounds, and would allow for setting breakpoints before an object has been loaded, etc., while the second option has the advantage of not requiring any changes to gdb. My suggestion/plan is to use both approaches, so that current releases of gdb can be used with Shadow, while at some point in the future we get full gdb support.

PR #399, opened 2018-02-17
gdb performance improvements

jtracey:
This partially fixes the problems in #381 by using the second option there. That is, with 
1aef9b0 in particular, elf-loader doesn't add dynamically loaded objects to the linkmap, preventing gdb from finding them. For shadow, this means that the performance of running in gdb or attaching gdb to a running experiment will be considerably faster, with the drawback that no debug symbols will be loaded for plugins and their dependencies. While there isn't a full solution in place (that would require patching gdb), there are some helper functions to address plugin debug symbols:

 - `vdl_linkmap_abi_update()` will add all objects to the linkmap and notify gdb, at what is probably only slightly faster than what it was before. This is useful for smaller experiments, or experiments where you're willing to be very patient.
 - `vdl_linkmap_abi_from_addr (addr)` will add to the linkmap the object either pointed to by `addr` or which has a map containing `addr`, and notify gdb. This is useful for things like stack traces, where you take the address of any frame without debug symbols and call this function on it.
- `vdl_linkmap_abi_from_addrs (count, ...)` will do the same, for `count` addresses passed in. It's useful for making gdb scripts a little shorter.
- `vdl_linkmap_shadow_print()` and `vdl_linkmap_abi_print()` print all the objects that have been dynamically linked/loaded, or have been added to the linkmap, respectively. The former is useful for seeing what is available to load debug symbols for, and the latter is to check what already has. The handles printed by these can be used in `vdl_linkmap_abi_from_addr (addr)`.

I'll need to document this on the wiki, along with some example gdb scripts for doing things like automatically loading all debug symbols for a segfault stacktrace.

Something this doesn't implement which might be useful is the ability to optionally run with the old behavior (update the linkmap on every load), since with this patch it's impossible to put a breakpoint in any object which hasn't been loaded yet.

### DIFF
--- a/src/external/elf-loader/gdb.c
+++ b/src/external/elf-loader/gdb.c
@@ -5,7 +5,7 @@
 #include <link.h>
 
 static
-ElfW (Dyn) *
+ElfW(Dyn) *
 file_get_dynamic (const struct VdlFile *file, long tag)
 {
   ElfW (Dyn) * cur = (ElfW (Dyn) *) file->dynamic;
@@ -43,7 +43,7 @@ gdb_initialize (struct VdlFile *file)
   // in the DT_DEBUG entry of the main executable dynamic section
   // because this is where gdb goes to look to find the pointer and
   // lookup an inferior's linkmap.
-  ElfW (Dyn) * dt_debug = file_get_dynamic (file, DT_DEBUG);
+  ElfW(Dyn) *dt_debug = file_get_dynamic (file, DT_DEBUG);
   unsigned long *p = (unsigned long *) &(dt_debug->d_un.d_ptr);
   *p = (unsigned long) &g_vdl;
 }
@@ -52,5 +52,8 @@ void
 gdb_notify (void)
 {
   g_vdl.state = VDL_CONSISTENT;
-  g_vdl.breakpoint ();
+  if (g_vdl.breakpoint)
+    {
+      g_vdl.breakpoint ();
+    }
 }
--- a/src/external/elf-loader/stage1.c
+++ b/src/external/elf-loader/stage1.c
@@ -98,6 +98,7 @@ global_initialize (unsigned long interpreter_load_base)
 
   vdl->version = 1;
   vdl->link_map = 0;
+  vdl->shadow_link_map = vdl_list_new();
   vdl->link_map_lock = rwlock_new ();
   vdl->breakpoint = 0;
   vdl->state = VDL_CONSISTENT;
@@ -241,6 +242,7 @@ stage1_freeres (void)
   vdl_utils_str_list_delete (g_vdl.search_dirs);
   vdl_hashmap_delete (g_vdl.files);
   vdl_hashmap_delete (g_vdl.contexts);
+  vdl_list_delete (g_vdl.shadow_link_map);
   futex_delete (g_vdl.ro_cache_futex);
   rwlock_delete (g_vdl.global_lock);
   rwlock_delete (g_vdl.tls_lock);
@@ -280,7 +282,7 @@ stage1 (struct Stage1InputOutput *input_output)
   extern ElfW(Dyn) _DYNAMIC[] __attribute__ ((visibility ("hidden")));
   extern const ElfW(Addr) _GLOBAL_OFFSET_TABLE_[] __attribute__ ((visibility ("hidden")));
   ElfW(Addr) load_base = (ElfW(Addr)) &_DYNAMIC - (unsigned long)_GLOBAL_OFFSET_TABLE_[0];
-  
+
   relocate_dt_rel (load_base);
 
   // Now that access to global variables is possible, we initialize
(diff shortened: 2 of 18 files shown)


jtracey:
Another possibility here is to expose these functions to Shadow, so that Shadow can load the debug symbols for every plugin, but not its dependencies. The tgen plugin has 7 dependencies, the tor plugin has 13, the tor-specific preload has 1, and the tor-ctl plugin has 4, so that's 25 per Tor client. It wouldn't make the loads less quadratic, but it would be a significant speedup in absolute terms (~29^2/4^2=53x speedup over the original performance, ~16x slowdown per Tor client vs this branch).

robgjansen:
This is great!

I'm not sure about this approach vs. having shadow automatically load all plugins but not their dependencies. It seems like this approach is the most efficient, and so probably the better approach in the general case as long as there is good documentation on how to get the symbols loaded. The automatic load case seems like it may be useful in specific situations, like on small networks or when you really don't want to manually load everything.

So, how about we merge this and use it as the default case, and then add support for shadow auto-loading plugin symbols but not dependencies, but only in the case that the user turns it on via a new shadow cli config switch. Does that seem reasonable? If so, should I merge this now and do the auto-load as a separate PR?

jtracey:
Yup, that sounds reasonable to me.

To be clear, the biggest use case I can see for loading plugin symbols by default is for when you want to put a breakpoint on all plugins, even ones that haven't been loaded yet. As in, suppose you know that the segfault you're seeing happens when you get to some code path in file.c, line 100, while x=0. So you want to put a breakpoint on "file.c:100 if x==0" to break *before* the segfault occurs (or just generally, any time you want a breakpoint on some non-backtrace inducing buggy behavior). With this PR, you can't do that, unless you know exactly which instance of the plugin it's going to occur in, that instance has already been dlmopen'd, and you added it to the linkmap.

Now that I think of it, you could actually implement the ability to load debug symbols every plugin as a gdb script that executes on a breakpoint immediately after the dlmopen call, so technically there's no reason we'd have to implement it in Shadow, other than convenience.

robgjansen:
> Now that I think of it, you could actually implement the ability to load debug symbols every plugin as a gdb script that executes on a breakpoint immediately after the dlmopen call, so technically there's no reason we'd have to implement it in Shadow, other than convenience.

Do you mean, when the breakpoint hits, elf-loader automatically loads the symbols "just in time"? If you didn't mean that, is there a reason that would not be possible?

Also, couldn't we do something similar on segfault? I.e., Shadow already sets a custom signal handler for segfault - it's in shd-process.c, and it just immediately calls abort in the plugin context so that the pth threads gets cleaned up properly. I think GDB automatically stops on segfault too. Could we load the symbols "just in time" from the elf-loader? Otherwise, I think we could do it from the Shadow custom segfault handler (though I'm not sure how Shadow would know that we are debugging with gdb).

jtracey:
>Do you mean, when the breakpoint hits, elf-loader automatically loads the symbols "just in time"? If you didn't mean that, is there a reason that would not be possible?

No, unfortunately that is impossible. There are, AFAIK, only two things you can put a breakpoint on: an absolute address, or an offset from a symbol. Absolute addresses can only apply to one instance of an object (which typically isn't what you want when you have a bunch of plugins loaded -- I want to stop here when *any* tor plugin hits this line, not just when tor plugin 72 hits that line), and can only be placed after the object's location in memory is known. A breakpoint at an offset from a symbol can only be placed if debug symbols for that object are already loaded, since gdb has no knowledge of symbols outside of debug symbols (a function is just an address in memory without them, for example). So there wouldn't be any way to stop gdb and load the symbols, because gdb doesn't know where to break.


>Also, couldn't we do something similar on segfault?

Yes, I don't have it on hand but I wrote a gdb script that does this. Since with a segfault you have access to the backtrace for the location you're interested in, you can know which symbols to load. To do it from elf-loader it would have to implement a signal handler for SIGSEGV, which wouldn't be too hard because we already have the SIGPROF handler implemented. The thing that I don't know is, is there some way to "forward" a signal as though it hadn't been caught? As I understand it, if elf-loader caught the SIGSEGV, Shadow wouldn't be able to also catch it and clean up. 

>Otherwise, I think we could do it from the Shadow custom segfault handler

Yeah, if there's no way to pass the signal along, we can just expose some of these functions in libvdl and let Shadow call them.

>(though I'm not sure how Shadow would know that we are debugging with gdb)

I don't think it matters for segaults, they should presumably be rare enough that we don't need to concern ourselves with gdb performance for them, right? So just every time there's a segfault, add the entire stack trace into the linkmap -- if you're in gdb you'll probably just look at the first one, which will be pretty fast, and if you're not in gdb, then there's no slowdown, it's just doing the same operations as we were doing before.

jtracey:
Oh, and to clarify what I meant: I was just saying that instead of having an option to load debug symbols for plugins by default, the user can just run a gdb script that does that, since the function can be called after dlmopen returns the plugin handle.

jtracey:
hold off on merging this, I think there's a bug

jtracey:
okay should be good now

PR #403, opened 2018-02-23
Add support for Debian testing

ahf:
I'm moving to using Shadow on my workstation, which is running Debian testing. I don't know if it makes sense to include this patch into Shadow itself as the ld version will probably change before Debian testing becomes the next stable.

Maybe it would make more sense to start matching on a glob here instead of matching on specific versions for the different distributions?


robgjansen:
> Maybe it would make more sense to start matching on a glob here instead of matching on specific versions for the different distributions?

Yes, I think it does make sense. I'll accept the PR for now while a long-term solution can be tracked in #408.

PR #409, opened 2018-03-06
minor elf-loader changes

### LINKED ISSUE #408: glob matching when searching for elf loader ld debug file
In `src/external/elf-loader/extract_system_config.py`, the `search_debug_file` function hard codes ld debug file paths. Adding glob matching to `find_build_id` would avoid the need to patch the file every time the ld version changes.

jtracey:
The first commit 17817c1  switches to using regex to find ld.so debug files, which fixes #408 (I opted for regex over globs to be a bit stricter in what we match).

The second commit b40578a fixes what should be a minor performance problem (some code we inherited used memcpy when it could have just used assignment), but on closer inspection might be a non-trivial boost since it occurs in some of the most frequently used elf-loader code (outside of dynamic loading), is code that is often called from within a global lock, and this should allow for more significant compiler optimizations. I would have to do some measurements to say for sure though, and I don't think it will be *that* significant.

I was going to add a commit to fix a very minor memory leak as well, but since it involves switching out a linked list for a hash table and I don't have enough time today to give it enough testing to feel comfortable committing just yet, I'll hold off on that for now.

### DIFF
--- a/src/external/elf-loader/extract-system-config.py
+++ b/src/external/elf-loader/extract-system-config.py
@@ -5,24 +5,6 @@ import re
 import getopt
 import os
 
-
-def find_build_id(path):
-    if os.path.exists(path):
-        file = os.popen('readelf -n {}'.format(path), 'r')
-        lines = file.readlines()
-        regex = re.compile(r'^    Build ID: (\w+)$')
-
-        for line in lines:
-            result = regex.search(line)
-
-            if not result:
-                continue
-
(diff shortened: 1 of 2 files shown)


robgjansen:
@jtracey would you like me to wait for the linked list -> hash table change before tagging stable release v1.12.1?

jtracey:
I'll see if I can find time to do it today, otherwise I wouldn't worry about it.

PR #410, opened 2018-03-07
use /usr/bin/env in scripts

jtracey:
Some distros, e.g. Ubuntu, don't put crucial executables like bash in /usr/bin.


robgjansen:
Sorry about that! Thanks for the fix!

PR #416, opened 2018-03-18
various bug fixes

jtracey:
While working on getting shadow running on the new machine in CrySP, I've encountered various bugs. The most significant of these is that a typo was preventing the work-stealing scheduler from ever being used (when I was originally implementing/measuring it, my local shadow copy implemented it over the hostsingle policy). This also means the TLS swapping code hasn't been used with recent changes. I've done some pretty thorough testing on an Ubuntu 16.04 machine, as well as some testing on Ubuntu 17.10 and a Solus VM, so hopefully it doesn't introduce any new bugs, but you might want to test it before merging too.


robgjansen:
This builds correctly for me on Cent OS 7, and all of the test cases pass.

PR #417, opened 2018-03-19
fix build warning

frochet:
Hey,

This should remove a compiler warning on ./setup build.

Best,


frochet:
Hm, just notices that one of #416 commits feature the same one. You can safely ignore mine :)

robgjansen:
Great, thanks for reporting, and for paying attention and noticing that this has been fixed in #416!

PR #706, opened 2020-01-27
Undefined name: 'e' --> Exception

cclauss:
`e` is an undefined name in this context so a NameError will always be raised instead of the actual exception.

Discovered by #704


robgjansen:
Thank you!

PR #707, opened 2020-01-29
Add Github workflow for continuous integration testing

sporksmith:
Adds a workflow to build and test shadow on every pull request or commit to master.

### DIFF
--- a/.github/workflows/build_shadow.yml
+++ b/.github/workflows/build_shadow.yml
@@ -1,0 +1,33 @@
+# Syntax reference:
+# https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions
+
+name: Tests
+
+on:
+  push:
+    paths-ignore:
+      - '**.md'
+      - 'LICENSE'
+  pull_request:
+    paths-ignore:
+      - '**.md'
+      - 'LICENSE'
+
+jobs:
+  testing:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: sudo apt install -y gcc g++ cmake make xz-utils python libglib2.0-0 libglib2.0-dev libigraph0v5 libigraph0-dev libc-dbg python-pyelftools
+
+      - name: Build
+        run: python setup build --clean --test
+
+      - name: Install
+        run: python setup install
+
+      - name: Test
+        run: python setup test
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -1,5 +1,16 @@
-name: lint_python
-on: [push, pull_request]
+# Syntax reference:
+# https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions
+
+name: PythonLint
+on:
+  push:
+    paths:
+      - 'setup'
+      - '**.py'
+  pull_request:
+    paths:
+      - 'setup'
+      - '**.py'
 jobs:
   lint_python:
     runs-on: ubuntu-latest
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # The Shadow Simulator
 
+![](https://github.com/shadow/shadow/workflows/Tests/badge.svg)
+
 Shadow is a unique discrete-event network simulator that runs real 
 applications like Tor and Bitcoin, and distributed systems of thousands of
 nodes on a single machine. Shadow combines the accuracy of emulation with the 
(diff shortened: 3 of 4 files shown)


sporksmith:
I need to change the badge path in the README before final merge; it's currently pointing into my repo for testing/validation. 

Probably also makes sense to squash this into a single commit before merging. Left unsquashed for review in case that's useful.

.github/workflows/build_shadow.yml
```diff
@@ -0,0 +1,24 @@
+# Syntax reference:
+# https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions
+
+name: Test
```
robgjansen: This is what appears on the badge. Should we use the plural form so it reads `Tests: Passing`?

An aside: [here are some other badges that are cool](https://github.com/badges/shields) :)

.github/workflows/build_shadow.yml
```diff
@@ -0,0 +1,24 @@
...
+on: [push, pull_request]
+
+jobs:
+  testing:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
```
robgjansen: Super nit-pick: Checkout is capitalized, but the other steps are not.

sporksmith: Done. Agreed it should be consistent. I could go either way but arbitrarily capitalized them all since that's what the examples do.

.github/workflows/build_shadow.yml
sporksmith: Done.

The custom badges are neat. It's not clear to me how to easily wire these up to the test results. Maybe something to poke at later, but I'd rather get this in first :).

sporksmith: For the tests it seems safer to use an ignore-list than an include-list. I set it to ignore markdown files, which seems pretty safe.

I changed the python action to only trigger and setup and *.py.

Unfortunately I don't see a way to set up file filters for multiple triggers (push and pull_request) without duplicating. (I suppose we could use another tool to generate these files and check in the generated output when we change the source, but seems a bit overkill for now).

sporksmith:
Ok, I think I'm ready for another look when you get a chance.

Btw still learning github workflow and review conventions - let me know if there are any issues. e.g. in the tools I'm used to, the convention is for the reviewee to mark comments closed if they believe they've addressed them; the reviewer can reopen them if they still have further feedback on that particular issue. If it's better in these tools to let the reviewer mark comments resolved, I'm fine with that too.

.github/workflows/build_shadow.yml
sporksmith: Tried using YAML anchors to deduplicate the filter paths, but github doesn't support them: https://github.com/sporksmith/shadow/runs/417939170?check_suite_focus=true

robgjansen:
> in the tools I'm used to, the convention is for the reviewee to mark comments closed if they believe they've addressed them; the reviewer can reopen them if they still have further feedback on that particular issue. 

This seems fine to me.

Also, I noticed that there is a "request review" feature and github (in the reviewers section at the top right of the PR), where you can request a review from me when you are ready. (It looks like I self-requested a review on this one.)

robgjansen:
Oh, it might also be good to set the "assignee" as the person who should merge it. Usually that should be the person who created the PR, probably, but if for some reason we want to change that we can use the assignee field for that.

sporksmith:
Thanks! I've squashed this down to one commit, but i don't think I can (fast-forward) merge since I don't have write permission.

sporksmith:
It looks like you could give me write permission, but configure it so that you need to approve requests before I'd be able to merge it in: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-reviews#required-reviews

robgjansen:
I tweaked the settings to give you write and require a review, tests to pass, and rebase in order to push. Are you allowed to rebase now?

sporksmith:
Thanks, done!

There are some strange workflow errors, but I suspect those are an artifact of pushing the workflows themselves.

`- Only actions in "shadow" are allowed for this repository`

I guess we'll see what happens with the next non-workflow-commit to master.

sporksmith:
Actually I'm hitting the same errors again. I searched some more and may have found the issue:

https://github.community/t5/GitHub-Actions/How-can-I-check-out-code-with-quot-Enable-local-Actions-only-for/td-p/36124

When you were tweaking the settings might you have turned on "Enable local Actions only for this repository"?

robgjansen:
> When you were tweaking the settings might you have turned on "Enable local Actions only for this repository"?

Yes! Sorry about that, I just undid that setting.

Issue #705, opened 2020-01-27
Migrate wiki into docs subdirectory


sporksmith:
The wiki is currently a separate git repo from the code itself. Migrating it into the same repo would create a tighter coupling between the code and documentation, e.g. allowing us to make atomic changes to both simultaneously. It'd also force updates to be in the form of pull requests which can be reviewed.

- [x] Port wiki with history into the code repo.
- [x] Maybe/TBD: Update wiki pages to point to the new locations, in the code repo.
- Deferred ~[ ] Delete the wiki entirely. (Possibly after some time period of leaving the stubs up)~

sporksmith:
Moving discussion from https://github.com/shadow/shadow/pull/708#issuecomment-580770139

Let's move discussion to #705 for posterity :)

> Now that this is merged, we have 3 options:
> 
>     1. Leave the wiki as-is
> 
>     2. Disable the wiki, removing the pages and breaking and links pointing to it
> 
>     3. Leave as-is, but restrict editing to those with push access (i.e. me and @sporksmith) - with the policy that the wiki is read-only
> 
> 
> I think 3. is the best answer, with the goal of eventually disabling the wiki after some time. But I could be talked into doing 2. now.

I was thinking roughly 3, but also to replace the content of the individual pages with links to their individual new homes, so that people don't unknowingly consume stale docs. Preparing a pull request to do so now.

sporksmith:
@robgjansen 

It turns out there's no way to create a pull request for a wiki (more reason to move the docs into the code repo!). Instead I pushed the proposed rewrite to here: https://github.com/sporksmith/Temp-Shadow-Wiki/wiki

PTAL, and if it looks good I'll push directly to the shadow wiki.

robgjansen:
Looks good to me.

Also, FYI, I noticed that if you to visit a page on the wiki that does not exist, [like this one](https://github.com/sporksmith/Temp-Shadow-Wiki/wiki/5-Developer-Guide), then it redirects you to the wiki home page. But that doesn't give us a one-to-one mapping from the old page to the new one, so it's probably best to do it the way you have done. (Maybe at the point when we are ready to delete the wiki, we could instead delete all pages except the Home.md page, and then add a short blurb on it and a link to the docs/README.md?)

sporksmith:
Sounds good

PR #708, opened 2020-01-30
Merge wiki

sporksmith:
This imports the wiki content and history (into the code repository).

I rewrote the wiki's history with git filter-branch so that all commits retroactively happen on the `docs` subdirectory, created a merge commit, and then did a few fixups.

So far the main disadvantage I see of this approach over the wiki is that the wiki has an auto-generated site-map in the sidebar, and has an option for a custom sidebar. We could create our own sitemap easily enough, but AFAICT can't get github to render it in a sidebar.

Given that there aren't that many pages here maybe that's not such a big deal.


docs/README.md
```diff
@@ -0,0 +1,37 @@
+# The Shadow Simulator
```
robgjansen: I think we need a `README.md` in the base of the repo, because many people will land on https://github.com/shadow/shadow and we want a file there to explain what the project is about. It's an extra click to get to `docs/`.

I think we should move this README back to the base, move the `contents` section to a new docs/README.md file, and then have the base README link to the docs README.

sporksmith: > Is it possible to keep consistent with our rebase strategy to avoid merge commits and keep a clean commit history? Or is there some reason that won't work here?

Yeah, I'd previously tried rebasing everything on top or even cherry picking everything on top, but kept running into conflicts. I think part of the issue might be that the wiki itself has a branches and merges in the history. I'll poke a bit more though and see what I can do.

> I think we need a README.md in the base of the repo

If there's no README.md in the root, github will render docs/README.md instead. e.g. see https://github.com/sporksmith/shadow/tree/merge-rewritten-wiki. You're proposed structure sounds fine to me too, though. I'll go ahead and give that a shot.

sporksmith: Took another stab at this.

Indeed the tricky part is that rebase seems to want to serialize the wiki history, which results in having to (re)resolve some merge conflicts. I went through and made a best effort of doing so. I'm not 100% sure I got it right, but at least the history is mostly preserved. I also validated that the end-state matched the wiki repo before re-applying my fixups.

I also left the two READMEs separate as-requested.

sporksmith:
I squashed my fixup commits, which retriggered the workflows, and I'm now able to merge.

Before I do just want to double check: We want to "rebase and merge", right? With the expected outcome that this will add 472 commits to the tip of shadow's history. (As opposed to squash and merge, which would effectively lose the history of the individual commits and just add 1 commit to the tip of shadow).

robgjansen:
It looks like your new commit triggered the correct status checks (now that we have the github actions settings correct). :+1:

robgjansen:
Rebase and merge seems OK to me; yeah, otherwise we lose the history as you described.

robgjansen:
Now that this is merged, we have 3 options:
1. Leave the wiki as-is
1. Disable the wiki, removing the pages and breaking and links pointing to it
1. Leave as-is, but restrict editing to those with push access (i.e. me and @sporksmith) - with the policy that the wiki is read-only

I think 3. is the best answer, with the goal of eventually disabling the wiki after some time. But I could be talked into doing 2. now.

sporksmith:
Let's move discussion to #705 for posterity :)

PR #709, opened 2020-01-31
Use dynamic port assignment for server ports in tests.

sporksmith:
This fixes the bug where the shutdown test fails if run multiple
times, and generally prevents the possibility of tests failing
in the case that a hard-coded port number happens to already
be in use.

### DIFF
--- a/src/test/shd-test-common.c
+++ b/src/test/shd-test-common.c
@@ -9,7 +9,7 @@
 #include <stdio.h>
 #include <string.h>
 
-int common_setup_tcp_sockets(int* server_listener_fd_out, int* client_fd_out, in_port_t server_listener_port) {
+int common_setup_tcp_sockets(int* server_listener_fd_out, int* client_fd_out, in_port_t* server_listener_port_out) {
     /* set up server */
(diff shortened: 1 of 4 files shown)


sporksmith:
> I'm really glad we have that github test action in place as it increases my confidence in the code changes :)

Yup! Once you get used to having some sort of automated testing in place, working without it is a scary prospect :)

Have a good weekend!

sporksmith:
Btw it occurred to me later - actually shouldn't hard-coded ports be fine under Shadow? Shouldn't the set of available ports available at the start of the simulation be ensured to always be the same?

robgjansen:
That is true for Shadow, yes. Hard-coded ports would affect the versions of the tests that are run outside of Shadow, though.

Do you recall which version of the tests were failing - the Shadow version or the Linux version?

sporksmith:
Thanks; I didn't realize the tests were run both natively and under shadow. Confirmed that it was the native version failing:

```
The following tests FAILED:
	 44 - shutdown (Failed)
Errors while running CTest
```

Issue #710, opened 2020-01-31
Some noreturn functions return


sporksmith:
It looks like several functions whose declarations have the "noreturn" attribute have shims that *do* return. This not only generates compile-time warnings, but is technically [undefined behavior](https://en.cppreference.com/w/c/language/_Noreturn) so could in principle result in subtle misbehavior in the emulator:

>  If the function declared _Noreturn returns, the behavior is undefined.

I'm digging into how difficult it'd be make these *actually* not return (and e.g. instead directly longjmp back into the emulator).

Here are the corresponding warnings (with -Werror enabled):

```
In file included from /home/jnewsome/src/shadow/src/preload/shd-interposer.c:222:0:
/usr/include/pthread.h: In function ‘__pthread_unwind_next’:
/home/jnewsome/src/shadow/src/preload/shd-preload-defs.h:281:1: error: ‘noreturn’ function does return [-Werror]
 PRELOADDEF(      , void, __pthread_unwind_next, (__pthread_unwind_buf_t *a), a);
 ^~~~~~~~~~
/usr/include/stdlib.h: In function ‘exit’:
/home/jnewsome/src/shadow/src/preload/shd-interposer.c:417:1: error: ‘noreturn’ function does return [-Werror]
 }
 ^
/usr/include/pthread.h: In function ‘pthread_exit’:
/home/jnewsome/src/shadow/src/preload/shd-interposer.c:427:1: error: ‘noreturn’ function does return [-Werror]
 }
 ^
/usr/include/stdlib.h: In function ‘abort’:
/home/jnewsome/src/shadow/src/preload/shd-interposer.c:437:1: error: ‘noreturn’ function does return [-Werror]
 }
```

PR #713, opened 2020-02-05
Ensure functions declared noreturn will never return.

sporksmith:
Resolves GH-710.


sporksmith:
Could you take another look? I ended up tweaking it a bit more - notably in __pthread_unwind_next, in non-emulated code it calls the real instead of blowing up, and in emulated code it tries to just abort the emulated process instead of the whole process.

PR #720, opened 2020-02-11
Set up a strategy matrix to test on both gcc and clang

sporksmith:


sporksmith:
Not entirely sure what's going on with the checks, but I *suspect* it's running the actions specified in my PR, but looking for successful completion of the actions specified in the shadow repo.

Might need you to merge this one.

Issue #711, opened 2020-01-31
Use -Werror compiler flag


sporksmith:
It'd be good to use the -Werror compiler flag to help prevent new subtle bugs from slipping in.

Enabling it is currently blocked on fixing (or maybe suppressing) the warnings that are already being generated.

sporksmith:
Broke "noreturn" warnings out into #710 .

PR #722, opened 2020-02-12
Fix one more clang warning and add -Werror.

sporksmith:
This can be reviewed now but shouldn't be merged until after clang is added to
the CI matrix (at which point CI will fail for this PR), and the other clang
warning fixes are merged in (after which it will pass again).

https://github.com/shadow/shadow/issues/711


sporksmith:
Hah, I'd forgotten to actually run the tests after compiling with gcc. Looks like doing it this way fixes the warning in clang, but breaks the tests in gcc. Poking at it a bit more, but may resort to just suppressing the unused argument warning.

sporksmith:
Replacing this with two new pull requests:
* #725 fixes the warning without breaking the build
* Another to enable Werror only for CI builds. Not by default since that'd needlessly break the build on a lot of configurations that we don't test.

PR #723, opened 2020-02-12
Add yum-utils dependency for Fedora

sporksmith:
In particular noticed this was needed on Fedora 31.
Also split dependency lists across lines and sort.


sporksmith:
It looks like I can't merge this one because it wants a passing 'testing' workflow, but the 'testing' workflow doesn't run on a markdown-only change

PR #725, opened 2020-02-20
Pass link flags only to link steps.

sporksmith:
* Adds a macro add_ldflags analagous to add_cflags.
* Uses that macro to fix a "unused-command-line-argument" warning in
  clang, where link flags were being passed to compile steps.

Progress on GH-711.


sporksmith:
Thanks!

>  I wonder if there are other places where we were adding linker flags with the add_cflags macro
`git grep add_cflags` doesn't turn up any; and presumably we'd be getting warnings in those places as well if there were.

PR #726, opened 2020-02-20
Enable Werror in continuous integration workflow.

sporksmith:
Fixes GH-711


CMakeLists.txt
```diff
@@ -101,6 +102,7 @@ set(CMAKE_C_FLAGS_DEBUG "")
 set(CMAKE_C_FLAGS_RELEASE "")
 add_compile_options(-ggdb)
 add_compile_options(-fno-omit-frame-pointer)
+add_compile_options(-Werror)
```
robgjansen: Do you really want to add `-Werror` here? It seems to me you should only be adding it below in the `if(SHADOW_WERROR STREQUAL ON)` block.

sporksmith: Oops, good catch!

PR #730, opened 2020-03-04
New router and netiface token buckets

robgjansen:
This adds a new router component between hosts that allows us to implement multiple queue management algorithms. I've implemented some initial queue management algorithms, and the default is CoDel. This has significantly improved congestion control in Shadow's network stack and we have observed much more consistent bandwidth utilization at high transfer rates when using the new code.

The other major change is a refactoring of the network interface so that bandwidth limits are enforced via token buckets rather than the previous dynamic time based heuristic.

Some TODO items before merging:
- [x] I need to update the leakcheck test so that it includes the printout of the router object counts.
- [ ] I would like to bump the version number to v1.14.0 tag v1.14.0 after merging.


robgjansen:
Do we want the shadow-plugin-tor-ci to also be required, like the others?

src/main/host/descriptor/shd-tcp.c
```diff
@@ -499,8 +499,8 @@ static void _tcp_tuneInitialBufferSizes(TCP* tcp) {
     /* make sure the user hasnt already written to the buffer, because if we
      * shrink it, our buffer math would overflow the size variable
      */
-    utility_assert(socket_getInputBufferLength(&(tcp->super)) == 0);
-    utility_assert(socket_getOutputBufferLength(&(tcp->super)) == 0);
+//    utility_assert(socket_getInputBufferLength(&(tcp->super)) == 0);
+//    utility_assert(socket_getOutputBufferLength(&(tcp->super)) == 0);
```
sporksmith: Are these no longer expected to hold? If so should they just be deleted?

src/main/host/shd-network-interface.h
```diff
@@ -7,10 +7,10 @@
 #ifndef SHD_NETWORK_INTERFACE_H_
 #define SHD_NETWORK_INTERFACE_H_
 
-#include "shadow.h"
-
 typedef struct _NetworkInterface NetworkInterface;
 
+#include "shadow.h"
```
sporksmith: I think the usual convention is for #includes to be the first thing after the include-guard? If there's a specific reason why NetworkInterface needs to be defined first, probably worth a comment?

src/main/routing/shd-router-queue-codel.c
```diff
@@ -0,0 +1,276 @@
...
+#include <math.h>
+
+#include "shadow.h"
+
+/* hard limit of queue size, in number of packets */
+//#define CODEL_PARAM_LIMIT 1000
+#define CODEL_PARAM_LIMIT G_MAXUINT
+
+/* target minimum standing queue delay time (5 milliseconds) */
+//#define CODEL_PARAM_TARGET (5*SIMTIME_ONE_MILLISECOND)
```
sporksmith: Delete this line?

src/main/routing/shd-router-queue-codel.c
```diff
@@ -0,0 +1,276 @@
...
+
+#include <math.h>
+
+#include "shadow.h"
+
+/* hard limit of queue size, in number of packets */
+//#define CODEL_PARAM_LIMIT 1000
+#define CODEL_PARAM_LIMIT G_MAXUINT
+
+/* target minimum standing queue delay time (5 milliseconds) */
```
sporksmith: Delete "5 milliseconds", since the ground truth of the value can be seen below and is already out of date? :)

src/main/routing/shd-router.c
```diff
@@ -0,0 +1,122 @@
...
+ * packet bursts from fast networks onto slow networks.
+ *
+ *  Created on: Jan 7, 2020
+ *      Author: rjansen
+ */
+
+#include "shadow.h"
+
+
+struct _Router {
```
sporksmith: Probably not a reason to deviate from this convention in this PR, but I believe in general names starting with _[A-Z] are reserved. See e.g. https://stackoverflow.com/a/228797, or more authoritatively [7.1.3 Reserved identifiers](https://port70.net/~nsz/c/c11/n1570.html#7.1.3) "All identifiers that begin with an underscore and either an uppercase letter or another underscore are always reserved for any use. "

src/main/host/shd-network-interface.c
```diff
@@ -105,7 +106,9 @@ static void _networkinterface_refillTokenBucketsBoth(NetworkInterface* interface
             _networkinterface_getRefillIntervalReceive());
 
     /* the refill may have caused us to be able to receive and send again */
-    networkinterface_receivePackets(interface);
+    if(interface->router) {
+        networkinterface_receivePackets(interface);
+    }
```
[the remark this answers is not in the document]
robgjansen: Null router is valid, it means we are a loopback interface and are not internet-connected. I'll add some more comments in the code to make that more clear.

src/main/routing/shd-router.c
robgjansen: Noted, thanks. Another thing to decide on for the contributors doc ;)

PR #731, opened 2020-03-05
Add a .clang-format configuration

sporksmith:
This specifies the formatting preferences for the clang-format tool,
which should pick up this configuration automatically.  I started with
clang-format's defaults and made a few tweaks in accordance with the
preferences observed in the existing code.

I don't plan or recommend mass reformatting, but this is useful
for keeping modified code tidy and formatted consistently.
A mass reformat might make sense down the line once we've had time
to get comfortable with the options, and don't have any large
pending changes (since it'd make merging a huge headache).

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,123 @@
+# Options reference:
+# https://clang.llvm.org/docs/ClangFormatStyleOptions.html
+#
+# To add Ctrl-k as a "format region" in visual and select modes of vim:
+# vmap <C-K> :py3f /usr/share/vim/addons/syntax/clang-format.py<cr>
+#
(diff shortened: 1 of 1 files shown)


sporksmith:
> We should probably provide instructions somewhere (in the new contributors file when we have one) about how to run the format tool.

Added to #733 

> Also, can/should we enforce formatting through a Github action?

I think we *could* e.g. have a github action that runs clang-format on the modified lines and fails if it produces a diff. We should probably get a little more experience with this config before we completely mandate it though.

In the shorter term it might be helpful to add a client-side pre-commit hook that runs it. (By design developers would need to opt in by copying a checked in hook into their .git directory)

PR #734, opened 2020-03-09
Better version management

robgjansen:
Making it easier to track and tag versions using bumpversion. Also adding the major, minor, and patch versions to `shd-config.h` so they can be accessed from the source code.


.bumpversion.cfg
```diff
@@ -0,0 +1,6 @@
+[bumpversion]
```
sporksmith: Maybe add a comment to this file with a pointer to the bumpversion tool?

robgjansen: Tried that, but it appears that bumpversion removes all comments from the file after you execute it. So we should add more details to the contributing file when we have one (i.e., a link to the tool and the process we follow to tag new versions).

PR #743, opened 2020-03-21
Update a dead link on the Getting Started Tutorial

florian-vuillemot:


robgjansen:
@sporksmith looks like we can't merge doc-only changes, since it doesn't trigger the tests to run, which we require for merging :/ I can override in this case, but how should we generally deal with this case? Just remove the action filters and run the tests always?

sporksmith:
> @sporksmith looks like we can't merge doc-only changes, since it doesn't trigger the tests to run, which we require for merging :/ I can override in this case, but how should we generally deal with this issue? Just remove the action filters and run the tests always?

Yeah, that's probably the best solution. Better to waste some compute resources running unnecessary tests than to waste a human's time trying to override the check.

PR #746, opened 2020-03-26
Build and test debug mode too

sporksmith:
Just came across the --werror --debug build getting broken in res/phantom. While that particular breakage was fairly innocuous, in general it's a good idea to test the debug build too.

### DIFF
--- a/.github/workflows/build_shadow.yml
+++ b/.github/workflows/build_shadow.yml
@@ -18,6 +18,7 @@ jobs:
     strategy:
       matrix:
         cc: ['gcc', 'clang']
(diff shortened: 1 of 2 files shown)


sporksmith:
Oh, you'll need to update the required actions again before I can merge this, since adding a dimension to the configuration matrix changes all the names

robgjansen:
OK, I went in and checked all of the boxes.

sporksmith:
You made the new ones required but still need to make `testing (clang)` and `testing (gcc)` not required :)

robgjansen:
Whee!

PR #747, opened 2020-03-29
Allow to run Shadow scripts with python2 and 3

florian-vuillemot:
With the end of life of python2, user should be able to use python3 to run scripts. 
For compatibility reasons, this PR keeps the code functional with python2.


florian-vuillemot:
I would like to continue on the directory https://github.com/shadow/shadow/tree/master/src/tools/topology but I can't find dependencies (as in https://github.com/shadow/shadow/blob/master/src/tools/topology/generate-topology.py).
@robgjansen , do you have an idea where I can find all that stuff ?

robgjansen:
Hmm... I think all of the topology scripts are obsolete and should probably be removed, rather than wasting time updating them for python3.

I added #748 for removing the old scripts and adding some new ones.

florian-vuillemot:
So I consider this PR done. Ready to merge at your initiative.
About the new issue, I agree to help. Just give me all the information or where to find it.

robgjansen:
I don't think we can merge this until after we remove the topology scripts (as part of #748), because otherwise the python 3 linter will cause the CI tests to fail.

florian-vuillemot:
Ok. I removed the topology scripts. I hope that I understood correctly as previously I though you wanted to remove by replacing (in the same time).
I also excluded src/external from the pylint analysis.

robgjansen:
Oh, I'm sorry for not being clear. I do want to remove and replace the topology scripts at the same time. But that should be done in a separate pull request that only addresses #748. I don't think we should remove the scripts as part of this pull request.

Could you please remove your last two commits? You can do that by undoing your last two commits in your local branch, and then force pushing the branch will update this pull request correctly.

Then I'll see about removing and replacing the scripts (I will not ask you for help with that part).

We might eventually want to add commit 3baa0839c4622a9a6b2495348a07319910876b70 again, so I am coping the diff here for posterity:

```diff
 .github/workflows/lint_python.yml
@@ -23,4 +23,4 @@ jobs:
        with:
          python-version: ${{ matrix.python-version }}
      - run: pip install flake8
-      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics --exclude src/external
```

robgjansen:
I've made the requested change and open #750, which obsoletes this PR. Thanks for the help @florian-vuillemot !

PR #752, opened 2020-04-03
Fix pthread test bugs

florian-vuillemot:
Allow to run the pthread test with one or more then 6 threads

### DIFF
--- a/src/test/pthreads/shd-test-pthreads.c
+++ b/src/test/pthreads/shd-test-pthreads.c
@@ -71,7 +71,7 @@ static void* _test_thread_muxtrylock(void* mx) {
 
         muxes->numlocked++;
 
-        if(muxes->numnolocked == 0) {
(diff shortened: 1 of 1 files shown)


florian-vuillemot:
If I understand correctly this error doesn't come from my code but from the openssl dependancy that moved from https://www.openssl.org/source/openssl-1.0.1e.tar.gz to https://openssl.org/source/old/1.0.1/openssl-1.0.1.tar.gz in the shadow-plugin-tor .
Issue: https://github.com/shadow/shadow-plugin-tor/issues/92

sporksmith:
I have #754 out to fix the integration test.

florian-vuillemot:
I'm not quite sure about what I have just done ! I think I started by making a mistake right before doing what you suggested @robgjansen 
Is it correct now?

robgjansen:
@florian-vuillemot I think this looks OK. Thank you!

Issue #740, opened 2020-03-12
Help with log file interpretation.


jppr12:
I'm trying to have a full understand of the messages present in the log file generate by shadow, and i'm not 100% sure of what is the difference between the:
inbound-localhost-counters, outbound-localhost-counters, inbound-remote-counters and the outbound-remote-counters.

I would be very grateful if you could help me with this issue.
By the way, and a small aside, there are any new examples of simple simulations compatible with the updates made in tgen?

robgjansen:
Inbound/outbound tracks packets, bytes, etc. received/sent on a particular network interface. Localhost are for connections over localhost, whereas remote are for connections to a host on a different machine.

The inbound/outbound counters that are printed for localhost are also printed for remote connections. So there is a lot of data present in the log file. In case you want to read the code that formats the string that is printed to the log file, that is here:

https://github.com/shadow/shadow/blob/d54fa6321b293ac03856777cb805990691d9149e/src/main/host/shd-tracker.c#L385-L403

The new TGen that we just released has examples of running in Shadow here:
https://github.com/shadow/tgen/tree/master/resource

jppr12:
Hi, thanks for the help. I'm not yet understanding 100% how this works but my problem now is with the relation between what is configured in the shadow.config.xml and the results of the simulation.

So, in the example resource/examples/shadow.config.xml, for what i understand, we have 2 servers and 1 client, so i can't understand how that hosts are related with what is configured in the tophology element. Who is id="poi-1" and why are both source and target on the edge part "poi-1".

The other question, executing the example on that folder, is why the log file only have representative values on the remote side of the counter? i suppose the connection is directed from client to servers, why in any point the are no values on the localhost side? When a client recieve packets destined for him the values shoulden't be on that side?

Thanks for the attention.

robgjansen:
> So, in the example resource/examples/shadow.config.xml, for what i understand, we have 2 servers and 1 client, so i can't understand how that hosts are related with what is configured in the tophology element. Who is id="poi-1" and why are both source and target on the edge part "poi-1".

The topology section represents the Internet map. Think of "poi-1" as the Internet service provider, or the router that hosts can get attached to. The edge between "poi-1" and "poi-1" shows the properties that will be enforced when any two hosts (client and servers) that are attached to "poi-1" attempt to communicate with each other. Since "poi-1" is the only "router", all hosts will get attached to it and have e.g. 50 milliseconds of latency between them. (Probably a better name for "poi-1" is "isp".)

[...]

PR #753, opened 2020-04-06
Remove deprecated element names in simulation example

robgjansen:
Found from question in GH-740


robgjansen:
@sporksmith there are some issues in the CI with failing to install/update packages in ubuntu. I think those should probably be fixed in a separate PR before merging this one.

sporksmith:
> there are some issues in the CI with failing to install/update packages in ubuntu.

Those looked like some transient error with the package server; I retried and those ones passed.

The CI test is failing though because the openssl we're trying to download isn't there anymore. I think I saw a fix get proposed or committed for that; I'll look into it.

https://github.com/shadow/shadow/pull/753/checks?check_run_id=564638090#step:6:659

sporksmith:
Ah right, [here](https://github.com/shadow/shadow-plugin-tor/commit/391f85f26bcaf09dd6b9f06f7331b6b73b47db17)'s the fix. I'll send a PR to bump the version of shadow-plugin-tor to include the fix.

Issue #742, opened 2020-03-21
Error on running Getting Started Example


florian-vuillemot:
**Describe the issue**
I'm running the Getting Started Example and obtaining the following error:
> ** Stopping Shadow, returning code -1 (error)

If I look in the client logs (shadow.data/hosts/client/stdout-client.tgen.1000.log) I can see the following line
> 2000-01-01 00:00:02 946684802.000000 [critical] [tgen-graph.c:1299] [tgengraph_new] error (4) while loading graph: Vertex attribute 'type' is unknown, please check your config and try again.

**To Reproduce**
Clone and install the latest Shadow and Tgen version then go in resource/examples and run `shadow shadow.config.xml > shadow.log`.

**Operating System (please complete the following information):**
 - OS and version: post the output of `lsb_release -d`  
Linux Mint 19.1 Tessa
 - Kernel version: post the output of `uname -a`  
Linux user 4.15.0-91-generic #92-Ubuntu SMP Fri Feb 28 11:09:48 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

**Shadow (please complete the following information):**
 - Version: post the output of `shadow --version`  
Shadow v1.13.2-510-gd54fa632 2020-03-13 (built 2020-03-21) running GLib v2.56.4 and IGraph v0.7.1

 - Which plug-ins you are using:  
No plug-in used

**Additional context**
I installed Shadow and Tgen from by clone the repository. I installed dependencies by following the 1.1-Shadow.md documentation.
No files were changed, I checked with a `git status`.
Files `shadow.data/hosts/server1` and `shadow.data/hosts/server2` contain only logs messages with the following pattern:
> 2000-01-01 00:59:59 946688399.000000 [message] [tgen-driver.c:254] [_tgendriver_onHeartbeat] [driver-heartbeat] [bytes-read=0,bytes-written=0]
Moreover, I can see in the beginning of the logs the files
> successfully loaded graphml file 'tgen.server.graphml.xml' and validated acti
ons: graph is weakly connected with 1 cluster, 1 vertex, and 0 edges

My shadow.data/hosts/client/stdout-client.tgen.1000.log:
> 2000-01-01 00:00:02 946684802.000000 [message] [tgen-main.c:45] [_tgenmain_run] Initializing TGen v1.0.0 running GLib v2.56.4 and IGraph v0.7.1 on host client with process id 1000
[...]

Issue #703, opened 2020-01-27
Failed tgen streams (STALLOUT) running resource/examples/shadow.config.xml


sporksmith:
**Describe the issue**

Running the example configuration as described in the [tutorial](https://github.com/shadow/shadow/wiki/2-Getting-Started-Tutorial#basic-functional-tests), I only get 99 completed streams. The tutorial says to expect 1000.

**To Reproduce**

    $ cd resource/examples
    $ shadow shadow.config.xml > shadow2.log
    $ for d in shadow.data/hosts/*client*; do grep "transfer-complete" ${d}/* ; done | tee clients.log | wc -l
    99

    $ grep transfer-error shadow.data/hosts/client/stdout-client.tgen.1000.log
    2000-01-01 00:05:45 946685145.000000 [message] [tgen-transfer.c:1508] [_tgentransfer_log] [transfer-error] transport TCP,4925,NULL:11.0.0.3:37500,NULL:0.0.0.0:0,server1:11.0.0.1:8888,state=SUCCESS,error=NONE transfer transfer,15,client,GET,1048576,server1,10,state=ERROR,error=STALLOUT total-bytes-read=15806 total-bytes-write=52 payload-bytes-read=15774/1048576 (1.50%) usecs-to-socket-create=0 usecs-to-socket-connect=100000 usecs-to-proxy-init=-1 usecs-to-proxy-choice=-1 usecs-to-proxy-request=-1 usecs-to-proxy-response=-1 usecs-to-command=100000 usecs-to-response=200000 usecs-to-first-byte=301000 usecs-to-last-byte=-1 usecs-to-checksum=-1
    ...

Full logs:

 *  [stdout-client.tgen.1000.log](https://github.com/shadow/shadow/files/4119383/stdout-client.tgen.1000.log)
 * [stdout-server1.tgen.1000.log](https://github.com/shadow/shadow/files/4119385/stdout-server1.tgen.1000.log)
 * [stdout-server2.tgen.1000.log](https://github.com/shadow/shadow/files/4119386/stdout-server2.tgen.1000.log)

**Operating System (please complete the following information):**

Description:	Ubuntu 18.04.3 LTS

 - Kernel version: post the output of `uname -a`  

Linux vusiri 5.3.0-26-generic #28~18.04.1-Ubuntu SMP Wed Dec 18 16:40:14 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

**Shadow (please complete the following information):**

[...]

PR #757, opened 2020-04-10
Updated example to follow the TGEN config format

florian-vuillemot:
Issue https://github.com/shadow/shadow/issues/742


florian-vuillemot:
@robgjansen should we also update the Getting Started Tutorial to inform users that the result is 99 due to an error in TGEN (https://github.com/shadow/shadow/issues/703) ?

robgjansen:
@florian-vuillemot I've incorporated your changes into a new PR at GH-759. We'll deal with the problem that only 99/100 streams complete successfully in GH-703. Thank you for contributing!

Issue #733, opened 2020-03-05
Create a CONTRIBUTING file documenting header-include strategy and other policies


sporksmith:
We should document the header-include strategy (see #732) and other relevant development practices that contributors need to know. Github's best practices for such guidelines: https://help.github.com/en/github/building-a-strong-community/setting-guidelines-for-repository-contributors

sporksmith:
Also cover version bump procedure and bumpversion tool. https://github.com/shadow/shadow/pull/734#discussion_r389771111

Issue #732, opened 2020-03-05
Revise header-include strategy for more modular builds


sporksmith:
Today the strategy is:

> [We] have a single shadow.h file that includes all of the other shadow header files. Then, in all of the individual header files, I only include system headers that are necessary for that file. Then, in all c files, I include shadow.h and only include other headers as necessary (that were not already included).

An alternative strategy under discussion is to have very header file be self-contained, such that they can be included individually rather than including the monolithic `shadow.h`. Modules inside Shadow could/would then include only the specific headers they need; we could still have a public shadow.h for external code (e.g. plugins) to use. e.g. this is roughly the current [for glib](https://wiki.gnome.org/Projects/GTK/BestPractices/GlibIncludes), and supported by Google's [Include What You Use](https://github.com/include-what-you-use/include-what-you-use/blob/master/README.md)

Admittedly a large part of my motivation here is because that's what I'm used to. I don't plan to move forward with this migration before working within the current strategy a bit more to understand whether it's worth migrating. Generally it'd make the code more modular, and make it easier to understand the dependencies between code modules inside Shadow. The only short term practical benefit I see though is faster build times, which doesn't seem like sufficient motivation on its own.

PR #758, opened 2020-04-10
Jnewsome/explicit deps

sporksmith:
This moves to a model of granular, self-contained, header files as proposed in #732. I started detangling some of this while trying to extract the logger as a separate library, and decided to try ripping the whole bandaid off.

Besides making the code more amenable modularization and refactoring, incremental build times are substantially improved since previously every C file effectively had a build dependency on every header file via shadow.h.

On my machine this has negligible effect on a clean build: From 21.5s to
19.9s using -j8. It significantly improves incremental builds:
From 5m to 2m40s to run
`find src/main -name '*.h' -exec touch \{\} \; -exec ./setup build -j8 \;`

I also added the beginning of a CONTRIBUTING file covering the header strategy and use of clang-format, per #733.


sporksmith:
> Since this is a large change, are we going to want to rebase the phantom branch with these changes?

Yes, I think so. I'll create a rebased branch in the phantom repo (sorting through the merge conflicts I'm sure will be there) and ask you to force push :).

PR #761, opened 2020-04-14
File glibtest

sporksmith:
Enables individual file tests to be turned on and off independently, so that we can enable them incrementally in the phantom branch.

Migrates to glib's test framework so that we can filter which tests run from the command-line. This also makes the test code more succinct and give better failure reporting.

Modifies subtests to be independent of each-other; i.e. not depend on earlier tests to have created or modified a test file.

### DIFF
--- a/src/test/file/CMakeLists.txt
+++ b/src/test/file/CMakeLists.txt
@@ -1,3 +1,7 @@
+find_package(GLIB REQUIRED)
+include_directories(${GLIB_INCLUDES})
+link_libraries(${GLIB_LIBRARIES})
+
 ## build the test as a dynamic executable that plugs into shadow
 add_shadow_plugin(shadow-plugin-test-file shd-test-file.c)
 
--- a/src/test/file/shd-test-file.c
+++ b/src/test/file/shd-test-file.c
@@ -5,6 +5,8 @@
 
 //TODO: Implement fwrite
 
+#include <errno.h>
+#include <glib.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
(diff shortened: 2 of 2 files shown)


src/test/file/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+find_package(GLIB REQUIRED)
```
robgjansen, line 1: I'm pretty sure we already search for the GLIB package in `src/main/CmakeLists.txt`, so adding this here will cause us to search twice (and the GLib paths will be printed twice) when we build the tests. I think this is still better than moving this search to the toplevel `CMakeLists.txt` (so that the result can be inherited everywhere) and then removing the `find_package()` here and in `src/main/CmakeLists.txt`. Do you agree?

src/test/file/shd-test-file.c
```diff
@@ -15,188 +17,137 @@
...
+#define assert_errno_is(e)                                                     \
+    {                                                                          \
+        int _errno = errno;                                                    \
+        int _e = e;                                                            \
+        if (_e != _errno) {                                                    \
+            g_error("Got errno %d (%s) instead of %d (%s)", _errno,            \
+                    strerror(_errno), _e, strerror(_e));                       \
+            g_test_fail();                                                     \
+        }                                                                      \
+    }
```
robgjansen, line 40: I'm guessing some of this would be useful if we continue converting other tests to the glibtest framework, but I suppose we can refactor it to a common file then.

src/test/file/CMakeLists.txt
sporksmith: Yup, good idea

src/test/file/shd-test-file.c
sporksmith: Yeah, I was thinking of waiting until they're needed again, but I agree it seems likely if we continue using glib testing.

sporksmith:
> My main concern is that we'll be rewriting some tests twice, once for glibtest and again for rust. But I suppose that's OK if we don't want to pull in Rust yet at this point.

Yeah, tweaking the tests to use the glib framework is probably faster than rewriting them in Rust; at least faster than adding the first Rust code/test, but I agree overall it's adding an extra step if we eventually want to migrate to Rust.

Maybe with this test merged so that you have something to work with, I'll take a look at using Rust for the udp test.

src/test/file/CMakeLists.txt
sporksmith: Heh, I went to move the search to the top-level, had second thoughts, and realized I'd misread your comment.

Unless we have reason to think the package search has some significant cost, I'm leaning slightly towards keeping it 'locally' in the places where we're actually using GLib. Since it sounds like you're also leaning that way, I'll leave it.

robgjansen: 👍

robgjansen:
> Maybe with this test merged so that you have something to work with, I'll take a look at using Rust for the udp test.

I guess this introduces a larger question about plans for our next release. Do we intend to release phantom changes before we start migrating the wider Shadow code-case to Rust? Or should we keep the phantom changes in a [public] dev branch as we continue improving Shadow in other ways, but without making a formal release? If we do the former, we may not want to introduce Rust as yet another dependency yet. But if we do the latter, which I'm in favor of, I would say we could go ahead with Rust whenever we feel ready.

sporksmith:
Suppose we create a `shadow/shadow:v1` branch from the current `shadow/shadow:master`, and then go ahead with adding Rust to `master`? That lets us keep non-Phantom-specific development public and leaves us a stable branch where we can backport fixes/improvements if needed. It also leaves our options open as to whether to backport the Rust migration into the v1 series or not. (I'm assuming an eventual Phantom release would be v2.0.0)

robgjansen:
Seems reasonable, but let's discuss at our meeting next week before taking action. That way we can include Ryan in the discussion.

Issue #402, opened 2018-02-22
segfault in pth_mutex_release in shadow thread


dhuseby:
```
Thread 1 "shadow" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7876b92f3ea0 (LWP 27191)]
0x00007876b172adcf in ?? ()
(gdb) bt
#0  0x00007876b172adcf in ?? ()
#1  0x000061e340240e00 in ?? ()
#2  0x000000000000000f in ?? ()
#3  0x00007876b0acac48 in ?? ()
#4  0x00007876b9133c04 in get_current_dtv (tp=<optimized out>) at /home/user/Projects/shadow/src/external/elf-loader/vdl-tls.c:306
#5  vdl_tls_get_addr_fast (module=132451165579872, offset=132451307981912) at /home/user/Projects/shadow/src/external/elf-loader/vdl-tls.c:615
#6  0x00007876b0acab98 in ?? ()
#7  0x00007876b0acab40 in ?? ()
#8  0x00007876b17376e2 in ?? ()
#9  0x000061e34023bb30 in ?? ()
#10 0x000061e33ece2e7b in pth_mutex_release (mutex=0x61e3402420d0) at /home/user/Projects/shadow/src/external/rpth/pth_sync.c:122
#11 0x0000000000000019 in ?? ()
#12 0x0000000000000000 in ?? ()
(gdb) info thread
  Id   Target Id         Frame 
* 1    Thread 0x7876b92f3ea0 (LWP 27191) "shadow" 0x00007876b172adcf in ?? ()
  2    Thread 0x7876b23ce700 (LWP 27193) "logger-helper" syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
(gdb) info registers
rax            0x7876b9298c58   132451307981912
rbx            0x7876b9298c58   132451307981912
rcx            0x3      3
rdx            0x7876b933c948   132451308652872
rsi            0x0      0
rdi            0x0      0
rbp            0x7876b0acaa60   0x7876b0acaa60
rsp            0x7876b0acaa10   0x7876b0acaa10
r8             0x0      0
r9             0x30     48
r10            0x30     48
r11            0x202    514
r12            0x7876b9298c58   132451307981912
r13            0x19     25
r14            0x7876b0acab98   132451165580184
r15            0xd      13
rip            0x7876b172adcf   0x7876b172adcf
eflags         0x10246  [ PF ZF IF RF ]
cs             0xe033   57395
[...]

PR #762, opened 2020-04-15
Rust tests

sporksmith:
Not intended to be merged in its current state; just for discussion. Follow-up to discussion in https://github.com/shadow/shadow/pull/761

Pulls in CMakeRust as a submodule, and creates a hello-world "test" using Rust. This version of it (directly calling libc's `write` from Rust) works, but aborts or segfaults if we call Rust's `println!` instead.

Applied these to the Phantom branch, and it works fine there (including using `println!`)

My inclination is to keep the integration tests (code meant to be run under Shadow) in C for now so that we don't end up writing phantom-only integration tests (nor have to spend time figuring out how to get Rust code running reliably under the elf loader).

The CMakeRust tool used here could be useful though as a way to start to writing Shadow code itself in Rust; especially code that we intend to extract into its own library anyway.


sporksmith:
@robgjansen @RWails for discussion ^

jtracey:
Back when we were working on elf-loader, there were some attempts to get a plugin written in Rust working:
https://github.com/dhuseby/shadow-rust-test/
(The last bug that I never got around to addressing is #402, almost certainly an elf-loader issue.)

sporksmith:
Interesting, thanks! #402 indeed might explain the segfault under gdb; maybe we're "luckily" getting an aligned stack in this case when not running under gdb.

PR #763, opened 2020-04-16
Test udp

sporksmith:
This is mostly getting in place a framework for UDP tests; particularly
for using dynamically assigned ports when not running under Shadow and
communicating them via a fifo(7). I'll flesh out some more tests in a
follow-up.

Initially I tried doing the same under Shadow as well, but it doesn't
deal well with blocking on an `open` call. I got it to sort of work if I
opened the fifo in nonblocking mode and looped, but decided it'd be
cleaner to just use hard-coded ports under Shadow for now. I expect that
the blocking `open` call won't be an issue under Phantom, so we can
perhaps switch it over there.

### DIFF
--- a/src/test/udp/CMakeLists.txt
+++ b/src/test/udp/CMakeLists.txt
@@ -1,0 +1,10 @@
+find_package(GLIB REQUIRED)
+include_directories(${GLIB_INCLUDES})
+link_libraries(${GLIB_LIBRARIES})
+
+## build the test as a dynamic executable that plugs into shadow
+add_shadow_exe(test-udp shd-test-udp.c)
+
+## register the tests
+add_test(NAME udp COMMAND /bin/bash -c "rm -f udp-fifo && mkfifo udp-fifo && ../shadow-test-launcher test-udp client 0 udp-fifo : test-udp server 0 udp-fifo")
+add_test(NAME udp-shadow COMMAND /bin/bash -c "rm -f udp-shadow-fifo && mkfifo udp-shadow-fifo && ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d udp.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/udp.test.shadow.config.xml")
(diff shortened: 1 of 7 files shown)


src/test/udp/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+find_package(GLIB REQUIRED)
+include_directories(${GLIB_INCLUDES})
+link_libraries(${GLIB_LIBRARIES})
+
+## build the test as a dynamic executable that plugs into shadow
+add_shadow_plugin(shadow-plugin-test-udp shd-test-udp.c)
+
+## create and install an executable that can run outside of shadow
+add_executable(test-udp shd-test-udp.c)
```
robgjansen, line 9: Can/should we just run the `test-udp` binary in Shadow too and that way we don't have to build `libshadow-plugin-test-udp.so` too?

I think we can do that like:

```diff
- ## build the test as a dynamic executable that plugs into shadow
- add_shadow_plugin(shadow-plugin-test-udp shd-test-udp.c)
-
- ## create and install an executable that can run outside of shadow
- add_executable(test-udp shd-test-udp.c)
+ ## build a position-independent executable that can run inside or outside of shadow
+ add_shadow_exe(test-udp shd-test-udp.c)
```

src/test/udp/udp.test.shadow.config.xml
```diff
@@ -0,0 +1,30 @@
...
+    </node>
+    <edge source="poi-1" target="poi-1">
+      <data key="d3">50.0</data>
+      <data key="d4">0.0</data>
+    </edge>
+  </graph>
+</graphml>
+]]></topology>
+  <kill time="5"/>
+  <plugin id="testudp" path="libshadow-plugin-test-udp.so"/>
```
robgjansen, line 22: If we go with the position-independent executable, we need to update this path.

src/test/udp/CMakeLists.txt
sporksmith: Done

src/test/udp/udp.test.shadow.config.xml
sporksmith: Done

PR #770, opened 2020-04-21
Rename files and include relative to src

sporksmith:
As discussed, this:
* Drops shd- prefix
* Replaces - with _ in names (outside of `external`, where I thought it best to minimize churn)
* Makes all local includes relative to `src`


sporksmith:
> It would be good to rebase phantom soon, as I'm guessing this may cause some conflicts...

Definitely; I'll do that next.

Huh, it wasn't registering the approval at first, but after re-requesting review it did. Merging now.

PR #780, opened 2020-04-25
Made Elf Loader executable by python2 and python3

florian-vuillemot:

PR #782, opened 2020-04-27
Allow compiler to validate  format strings, and fix errors

sporksmith:


src/support/logger/logger.h
```diff
@@ -51,6 +51,7 @@ Logger* logger_getDefault();
 
 // Thread safe. `logger` may be NULL, in which case glib's logging
 // functionality will be used.
+__attribute__((__format__ (__printf__, 6, 7)))
```
robgjansen, line 54: A comment about what this does would be helpful.

sporksmith: Done

sporksmith:
> Do we want to add a compiler flag to help us identify issues like this in the future?

Adding the `format` attributes causes the compiler to generate warnings for such bugs, which are turned into errors with -Werror (as is done in our CI).

> I'd like to give Florian a chance to rebase and then merge #780 before this one. Otherwise we'll have to ask to rebase again. So please hold off until #780 is in (asuming that happens soon), thanks!

Ok, will do!

PR #786, opened 2020-05-01
Migrates bind tests to glib

sporksmith:


src/test/bind/test_bind.c
```diff
@@ -291,167 +172,83 @@ static int _check_matching_addresses(int fd_server_listen, int fd_server_accept,
      *   + client socket pot != accepted peer ports
```
robgjansen, line 172: I know you didn't introduce it, but there is a spelling error here: pot -> port.

src/test/bind/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+include_directories(${GLIB_INCLUDES})
+link_libraries(${GLIB_LIBRARIES})
+
 ## build the test as a dynamic executable that plugs into shadow
 add_library(shadow-plugin-test-bind SHARED test_bind.c)
```
robgjansen, line 5: FYI: I think we don't want to change this to `add_shadow_exe` because the library version works well enough and we want to minimize conflicts with the phantom branch.

src/test/bind/test_bind.c
```diff
@@ -58,175 +26,94 @@ static int _do_bind(int fd, in_addr_t address, in_port_t port) {
...
+        usleep((useconds_t)1000);
     }
-
-    return EXIT_SUCCESS;
+    errno = saved_errno;
+    return result;
 }
 
-static int _do_accept(int fd, int* outfd) {
+static int _do_accept(int fd) {
```
robgjansen, line 55: I wonder if any of the bind/connect/accept helpers can or should be extracted to the `src/test/test_common.h` file?

src/test/bind/test_bind.c
sporksmith: Done

src/test/bind/CMakeLists.txt
sporksmith: Agreed. I didn't touch the targets here; just added the GLIB includes and libraries.

src/test/bind/test_bind.c
sporksmith: I looked a bit at doing this, but I'm not sure they make sense in a wider context. Leaving it for now.

PR #787, opened 2020-05-02
Remove the hard-coded tcp port and let the os automatically select it

florian-vuillemot:
Remove the hard-coded tcp port and let the os select it automatically. Once the port is selected we send it by a mqueue file. The client can read it and open a connection on the server.

### DIFF
--- a/src/test/tcp/CMakeLists.txt
+++ b/src/test/tcp/CMakeLists.txt
@@ -7,12 +7,6 @@ add_shadow_plugin(shadow-plugin-test-tcp test_tcp.c)
 ## create and install an executable that can run outside of shadow
 add_executable(test-tcp test_tcp.c)
 
-## add mqueue library
-find_library(LIBRT rt)
-if(LIBRT)
-    target_link_libraries(test-tcp ${LIBRT})
-endif()
-
 ## register the tests
 
 ## tcp blocking - loopback, lossless and lossy
--- a/src/test/tcp/CMakeLists.txt
+++ b/src/test/tcp/CMakeLists.txt
@@ -1,8 +1,18 @@
+find_package(GLIB REQUIRED)
+include_directories(${GLIB_INCLUDES})
+link_libraries(${GLIB_LIBRARIES})
+
 ## build the test as a dynamic executable that plugs into shadow
 add_shadow_plugin(shadow-plugin-test-tcp test_tcp.c)
 
 ## create and install an executable that can run outside of shadow
 add_executable(test-tcp test_tcp.c)
+
+## add mqueue library
+find_library(LIBRT rt)
+if(LIBRT)
+    target_link_libraries(test-tcp ${LIBRT})
+endif()
 
 ## register the tests
 
--- a/src/test/tcp/CMakeLists.txt
+++ b/src/test/tcp/CMakeLists.txt
@@ -10,85 +10,85 @@ add_executable(test-tcp test_tcp.c)
 ## register the tests
 
 ## tcp blocking - loopback, lossless and lossy
-## these also test "localhost" instead of "127.0.0.1" in the loopback tests
+## these also test localhost instead of 127.0.0.1 in the loopback tests
 add_test(
     NAME tcp-blocking-loopback
-    COMMAND /bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ../shadow-test-launcher test-tcp blocking $QUEUE server : test-tcp blocking $QUEUE client localhost; ipcrm -q $QUEUE"
+    COMMAND ${CMAKE_SOURCE_DIR}/src/test/tcp/with_q.sh ../shadow-test-launcher test-tcp blocking $QUEUE server : test-tcp blocking $QUEUE client localhost
 )
 add_test(
     NAME tcp-blocking-loopback-shadow
-    COMMAND /bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d blocking-loopback.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-blocking-loopback.test.shadow.config.xml; ipcrm -q $QUEUE"
+    COMMAND ${CMAKE_SOURCE_DIR}/src/test/tcp/with_q.sh ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d blocking-loopback.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-blocking-loopback.test.shadow.config.xml
 )
 add_test(
     NAME tcp-blocking-lossless-shadow
-    COMMAND /bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d blocking-lossless.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-blocking-lossless.test.shadow.config.xml; ipcrm -q $QUEUE"
+    COMMAND ${CMAKE_SOURCE_DIR}/src/test/tcp/with_q.sh ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d blocking-lossless.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-blocking-lossless.test.shadow.config.xml
 )
 add_test(
     NAME tcp-blocking-lossy-shadow
-    COMMAND /bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d blocking-lossy.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-blocking-lossy.test.shadow.config.xml; ipcrm -q $QUEUE"
+    COMMAND ${CMAKE_SOURCE_DIR}/src/test/tcp/with_q.sh ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d blocking-lossy.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-blocking-lossy.test.shadow.config.xml
 )
 
 ## tcp nonblocking poll - loopback, lossless and lossy
 add_test(
     NAME tcp-nonblocking-poll-loopback
-    COMMAND /bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ../shadow-test-launcher test-tcp nonblocking-poll $QUEUE server : test-tcp nonblocking-poll $QUEUE client 127.0.0.1; ipcrm -q $QUEUE"
+    COMMAND ${CMAKE_SOURCE_DIR}/src/test/tcp/with_q.sh ../shadow-test-launcher test-tcp nonblocking-poll $QUEUE server : test-tcp nonblocking-poll $QUEUE client 127.0.0.1
 )
 add_test(
     NAME tcp-nonblocking-poll-loopback-shadow
-    COMMAND /bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-poll-loopback.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-poll-loopback.test.shadow.config.xml; ipcrm -q $QUEUE"
+    COMMAND ${CMAKE_SOURCE_DIR}/src/test/tcp/with_q.sh ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-poll-loopback.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-poll-loopback.test.shadow.config.xml
 )
 add_test(
     NAME tcp-nonblocking-poll-lossless-shadow
-    COMMAND /bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-poll-lossless.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-poll-lossless.test.shadow.config.xml; ipcrm -q $QUEUE"
+    COMMAND ${CMAKE_SOURCE_DIR}/src/test/tcp/with_q.sh ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-poll-lossless.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-poll-lossless.test.shadow.config.xml
 )
 add_test(
     NAME tcp-nonblocking-poll-lossy-shadow
-    COMMAND /bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-poll-lossy.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-poll-lossy.test.shadow.config.xml; ipcrm -q $QUEUE"
+    COMMAND ${CMAKE_SOURCE_DIR}/src/test/tcp/with_q.sh ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-poll-lossy.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-poll-lossy.test.shadow.config.xml
 )
 
 ## tcp nonblocking epoll - loopback, lossless and lossy
 add_test(
     NAME tcp-nonblocking-epoll-loopback
-    COMMAND /bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ../shadow-test-launcher test-tcp nonblocking-epoll $QUEUE server : test-tcp nonblocking-epoll $QUEUE client 127.0.0.1; ipcrm -q $QUEUE"
+    COMMAND ${CMAKE_SOURCE_DIR}/src/test/tcp/with_q.sh ../shadow-test-launcher test-tcp nonblocking-epoll $QUEUE server : test-tcp nonblocking-epoll $QUEUE client 127.0.0.1
 )
 add_test(
     NAME tcp-nonblocking-epoll-loopback-shadow
-    COMMAND /bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-epoll-loopback.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-epoll-loopback.test.shadow.config.xml; ipcrm -q $QUEUE"
+    COMMAND ${CMAKE_SOURCE_DIR}/src/test/tcp/with_q.sh ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-epoll-loopback.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-epoll-loopback.test.shadow.config.xml
 )
 add_test(
     NAME tcp-nonblocking-epoll-lossless-shadow
-    COMMAND /bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-epoll-lossless.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-epoll-lossless.test.shadow.config.xml; ipcrm -q $QUEUE"
+    COMMAND ${CMAKE_SOURCE_DIR}/src/test/tcp/with_q.sh ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-epoll-lossless.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-epoll-lossless.test.shadow.config.xml
 )
 add_test(
     NAME tcp-nonblocking-epoll-lossy-shadow
-    COMMAND /bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-epoll-lossy.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-epoll-lossy.test.shadow.config.xml; ipcrm -q $QUEUE"
+    COMMAND ${CMAKE_SOURCE_DIR}/src/test/tcp/with_q.sh ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-epoll-lossy.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-epoll-lossy.test.shadow.config.xml
 )
 
 ## tcp nonblocking select - loopback, lossless and lossy
 add_test(
     NAME tcp-nonblocking-select-loopback
-    COMMAND /bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ../shadow-test-launcher test-tcp nonblocking-select $QUEUE server : test-tcp nonblocking-select $QUEUE client 127.0.0.1; ipcrm -q $QUEUE"
+    COMMAND ${CMAKE_SOURCE_DIR}/src/test/tcp/with_q.sh ../shadow-test-launcher test-tcp nonblocking-select $QUEUE server : test-tcp nonblocking-select $QUEUE client 127.0.0.1
 )
 add_test(
     NAME tcp-nonblocking-select-loopback-shadow
-    COMMAND /bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-select-loopback.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-select-loopback.test.shadow.config.xml; ipcrm -q $QUEUE"
+    COMMAND ${CMAKE_SOURCE_DIR}/src/test/tcp/with_q.sh ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-select-loopback.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-select-loopback.test.shadow.config.xml
 )
 add_test(
     NAME tcp-nonblocking-select-lossless-shadow
-    COMMAND /bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-select-lossless.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-select-lossless.test.shadow.config.xml; ipcrm -q $QUEUE"
+    COMMAND ${CMAKE_SOURCE_DIR}/src/test/tcp/with_q.sh ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-select-lossless.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-select-lossless.test.shadow.config.xml
 )
 add_test(
     NAME tcp-nonblocking-select-lossy-shadow
-    COMMAND /bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-select-lossy.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-select-lossy.test.shadow.config.xml; ipcrm -q $QUEUE"
+    COMMAND ${CMAKE_SOURCE_DIR}/src/test/tcp/with_q.sh ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d nonblocking-select-lossy.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-nonblocking-select-lossy.test.shadow.config.xml
 )
 
 add_test(
     NAME tcp-iov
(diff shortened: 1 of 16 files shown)


src/test/tcp/test_tcp.c
```diff
@@ -49,6 +52,45 @@ static void _mylog(const char* fileName, const int lineNum, const char* funcName
...
+static short queue_recv_u16(const char* queuename) {
+    mqd_t mq;
+    guint64 val;
+    GError *error = NULL;
+    struct mq_attr attr;
+    unsigned int priorite;
+    char buf[10] = {0};
+
+    // let the server write in the queue
+    usleep(1000);
```
sporksmith, line 76: Sleeping in tests is almost never the right solution for resolving race conditions. It makes the test slower, and still leaves us with a race condition.

We should find a way to properly synchronize (such as by creating the ipc mechanism before we kick off the tests, as mentioned above)

src/test/tcp/test_tcp.c
```diff
@@ -49,6 +52,45 @@ static void _mylog(const char* fileName, const int lineNum, const char* funcName
...
+
+    mq = mq_open(queuename, O_RDONLY);
+    g_assert(mq != (mqd_t)-1);
```
sporksmith, line 79: assert (and g_assert) are generally for catching programming errors (such as a function receiving NULL that is documented as not accepting NULL). It's generally not safe to use them for catching runtime errors, since they can be disabled in some builds.

It'd also be good to output the more specific error from errno, e.g. using `perror` or `strerror`. Alternatively you can use `assert_true_errno` from https://github.com/shadow/shadow/blob/master/src/test/test_glib_helpers.h

src/test/tcp/test_tcp.c
```diff
@@ -49,6 +52,45 @@ static void _mylog(const char* fileName, const int lineNum, const char* funcName
...
+
+    // get the queue attributes to obtain the message length
+    g_assert(mq_getattr(mq, &attr) == 0);
```
sporksmith, line 82: Use assert_true_errno or equivalent

src/test/tcp/test_tcp.c
```diff
@@ -49,6 +52,45 @@ static void _mylog(const char* fileName, const int lineNum, const char* funcName
...
+
+    mq_receive(mq, buf, attr.mq_msgsize, &priorite);
```
sporksmith, line 84: Check return value, and use assert_true_errno or equivalent

src/test/tcp/test_tcp.c
```diff
@@ -212,9 +257,18 @@ static int _do_connect(int fd, struct sockaddr_in* serveraddr, iowait_func iowai
     return 0;
 }
 
-static int _do_serve(int fd, struct sockaddr_in* bindaddr, iowait_func iowait, int* clientsdout) {
+static void send_port_in_queue(int fd, struct sockaddr_in* bindaddr, const char *queuename){
+    socklen_t len = sizeof(bindaddr);
+
+    g_assert(getsockname(fd, (struct sockaddr *)bindaddr, &len) != -1);
```
sporksmith, line 263: Use assert_true_errno or equivalent

src/test/tcp/test_tcp.c
```diff
@@ -212,9 +257,18 @@ static int _do_connect(int fd, struct sockaddr_in* serveraddr, iowait_func iowai
...
+    queue_send_u16(queuename, ntohs(bindaddr->sin_port));
```
sporksmith, line 264: Use assert_true_errno or equivalent

src/test/tcp/test_tcp.c
```diff
@@ -49,6 +52,45 @@ static void _mylog(const char* fileName, const int lineNum, const char* funcName
     fflush(stdout);
 }
 
+static void queue_send_u16(const char* queuename, uint16_t i) {
+    mqd_t mq = mq_open(queuename, O_WRONLY|O_CREAT, 0644, NULL);
```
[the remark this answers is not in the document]
sporksmith: I dug into this a bit more. It looks like the posix mq namespace is conventionally mounted on /dev/mqueue, and you can create queues with `touch` and remove them with `rm`. That removes the biggest blocker for using them IMO; it means we should be able to ensure the queue is freshly created when starting the test.

src/test/tcp/CMakeLists.txt
```diff
@@ -1,9 +1,19 @@
...
+include_directories(${GLIB_INCLUDES})
+link_libraries(${GLIB_LIBRARIES})
+
 ## build the test as a dynamic executable that plugs into shadow
 add_shadow_plugin(shadow-plugin-test-tcp test_tcp.c)
 
 ## create and install an executable that can run outside of shadow
 add_executable(test-tcp test_tcp.c)
 
+## add mqueue library
```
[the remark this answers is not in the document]
florian-vuillemot: I tried with a fifo like you did for UDP. And as you mentioned in your PR this didn't work ^^
To be honest, I chose this implementation because I had already used it, without thinking about Posix or System V. Your point on the command line it's really good I will check it :)

src/test/tcp/test_tcp.c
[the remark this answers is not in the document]
florian-vuillemot: Really interesting I didn't know it thank you !

src/test/tcp/test_tcp.c
florian-vuillemot: Yes, this was a first step. mqueue allows to do async and cleans this stuff. I don't know if I will be able to do the same thing with system v, what do you think ? But maybe the problem that I encountered came from the Posix implementation and I will not get it with system v ?

sporksmith: I'm fine with sticking with posix mqueues now that I know how to manipulate them from the command-line. It's also how we solve this problem: in the test command in the CMakeLists.txt file, we should delete the queue if it already exists, create a fresh queue, and *then* start the test.

That way when we get here the queue is guaranteed to already exist. The other side might not have written to it yet, but that's ok - the read will just block until there's a message there, which is what we want.

sporksmith:
Hi Florian - is this ready for another look or are you still working on it? If it is ready, please hit the "re-request review button". Thanks!

src/test/tcp/test_tcp.c
```diff
@@ -20,12 +20,19 @@
...
+#include "test/test_glib_helpers.h"
 
-#define USAGE "USAGE: 'shd-test-tcp iomode type'; iomode=('blocking'|'nonblocking-poll'|'nonblocking-epoll'|'nonblocking-select') type=('client' server_ip|'server')"
+#define USAGE "USAGE: 'shd-test-tcp iomode type [queuename]'; iomode=('blocking'|'nonblocking-poll'|'nonblocking-epoll'|'nonblocking-select') type=('client' server_ip|'server') [queuname=(default='/tcp_shadow_queue')]"
 #define MYLOG(...) _mylog(__FILE__, __LINE__, __FUNCTION__, __VA_ARGS__)
-#define SERVER_PORT 58333
 #define BUFFERSIZE 20000
 #define ARRAY_LENGTH(arr)  (sizeof (arr) / sizeof ((arr)[0]))
+// default queue filename for IPC communication
+#define QUEUENAME "./test-tcp"
```
sporksmith, line 33: Unused?

src/test/tcp/test_tcp.c
```diff
@@ -49,6 +61,37 @@ static void _mylog(const char* fileName, const int lineNum, const char* funcName
     fflush(stdout);
 }
 
+// create a System V queue with key based on a file queue name
+static int get_queue(const char* queuename) {
+    key_t key = ftok(queuename, 0);
```
sporksmith, line 66: Doing the assignment inside the check gives better error messages on failures. e.g.
```
key_t key;
assert_true_errno((key = ftok(queuename, 0) != -1);
```
On failure will output the expression `(key = ftok(queuename, 0) != -1` instead of just `key != -1`

src/test/tcp/test_tcp.c
```diff
@@ -49,6 +61,37 @@ static void _mylog(const char* fileName, const int lineNum, const char* funcName
...
+    assert_true_errno(key != -1);
+
+    int fd = msgget(key, IPC_CREAT | 0666);
```
sporksmith, line 69: (No action needed) - Interesting; I suppose we can let the reader and writer race to create the queue here rather than creating it from the CMakeLists.txt.

src/test/tcp/test_tcp.c
```diff
@@ -49,6 +61,37 @@ static void _mylog(const char* fileName, const int lineNum, const char* funcName
...
+    assert_true_errno(fd != -1);
```
sporksmith, line 70: Assignment inside the check, please

src/test/tcp/test_tcp.c
```diff
@@ -49,6 +61,37 @@ static void _mylog(const char* fileName, const int lineNum, const char* funcName
...
+    return fd;
+}
+
+static void queue_send_u16(const char* queuename, uint16_t i) {
+    int queue, r;
+    IntegerMessage msg = {QUEUE_MTYPE, i};
+
+    queue = get_queue(queuename);
+    r = msgsnd(queue, &msg, sizeof(msg.msg), IPC_NOWAIT);
+    assert_true_errno(r != -1);
```
sporksmith, line 81: Assignment inside the check, please

src/test/tcp/test_tcp.c
```diff
@@ -49,6 +61,37 @@ static void _mylog(const char* fileName, const int lineNum, const char* funcName
...
+
+    return fd;
+}
+
+static void queue_send_u16(const char* queuename, uint16_t i) {
+    int queue, r;
+    IntegerMessage msg = {QUEUE_MTYPE, i};
+
+    queue = get_queue(queuename);
+    r = msgsnd(queue, &msg, sizeof(msg.msg), IPC_NOWAIT);
```
sporksmith, line 80: We don't want `IPC_NOWAIT` here. Better to block. We have timeouts for the test as a whole that will prevent the test from blocking forever if something goes wrong.

src/test/tcp/test_tcp.c
```diff
@@ -49,6 +61,37 @@ static void _mylog(const char* fileName, const int lineNum, const char* funcName
...
+    assert_true_errno(r != -1);
+}
+
+static short queue_recv_u16(const char* queuename) {
+    int queue, r;
+    IntegerMessage msg;
+
+    queue = get_queue(queuename);
+    r = msgrcv(queue, &msg, sizeof(msg.msg), QUEUE_MTYPE, MSG_NOERROR | IPC_NOWAIT);
+    assert_true_errno(r != -1);
```
sporksmith, line 90: Assignment inside the check, please

src/test/tcp/test_tcp.c
```diff
@@ -49,6 +61,37 @@ static void _mylog(const char* fileName, const int lineNum, const char* funcName
...
+    r = msgsnd(queue, &msg, sizeof(msg.msg), IPC_NOWAIT);
+    assert_true_errno(r != -1);
+}
+
+static short queue_recv_u16(const char* queuename) {
+    int queue, r;
+    IntegerMessage msg;
+
+    queue = get_queue(queuename);
+    r = msgrcv(queue, &msg, sizeof(msg.msg), QUEUE_MTYPE, MSG_NOERROR | IPC_NOWAIT);
```
sporksmith, line 89: We don't want `MSG_NOERROR` - we'd rather have an error than silent runcation.

We don't want `IPC_NOWAIT` - we'd rather wait for the writer to send than fail if this executes first.

src/test/tcp/test_tcp.c
```diff
@@ -212,9 +258,18 @@ static int _do_connect(int fd, struct sockaddr_in* serveraddr, iowait_func iowai
     return 0;
 }
 
-static int _do_serve(int fd, struct sockaddr_in* bindaddr, iowait_func iowait, int* clientsdout) {
+static void send_port_in_queue(int fd, struct sockaddr_in* bindaddr, const char *queuename){
+    socklen_t len = sizeof(bindaddr);
```
sporksmith, line 262: You're getting the size of the pointer, not the size of the struct. This "luckily" happens to work since pointers are 8 bytes long here :).

Recommend changing to `sizeof(*bindaddr)`

src/test/tcp/test_tcp.c
```diff
@@ -759,6 +814,7 @@ int main(int argc, char *argv[]) {
 
     iowait_func wait = NULL;
     int use_iov = 0;
+    char *queuename = QUEUENAME;
```
sporksmith, line 817: We should try to remove the queue at the end, using `msgctl`. This should probably happen in an `atexit` handler so that it still happens on test failures.

Also, slight preference for getting the queue handle here and passing the handle around rather than passing the string around. That keeps the queue management at roughly a single (outer) scope, and lets us pass around a more strongly-typed handle rather than just a string.

src/test/tcp/CMakeLists.txt
```diff
@@ -1,9 +1,18 @@
+include_directories(${GLIB_INCLUDES})
+link_libraries(${GLIB_LIBRARIES})
```
sporksmith, line 2: It doesn't look like you're actually using any glib functionality. If that's the case better to leave off the dependency for now.

src/test/tcp/test_tcp.c
```diff
@@ -20,12 +20,19 @@
 #include <sys/uio.h>
 #include <time.h>
 #include <unistd.h>
+#include <glib.h>
```
sporksmith, line 23: It doesn't look like you're actually using any glib functionality. If that's the case better to leave off the dependency for now.

src/test/tcp/test_tcp.c
```diff
@@ -20,12 +20,19 @@
...
+#include <sys/msg.h>
+#include <sys/types.h>
+#include "test/test_glib_helpers.h"
```
sporksmith, line 26: Please put a line-break between the system headers and our internal includes. This helps keep them in the right order when using e.g. `clang-format`.

src/test/tcp/test_tcp.c
sporksmith, line 69: No need to make this world accessible, especially since even a read from an outside process would confuse the test. 0066 is probably what we want here.

florian-vuillemot: 0666 didn't work. But 0600 works. Is it ok?

src/test/tcp/CMakeLists.txt
florian-vuillemot: We need those dependencies with test_glib_helpers.h

src/test/tcp/test_tcp.c
florian-vuillemot: hum, I tried with other options but I think I should at least let `MSG_NOERROR`.. But, I don't think it's a big deal because if the message is not correct the server port will not be correct either and tests will fail. On the other hand, this will be more difficult to troubleshoot because the error will not be correctly located

src/test/tcp/test_tcp.c
florian-vuillemot: oh yes of course !

src/test/tcp/test_tcp.c
florian-vuillemot: We need a filename with ftok

src/test/tcp/CMakeLists.txt
sporksmith: Oops, right, my mistake :)

src/test/tcp/test_tcp.c
```diff
@@ -782,9 +844,17 @@ int main(int argc, char *argv[]) {
             MYLOG("error, client mode also needs a server ip address; see usage");
             return -1;
         }
+        if (argc == 5) {
+            queuename = argv[4];
```
sporksmith, line 848: We should make this argument mandatory rather than having a default. We want to be careful to avoid diffiult-to-debug errors in the case that multiple concurrent tests accidentally try to use the same queue name

src/test/tcp/test_tcp.c
sporksmith: I think we want to make it mandatory to get this from the command-line (see below), in which case we can drop this.

src/test/tcp/test_tcp.c
florian-vuillemot: To do that I should provide a unique filename for each test. Which files can I take ?

florian-vuillemot:
@sporksmith, I'm not sure to understand your question. We have to use at least one flag with `msgrcv` and `msgsnd`. Flags allowed are limited and if we don't want to use `MSG_NOWAIT` we have to use `MSG_NOERROR` to provide at leat one flag if I understand correctly.

src/test/tcp/test_tcp.c
sporksmith: We can use an empty file named after the test-name. e.g. something like (pseudo-code; might need some escaping to embed a shell variable)

```
add_test(
    NAME tcp-blocking-loopback
- COMMAND shadow-test-launcher test-tcp blocking ./test-tcp server : test-tcp blocking ./test-tcp client localhost
+ COMMAND /bin/bash -c "QFILE=tcp-blocking-loopback-qfile && shadow-test-launcher test-tcp blocking ./test-tcp server $QFILE : test-tcp blocking ./test-tcp client localhost $QFILE && rm $QFILE"
```

It'd be nice if `ctest` set some environment variable to the test-name, in which case we could just use that instead of having to pipe it through ourselves, but as far as I can tell it doesn't.

sporksmith:
> @sporksmith, I'm not sure to understand your question. We have to use at least one flag with msgrcv and msgsnd. Flags allowed are limited and if we don't want to use MSG_NOWAIT we have to use MSG_NOERROR to provide at leat one flag if I understand correctly.

You should be able to pass `0` to mean "no flags".

src/test/tcp/test_tcp.c
sporksmith: Actually, since this can be a regular file, we can just use `mktemp` to get us a guaranteed-unique filename in the `tmp` directory. e.g.

```
COMMAND /bin/bash -c 'QFILE=`mktemp` && ...
```

florian-vuillemot:
Humm, I tried without success but now it seems to be working !

src/test/tcp/test_tcp.c
sporksmith: Alternatively, since we're forced to handle this in the wrapper script anyways, we could use `ipcmk` to generate a distinct message queue for each test, and pass the generated key directly instead of using ftok.

florian-vuillemot: The only problem with one of those technics is the difficulty to provide dynamic arguments when we run tests with Shadow. I will let you take a look at what I did. I hope there is a better and simpler way !

sporksmith: Yes, I'd overlooked the difficulty of passing a dynamic command-line argument to Shadow. I don't know of an easy way to do this, currently.

I was surprised what you did seemed to work; I checked it out locally and found that actually the tests are failing, but the failures are being suppressed. In:

```
/bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ${CMAKE_BINARY_DIR}/src/main/shadow -l ...; ipcrm -q $QUEUE"
```

The bash process ends up returning the exit code of `ipcrm` rather than shadow's exit code. If we wanted to ensure cleanup while returning the correct exit code it'd need to be something like:

```
/bin/bash -c "QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ${CMAKE_BINARY_DIR}/src/main/shadow -l ...; RV=$?; ipcrm -q $QUEUE; exit $RV"
```

As for dealing with the dynamic argument issue... the cleanest workaround I can think of is to use an environment variable instead of a command-line argument. Then you can grab it from inside the program using `getenv`. You'd need to add an `export` to the shell snippet. e.g.

```
/bin/bash -c "export QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'` && ${CMAKE_BINARY_DIR}/src/main/shadow -l ...; RV=$?; ipcrm -q $QUEUE; exit $RV"
```

Perhaps it'd be better to encapsulate this in a small shell script. I think this should do it:

```
#!/bin/bash

# Catch failures
set -euo pipefail

# Create Q, and export into environment of child processes
export QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'`

# Ensure Q is removed on exit
trap "ipcrm -q $QUEUE" EXIT

# Interpret remaining args as a command to run:
$@
```

The test commands could then be something like:
```
COMMAND ${CMAKE_SOURCE_DIR}/with_q.sh ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d blocking-loopback.shadow.data ${CMAKE_CURRENT_SOURCE_DIR}/tcp-blocking-loopback.test.shadow.config.xml
```

florian-vuillemot: Creating the queue from script seem ok, but not destructing. Even with the atexit method, I see queues in my IPC after tests.

src/test/tcp/with_q.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# Catch failures
+set -euo pipefail
+
+# Create Q, and export into environment of child processes
+export QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'`
```
sporksmith, line 7: You left off the bits of my suggested script that a) clean up the queue. It makes sense to clean it up in the script instead of inside the test to keep creation and deletion in the same scope. b) Actually run the command provided as arguments; i.e. the actual test :).

```
# Remove the queue on script exit
trap "ipcrm -q $QUEUE" EXIT

# Run the specified command
$@
```

sporksmith: P.S. In general I suggest checking the test output when you're not sure what's going on. That should make it clear in this case that the tests aren't actually running.

I find it easiest to use the `ctest` command directly when debugging an individual test, and run only the tests I'm working on. Turning on test output, the command would be something like: `ctest --tests-regex tcp --verbose`

To see the output of the test code running under shadow, you'll need to check the logs in the data directories specified in the test commands. e.g. `blocking-lossy.shadow.data`

florian-vuillemot: I forgot to include it, but of course I tried with your full script but tests failed. I will try with the ctest  tomorrow :)

florian-vuillemot: With `trap` I obtained the following error
```
3: ** (process:1000): ERROR **: 00:00:01.000: !((msgqueue = msgget(key, 0)) != -1): No such file or directory
3: shadow/src/test/tcp/with_q.sh: line 12:  7419 Trace/breakpoint trap   (core dumped) $@
```

florian-vuillemot: I searched again but I keep getting errors. I think the better way is to provide the filename from the env argument, then create and delete the queue in the test with `ftok`. It's not the best but should work ! What do you think ?

sporksmith: I don't think the `trap` is what's causing the error. It's just reporting it. Assuming you also added the `$@` at the end, the test is now being run (and failing).

I got the tests working in my local copy of this branch but can't seem to push the commits. Did you check the ['allow edits from maintainers'](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) box? It's also my first time trying to use that feature, so I might be doing something wrong.

Anyway, here's my patch that gets the tests working. Feel free to apply these changes locally, or we can try to figure out the 'allow edits from maintainers' thing.

```
diff --git a/src/test/tcp/test_tcp.c b/src/test/tcp/test_tcp.c
index 2b7a6d1f..ab675a1a 100644
--- a/src/test/tcp/test_tcp.c
+++ b/src/test/tcp/test_tcp.c
@@ -62,32 +62,12 @@ static void _mylog(const char* fileName, const int lineNum, const char* funcName
     fflush(stdout);
 }
 
-static void remove_queue() {
-    assert_true_errno(msgctl(msgqueue, IPC_RMID, NULL) != -1);
-}
-
-// create a System V queue
-// the key can be provide directly as number or put in env variable
-// if the program is running as client, we should remove the message queue at the end
-static void create_msgqueue(const char *queueid) {
-    char *endptr;
-    long int key;
-
-    if (queueid[0] == '$'){
-        queueid = getenv(queueid + 1);
-    }
-
-    errno = 0;
-    key = strtol(queueid, &endptr, 10);
-    // Check for various possible errors
-    if ((errno == ERANGE && (key == LONG_MAX || key == LONG_MIN))
-        || (errno != 0 && key == 0)
-        || endptr == queueid) {
-        perror("strtol");
-        exit(EXIT_FAILURE);
-    }
-
-    assert_true_errno((msgqueue = msgget(key, 0)) != -1);
+// Get the message queue id.
+static void get_msgqueue() {
+    char* id_string = getenv("QUEUE");
+    // FIXME: Validate that integer conversion worked.
+    // FIXME: Should return the queue id. We don't need to use a global anymore.
+    msgqueue = atoi(id_string);
 }
 
 static void queue_send_u16(uint16_t i) {
@@ -824,6 +804,7 @@ int main(int argc, char *argv[]) {
     }
 
     const char *io_mode = argv[1];
+    // FIXME: We don't need this argument anymore.
     const char *queuename = argv[2];
     const char *execution_mode = argv[3];
     iowait_func wait = NULL;
@@ -851,11 +832,11 @@ int main(int argc, char *argv[]) {
             MYLOG("error, client mode also needs a server ip address; see usage");
             return -1;
         }
-        create_msgqueue(queuename);
+        get_msgqueue();
         MYLOG("running client in mode %s", io_mode);
         result = _run_client(wait, argv[4], use_iov);
     } else if(strncasecmp(execution_mode, "server", 6) == 0) {
-        create_msgqueue(queuename);
+        get_msgqueue();
         MYLOG("running server in mode %s", io_mode);
         result = _run_server(wait, use_iov);
     } else {
diff --git a/src/test/tcp/with_q.sh b/src/test/tcp/with_q.sh
index c5b67126..73dbd59b 100755
--- a/src/test/tcp/with_q.sh
+++ b/src/test/tcp/with_q.sh
@@ -5,3 +5,9 @@ set -euo pipefail
 
 # Create Q, and export into environment of child processes
 export QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'`
+
+# Ensure Q is removed on exit
+trap "ipcrm -q $QUEUE" EXIT
+
+# Interpret remaining args as a command to run:
+$@

```

florian-vuillemot: Ah ! Ok I understand now, my usage of the IPC interface were really bad ! Thank you @sporksmith 
I allowed maintainers to edit my PR, but I don't know why you can't. Maybe because the PR comes from a fork ?

sporksmith: Ok; it might just be that I don't have sufficient permission in the shadow repo to be considered a 'maintainer' for this purpose.

Attaching patches created with `git format-patch` for convenience. Feel free to apply them with `git am` if you like. If it's easier to just manually make the changes again yourself manually that's fine too.

0001-with_q.sh-Run-the-provided-command-and-destroy-q-on-.patch
```
From 3bf59bbe90f93c04fe8846c4a2a7829f171150f7 Mon Sep 17 00:00:00 2001
From: Jim Newsome <jnewsome@torproject.org>
Date: Fri, 15 May 2020 09:24:56 -0500
Subject: [PATCH 1/2] with_q.sh: Run the provided command, and destroy q on
 exit

---
 src/test/tcp/with_q.sh | 6 ++++++
 1 file changed, 6 insertions(+)

diff --git a/src/test/tcp/with_q.sh b/src/test/tcp/with_q.sh
index c5b67126..73dbd59b 100755
--- a/src/test/tcp/with_q.sh
+++ b/src/test/tcp/with_q.sh
@@ -5,3 +5,9 @@ set -euo pipefail
 
 # Create Q, and export into environment of child processes
 export QUEUE=`ipcmk --queue | sed 's/[^0-9]*//g'`
+
+# Ensure Q is removed on exit
+trap "ipcrm -q $QUEUE" EXIT
+
+# Interpret remaining args as a command to run:
+$@
-- 
2.26.2
```

0002-test_tcp-Use-queue-ID-from-environment-variable.patch
```From bae2ac7a7489f1aebbc473789069176a029bfe94 Mon Sep 17 00:00:00 2001
From: Jim Newsome <jnewsome@torproject.org>
Date: Fri, 15 May 2020 09:25:19 -0500
Subject: [PATCH 2/2] test_tcp: Use queue ID from environment variable

---
 src/test/tcp/test_tcp.c | 37 +++++++++----------------------------
 1 file changed, 9 insertions(+), 28 deletions(-)

diff --git a/src/test/tcp/test_tcp.c b/src/test/tcp/test_tcp.c
index 2b7a6d1f..ab675a1a 100644
--- a/src/test/tcp/test_tcp.c
+++ b/src/test/tcp/test_tcp.c
@@ -62,32 +62,12 @@ static void _mylog(const char* fileName, const int lineNum, const char* funcName
     fflush(stdout);
 }
 
-static void remove_queue() {
-    assert_true_errno(msgctl(msgqueue, IPC_RMID, NULL) != -1);
-}
-
-// create a System V queue
-// the key can be provide directly as number or put in env variable
-// if the program is running as client, we should remove the message queue at the end
-static void create_msgqueue(const char *queueid) {
-    char *endptr;
-    long int key;
-
-    if (queueid[0] == '$'){
-        queueid = getenv(queueid + 1);
-    }
-
-    errno = 0;
-    key = strtol(queueid, &endptr, 10);
-    // Check for various possible errors
-    if ((errno == ERANGE && (key == LONG_MAX || key == LONG_MIN))
-        || (errno != 0 && key == 0)
-        || endptr == queueid) {
-        perror("strtol");
-        exit(EXIT_FAILURE);
-    }
-
-    assert_true_errno((msgqueue = msgget(key, 0)) != -1);
+// Get the message queue id.
+static void get_msgqueue() {
+    char* id_string = getenv("QUEUE");
+    // FIXME: Validate that integer conversion worked.
+    // FIXME: Should return the queue id. We don't need to use a global anymore.
+    msgqueue = atoi(id_string);
 }
 
 static void queue_send_u16(uint16_t i) {
@@ -824,6 +804,7 @@ int main(int argc, char *argv[]) {
     }
 
     const char *io_mode = argv[1];
+    // FIXME: We don't need this argument anymore.
     const char *queuename = argv[2];
     const char *execution_mode = argv[3];
     iowait_func wait = NULL;
@@ -851,11 +832,11 @@ int main(int argc, char *argv[]) {
             MYLOG("error, client mode also needs a server ip address; see usage");
             return -1;
         }
-        create_msgqueue(queuename);
+        get_msgqueue();
         MYLOG("running client in mode %s", io_mode);
         result = _run_client(wait, argv[4], use_iov);
     } else if(strncasecmp(execution_mode, "server", 6) == 0) {
-        create_msgqueue(queuename);
+        get_msgqueue();
         MYLOG("running server in mode %s", io_mode);
         result = _run_server(wait, use_iov);
     } else {
-- 
2.26.2
```

src/test/tcp/tcp-nonblocking-epoll-loopback.test.shadow.config.xml
```diff
@@ -22,6 +22,6 @@
   <plugin id="testtcp" path="libshadow-plugin-test-tcp.so"/>
   <node id="tcptestnode" >
     <application plugin="testtcp" time="1" arguments="nonblocking-epoll server" />
-    <application plugin="testtcp" time="2" arguments="nonblocking-epoll client 127.0.0.1" />
+    <application plugin="testtcp" time="2" arguments="nonblocking-epoll client 127.0.0.1 " />
```
sporksmith, line 25: Added a stray space here

src/test/tcp/with_q.sh
florian-vuillemot: I just see your update, sorry. I think I can change the commit owner and force push to let you as owner of your modifications if you want

Issue #756, opened 2020-04-09
epoll-writeable-shadow test case takes too long


robgjansen:
There must be some kind of failure occuring in the `epoll-writeable-shadow` test since it is taking longer than 1 minute to complete:

https://github.com/shadow/shadow/runs/571228243#step:8:208

florian-vuillemot:
Maybe the problem comes from the topology config? If you change `<kill time="10000"/>` by for example `<kill time="1000"/>` the execution time is divided by 10.

```
1/3 Test #1: epoll ............................   Passed    1.10 sec
    Start 2: epoll-shadow
2/3 Test #2: epoll-shadow .....................   Passed    0.11 sec
    Start 3: epoll-writeable-shadow
3/3 Test #3: epoll-writeable-shadow ...........   Passed    6.57 sec
100% tests passed, 0 tests failed out of 3
Total Test time (real) =   7.85 sec
```

And less for `<kill time="100"/>`
```
1/3 Test #1: epoll ............................   Passed    1.10 sec
    Start 2: epoll-shadow
2/3 Test #2: epoll-shadow .....................   Passed    0.10 sec
    Start 3: epoll-writeable-shadow
3/3 Test #3: epoll-writeable-shadow ...........   Passed    0.80 sec
100% tests passed, 0 tests failed out of 3
Total Test time (real) =   2.07 sec
```

I tried to remove the for loop in the client (for(;;)) in shd-test-epoll-writeable.c, but I encountered the same time error. As if Shadow continues to work even if the 2 nodes are finished.
Is it normal ?

sporksmith:
`git bisect` identifies the culprit as https://github.com/shadow/shadow/commit/58224fd5187e4a603567639fd3bd6fedff0ed1fb, using the test script:

```
#!/bin/bash
./setup build --clean --test -j8
cd build
timeout 5 ctest -R epoll-writeable-shadow`
```

sporksmith:
@florian-vuillemot Yeah, the issue seems to be shadow no longer deciding to exit earlier; in the commit before 58224fd it exits quickly; at and after 58224fd shadow keeps running until it hits that timeout. I poked a bit to add more debugging output from the test itself and do more error-checking; the test-code all seems to be WAI. I even changed the client to exit after the expected amount of data was received, and shadow keeps running long after both client and server have exited:

```
[...]

PR #788, opened 2020-05-02
Skip unneccesary token bucket refills

robgjansen:
Don't schedule events to refill token buckets that are already full.

Fixes GH-756

### DIFF
--- a/src/main/host/network_interface.c
+++ b/src/main/host/network_interface.c
@@ -66,14 +66,22 @@ struct _NetworkInterface {
      * packets that do not conform to incoming rate limits are dropped. */
     NetworkInterfaceTokenBucket receiveBucket;
 
+    /* Store the time we started refilling our token buckets. This is used to
+     * help us compute when refills should occur following an idle period. */
+    SimulationTime timeStartedRefillingBuckets;
+
+    /* If we have scheduled a refill task but it has not yet executed. */
+    gboolean isRefillPending;
+
     /* To support capturing incoming and outgoing packets */
     PCapWriter* pcap;
 
     MAGIC_DECLARE;
 };
 
-/* forward declaration */
+/* forward declarations */
 static void _networkinterface_sendPackets(NetworkInterface* interface);
+static void _networkinterface_refillTokenBucketsCB(NetworkInterface* interface, gpointer userData);
 
 static gint _networkinterface_compareSocket(const Socket* sa, const Socket* sb, gpointer userData) {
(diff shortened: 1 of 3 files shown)


src/test/epoll/test_epoll_writeable.c
```diff
@@ -2,7 +2,7 @@
  * The Shadow Simulator
  * See LICENSE for licensing information
  */
-#define _POSIX_C_SOURCE 199309L
+//#define _POSIX_C_SOURCE 199309L
```
sporksmith, line 5: I know this is from my commit, but - maybe just delete? Any idea why this was here in the first place or why we might need it again?

IIRC I had to comment it out to enable some feature, but I wasn't sure commenting it out was the right fix.

src/main/host/network_interface.c
```diff
@@ -66,154 +66,163 @@ struct _NetworkInterface {
...
 
-    /* add more bytes to the token buckets */
-    _networkinterface_refillTokenBucket(&interface->receiveBucket);
-    _networkinterface_refillTokenBucket(&interface->sendBucket);
+    /* This computes the time that the next refill should have occurred if we
+     * were to always call the refill function every interval, even though in
+     * practice we stop scheduling refill tasks once we notice we are idle. */
+    SimulationTime offset = now - interface->timeStartedRefillingBuckets;
+    SimulationTime relTimeSincelastRefill = offset % interval;
+    SimulationTime relTimeUntilNextRefill = interval - relTimeSincelastRefill;
```
sporksmith, line 139: Looked over this again per offline discussion. I think it's right.

src/test/epoll/test_epoll_writeable.c
robgjansen: > Any idea why this was here in the first place or why we might need it again?

According to [the libc manual on Feature Test Macros](https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html):

> Macro: _POSIX_C_SOURCE
>
>    Define this macro to a positive integer to control which POSIX functionality is made available. The greater the value of this macro, the more functionality is made available.
>
> ...
>
>    If you define this macro to a value greater than or equal to 199309L, then the functionality from the 1993 edition of the POSIX.1b standard (IEEE Standard 1003.1b-1993) is made available. 
> ...

There are codes for 1990, 1992, 1993, 1995, 2001, 2008. Since it works without defining it, I think deleting the line should be fine.

PR #789, opened 2020-05-06
Documentation updates: Site map, reorg, version control

sporksmith:


sporksmith:
Was going to write up the intended merge workflow for the Phantom branch, but thought it'd be better to document the current state as a starting point. Once this is in I'll merge it over to Phantom and send a PR to update it for the proposed merge strategy.

PR #801, opened 2020-05-15
Logger improvements

sporksmith:
In the default logger:
  * Log elapsed time
  * Log the source file basename instead of the whole path
  * Log using g_print (as shadow logger does) ~directly to stderr~ instead of to glib's logger
  * In debug builds, abort the program on errors.

As I was testing, dealing with glib's logger underneath was a bit of a headache. While it has some nice features, the defaults aren't want we want. Easier to just bypass it for now.

Ported bind test to use our logger macros instead of glib's to help test and to get the additional fields (code location, time, etc).

### DIFF
--- a/src/support/logger/logger.c
+++ b/src/support/logger/logger.c
@@ -1,6 +1,7 @@
 #include "support/logger/logger.h"
 
 #include <pthread.h>
+#include <stdio.h>
 #include <sys/time.h>
 
 static Logger* defaultLogger = NULL;
@@ -13,19 +14,6 @@ void logger_setDefault(Logger* logger) {
 }
 
 Logger* logger_getDefault() { return defaultLogger; }
-
-static GLogLevelFlags _gloglevelflags(LogLevel level) {
-    switch (level) {
-        case LOGLEVEL_CRITICAL: return G_LOG_LEVEL_CRITICAL;
-        case LOGLEVEL_WARNING: return G_LOG_LEVEL_WARNING;
-        case LOGLEVEL_MESSAGE: return G_LOG_LEVEL_MESSAGE;
-        case LOGLEVEL_INFO: return G_LOG_LEVEL_INFO;
-        case LOGLEVEL_DEBUG: return G_LOG_LEVEL_DEBUG;
-        case LOGLEVEL_UNSET:
-        case LOGLEVEL_ERROR:
-        default: return G_LOG_LEVEL_ERROR;
-    }
-}
 
 // Process start time, initialized on first use.
 // TODO: Parse out of /proc/$$/stat instead to get the true process start time.
@@ -58,8 +46,8 @@ static void _logger_default_log(LogLevel level, const gchar* fileName,
     gmtime_r(&tv.tv_sec, &tm);
     gchar* timeString = g_strdup_printf("%02d:%02d:%02d.%06ld", tm.tm_hour,
                                         tm.tm_min, tm.tm_sec, tv.tv_usec);
-    g_log("shadow", _gloglevelflags(level), "%s [%s:%i] [%s] %s", timeString,
-          baseName, lineNumber, functionName, message);
+    g_print("%s %s [%s:%i] [%s] %s\n", timeString, loglevel_toStr(level),
+            baseName, lineNumber, functionName, message);
     g_free(message);
     g_free(timeString);
     g_free(baseName);
--- a/src/support/logger/logger.c
+++ b/src/support/logger/logger.c
@@ -19,35 +19,37 @@ Logger* logger_getDefault() { return defaultLogger; }
 // Process start time, initialized explicitly or on first use.
 static pthread_mutex_t _start_time_mutex = PTHREAD_MUTEX_INITIALIZER;
 static bool _start_time_initd = false;
-static struct timeval _start_time;
+static int64_t _monotonic_start_time_micros;
 
-struct timeval logger_get_global_start_time() {
+int64_t logger_now_micros() {
+    return g_get_monotonic_time();
+}
+
+int64_t logger_get_global_start_time_micros() {
     pthread_mutex_lock(&_start_time_mutex);
     if (!_start_time_initd) {
         // TODO: Parse out of /proc/$$/stat instead to get the true process
-        // start time.
-        gettimeofday(&_start_time, NULL);
+        // start time?
+        _monotonic_start_time_micros = logger_now_micros();
         _start_time_initd = true;
     }
-    struct timeval start_time = _start_time;
+    int64_t t = _monotonic_start_time_micros;
     pthread_mutex_unlock(&_start_time_mutex);
-    return start_time;
+    return t;
 }
 
-void logger_set_global_start_time(const struct timeval* t) {
+void logger_set_global_start_time_micros(int64_t t) {
     pthread_mutex_lock(&_start_time_mutex);
-    _start_time = *t;
+    _monotonic_start_time_micros = t;
     _start_time_initd = true;
     pthread_mutex_unlock(&_start_time_mutex);
 }
 
-struct timeval logger_get_global_elapsed_time() {
-    const struct timeval start_time = logger_get_global_start_time();
-    struct timeval now;
-    gettimeofday(&now, NULL);
-    struct timeval elapsed;
-    timersub(&now, &start_time, &elapsed);
-    return elapsed;
+int64_t logger_elapsed_micros() {
+    // We need to be careful here to get t0 first, since the first time this
+    // function is called it will cause the start time to be lazily initialized.
+    int64_t t0 = logger_get_global_start_time_micros();
+    return logger_now_micros() - t0;
 }
 
 static void _logger_default_log(LogLevel level, const gchar* fileName,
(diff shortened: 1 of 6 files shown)


sporksmith:
Thinking some more about how this will be used in the Phantom branch, I'm going to make a few changes here first...

sporksmith:
I'm a bit mystified why the CI is failing (only getting 30 successful transfers instead of 40).

I'm able to reproduce locally though, and verified that it still passes at master, so it's definitely something in this PR. Going to run a git bisect to see if I can narrow it down, but probably reasonable to review in the meantime if you get a chance.

sporksmith:
After bisecting to the commit that broke the shadow-plugin-tor CI (f836442447859a478f5016c5244fcb43143b459c) and breaking it into smaller commits, I eventually found I can reproduce the failure by just *calling* gettimeofday from within the shadow logger. See #804

sporksmith:
Digging into the simulation results, it seems to be the "torhiddenclient" transfers that fail.

All successful transfers on a run on master:
```
$ find simulation-succeeded/shadowtor-minimal/shadow.data/hosts -name '*client*.log' | xargs grep -c transfer-complete | grep -vE ':0$'
simulation-succeeded/shadowtor-minimal/shadow.data/hosts/torbridgeclient/stdout-torbridgeclient.tgen.1002.log:10
simulation-succeeded/shadowtor-minimal/shadow.data/hosts/torhiddenclient/stdout-torhiddenclient.tgen.1002.log:10
simulation-succeeded/shadowtor-minimal/shadow.data/hosts/client/stdout-client.tgen.1000.log:10
simulation-succeeded/shadowtor-minimal/shadow.data/hosts/torclient/stdout-torclient.tgen.1002.log:10
```

And the failed run on #804:
```
$ find simulation/shadowtor-minimal/shadow.data/hosts -name '*client*.log' | xargs grep -c transfer-complete | grep -vE ':0$'
simulation/shadowtor-minimal/shadow.data/hosts/torbridgeclient/stdout-torbridgeclient.tgen.1002.log:10
simulation/shadowtor-minimal/shadow.data/hosts/client/stdout-client.tgen.1000.log:10
simulation/shadowtor-minimal/shadow.data/hosts/torclient/stdout-torclient.tgen.1002.log:10
```

sporksmith:
Ready for review. (I'll rebase to fix the block-fixup check before merging).

I've successfully kicked the can for now by using a different time API, which doesn't seem to trigger the bug.

src/support/logger/logger.c
```diff
@@ -11,27 +16,69 @@ void logger_setDefault(Logger* logger) {
...
                                 const gchar* functionName,
                                 const gint lineNumber, const gchar* format,
                                 va_list vargs) {
     gchar* message = g_strdup_vprintf(format, vargs);
-    g_log("shadow", _gloglevelflags(level), "[%s:%i] [%s] %s", fileName,
-          lineNumber, functionName, message);
+    gchar* baseName = g_path_get_basename(fileName);
+
+    int64_t elapsed_micros =
+        logger_now_micros() - logger_get_global_start_time_micros();
```
robgjansen, line 63: This line should be replaced with `logger_elapsed_micros()`.

src/support/logger/logger.c
```diff
@@ -11,27 +16,69 @@ void logger_setDefault(Logger* logger) {
...
+    if (!_start_time_initd) {
+        // TODO: Parse out of /proc/$$/stat instead to get the true process
+        // start time?
+        _monotonic_start_time_micros = logger_now_micros();
+        _start_time_initd = true;
     }
+    int64_t t = _monotonic_start_time_micros;
+    pthread_mutex_unlock(&_start_time_mutex);
+    return t;
+}
```
robgjansen, line 39: Should we really lock a mutex on every single log message, just to be able to set the start time? I think we can avoid it if we use atomic operations (`#include <stdatomic.h>` in the C11 standard).

Assuming you have defined `volatile atomic_flag _start_time_initd = ATOMIC_FLAG_INIT;`, then something like:

```
int64_t logger_get_global_start_time_micros() {
    if(!atomic_flag_test_and_set(&_start_time_initd)) {
        logger_set_global_start_time_micros(logger_now_micros());
    }
    return _monotonic_start_time_micros;
}
```

See also: https://en.cppreference.com/w/c/atomic

src/support/logger/logger.h
```diff
@@ -4,11 +4,10 @@
 /*
  * A simple logger API.
  *
- * By default this simply wraps glib's logging functionality. However, it also
- * supports overriding with a custom Logger.  Unlike in glib, when a custom
- * Logger is supplied, it's that logger's job to do any necessary
- * synchronization. This allows us to use a custom Logger in Shadow that avoids
- * a global lock.
+ * By default this simply writes to stderr. However, it also supports
```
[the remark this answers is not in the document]
sporksmith: Oops, yes I'd switched it but missed updating this comment.

src/support/logger/logger.h
```diff
@@ -61,4 +60,21 @@ void logger_log(Logger* logger, LogLevel level, const gchar* fileName,
...
+
+// Returns "now" according to a monotonic system clock.
+int64_t logger_now_micros();
+
+// Returns elapsed micros since agreed-upon start time.
+int64_t logger_elapsed_micros();
+
+// Set the global start time used in log messages. If this isn't called, the
+// start time will be set to the current time the first time it's accessed.
+void logger_set_global_start_time_micros(int64_t);
```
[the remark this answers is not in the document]
sporksmith: Oops, missed this comment.

This is a matter of some debate, but in general I agree with the Google style guide that it's generally preferable to use signed integers to represent integers

https://google.github.io/styleguide/cppguide.html#Integer_Types

> Unsigned integers are good for representing bitfields and modular arithmetic. Because of historical accident, the C++ standard also uses unsigned integers to represent the size of containers - many members of the standards body believe this to be a mistake, but it is effectively impossible to fix at this point. The fact that unsigned arithmetic doesn't model the behavior of a simple integer, but is instead defined by the standard to model modular arithmetic (wrapping around on overflow/underflow), means that a significant class of bugs cannot be diagnosed by the compiler. In other cases, the defined behavior impedes optimization.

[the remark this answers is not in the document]
robgjansen: OK. I don't feel strongly.

src/support/logger/logger.c
sporksmith: Thinking about this some more, we probably don't need to call the setter after any additional threads are started. I'll change it and note that restriction.

robgjansen: > More generally, grabbing a mutex is cheap anyway as long there's not actually any contention; in the common case we're only executing a couple instructions with the mutex held so there's unlikely to be any contention.

It becomes more likely as the number of Shadow worker threads increases, and especially as we increase the logging verbosity. I currently run some experiments on an 80 hyperthread machine, and all of them could be trying to log at the same time.

> I'd originally used pthread_once, which probably does something like this under the hood, but changed it when I realized I needed a setter for use inside the shim. Using atomic/pthread once primitives won't address parallel access via the setter.

Is your argument that using `pthread_once` in the getter and keeping the mutex for the setter is too much added complexity? It does seem worth it to me in order to **guarantee** that contention doesn't reduce performance.

robgjansen: (Whoops, it looks like your comment came in as I was typing mine and I either missed it or Github didn't load it until I just refreshed the page.)

sporksmith: Actually just using an atomic wouldn't give us safety since if the timer isn't initialized explicitly, and two threads enter this code simultaneously, the second thread needs to wait for the first thread to finish initializing.

We can handle that case with pthread_once instead (with the caveat that the setter is no longer thread safe).

> pthread_once in the getter and keeping the mutex for the setter is too much added complexity?

I don't think that would quite do what we want. If we want the setter to be thread-safe, then the getter would also need to lock the mutex.

I think the possibilities here are:

1. Just use a mutex as the code is now. This is the simplest, and IMO unlikely to be a performance issue even in the 80 hyperthread case due to only executing a few instructions while holding the mutex. I don't think it's worth spending the time trying to validate that now though, so alternatively:
2. Use [pthread read/write locks](https://linux.die.net/man/3/pthread_rwlock_init), which allow parallel "readers" under the lock. I think this would address your concern *and* let us have a thread-safe setter. What I've "heard" though is that the extra complexity in read/write locks actually make them less efficient in cases like this one where the lock isn't held anyway, though.
3. Use pthread_once and declare that the setter is not thread safe. I think we can agree that this is the most performant option, and that we don't actually need the setter to be thread safe?

I'm leaning towards 3 and will implement it shortly; LMK if you disagree :)

robgjansen: I don't know much about mutex performance, so I'm willing to defer to your judgement and go with option 1.

But if you are now convinced about option 3, that's fine too :)

sporksmith: Unless we have a reason to make the setter thread safe, I think no reason not to go with #3 :)

jtracey: For future reference: Even without contention, lots of atomic operations (including locking) can slow compute throughput down, especially on large multi-CPU machines, because every call to an atomic operation requires cache coherence across all cores, even when there's no data dependence.

robgjansen: I wonder if it's possible to use a constructor to set the init time and then avoid the `pthread_once` call? Something like `__attribute__((constructor(PRIORITY))) static void _init_func() {}`.

Not sure if it's worth it at this point, but wanted to mention it as another possible option here.

sporksmith: > I wonder if it's possible to use a constructor to set the init time and then avoid the pthread_once call? Something like __attribute__((constructor(PRIORITY))) static void _init_func() {}.

It's possible, but using constructors can be a bit frought to ensure that they get called in the correct order.

Btw re @jtracey's comment I looked a bit closer at the disassembly for pthread_once and  that once initialization has happened there's no atomic operation. It just compares against a plain bool first and only if that hasn't been set yet gets into the synchronization bits. i.e. after initialization the pthread_once overhead should be negligible - just a function call and a bool comparison.

sporksmith: I was actually thinking about going the other direction and switching over some of the other constructor usage in the shim to use pthread_once as well. I think the only constructor we need is the main shim startup "entry point"; it's probably safer and clearer to use lazy initialization anywhere else we need something init'd.

robgjansen: Thanks for looking into `pthread_once` - it seems like the best way to go!

Issue #773, opened 2020-04-22
More human-readable Shadow config


robgjansen:
Support for a more human-readable format such as YAML would be nice as it would be easier to read/write/modify config files and would help us eventually move away from XML configuration files. This work could be split into multiple steps:

  - [x] (Merged in #803) - Create an external tool that converts between `shadow.config.xml` XML files and `shadow.config.yaml` YAML files (i.e., converts XML to YAML, and YAML to XML). 
  - [x] (Merged in #895) - Modify Shadow so that it will accept its XML configuration on stdin rather than requiring that we read it from the `shadow.config.xml` file. (It should be able to read from stdin or from a file.)
  - [x] (Merged in #972) - Convert existing tests from xml to yaml, and use the yaml2xml converter in the test cases.
  - [ ] Long term: add support to Shadow so it can natively read and understand YAML files.
  - [ ] Long term: new "options" section in the config file. Here we could specify any global options, and make it so that command line options could also instead be added to the yaml file (and Shadow internally would process the options using the same code for consistency whether the options are specified on the command line or in the config file).

As an example, the [small example in the shadow repo](https://github.com/shadow/shadow/blob/master/resource/examples/shadow.config.xml) currently looks like this in XML:

```xml
<shadow stoptime="3600">
  <topology><![CDATA[<?xml version="1.0" encoding="utf-8"?><graphml xmlns="http://graphml.graphdrawing.org/xmlns" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://graphml.graphdrawing.org/xmlns http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd">
  <key attr.name="packetloss" attr.type="double" for="edge" id="d6" />
  <key attr.name="latency" attr.type="double" for="edge" id="d5" />
  <key attr.name="packetloss" attr.type="double" for="node" id="d4" />
  <key attr.name="countrycode" attr.type="string" for="node" id="d3" />
  <key attr.name="bandwidthdown" attr.type="int" for="node" id="d2" />
  <key attr.name="bandwidthup" attr.type="int" for="node" id="d1" />
  <key attr.name="ip" attr.type="string" for="node" id="d0" />
  <graph edgedefault="undirected">
    <node id="isp">
      <data key="d0">0.0.0.0</data>
      <data key="d1">2251</data>
      <data key="d2">17038</data>
      <data key="d3">US</data>
      <data key="d4">0.0</data>
    </node>
[...]

PR #803, opened 2020-05-20
Tool that converts XML shadow config file to YAML.

florian-vuillemot:
https://github.com/shadow/shadow/issues/773
Example files included are here to simplify discussion on the script result


florian-vuillemot:
@robgjansen @sporksmith should I keep the compatibility with Python2 ?

robgjansen:
No, I think we've already decided to no longer support python2, so as long as it supports python3 I think that's OK.

robgjansen:
Does the python2 linter test add value, even though we don't require it to pass? If not, perhaps it's time to remove it. What do you think @sporksmith ?

sporksmith:
> Does the python2 linter test add value, even though we don't require it to pass? If not, perhaps it's time to remove it. What do you think @sporksmith ?

Dropping it SGTM :)

florian-vuillemot:
Great ! About the code, the output is not exactly like in your example @robgjansen , the format is like the yaml package provided. Would you want to improve it ? And if yes, how ?

src/test/tools/convert_xml_to_yaml/file.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - arguments: tgen.server.graphml.xml
+    plugin: tgen
+    starttime: '1'
+- id: client
+  process:
+  - arguments: tgen.client.graphml.xml
+    plugin: tgen
+    starttime: '2'
+option:
+  stoptime: '3600'
```
robgjansen, line 13: This is sort of backwards from what I would expect. The numbers are given in quotes, and the strings are not. Is that a standard YAML thing?

src/test/tools/convert_xml_to_yaml/file.yaml
```diff
@@ -0,0 +1,32 @@
...
+    \ />\n  <key attr.name=\"packetloss\" attr.type=\"double\" for=\"node\" id=\"\
+    d4\" />\n  <key attr.name=\"countrycode\" attr.type=\"string\" for=\"node\" id=\"\
+    d3\" />\n  <key attr.name=\"bandwidthdown\" attr.type=\"int\" for=\"node\" id=\"\
+    d2\" />\n  <key attr.name=\"bandwidthup\" attr.type=\"int\" for=\"node\" id=\"\
+    d1\" />\n  <key attr.name=\"ip\" attr.type=\"string\" for=\"node\" id=\"d0\" />\n\
+    \  <graph edgedefault=\"undirected\">\n    <node id=\"isp\">\n      <data key=\"\
+    d0\">0.0.0.0</data>\n      <data key=\"d1\">2251</data>\n      <data key=\"d2\"\
+    >17038</data>\n      <data key=\"d3\">US</data>\n      <data key=\"d4\">0.0</data>\n\
+    \    </node>\n    <edge source=\"isp\" target=\"isp\">\n      <data key=\"d5\"\
+    >50.0</data>\n      <data key=\"d6\">0.01</data>\n    </edge>\n  </graph>\n</graphml>\n"
```
robgjansen, line 32: This is MUCH harder to read than it was before. It's basically unreadable. Is there a way to automatically format this so it is pretty-printed?

src/tools/convert_xml_to_yaml.py
```diff
@@ -0,0 +1,69 @@
...
+    return tree.getroot()
+
+
+if __name__ == '__main__':
+    if len(sys.argv) != 3:
+        print(f'Usage: {sys.argv[0]} <xml_filename> <yaml_filename>', file=sys.stderr)
+        exit(1)
+
+    xml_filename = sys.argv[1]
+    yaml_filename = sys.argv[2]
```
robgjansen, line 65: Would it be useful to use the `argparse` package here to make it easier for people to use to convert their older files? Or is that a waste of time? (I find it useful whenever tools have help menus, which you get almost for free with argparse.)

robgjansen (commented):
We should also get @sporksmith's comments here, since I think he has more experience with YAML than me and probably also a stronger opinion about how we ought to format things to improve usability.

src/test/tools/convert_xml_to_yaml/file.yaml
sporksmith: Bare (unquoted) strings are allowed in YAML, but I think can't begin with digits. It looks like the conversion is forcing everything to be a string, and leaving off the quotes where they aren't needed.

I agree it'd be better for it to make the integer-valued fields integers instead of forcing them to be strings.

src/test/tools/convert_xml_to_yaml/file.yaml
sporksmith: +1. In addition to keeping or improving the internal formatting of the XML from the source, I think we can get rid of the `\n`s and escaped quotes by using a multiline string. https://yaml-multiline.info/

sporksmith (commented):
I think we'd discussed creating a YAML->XML conversion tool first, so that the config can be written and maintained in YAML and then converted to shadow's current format "just in time"?

This tool could also be useful for rewriting old configs, but I think that might be getting ahead of ourselves a bit.

src/test/tools/convert_xml_to_yaml/file.yaml
florian-vuillemot: Good advice, it's much better now ! Thank you

src/tools/convert_xml_to_yaml.py
florian-vuillemot: Done

florian-vuillemot:
> I think we'd discussed creating a YAML->XML conversion tool first, so that the config can be written and maintained in YAML and then converted to shadow's current format "just in time"?
> 
> This tool could also be useful for rewriting old configs, but I think that might be getting ahead of ourselves a bit.

In the same tool or another one ?

robgjansen (changes requested):
Sorry for the delay. Here is some more guidance. I hope this isn't too specific, but I wanted to give you a better idea of what I was thinking to help you make progress! :)

---

I think I would make both the YAML->XML and XML->YAML part of the same tool. You can use sub-commands for this in argparse using the `add_subparsers` functions.

There would be two sub-parsers for the two sub-commands: `yaml2xml` and `xml2yaml`, each would contain 1 *positional* argument, the input filename, and 1 *optional* argument, the output filename. You would invoke the tool like this:

    convert yaml2xml shadow.config.yaml
    convert xml2yaml shadow.config.xml

By default, the tool would do the requested conversion, strip off the filename suffix and append the the suffix corresponding to the new format. So this:

    convert yaml2xml shadow.config.yaml

would write a file called `shadow.config.xml`.

But if the optional output filename was given:

    convert yaml2xml shadow.config.yaml --output myconf.xml

then the tool would just write the file to `myconf.xml` instead of doing the suffix conversion.

Some things to consider:
  - Be sure to check for errors, so that you gracefully handle the cases where the tool in invoked with a sub-command and file of the incorrect format (i.e., `convert yaml2xml shadow.config.xml` or `convert xml2yaml shadow.config.yaml`).
  - You should test some real `convert yaml2xml shadow.config.yaml` cases to make sure that the `shadow.config.xml` file that is produced actually works in Shadow. (Run a Shadow experiment with the XML config file and make sure it works correctly.)

robgjansen:
I'm assuming this isn't ready yet since it's missing the `yaml2xml` mode that @sporksmith mentioned should be done first. Let me know if that is incorrect.

florian-vuillemot:
Yes ! I will ask you for a review when this will be done :)

sporksmith (changes requested):
Submitting a place-holder review so I can get a clear signal when it's ready again. When it is, please use the 're-request' review button :)

florian-vuillemot:
@robgjansen @sporksmith could you please give me a feedback on the example files ? I don't want to miss something before cleaning the code and re requesting a review.

src/test/tools/convert_xml_to_yaml/xml2yaml.yaml
```diff
@@ -0,0 +1,40 @@
...
+  process:
+  - arguments: tgen.server.graphml.xml
+    plugin: tgen
+    starttime: 1
+- id: client
+  process:
+  - arguments: tgen.client.graphml.xml
+    plugin: tgen
+    starttime: 2
+option:
```
sporksmith, line 12: Maybe call this 'options' in case we add more

src/test/tools/convert_xml_to_yaml/xml2yaml.yaml
```diff
@@ -0,0 +1,40 @@
...
+  stoptime: 3600
+plugin:
```
sporksmith, line 14: Would be nice if we could match the typical order we use in xml files. In particular would be good to have the topology and the plugin definitions before the hosts, since those are both referenced from within the host process objects.

I think either hard-coding an order matching our "typical" order in the examples, or preserving the order in whatever source xml file being translated would be fine; whichever is easier.

src/test/tools/convert_xml_to_yaml/yaml2xml.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" ?>
+<shadow stoptime="3600">
```
sporksmith, line 2: Again would be nice to preserve our ~canonical ordering. Maybe less important here though since the idea is that generally humans won't be reading the xml file.

sporksmith (changes requested):
Only reviewed the example xml and yaml files

src/test/tools/convert_xml_to_yaml/xml2yaml.yaml
florian-vuillemot: Same with 'host' and 'plugin' ?

src/test/tools/convert_xml_to_yaml/xml2yaml.yaml
florian-vuillemot: Should be better :)

src/test/tools/convert_xml_to_yaml/yaml2xml.xml
florian-vuillemot: I didn't remove the  doctype `<?xml version="1.0" ?>`. I didn't find an elegant way to do it... What do you think ? Should I hard code a solution to remove it ?

sporksmith: As long as shadow still parses it, I think it's fine.

src/test/tools/convert_xml_to_yaml/xml2yaml.yaml
```diff
@@ -0,0 +1,40 @@
...
+          <data key="d3">US</data>
+          <data key="d4">0.0</data>
+        </node>
+        <edge source="isp" target="isp">
+          <data key="d5">50.0</data>
+          <data key="d6">0.01</data>
+        </edge>
+      </graph>
+    </graphml>
+plugin:
```
sporksmith, line 27: It's a little odd that we have list values with singular keys. e.g. `plugin` instead of `plugins`, `host` instead of `hosts`. The singular form makes sense in xml, where it's referring to a single item in that list, but is a little confusing in yaml, where it's referring to the whole list.

At the risk of making the mapping between xml and yaml a little less consistent, I'd lean towards remapping to/from the the singular/plural forms.

@robgjansen wdyt?
@florian-vuillemot how difficult would this be to do?

sporksmith (changes requested):
Thanks, Florian! The consistent ordering is definitely an improvement.

src/test/tools/convert_xml_to_yaml/xml2yaml.yaml
robgjansen: I agree with @sporksmith: plurals for the `yaml` lists makes more sense to me too.

florian-vuillemot: Hum 3 choices:
- Hard code each key and the plural
- Add an 's' to each key automatically
- Use both methods if there is a key which plural is different from adding a simple 's' at the end (process => processes)

What do you prefer ?

sporksmith: For the sake of simplicity and consistency, I'd lean towards just using a hard-coded explicit mapping for all of these; i.e. the first option.

florian-vuillemot: Where could I find all keys to map ?

robgjansen: I think you got them all:
- plugin -> plugins
- host -> hosts
- process -> processes

Also, "node" was renamed to "host", and "application" was renamed to "process", but we still technically accept "node" and "application" as valid inputs. So I suppose you should add these too:
- node -> nodes
- application -> applications

robgjansen: And if you're curious, the parsing code starting at root level is here:
https://github.com/shadow/shadow/blob/master/src/main/core/support/configuration.c#L733

florian-vuillemot: I could also do:
    node -> hosts
    application -> processes
This could help to remove old input.

florian-vuillemot:
If the examples files look good, could you please start to review the code ? :)

src/test/tools/convert_xml_to_yaml/xml2yaml.yaml
robgjansen: Sure, OK. But then you would have to spell out an explicit reverse mapping from YAML -> XML to make sure that "hosts" in YAML maps to "host" in XML (and not "node" in XML). So I think you will have to actually type out the `YAML_TO_XML` table rather than doing this:

```
YAML_TO_XML = {v: k for k, v in XML_TO_YAML.items()}
```

robgjansen:
Note that @sporksmith will review this but probably won't be able to until next week.

src/tools/convert.py
```diff
@@ -0,0 +1,220 @@
...
+    parser.add_argument('--output', help='Output filename', default=None, nargs='?')
+    args = parser.parse_args()
+
+    if args.operation == 'yaml2xml':
+        yaml_as_dict = get_yaml_from_filename(args.filename)
+        xml_root = create_xml_root(yaml_as_dict)
+        dict_to_xml(xml_root, yaml_as_dict)
+        filename_converted = get_output_filename(args, 'yaml', 'xml')
+        save_xml(xml_root, filename_converted)
+    else:
```
sporksmith, line 216: Let's explicitly check for 'xml2yaml' here, and return an error if the provided string here isn't one of the 2 supported operations. e.g. it'd be super-confusing to typo `yaml2xm`, and have it silently fall back to trying to do `xml2yaml`

src/tools/convert.py
```diff
@@ -0,0 +1,220 @@
...
+    parser.add_argument('operation', help='yaml2xml or xml2yaml')
+    parser.add_argument('filename', help='Filename to convert')
+    parser.add_argument('--output', help='Output filename', default=None, nargs='?')
+    args = parser.parse_args()
+
+    if args.operation == 'yaml2xml':
+        yaml_as_dict = get_yaml_from_filename(args.filename)
+        xml_root = create_xml_root(yaml_as_dict)
+        dict_to_xml(xml_root, yaml_as_dict)
+        filename_converted = get_output_filename(args, 'yaml', 'xml')
```
sporksmith, line 214: `--output` flag ignored?

src/tools/convert.py
```diff
@@ -0,0 +1,220 @@
...
+        save_xml(xml_root, filename_converted)
+    else:
+        xml_root = get_xml_root_from_filename(args.filename)
+        d = shadow_xml_to_dict(xml_root)
+        filename_converted = get_output_filename(args, 'xml', 'yaml')
```
sporksmith, line 219: Same here about this helper being a bit opaque

sporksmith (changes requested):
Looking good! Could you add a test? Just validating that your example files get converted to the expected result is probably sufficient.

I'd put the test in the same directory as the tool; you might need to tweak some CMakeLists.txt files for it to get picked up in the automated testing.

src/tools/convert.py
florian-vuillemot: As we do this operation in both cases I created a function `get_output_filename` to take care of the `--output` flag. I read you other comment. Should I change the name or remove this function ?

florian-vuillemot:
> Looking good! Could you add a test? Just validating that your example files get converted to the expected result is probably sufficient.
> 
> I'd put the test in the same directory as the tool; you might need to tweak some CMakeLists.txt files for it to get picked up in the automated testing.

robgjansen:
I think a good general strategy here would be to use the script to convert an XML file from an existing test case (e.g. the `udp` test) to YAML, and then back to XML again, and then run the test case using our ctest system to make sure it passes.

Here is more specific guidance on how I would write the test case:

- Create a new directory, e.g. `config`, in `src/test/config`
- Create a `src/test/config/CMakeLists.txt` file with something like

```cmake
add_test(
	NAME config-convert-xml2yaml
	COMMAND /usr/bin/env python xml2yaml --output udp.test.shadow.config.yaml ${CMAKE_SOURCE_DIR}/src/test/udp/udp.test.shadow.config.xml
)

add_test(
	NAME config-convert-yaml2xml
	COMMAND /usr/bin/env python yaml2xml  --output udp.test.shadow.config.xml udp.test.shadow.config.yaml
)
set_tests_properties(config-convert-yaml2xml PROPERTIES DEPENDS "config-convert-xml2yaml")

add_test(
        NAME config-convert-udp-shadow 
        COMMAND ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d udp.shadow.data udp.test.shadow.config.xml
)
set_tests_properties(config-convert-udp-shadow PROPERTIES DEPENDS "config-convert-yaml2xml;udp-shadow")
```

And then make sure those tests are passing. Note I did not test this but something similar should work (and maybe this code works as is!).

florian-vuillemot:
> I think a good general strategy here would be to use the script to convert an XML file from an existing test case (e.g. the `udp` test) to YAML, and then back to XML again, and then run the test case using our ctest system to make sure it passes.
> 
> Here is more specific guidance on how I would write the test case:
> 
>     * Create a new directory, e.g. `config`, in `src/test/config`
> 
>     * Create a `src/test/config/CMakeLists.txt` file with something like
> 
> 
> ```cmake
> add_test(
> 	NAME config-convert-xml2yaml
> 	COMMAND /usr/bin/env python xml2yaml --output udp.test.shadow.config.yaml ${CMAKE_SOURCE_DIR}/src/test/udp/udp.test.shadow.config.xml
> )
> 
> add_test(
> 	NAME config-convert-yaml2xml
> 	COMMAND /usr/bin/env python yaml2xml  --output udp.test.shadow.config.xml udp.test.shadow.config.yaml
> )
> set_tests_properties(config-convert-yaml2xml PROPERTIES DEPENDS "config-convert-xml2yaml")
> 
> add_test(
>         NAME config-convert-udp-shadow 
>         COMMAND ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d udp.shadow.data udp.test.shadow.config.xml
> )
> set_tests_properties(config-convert-udp-shadow PROPERTIES DEPENDS "config-convert-yaml2xml;udp-shadow")
> ```
> 
> And then make sure those tests are passing. Note I did not test this but something similar should work (and maybe this code works as is!).

Thank you this helped me a lot !

There is only one problem with this method, it's the `test-udp` dependancy that is not generating in the good place:

`attribute 'path': 'test-udp' is not a valid path to an existing regular file`

And If I add `add_shadow_exe(test-udp ${CMAKE_SOURCE_DIR}/src/test/udp/test_udp.c)` to this CMakeList.txt I obtain:

`add_executable cannot create target "test-udp" because another target with the same name already exists. `

I also tried by changing the test with a `diff` operation, and obtained an error because `node` keywords are replaced `host`...

robgjansen:
Oh, the `test-udp` dependency must not be accessible since you are not in or below the same directory where it's defined.

How about instead we just add a new test in `src/test/config/test_config.c` with something like:

```c
#include <unistd.h>
#include <stdio.h>
int main(int argc, char* argv[]) {
    char myname[HOST_NAME_MAX] = {0};
    if(gethostname(myname, HOST_NAME_MAX) == 0){
        printf("%s\n", myname);
        return 0;
    } else {
        return 1;
    }
}
```

And then add to the top of the CMakeLists.txt file the command to build `test-config` from this new source file, and also change the dependency from `test-udp` to `test-config`. You'll also want to create a new shadow.config.xml file to create a simulation with several nodes and hosts with that are running this new test plugin. See some of the other tests like `test/udp` for examples of how to structure this.

florian-vuillemot:
Hum, maybe one of the problems come from the PyYaml package.
Could you please give me the version of the python and this package ?

I found `sort_keys was added in PyYAML 5.1`

robgjansen:
We run tests using Github actions, so the default version of the yaml library that is installed will vary by the operating system. We officially support Ubuntu 16.04, 18.04, 20.04, Fedora 32, Debian 10, CentOS 7 and CentOS 8.

Is there a way to work around the sort issue, e.g., by supplying your own dumper class that sorts the keys and then passes it on to the default dumper?

sporksmith:
> We run tests using Github actions, so the default version of the yaml library that is installed will vary by the operating system. We officially support Ubuntu 16.04, 18.04, 20.04, Fedora 32, Debian 10, CentOS 7 and CentOS 8.

@robgjansen, now that we have the public dev branch, perhaps this and future PRs should target that branch, and be backported to master as-needed? In particular that will allow us to run CI on all of our supported platforms.

robgjansen:
> @robgjansen, now that we have the public dev branch, perhaps this and future PRs should target that branch, and be backported to master as-needed? In particular that will allow us to run CI on all of our supported platforms.

My thoughts:
- Running CI on all supported platforms is a definite positive.
- I'm worried that we might still want to backtrack (or sidetrack) on the approach that we're taking in the dev branch, so I'd like to gain a bit more confidence there before asking that all PRs are merged there.
- I feel bad changing the requirements on this PR.

So, maybe we let this one merge into master, and hopefully there won't be many more that come up before we decide that we've crossed the "point of no return" in the dev branch. What do you think @sporksmith?

sporksmith:
Well, a couple comments ago we asked Florian to ensure whatever library he uses is available on all of our supported platforms. The easiest way to do that is to use CI to test it on all of those platforms, which currently means targeting dev :).

Agreed about minimizing change requirements in the middle of the PR; I'm fine with letting this one continue targeting master if that's what florian prefers. In that case either he'll need to find some other way of demonstrating that it ought to work on all of our supported platforms, or we'll need to drop that requirement for now.

> I'm worried that we might still want to backtrack (or sidetrack) on the approach that we're taking in the dev branch, so I'd like to gain a bit more confidence there before asking that all PRs are merged there.

This would be an easy-enough one to backport (cherry-pick) to master after seeing it work on dev.

robgjansen:
OK. Maybe rebasing onto dev, force pushing, and changing the merge target on github will be smooth? Probably the best way of finding out is to give it a try :)

florian-vuillemot:
I hope it's ok :)

robgjansen:
For some reason the tests are not being triggered and I'm not sure why. @sporksmith ideas?

sporksmith:
> For some reason the tests are not being triggered and I'm not sure why. @sporksmith ideas?

It doesn't look like any changes were pushed since switching the PR base from master to dev; I think it needs something pushed to trigger it.

src/tools/convert.py
florian-vuillemot: @sporksmith do you have any idea of improvement before I push to run the pipeline again ?

src/test/config/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+        NAME config-convert-yaml2xml
+        COMMAND /usr/bin/env python3 ${CMAKE_SOURCE_DIR}/src/tools/convert.py yaml2xml  --output test.shadow.config.xml test.shadow.config.yaml
+)
+set_tests_properties(config-convert-yaml2xml PROPERTIES DEPENDS "config-convert-xml2yaml")
+
+add_test(
+        NAME config-convert-udp-shadow
+        COMMAND ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d convert.shadow.data test.shadow.config.xml
+)
+set_tests_properties(config-convert-udp-shadow PROPERTIES DEPENDS "config-convert-yaml2xml")
```
sporksmith, line 21: Can drop this line if we make the above changes

src/test/config/test_config.c
```diff
@@ -0,0 +1,13 @@
+#include <unistd.h>
+#include <stdio.h>
+
+
+int main(int argc, char* argv[]) {
```
sporksmith, line 5: It's not clear what the call to `gethostname` is testing. Assuming we're just trying to check that shadow parsed the config, I think we can just immediately exit. e.g. `{ return 0; }`

src/test/tools/convert_xml_to_yaml/original.xml
```diff
@@ -0,0 +1,32 @@
+<shadow stoptime="3600">
```
sporksmith, line 1: Unused? (And if not, perhaps add an xml comment explaining what this is for)

src/test/config/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+include_directories(${GLIB_INCLUDES})
+link_libraries(${GLIB_LIBRARIES})
+
+add_shadow_exe(test-config test_config.c)
+
+add_test(
+        NAME config-convert-xml2yaml
+        COMMAND /usr/bin/env python3 ${CMAKE_SOURCE_DIR}/src/tools/convert.py xml2yaml --output test.shadow.config.yaml ${CMAKE_CURRENT_SOURCE_DIR}/shadow.config.xml
```
sporksmith, line 8: I don't think you need for `/usr/bin/env`; that's just a hack for getting the path to the python executable in hash-bang lines in scripts.

src/test/config/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+)
+
+add_test(
```
sporksmith, line 11: Could use a comment briefly explaining that this is reversing the conversion of the previous test.

src/test/tools/convert_xml_to_yaml/xml2yaml.yaml
```diff
@@ -0,0 +1,40 @@
+options:
```
sporksmith, line 1: Unused?

src/test/tools/convert_xml_to_yaml/yaml2xml.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" ?>
```
sporksmith, line 1: Unused?

src/tools/convert.py
sporksmith: Resolving

src/tools/convert.py
sporksmith: Resolving

src/test/config/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+        NAME config-convert-yaml2xml
+        COMMAND /usr/bin/env python3 ${CMAKE_SOURCE_DIR}/src/tools/convert.py yaml2xml  --output test.shadow.config.xml test.shadow.config.yaml
+)
+set_tests_properties(config-convert-yaml2xml PROPERTIES DEPENDS "config-convert-xml2yaml")
+
+add_test(
```
sporksmith, line 17: Could use a comment that this is consuming the converted config from the previous test

src/test/config/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+include_directories(${GLIB_INCLUDES})
```
sporksmith, line 1: Optional, because I don't want to change the requirements on you again: :)

This sequence of tests validates that the config round-trips to *something* that shadow will parse, but wouldn't necessarily catch a lot of issues like dropping or changing parameters that the simple test here doesn't depend on.

It'd be nice to have tests that do a conversion and compare against an expected output. I see you have some extra yml and xml files still in here, that maybe are intended for such a purpose? If so, you could write such tests using `diff`, which returns a non-zero exit-code iff there is a diff. e.g.

```
COMMAND ${CMAKE_SOURCE_DIR}/src/tools/convert.py xml2yaml --output converted.yaml path/to/test-input.xml && diff path/to/test-expected-output.yaml converted.yaml)
```

src/tools/convert.py
```diff
@@ -0,0 +1,233 @@
...
+    '''
+    Write an XML element in a filename
+    '''
+    xmlstr = minidom.parseString(ET.tostring(xml_root)).toprettyxml(indent="   ")
+    with open(filename, "w", encoding='utf8') as f:
+        f.write(unescape(xmlstr, entities={'&quot;': '"'}))
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Convert a XML file into a YAML file.')
```
sporksmith, line 217: Should be a little more specific here; this doesn't convert arbitrary XML and YAML files :). Maybe 'Convert shadow config files between XML and YAML'

sporksmith (changes requested):

src/test/config/CMakeLists.txt
florian-vuillemot: I can't remove it because I need the `test.shadow.config.xml` file created by the `config-convert-yaml2xml` test

src/test/config/CMakeLists.txt
florian-vuillemot: Yes ! I totally agree with you, I think it's the best way.
Unfortunately we can't run CMake test with multiple tests easily. So I used the XML and YAML files created previously in the test

src/test/config/CMakeLists.txt
```diff
@@ -1,21 +1,38 @@
...
         COMMAND ${CMAKE_BINARY_DIR}/src/main/shadow -l debug -d convert.shadow.data test.shadow.config.xml
 )
-set_tests_properties(config-convert-udp-shadow PROPERTIES DEPENDS "config-convert-yaml2xml")
+set_tests_properties(config-convert-shadow PROPERTIES DEPENDS "config-convert-yaml2xml")
+
+# Ensure the format of the convertion is as expected from the XML to YAML convertion
+add_test(
+        NAME config-convert-yaml2xml-format
+        COMMAND diff ${CMAKE_CURRENT_SOURCE_DIR}/xml2yaml.yaml test.shadow.config.yaml
+)
```
sporksmith, line 32: Shouldn't this have a dependency on config-convert-xml2yaml?

src/test/config/CMakeLists.txt
```diff
@@ -1,21 +1,38 @@
...
+
+# Ensure the format of the convertion is as expected from the YAML to XML convertion
+add_test(
```
sporksmith, line 35: Shouldn't this have a dependency on config-convert-yaml2xml?

sporksmith (approved):
Hmm, looks like the CI workflow still isn't triggering :-/. Not sure what's going on here.

florian-vuillemot:
The build was triggered before I updated the `build_shadow.yml` file. The build failed because PyYaml is not installed on the host.

robgjansen:
I was able to get the tests to run in my repo [here](https://github.com/robgjansen/shadow/pull/2) and indeed the tests are failing on some systems because pip3 does not exist.

There is also now a conflict with the build_shadow file. You'll have to resolve the conflict, this is what I did:

```
git checkout dev
git pull # make sure i'm up to date
git checkout feature/convert_xml_file_to_yaml
git rebase dev
# fix conflixt in build_shadow
git add .github/workflows/build_shadow.yml
git rebase --continue
```

Then I hope if you `git push -f  feature/convert_xml_file_to_yaml` in your florian-vuillemot repo, the tests will run.

florian-vuillemot:
The error encountered on the ordering come from the python3.5 dict orderings. I converted my code to use OrderedDict but when I pass the yaml to the dumper PyYaml it looks to be using a dict internally and doesn't return the expected output.

Python3.5 will be supported until 2020-09-13. I think we have to consider skipping this test only for ubuntu16.04.

sporksmith:
Disabling tests on some platforms is effectively ceasing to support that platform (at least to some degree). Ubuntu 16.04 still has a ~year before EOL, so it'd be better if we didn't do that just yet.

Perhaps we can install a newer Python on that platform, following https://realpython.com/installing-python/, we can tweak the package install step for Ubuntu 16.04 to incorporate:
```
add-apt-repository ppa:deadsnakes/ppa
apt-get update
apt-get install python3.6
```

stevenengler:
A possible alternative to changing the Python version could be to use one of the pyyaml forks that are designed for ordered output, for example: https://pypi.org/project/oyaml/

sporksmith:
Hi Florian, we noticed it looks like you're forcing Python 3.6 on all platforms. We'd rather use the package-manager's default python version except where the custom Python is really needed (Ubuntu 16). We're also about to move as much of the CI logic as we can into shell scripts, so it'd be better not to use GH actions where direct shell commands could be used instead. (Alternatively, @stevenengler 's suggestion of using a different yaml package that explicitly handles ordering would work too)

Some of the other changes to the CI we're about to start merging will conflict with what you have here. I think you can either:

1. Comment out the line in the CMakelists.txt that invokes the config subdirectory for the moment, merge this PR, file an issue to get them working, and handle that issue as a follow-up PR (after our other CI-changes are in).
2. Continue working on getting the tests passing in this PR. Once they're passing either you could rebase again, or we could take it over at that point and resolve the conflicts.

florian-vuillemot:
Hum, sorry for the lack of information. Of course my goal is not to force Python3.6 on all platforms! I searched as I could for a first step that would allow me to pass tests on blocking platforms then iterate as near as possible of the actual configuration.
@stevenengler I tried with oyaml but I obtained nothing good. I also combined oyaml with OrderingDict but with python3.5 I keep my ordering error.. So I looked the code on github but didn't see anything that could help me.

@sporksmith I don't know what to do, I'm too implicated in this PR to know what is the best. In one hand I'm sure we can fix this error, but in other hand this issue has been opened for 2months, looks good and we have to create special condition in the GH action just to handle this nice to have test.

sporksmith:
Sorry, I know this must be frustrating!

> Hum, sorry for the lack of information. Of course my goal is not to force Python3.6 on all platforms! I searched as I could for a first step that would allow me to pass tests on blocking platforms then iterate as near as possible of the actual configuration.

Makes sense; just wanted to be be sure you weren't working solidifying it working that way.

Do you know what the issue is currently on Ubuntu 16? Is installing Python3.6 not working, or is Python3.6 not fixing the issue?

> @sporksmith I don't know what to do, I'm too implicated in this PR to know what is the best. In one hand I'm sure we can fix this error, but in other hand this issue has been opened for 2months, looks good and we have to create special condition in the GH action just to handle this nice to have test.

Thinking through this a bit more - the issue on Ubuntu 16 is that ordering isn't preserved or predictable (and therefore difficult to test the conversion details), but we expect that the conversions are still valid? If that's the case I think it'd be ok to [exclude](https://github.com/shadow/shadow/blob/dev/.github/workflows/build_shadow.yml#L141) the diff-tests on Ubuntu 16 as long as the other tests still pass (i.e. that the conversions run without error and that shadow accepts the converted result).

florian-vuillemot:
hum seems to work now. But I have no idea what I fixed during the rebase that allowed the test to be correct now...

ci/container_scripts/install_deps.sh
```diff
@@ -30,16 +31,26 @@ RPM_PACKAGES="
...
   yum-utils
+  diffutils
   "
 
 case "$CONTAINER" in
     ubuntu:*|debian:*)
         sed -i '/deb-src/s/^# //' /etc/apt/sources.list
         DEBIAN_FRONTEND=noninteractive apt-get update
         DEBIAN_FRONTEND=noninteractive apt-get install -y $APT_PACKAGES
+        if [[ `python3 --version` == *" 3.5"* ]]; then
```
sporksmith, line 46: Please leave a comment that this is for ordered dictionaries for `src/tools/convert.py`

sporksmith (approved):
Hurray! Looks good, and thanks for getting it rebased again. Just a couple small tweaks, please, and then I'll go ahead and merge

robgjansen (approved):
This is great! Thanks for sticking with us and getting this passing on all of our supported platforms! 😄

PR #804, opened 2020-05-20
Call gettimeofday from within shadow logger

sporksmith:
Reproduce integration test breakage from #801 by simply calling `gettimeofday` from within the shadow logger (and not doing anything with the result)


robgjansen:
I wonder if #805 has anything to do with this issue.

sporksmith:
I wouldn't *think* so. Whether the second parameter to gettimeofday is a pointer to a struct or pointer to void, the ABI for passing the pointer should be the same. It is a weird coincidence, though!

PR #809, opened 2020-05-28
Reinstate release-tagging instructions

### LINKED ISSUE #772: Create a release playbook
Broken out of https://github.com/shadow/shadow/issues/733#issuecomment-596605275. I think this belongs in a playbook-style documentation for core developers rather than the more general CONTRIBUTING file.

sporksmith:
Accidentally deleted instead of moving in
efd653a204d28499d02c92a478d0ad75299d17b6

Closes #772

### DIFF
--- a/docs/maintainer-playbook.md
+++ b/docs/maintainer-playbook.md
@@ -2,20 +2,27 @@
 
 ### Tagging Shadow releases
 
+We use [Semantic Versioning](https://semver.org/), and increment version
+numbers with the [bumpversion](https://pypi.org/project/bumpversion/) tool.
+
 The following commands can be used to tag a new version of Shadow, after which an
 archive will be available on github's [releases page](https://github.com/shadow/shadow/releases).
 
 ```bash
 git checkout master
-git tag -s v1.10.0
-git push origin v1.10.0
-```
-Our releases will then be tagged off of the master branch. Once tagged, a signed archive of a release can be created like this:
 
-```bash
-git archive --prefix=shadow-v1.10.0/ --format=tar v1.10.0 | gzip > shadow-v1.10.0.tar.gz
-gpg -a -b shadow-v1.10.0.tar.gz
-gpg --verify shadow-v1.10.0.tar.gz.asc
+# Bump the patch, minor, or major version number, and commit the change.
+bumpversion <patch|minor|major> --commit
+
+# Get the new version number.
+VERSION=`awk -F "=" '/current_version/ {print $2}' .bumpversion.cfg | tr -d ' '`
+
+# Create a signed tag for the new version.
+git tag -s v$VERSION
+
+# Push to GitHub.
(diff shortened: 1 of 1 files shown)


docs/maintainer-playbook.md
```diff
@@ -1,3 +1,21 @@
...
+git tag -s v1.10.0
+git push origin v1.10.0
+```
+Our releases will then be tagged off of the master branch. Once tagged, a signed archive of a release can be created like this:
+
+```bash
+git archive --prefix=shadow-v1.10.0/ --format=tar v1.10.0 | gzip > shadow-v1.10.0.tar.gz
+gpg -a -b shadow-v1.10.0.tar.gz
+gpg --verify shadow-v1.10.0.tar.gz.asc
+```
```
robgjansen, line 19: These instructions were first written back before Github automatically included the release assets on tagged releases. I don't think we need these anymore since it's no longer part of our release process.

docs/maintainer-playbook.md
```diff
@@ -1,3 +1,21 @@
...
+### Tagging Shadow releases
+
+The following commands can be used to tag a new version of Shadow, after which an
+archive will be available on github's [releases page](https://github.com/shadow/shadow/releases).
+
+```bash
+git checkout master
+git tag -s v1.10.0
+git push origin v1.10.0
+```
```
robgjansen, line 12: I recently started using the python module `bumpversion` for automatically bumping the version number. We could also make it automatically tag the release. (See the `.bumpversion.cfg` file in the root.)

Should we continue to use bumpversion?

It might still be good to keep the above commands as an alternative method, but if we don't use bumpversion we'll have to update the version number in the toplevel CMakeLists.txt file manually.

docs/maintainer-playbook.md
sporksmith: Done

docs/maintainer-playbook.md
sporksmith: I haven't used the bumpversion tool before, but poked at it a bit and took a stab at updating the instructions.

I think it makes sense to use the --commit flag with bumpversion to automatically generate a commit. I see we could also change the config to make that the default, but I think I prefer having to explicitly ask for it on the command-line.

bumpversion also has a -t option to automatically tag the new version, but it wouldn't be a signed tag. Assuming that matters (does it?), I left the instructions to create a signed tag instead. Unfortunately that means a slightly cludgey bit of shell script to grab the current version number. Surprisingly bumpversion doesn't seem to have an option to just grab and print the current version string.

sporksmith:
Modified to use bumpversion to create the tag.

Unfortunately we still need the bit of hacky shell to get the tag-string to push back up to GH. At least now we can count on bumpversion to create the tag name we expect. If the hacky bit of shell fails to construct the right tag name, the `git push` should also fail.

PR #823, opened 2020-06-04
Fix unquoted string containing colon

robgjansen:
Whoops! shadow/shadow@57d1ae3f1752e2c25597bdbee78458fc3e87449b introduced an error that caused the bug and question issue templates to not be presented upon creating a new issue. I think it was an error with missing quotes for a string that contains a colon.


robgjansen:
I think the only way to check this without merging would be to merge it into the master branch of a fork on Github.

sporksmith:
Ah yes, tricksy yaml letting unquoted strings work until they don't :)

PR #861, opened 2020-06-30
Re-enable some CI tests for other distros

stevenengler:
Re-enables CI tests for ubuntu:20.04, debian:10-slim, fedora:32, and centos:8. The 'findutils' package is for our cmake rust unit tests which require the 'find' tool. This PR also stops running linting tests with Python 2.7.

The tests do not currently complete on ubuntu:16.04 (shadow/shadow#864) or when using clang (shadow/shadow#865).

### DIFF
--- a/.github/workflows/build_shadow.yml
+++ b/.github/workflows/build_shadow.yml
@@ -5,7 +5,7 @@ name: Tests
 
 on:
   push:
-    branches: [master, 'rebases/**']
+    branches: [dev]
     paths-ignore:
       - '**.md'
       - 'LICENSE'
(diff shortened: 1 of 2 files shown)


.github/workflows/build_shadow.yml
```diff
@@ -22,24 +19,24 @@ jobs:
       options: '--shm-size="1g"'
     strategy:
       matrix:
-        container: ['ubuntu:18.04', 'centos:7'] #, 'ubuntu:16.04', 'ubuntu:20.04', 'debian:10-slim', 'fedora:32', 'centos:8']
-        cc: ['gcc'] #, 'clang']
+        container: ['ubuntu:18.04', 'ubuntu:20.04', 'debian:10-slim', 'fedora:32', 'centos:7', 'centos:8'] # 'ubuntu:16.04'
```
sporksmith, line 22: Could you add a comment with a link to the issue for ubuntu 16.04?

.github/workflows/build_shadow.yml
```diff
@@ -22,24 +19,24 @@ jobs:
...
+        cc: ['gcc'] # 'clang'
```
sporksmith, line 23: Can clang be enabled too?

.github/workflows/build_shadow.yml
stevenengler: Added

.github/workflows/build_shadow.yml
stevenengler: It doesn't yet build with clang for some platforms.

PR #866, opened 2020-07-01
Added socket() tests

stevenengler:
Added tests for `socket(...)` and `syscall(SYS_socket, ...)`.

There are also "extended tests", which are tests that fail as they are unsupported by Shadow, but they help us keep track of missing features in Shadow.


src/test/socket/socket/test_socket.rs
```diff
@@ -0,0 +1,501 @@
...
+        //libc::AF_PPPOX,
+        //libc::AF_LLC,
+        //libc::AF_IB,
+        //libc::AF_MPLS,
+        //libc::AF_CAN,
+        //libc::AF_TIPC,
+        //libc::AF_BLUETOOTH,
+        //libc::AF_ALG,
+        //libc::AF_VSOCK,
+        //libc::AF_XDP,
```
robgjansen, line 309: It's unclear why some of these are commented out. For example, are we supposed to uncomment them later if we add support? I suggest adding a comment to explain why.

src/test/socket/socket/test_socket.rs
```diff
@@ -0,0 +1,501 @@
...
+
+        print!("Testing {}...", socket_fn.format(args));
+
+        num_performed += 1;
+
+        let fd;
+        match check_socket_call(args, *socket_fn, should_error, expected_errno) {
+            // the test passed
+            Ok(result_fd) => {
+                println!(" ✓");
```
robgjansen, line 401: Neat check mark :)

src/test/socket/socket/socket.shadow.config.xml
```diff
@@ -0,0 +1,27 @@
...
+      <data key="d3">50.0</data>
+      <data key="d4">0.0</data>
+    </edge>
+  </graph>
+</graphml>
+]]></topology>
+  <kill time="5"/>
+  <plugin id="testsocket" path="../../target/debug/test_socket"/>
+  <node id="testnode" quantity="1">
+    <application plugin="testsocket" starttime="1" arguments="--shadow-passing"/>
```
robgjansen, line 24: 👍

src/test/socket/socket/test_socket.rs
```diff
@@ -0,0 +1,501 @@
...
+        SocketFn::Socket => unsafe {
+            libc::socket(args.domain, args.sock_type | args.flag, args.protocol)
+        },
+        SocketFn::Syscall => unsafe {
+            libc::syscall(
+                libc::SYS_socket,
+                args.domain,
+                args.sock_type | args.flag,
+                args.protocol,
+            ) as i32
```
robgjansen, line 462: I'm not sure what `as i32` does, but doesn't `syscall` return an i64 type on our supported platforms?

src/test/socket/socket/test_socket.rs
stevenengler: This was just to easily see what we aren't testing (I don't think Shadow will ever support `AF_IPX` for example), but are still technically supported by the socket system call. I think I'll remove these comments though as this really isn't that useful.

src/test/socket/socket/test_socket.rs
stevenengler: I'm sure this will break someone's editor at some point :)

src/test/socket/socket/test_socket.rs
stevenengler: `syscall` returns an i64, but calling it with `SYS_socket` should never return greater than 2^31-1. I'll add a check first before downcasting it just to be safe.

stevenengler:
Waiting on # 803 to avoid adding merge conflicts.

PR #874, opened 2020-07-07
Upgrade to CMake 3.2

stevenengler:
CMake >= 3.2 is available by default on all supported platforms except for CentOS 7, which requires the `cmake3` package from EPEL and for `cmake` to be linked to `cmake` using a method such as https://stackoverflow.com/a/48842999.

> I'm a bit worried about requiring CMake v3.2 because it might not be available in all of our supported distros. I see that you have a way to make it work in CentOS 7, but I don't know if it's easy to get that working all older distros.

All of the other supported distros support CMake 3.2 (for example Ubuntu 16.04 is 3.5, Fedora 32 is 3.17, CentOS 8 is 3.11, and Debian 10 is 3.13). CentOS 7 is the oldest distro that we support, which uses CMake 2.8.

> It would require an extra step that we would need to ask users to perform during setup.

I agree this isn't ideal. If we wanted to not require this step, we could modify the setup script to search for cmake3 and ctest3, and use those instead if found.

> 3.2 is mainly required to get the build working with Ninja. Is it worth it?

I don't use Ninja, but I think the main advantage of ninja vs make is the speed. But Shadow doesn't take long to compile, so I'm not sure if the advantage is significant here. On the other hand, we get a lot of new CMake features by upgrading to 3.2, which might make things easier for us going forward.

### DIFF
--- a/.github/workflows/build_shadow.yml
+++ b/.github/workflows/build_shadow.yml
@@ -92,9 +92,15 @@ jobs:
       - name: Install Dependencies (yum-centos-7)
         if: ${{ matrix.package_manager == 'yum-centos-7' }}
         run: |
-          # clang on centos 7 requires gcc and gcc-c++ (libgcc is not sufficient)
+          # Clang on centos 7 requires gcc and gcc-c++ (libgcc is not sufficient).
+          # Must also link cmake to cmake3 as an alternative.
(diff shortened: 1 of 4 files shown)


stevenengler:
@robgjansen If you're still unsure about this, we can leave it unmerged. But these changes are now in the shadow repository so we don't need to worry about losing them.

stevenengler:
Waiting on # 803 to avoid adding merge conflicts.

PR #875, opened 2020-07-08
Dockerize CI

sporksmith:


sporksmith:
Looks like a transient error uploading to the coverage server; it worked fine on previous pushes (and helped me catch that I'd accidentally left a test disabled!).

Leaving it as-is for the moment, but open to ideas in general. There's a button to rerun the whole CI matrix again, but AFAIK no way to run just the coverage config again. Maybe we should not fail the workflow if that step fails?

ci/run_one.sh
```diff
@@ -0,0 +1,49 @@
...
+#
+# Example:
+#   sudo CONTAINER=centos:8 CC=clang BUILDTYPE=debug ci/run_one.sh
+
+set -euo pipefail
+
+TAG="shadow:${CONTAINER/:/-}-$CC-$BUILDTYPE"
+echo "Running $TAG"
+
+# Build and tag a Docker image witht he given configuration.
```
robgjansen, line 17: `with the`

robgjansen:
> Maybe we should not fail the workflow if that step fails?

That's probably a good call - it appears there have been some reliability issues that are still being worked through. If this frequently fails, we'll have to consider a workaround.
https://github.com/codecov/codecov-action/issues/43

ci/run_one.sh
```diff
@@ -0,0 +1,49 @@
...
+
+RUN mkdir /root/shadow
+COPY . /root/shadow
+WORKDIR /root/shadow
+
+RUN . ci/container_scripts/build_and_install.sh
+EOF
+
+# Run the tests inside the image we just built.
+docker run --shm-size=1g $TAG /bin/bash ci/container_scripts/test.sh
```
stevenengler, line 40: Should we remove the container here when we're done? For debugging failed tests it might be useful to leave them, but the run_all script will create 29 new containers on each run, so for disk space purposes I think we should delete at least all passing containers.

ci/run_one.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# Build and tag a Docker image witht he given configuration.
+docker build -t "$TAG" -f- . <<EOF
+FROM $CONTAINER
+
+ENV CONTAINER $CONTAINER
+SHELL ["/bin/bash", "-c"]
+
+COPY ci/container_scripts/install_deps.sh /root/install_deps.sh
+RUN . /root/install_deps.sh
```
stevenengler, line 25: For these steps, is there an advantage to sourcing them rather than running them? I don't think the environment variables carry over between RUNs, so it might be good to remove the `.` from the RUN steps.

ci/container_scripts/install_deps.sh
```diff
@@ -0,0 +1,76 @@
...
+  xz-devel
+  yum-utils
+  "
+
+case "$CONTAINER" in
+    ubuntu:*|debian:*)
+        sed -i '/deb-src/s/^# //' /etc/apt/sources.list
+        DEBIAN_FRONTEND=noninteractive apt-get update
+        DEBIAN_FRONTEND=noninteractive apt-get install -y $APT_PACKAGES
+        ;;
```
stevenengler, line 43: At the end of each case (in this file and install_extra_deps), I think it would be good to clear the package repository cache (with `rm -rf /var/lib/apt/lists/*` or `dnf clean all`) to avoid saving stale repository information into the layers. This will add ~5-10 seconds to the build time, but I think is worthwhile to avoid problems where we rebuild an image from Docker's cached layers.

ci/run_one.sh
sporksmith: Added a command-line option

stevenengler: I'm not clear on how this works. You use `docker run` to make a container based on the image named `$TAG`, but then later it runs `docker rm $TAG`. Since `$TAG` is the image name and you don't assign a name to the container (so it's randomly generated), I think `docker rm` should fail since `$TAG` is not a valid container name?

ci/run_one.sh
stevenengler: Ah okay I didn't think of that. Feel free to run them however you think is best, not making them +x is a good reason.

ci/container_scripts/install_deps.sh
stevenengler: So a possible problem is you build the image, and in step `install_deps` the package manager creates a package cache, uses it, and then uses that same cache in the `install_extra_deps`. This is fine (other than the increased disk space in the image and its layers) until you edit something in the `install_extra_deps` step which will rebuild the image starting from that step. Now for apt will continue using the package cache saved into the image from the previous step, but this cache may be very old and `apt install` for example will fail. Usually it's recommended to delete the package cache at the end of each 'run' step, and re-download it at the start of the next step if needed. This might explain it better: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

ci/run_one.sh
sporksmith: Good catch. Fixed! Per offline discussion, containers are now removed on exit with `--rm`

ci/container_scripts/install_deps.sh
sporksmith: I see what you mean, but I'm still leaning towards the extra overhead of downloading and deleting the package cache twice being worth it. Having to remember to use `-n` / `--no-cache` after adding a new dependency doesn't seem too bad, and is hopefully a fairly obvious solution when there are errors about the package not existing.

PR #881, opened 2020-07-10
Don't fail CI tests on Codecov upload failure

stevenengler:
~As our coverage was being run in `shadow/src`, codecov didn't know how to match the files in the repository with the files in the coverage report. This PR tells codecov how to link them.~

Set the Codecov upload action to not fail the CI test if the upload fails.

Also set the dev branch as the main branch, but I don't think this will activate until we add it to the master branch as well.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -64,7 +64,7 @@ jobs:
           name: coverage 
           file: grcov.lcov.txt
           flags: tests
-          fail_ci_if_error: true
+          fail_ci_if_error: false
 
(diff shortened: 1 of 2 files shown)


stevenengler:
Something weird has happened that I don't understand, and now after I've made this PR, codecov for me is able to link files even on other PRs. So I may close this PR now that it magically started working.

robgjansen:
I'm not sure exactly how, but they scan the repo so it might only work because you had the codecov.yml here and it happened to scan it. It might be nice to merge this anyway since there might be other options we want to tweak in the future.

PR #884, opened 2020-07-13
Update installing dependencies in doc

florian-vuillemot:
All cases are not covered, for example, `yum install cargo` is not handled on centos7 but works on centos8. I think it is sufficient to understand what the user has to install.

### DIFF
--- a/docs/1.1-Shadow.md
+++ b/docs/1.1-Shadow.md
@@ -56,8 +56,10 @@ sudo debuginfo-install glibc
 sudo yum install -y \
     python3-numpy \
     python3-lxml \
+    python3-matplotlib \
     python3-networkx \
-    python3-scipy
+    python3-scipy \
+    python3-yaml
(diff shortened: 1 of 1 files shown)


florian-vuillemot:
I removed python3-matplotlib dependencies because `yum` can't find it. Could we maybe install the python3-pip dependencies and install dependencies with pip3 like in tests?

sporksmith:
> I removed python3-matplotlib dependencies because `yum` can't find it. Could we maybe install the python3-pip dependencies and install dependencies with pip3 like in tests?

On what platform wasn't it found? It might make sense to leave it there if it's present on some platforms, with the same reasoning as for cargo.

florian-vuillemot:
The package `python3-matplotlib` was not found on centos:7 and centos:8, but found on fedora:32.
The package `python3-yaml` was found on centos:7 and fedora:32 but not on centos:8

PR #885, opened 2020-07-13
Added a shadow config template for tests

stevenengler:
We don't need to merge this, but it's an idea for a way to avoid duplicating the shadow config xml for each test. Do you have any thoughts about this @sporksmith?


sporksmith:
> We don't need to merge this, but it's an idea for a way to avoid duplicating the shadow config xml for each test. Do you have any thoughts about this @sporksmith?

It's certainly tempting to deduplicate these configs! I think this template would cover quite a lot of them.

OTOH I'm generally hesitant to introduce layers of metaprogramming in tests. When a given test fails it makes it more difficult to understand what's happening. e.g. if there's a config-related failure, the developer would need to understand that the templating is happening, figure out how to look at the actual generated config, maybe understand/debug the m4 macro processing, etc.

I'd lean slightly towards *not* doing this, since the duplication generally isn't a problem in practice.

stevenengler:
> > We don't need to merge this, but it's an idea for a way to avoid duplicating the shadow config xml for each test. Do you have any thoughts about this @sporksmith?
> 
> It's certainly tempting to deduplicate these configs! I think this template would cover quite a lot of them.
> 
> OTOH I'm generally hesitant to introduce layers of metaprogramming in tests. When a given test fails it makes it more difficult to understand what's happening. e.g. if there's a config-related failure, the developer would need to understand that the templating is happening, figure out how to look at the actual generated config, maybe understand/debug the m4 macro processing, etc.
> 
> I'd lean slightly towards _not_ doing this, since the duplication generally isn't a problem in practice.

Thanks, that was my concern too, so I'll close this and if we decide to revisit this sometime in the future we'll know there is an example PR somewhere.

PR #886, opened 2020-07-13
Adds IntervalMap: map from a range to a value.

sporksmith:
This is intended to help keep track of plugins' memory mappings. Nothing
uses it yet, but since it's fairly general it probably makes sense to
split off and merge separately.

### DIFF
--- a/src/main/utility/interval_map.rs
+++ b/src/main/utility/interval_map.rs
@@ -24,13 +24,30 @@ impl Interval {
 /// Describes modifications of an IntervalMap after overwriting an interval.
 #[derive(PartialEq, Eq, Debug)]
 pub enum Mutation<V> {
-    /// (Original interval, new begin)
+    ///       b     e
+    /// from: |---v-|
+    /// to:     |-v-|
+    ///         b'
+    /// Contains: ((b,e), b')
     ModifiedBegin(Interval, usize),
-    /// (Original interval, new end)
+    ///       b     e
+    /// from: |---v-|
+    /// to:   |-v-|
+    ///           e'
+    /// Contains: ((b,e), e')
     ModifiedEnd(Interval, usize),
-    /// (Original interval, new lower interval, new higher interval)
+    ///        b             e
+    /// from:  |-----v-------|
+    /// to:    |-v--|  |--v--|
+    ///        b   e'  b'    e
+    ///
+    /// Contains: ((b,e), (b,e'), (b',e)
     Split(Interval, Interval, Interval),
-    /// (Removed interval, its value)
+    ///       b     e
+    /// from: |---v-|
+    /// to:   
+    ///
+    /// Contains: (b,e), v
     Removed(Interval, V),
 }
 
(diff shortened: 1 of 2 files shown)


src/main/utility/interval_map.rs
```diff
@@ -0,0 +1,610 @@
...
+
+                // Truncate the existing interval.
+                self.ends[i] = new1.end();
+
+                // Create a new interval, starting after the insertion interval.
+                begins_insertions.push(new2.begin());
+                ends_insertions.push(new2.end());
+                vals_insertions.push(self.vals[i].clone());
+                mutations.push(Mutation::Split(old, new1, new2));
+            } else {
```
stevenengler, line 161: A nitpick, but I think it would help readability if the shorter case came first, sort of as an introduction to all of the variables.

src/main/utility/interval_map.rs
```diff
@@ -0,0 +1,610 @@
...
+    pub fn get(&self, x: usize) -> Option<(Interval, &V)> {
+        match self.get_index(x) {
+            None => None,
+            Some(i) => Some(self.item_at(i)),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
stevenengler, line 247: There's one other test I think would be useful (unless I missed it). One where it adds an interval with a begin that matches a begin from another interval. Something like `m.insert(1, 5, "i1".to_string())`, and then `m.insert(1, 3, "i2".to_string())`. This would make sure the indexing works properly when the binary search returns the exact `splice_start`.

src/main/utility/interval_map.rs
sporksmith: Done

src/main/utility/interval_map.rs
```diff
@@ -0,0 +1,610 @@
...
+            }
+        }
+
+        // Find the end of the splice interval.
+        // TODO: Maybe do binary search here.
+        while splice_end < self.ends.len() && self.ends[splice_end] <= end {
+            splice_end += 1
+        }
+
+        // Check whether we need to clip the beginning splice_end's interval.
```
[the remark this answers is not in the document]
sporksmith: Done

src/main/utility/interval_map.rs
```diff
@@ -0,0 +1,610 @@
...
+            .collect();
+        let dropped_ends: Vec<_> = self
+            .ends
+            .splice(splice_start..splice_end, ends_insertions)
+            .collect();
+        {
+            // We use the dropped_vals iterator directly here to avoid extra copies.
+            // This is in a new scope to limit the lifetime of the mutable borrow from self.vals.
+            let mut dropped_vals = self.vals.splice(splice_start..splice_end, vals_insertions);
+            for i in 0..dropped_begins.len() {
```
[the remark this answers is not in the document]
sporksmith: Per offline discussion, let's hold off a bit on pulling in that dependency. I was able to simplify this a bit with the same idea though.

src/main/utility/interval_map.rs
```diff
@@ -0,0 +1,610 @@
...
+        mutations
+    }
+
+    // Returns the item at the given index.
+    fn item_at(&self, i: usize) -> (Interval, &V) {
+        (Interval::new(self.begins[i], self.ends[i]), &self.vals[i])
+    }
+
+    // Returns the index of the interval containing `x`.
+    fn get_index(&self, x: usize) -> Option<usize> {
```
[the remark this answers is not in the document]
sporksmith: I think of find* as typically being a linear search. This is a log_2(n) search, analagous to BTreeMap::get.

sporksmith:
> Looks good to me! It took me some time to work through the `splice()` function and the various indexes and interval positions, but I can't think of a way to simplify it further. Some sort of example visualization might help, for example maybe:
> 
> ```rust
> /*
>                  begin   end
> new interval:     |--val--|
> existing map:  |----|  |-----|
>                           ^
>                           └── splice_start/splice_end = 1
> resulting map: |--|--val--|--|
> */
> ```

Great idea! Done

src/main/utility/interval_map.rs
sporksmith: Good idea - done

PR #895, opened 2020-07-19
Allow to read config from stdin

florian-vuillemot:


florian-vuillemot:
@sporksmith @robgjansen do you have have an idea to improve the `utility_getFileContent` ? Waiting `"</shadow>"` string is not beautiful. Moreover I'm really surprised that I have to re code this function, do you know another way ?

sporksmith:
> @sporksmith @robgjansen do you have have an idea to improve the `utility_getFileContent` ? Waiting `"</shadow>"` string is not beautiful. Moreover I'm really surprised that I have to re code this function, do you know another way ?

Typically for this kind of usage you want to read until end-of-file, rather than looking for a specific "end" content pattern.

You might be able to use `scanf` with the `m` or `a` modifier to tell it to allocate memory for you. Unfortunately the `s` specifier stops at white-space, but you might be able to use an "all-character" character set. e.g. something like. `scanf("%a[^]", &buf)`

sporksmith:
Alternatively we might be able to just pass "/dev/stdin" as the filename to the existing helper function, but I'm not sure how portable that is.

florian-vuillemot:
In case we read until the EOF we have to consider that if a user enters the schema by copy/past in stdin, he also has to send a CTRL+D. It's cleaner in the code but I don't find it really "user friendly" even if I will be surprised if someone does a copy/past of a config file.

sporksmith:
Reading until EOF is the convention used by most command-line utilities that read from stdin. I agree it can be a little confusing for someone new to the command-line, but I don't think it's worth breaking convention.

Typically the user would be using shell redirection or a pipe to send the contents in anyway, rather than copy-pasting. e.g. `convert yml2xml config.xml | shadow -`

florian-vuillemot:
It seems that `scanf("%a[^]", &buf)` is not allowed by the compiler. I didn't find any easy and clean way to do the job with `scanf`.. So I let `fgets` read until EOF.

I will add tests now except if you have any other idea for cleaning. :)

src/main/core/master.c
```diff
@@ -175,7 +175,13 @@ static gboolean _master_loadConfiguration(Master* master) {
             return FALSE;
         }
 
-        file = utility_getFileContents(fileName->str);
+        // Read config from file or stdin
+        if (0 == g_strcmp0("-", fileName->str)) {
+            file = utility_getFileContent(stdin);
```
sporksmith, line 180: The names of this new helper function should give more of a clue of how it's different from the other. Maybe `utility_getFileStreamContents`

src/main/utility/utility.c
```diff
@@ -356,3 +356,16 @@ struct timespec utility_timespecFromMillis(int64_t millis) {
         .tv_nsec = (millis % 1000) * 1000000, // ms to ns
     };
 }
+
+
+GString* utility_getFileContent(FILE *stream) {
+    char buf[BUFSIZ];
```
sporksmith, line 362: I don't think we want to have a hard-coded buffer size. If we can't find some existing utility function that will do the buffer allocation for us, I think I'd lean towards just passing "/dev/stdin" to `utility_getFileContents` for now.

src/main/utility/utility.c
```diff
@@ -356,3 +356,16 @@ struct timespec utility_timespecFromMillis(int64_t millis) {
...
+    GString *content = g_string_new(NULL);
+
+    memset(buf, 0, sizeof(buf));
+    while (NULL != fgets(buf, sizeof(buf), stream)) {
+        g_string_append(content, buf);
+    }
+
+    return content;
```
sporksmith, line 370: This string is still empty.

src/main/utility/utility.c
robgjansen: Agreed with using "/dev/stdin". (The portability question was [asked on stackoverflow](https://unix.stackexchange.com/questions/36403/how-portable-are-dev-stdin-dev-stdout-and-dev-stderr).)

src/main/utility/utility.c
florian-vuillemot: In which case ? I didn't obtain this kind of error in my tests.

sporksmith: Ah I'd misread the code. This implementation does look correct, though I'd still lean towards just using "/dev/stdin" to the existing helper rather than adding a new one.

src/main/utility/utility.c
florian-vuillemot: Waoo I find totally amazing that we can open "/dev/stdin" :0
Really good advice thank you !

src/main/core/master.c
```diff
@@ -175,7 +175,13 @@ static gboolean _master_loadConfiguration(Master* master) {
             return FALSE;
         }
 
-        file = utility_getFileContents(fileName->str);
+        // Read config from file or stdin
+        if (0 == g_strcmp0("-", fileName->str)) {
+            file = utility_getFileContents("/dev/stdin");
```
florian-vuillemot, line 180: Maybe could I remove this condition by relocating `/dev/stdin` string in the `master->options` hydration ?

florian-vuillemot:
I would like to add some tests. I thought in the `config` directory by adding 2 sub directories in it (convertor, read_from_stdin). What do you think?

sporksmith:
> I would like to add some tests. I thought in the `config` directory by adding 2 sub directories in it (convertor, read_from_stdin). What do you think?

Is the idea to move the tests currently in the config directory to a new subdirectory `convert`, and add test(s) for this new functionality in `read_from_stdin`? Sounds good to me.

src/main/core/master.c
sporksmith: I don't follow what you're suggesting, but this looks fine as is.

sporksmith:
@robgjansen LMK whether you want to review; otherwise I can go ahead and (squash) and merge

PR #896, opened 2020-07-21
Added tests for the listen() call

### LINKED ISSUE #839: Thoroughly test the listen() syscall
See #832 for an overview of our approach.

stevenengler:
Closes #839.

There is some functionality that's duplicated with other tests, but I'm still waiting until I have a few more tests done to have a bigger picture about what functionality to move into a helper library.

### DIFF
--- a/src/test/socket/listen/test_listen.rs
+++ b/src/test/socket/listen/test_listen.rs
@@ -1,0 +1,434 @@
+/*
+ * The Shadow Simulator
+ * See LICENSE for licensing information
+ */
+
+struct ListenArguments {
+    fd: libc::c_int,
+    backlog: libc::c_int,
+}
+
+#[derive(Debug, Copy, Clone)]
+struct BindAddress {
+    address: libc::in_addr_t,
+    port: libc::in_port_t,
+}
+
+/// A boxed function to run as a test.
+type TestFn = Box<dyn Fn() -> Result<(), String>>;
+
+fn main() {
+    // should we run only tests that shadow supports
+    let run_only_passing_tests = match std::env::args().position(|x| x == "--shadow-passing") {
+        Some(_) => true,
+        None => false,
+    };
+    // should we summarize the results rather than exit on a failed test
+    let summarize = match std::env::args().position(|x| x == "--summarize") {
+        Some(_) => true,
+        None => false,
+    };
+
+    let tests = if run_only_passing_tests {
+        get_passing_tests()
+    } else {
+        get_all_tests()
+    };
+
+    if let Err(_) = run_tests(tests.iter(), summarize) {
+        println!("Failed.");
+        std::process::exit(1);
+    }
+
+    println!("Success.");
+}
+
+fn get_passing_tests() -> std::collections::BTreeMap<String, TestFn> {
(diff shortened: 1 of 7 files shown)


src/test/socket/listen/test_listen.rs
```diff
@@ -0,0 +1,434 @@
...
+type TestFn = Box<dyn Fn() -> Result<(), String>>;
+
+fn main() {
+    // should we run only tests that shadow supports
+    let run_only_passing_tests = match std::env::args().position(|x| x == "--shadow-passing") {
+        Some(_) => true,
+        None => false,
+    };
+    // should we summarize the results rather than exit on a failed test
+    let summarize = match std::env::args().position(|x| x == "--summarize") {
```
sporksmith, line 27: Consider using `any` instead

src/test/socket/listen/test_listen.rs
```diff
@@ -0,0 +1,434 @@
...
+        None => false,
+    };
+
+    let tests = if run_only_passing_tests {
+        get_passing_tests()
+    } else {
+        get_all_tests()
+    };
+
+    if let Err(_) = run_tests(tests.iter(), summarize) {
```
sporksmith, line 38: I haven't tried it, but I saw someone mentioning in passing that `main` can return a Result, in which case you just return the result from this call. (Not sure what the output would look like on error though; explicit handling like you have it might still be better)

src/test/socket/listen/test_listen.rs
```diff
@@ -0,0 +1,434 @@
...
+
+    // optionally bind to an address before listening
+    let bind_addresses = [
+        None,
+        Some(BindAddress {
+            address: libc::INADDR_LOOPBACK,
+            port: 0,
+        }),
+        Some(BindAddress {
+            address: libc::INADDR_ANY,
```
sporksmith, line 65: Consider storing in network order.

I see you do switch the byte order just before the bind call, but the convention I (and I think most) use is that a value of type `in_addr_t` should always be in network order.

src/test/socket/listen/test_listen.rs
```diff
@@ -0,0 +1,434 @@
...
+    address: libc::in_addr_t,
+    port: libc::in_port_t,
+}
+
+/// A boxed function to run as a test.
+type TestFn = Box<dyn Fn() -> Result<(), String>>;
+
+fn main() {
+    // should we run only tests that shadow supports
+    let run_only_passing_tests = match std::env::args().position(|x| x == "--shadow-passing") {
```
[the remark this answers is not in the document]
stevenengler: Done

src/test/socket/listen/test_listen.rs
stevenengler: Done

src/test/socket/listen/test_listen.rs
stevenengler: I tried this again and the output doesn't look too bad, so I made this change.

src/test/socket/listen/test_listen.rs
stevenengler: Done.

Issue #893, opened 2020-07-16
Support per-message cpu load in phold test


robgjansen:
Whenever a node receives a message, it currently sends a response immediately. Instead, we want to be able to add a configurable amount of CPU load before sending the response in order to test more types of workloads.

PR #897, opened 2020-07-21
Add cpu load generator in phold benchmark test

robgjansen:
refs #893

### DIFF
--- a/src/test/phold/test_phold.c
+++ b/src/test/phold/test_phold.c
@@ -321,9 +321,9 @@ static inline void _phold_logHeartbeatMessage(PHold* phold) {
 static void _phold_generateCPULoad(PHold* phold) {
     PHOLD_ASSERT(phold);
     // this is volatile to prevent the compiler from optimizing out the loop
-    double volatile result = 0;
+    guint64 volatile result = 0;
     for (guint64 i = 0; i < phold->cpuload; i++) {
-        result = sqrt((double)i);
+        result = i;
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/test/phold/test_phold.c
```diff
@@ -317,6 +318,15 @@ static inline void _phold_logHeartbeatMessage(PHold* phold) {
     phold->num_bytes_sent = 0;
 }
 
+static void _phold_generateCPULoad(PHold* phold) {
+    PHOLD_ASSERT(phold);
+    // this is volatile to prevent the compiler from optimizing out the loop
+    double volatile result = 0;
+    for (guint64 i = 0; i < phold->cpuload; i++) {
+        result = sqrt((double)i);
```
sporksmith, line 326: I'm not sure sqrt will have consistent performance characteristics. There's an x86 fsqrt instruction; will the implementation use it? Or do something clever to unroll the loop such that it can vectorize the computation even when it has to write the results into `result` one at a time? Or not use any of that and compute it in software? In any of those cases is the computation time constant, or does it depend on the value of `i`?

I think just making `result` an int64 and writing `i` into it would be more predictable, and sufficient.

robgjansen: The linux tool `stress-ng` spins on `sqrt()` as the default CPU load stress test. But you're right that I may not be setting things up correctly so that it works the same in all of our targets.

In case you are interested, stress-ng has [77 other ways to spin](https://github.com/ColinIanKing/stress-ng/blob/master/stress-cpu.c) :)

I'm ambivalent, so removing the `sqrt()` is fine with me.

sporksmith: Interesting; afaict that project doesn't care that much if the load factor is actually linear in `i` or not. Though they also use a random input to the sqrt instead of `i`, which would help keep it ~linear even if `sqrt`'s cost does depend on the value of its input.

But yeah, I'd lean towards keeping this as simple and predictable as possible :)

PR #906, opened 2020-08-04
Add Host and Process into Thread base and expose their IDs

sporksmith:


src/main/host/thread_preload.c
```diff
@@ -499,34 +496,33 @@ long threadpreload_nativeSyscall(Thread* base, long n, va_list args) {
...
+                                  .free = threadpreload_free,
+                                  .newClonedPtr = threadpreload_newClonedPtr,
+                                  .releaseClonedPtr = threadpreload_releaseClonedPtr,
+                                  .getReadablePtr = threadpreload_getReadablePtr,
+                                  .getReadableString = threadpreload_getReadableString,
+                                  .getWriteablePtr = threadpreload_getWriteablePtr,
+                                  .flushPtrs = threadpreload_flushPtrs,
+                                  .nativeSyscall = threadpreload_nativeSyscall,
+                              }),
+        .ptr_to_block = g_hash_table_new(g_int64_hash, g_int64_equal),
```
robgjansen, line 518: The indentation of the ThreadMethods seems odd. Did you run clang-format yet?

src/main/host/thread_ptrace.c
```diff
@@ -715,31 +713,28 @@ Thread* threadptrace_new(Host* host, Process* process, gint threadID) {
...
+                                  .newClonedPtr = threadptrace_newClonedPtr,
+                                  .releaseClonedPtr = threadptrace_releaseClonedPtr,
+                                  .getReadablePtr = threadptrace_getReadablePtr,
+                                  .getReadableString = threadptrace_getReadableString,
+                                  .getWriteablePtr = threadptrace_getWriteablePtr,
+                                  .flushPtrs = threadptrace_flushPtrs,
+                                  .nativeSyscall = threadptrace_nativeSyscall,
+                              }),
         // FIXME: This should the emulated CPU's frequency
         .tsc = {.cyclesPerSecond = 2000000000UL},
```
robgjansen, line 733: Indentation again. I think it might be because the following line was not edited so clang-format isn't able to keep the entire initializer consistent?

src/main/host/thread_preload.c
sporksmith: I hadn't, but doing so didn't change this. It just ends up getting indented more due the extra level of nesting here (initializing the struct that we pass to `thread_create`)

src/main/host/thread_ptrace.c
sporksmith: Same here

sporksmith:
> It looks like you are not really "moving" host and process into thread base so much as adding additional references to them in the base. I assume that you intended it that way, and you will be calling the getID functions from rust in later PRs.

Oops, right.

> We are still storing the host and process references in syscall handler. I suppose we'll have to continue to live with that since those are needed to handle syscalls.

Yeah; I think we're going to want to move the syscall handler to be owned by the process rather than by the thread, since it tracks process-wide state. When we do that it'll no longer have a pointer to the thread so will need its own pointers to the process and host. (Though I suppose it could drop the host pointer and get it via the process when needed)

PR #907, opened 2020-08-04
IntervalMap add fuzz tests and fix insertions over exact interval

sporksmith:


src/main/utility/interval_map.rs
```diff
@@ -286,43 +288,89 @@ mod tests {
...
+                "length-sum {} is smaller than inserted interval length {}",
+                new_len_sum,
+                end - begin + 1
+            ));
+        }
+        if new_len == 0 {
+            return Err(format!("new length is zero"));
+        }
 
-        assert!(new_len >= old_len);
```
stevenengler, line 322: Was this `assert!(new_len >= old_len)` check intentionally removed? (just making sure it wasn't accidentally missed if it's still wanted)

src/main/utility/interval_map.rs
```diff
@@ -286,43 +288,89 @@ mod tests {
...
+        Ok(mutations)
+    }
 
-        mutations
+    #[test]
+    fn test_insert_random() {
+        use rand::thread_rng;
+        use rand::Rng;
+        let dist = rand::distributions::Uniform::new_inclusive(0, 10);
+        let mut rng = thread_rng();
```
stevenengler, line 350: What do you think about using a seed here (and for the other test) so that the tests are reproducible?

src/main/utility/interval_map.rs
```diff
@@ -197,6 +198,7 @@ impl<V: Clone> IntervalMap<V> {
         // after the splice end (after having clipped the end of any existing interval contained in
         // the range, above).
         let splice_end = match self.ends.binary_search(&end) {
+            Ok(i) if self.ends[i] <= end => i + 1,
```
stevenengler, line 201: Maybe put the `self.ends[i] <= end` in brackets? I was really confused for a minute wondering what a backward arrow means in a rust match expression :)

src/main/utility/interval_map.rs
sporksmith: Done

src/main/utility/interval_map.rs
sporksmith: Yup! The fuzz testing tripped it, and I realized this was an incorrect assertion. the new number of intervals can be fewer than the old one, when the inserted interval removes multiple old intervals.

src/main/utility/interval_map.rs
sporksmith: Done

src/main/utility/interval_map.rs
stevenengler: Ah okay that makes sense!

Issue #902, opened 2020-07-31
Tests show as passing even if they do not complete before the stop time


stevenengler:
**Describe the issue**
If the test takes longer than the stop time, or if the test gets stuck, Shadow will eventually kill the process but still return 0 as the exit code. This causes ctest to think that the test passed, even though it didn't complete.

**To Reproduce**
Run a test with a blocking call and a timeout. For example, run a test that calls `sleep(10);` but with `<kill time="5"/>`.

**Shadow (please complete the following information):**
 - Version: post the output of `shadow --version`  
The dev branch as of at least `17b9e45430c06`.

sporksmith:
This is a good observation, but a bit tricky to fix because some tests are designed to run until the simulation end time is reached. I think this is also the norm for most "real" simulations - e.g. tor nodes will still be running when the simulation ends.

stevenengler:
Tor I think should be fine since it returns 0 on SIGTERM, but for other software it might be unexpected. I think it would be useful to have two options here, "stop time" to send a SIGTERM and continue to run Shadow until the processes exit, and a "kill time" to send a SIGKILL and return the exit code.

This is causing our tests to think that `timerfd-shadow-ptrace` passed when it's actually timing out. And `phold-shadow-ptrace` will fail as well, but it's designed to run forever.

sporksmith:
Seems like a promising approach. I suppose then we'd modify phold to catch SIGTERM and shutdown gracefully.

@robgjansen wdyt?

stevenengler:
> Seems like a promising approach. I suppose then we'd modify phold to catch SIGTERM and shutdown gracefully.

I did a little more testing, and it seems that shadow-ptrace doesn't yet allow you to catch signals, so that will need to be supported first. And shadow-preload does work, but I'm not convinced it will work properly all of the time because of the read() loop within the shim. But as long as the signal is sent when the program is not within a system call (or waiting on an event in general), I think it will be fine.

robgjansen:
Updating tests to use macros so that we always arrange to check the plugin return code; see #1114, #1116, #1117, #1123, and #1131.

PR #908, opened 2020-08-04
If stoptime is reached, exit with an error code

stevenengler:
When running tests, if the test doesn't end before stoptime, shadow will exit with an exit code of 0, which ctest will think is a passing test. Instead, this PR will collect the exit code of the process and return that instead.

I'm not sure how this will affect experiments where the user doesn't care about the exit code at stoptime, where Shadow might now return an error code. Tor handles SIGTERM with libevent and returns an exit code of 0, so this shouldn't affect Tor experiments.

Closes shadow/shadow#902.

### DIFF
--- a/src/main/core/main.c
+++ b/src/main/core/main.c
@@ -188,6 +188,11 @@ gint main_runShadow(gint argc, gchar* argv[]) {
         return EXIT_FAILURE;
     }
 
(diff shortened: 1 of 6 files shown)


stevenengler:
Closing for now. Some of these changes have been added in other PRs.

PR #910, opened 2020-08-06
Parser for proc/maps

sporksmith:


src/main/utility/proc_maps.rs
```diff
@@ -0,0 +1,470 @@
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s == "p" {
+            Ok(Sharing::Private)
+        } else if s == "s" {
+            Ok(Sharing::Shared)
+        } else {
+            Err(format!("Bad sharing specifier {}", s))?
+        }
+    }
+}
```
stevenengler, line 24: I've been trying to understand how this works but I'm confused. Supposedly (using the `print_type_of` function in https://stackoverflow.com/a/58119924) the `Err(format!("Bad sharing specifier {}", s))` is of type `Result<Result<Sharing, Box<dyn Error>>, String>`, but I don't see how the `?` turns it into a `Result<Sharing, Box<dyn Error>>`. And since `String` does not implement `Error`, I don't see how the `String` can be returned in the result. It works, but I have no idea why :)

sporksmith: There's an implementation of `From<String> for Box<dyn Error>`: https://doc.rust-lang.org/std/string/struct.String.html#impl-From%3CString%3E-6

The `?` operator uses that to do the conversion :)

stevenengler: Ah okay that makes things much clearer, thanks :) And it's great how the Rust documentation links to the source code, so you can see that the conversion creates a `StringError` type.

src/main/utility/proc_maps.rs
```diff
@@ -0,0 +1,470 @@
...
+
+/// Reads and parses the contents of a /proc/[pid]/maps file
+pub fn mappings_for_pid(pid: libc::pid_t) -> Result<Vec<Mapping>, Box<dyn Error>> {
+    use std::fs::File;
+    use std::io::Read;
+
+    let mut file = File::open(format!("/proc/{}/maps", pid))?;
+    let mut contents = String::new();
+    file.read_to_string(&mut contents)?;
+    Ok(parse_file_contents(&contents)?)
```
stevenengler, line 190: This can be just `parse_file_contents(&contents)` I think.

src/main/lib.rs
```diff
@@ -2,6 +2,9 @@
  * The Shadow Simulator
  * See LICENSE for licensing information
  */
+#[macro_use]
+extern crate lazy_static;
```
stevenengler, line 6: Really small thing, but I think the [recommended](https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html) way is now to just `use lazy_static::lazy_static;`, and they plan to deprecate (but not remove) `#[macro_use]` at some point.

src/main/utility/proc_maps.rs
sporksmith: Done

src/main/lib.rs
sporksmith: Ah good to know :)

PR #913, opened 2020-08-07
Memory manager

sporksmith:
Manages efficient access to plugin memory, primarily by remapping plugin's memory into a shared memory file.


sporksmith:
Oops, left in `#![allow(unused)]`; handy to cut down noise while writing, but shouldn't be merged. Pushing a fixup shortly.

sporksmith:
> Oops, left in `#![allow(unused)]`; handy to cut down noise while writing, but shouldn't be merged. Pushing a fixup shortly.

Fixed; ready for review.

src/main/host/memory_manager.rs
```diff
@@ -0,0 +1,769 @@
+use super::syscall_types::*;
```
stevenengler, line 1: Using the wildcard here is a little confusing since there are types with the same names in other modules. For example, this file uses both a `PluginPtr` and a `c::PluginPtr`, which isn't obvious at first glance that they're different types. Rather than the wildcard, I think we should either `use super::syscall_types;` or `use super::syscall_types as <something shorter>`.

src/main/host/memory_manager.rs
```diff
@@ -0,0 +1,769 @@
...
+    }
+
+    /// Gets a reference to an object in the plugin's memory.
+    /// SAFETY
+    /// * The pointer must point to readable value of type T.
+    /// * Returned ref mustn't be accessed after Thread runs again or flush is called.
+    #[allow(dead_code)]
+    pub unsafe fn _get_ref<T>(&self, thread: &impl Thread, src: PluginPtr) -> Result<&T, i32> {
+        let raw = self.get_readable_ptr(thread, src, std::mem::size_of::<T>())?;
+        Ok(&*(raw as *const T))
```
stevenengler, line 411: I think there should be a check here that raw is not null, since as far as I can tell `get_readable_ptr()` can return a null pointer, which shouldn't be cast as a rust reference. (And the same for functions below, including the `from_raw_parts()`.)

src/main/host/memory_manager.rs
```diff
@@ -0,0 +1,769 @@
...
+    // We initially mmap enough *address space* in Shadow to accomodate a large stack, but we only
+    // lazily allocate it. This is both to prevent wasting memory, and to handle that we can't map
+    // the current "working area" of the stack in thread-preload.
+    stack_copied: RefCell<Interval>,
+
+    // The bounds of the heap. Note that before the plugin's first `brk` syscall this will be a
+    // zero-sized interval (though in the case of thread-preload that'll have already happened
+    // before we get control).
+    heap: RefCell<Interval>,
+}
```
stevenengler, line 67: I don't understand the reason for all of the RefCells here (except maybe `misses_by_path`). Why is interior mutability needed, rather than just taking the entire memory manager as mutable? Rust's ownership checks seem like they would be extra useful for an "unsafe" memory manager.

src/main/host/memory_manager.rs
```diff
@@ -0,0 +1,769 @@
...
+        thread: &impl Thread,
+        src: PluginPtr,
+        len: usize,
+    ) -> Result<&mut [T], i32> {
+        let raw = self.get_writeable_ptr(thread, src, std::mem::size_of::<T>() * len)?;
+        Ok(std::slice::from_raw_parts_mut(raw as *mut T, len))
+    }
+
+    /// We take a mutable reference here to ensure there are no outstanding borrowed
+    /// references/pointers, which might otherwise become invalidated if shadow's mapping of the
```
stevenengler, line 459: How does taking a mutable reference ensure there are no borrowed references/pointers?

src/main/host/memory_manager.rs
sporksmith: If the methods that return immutable references took a mutable reference to the MemoryManager, then we'd never be able to have multiple outstanding immutable refs. https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5b2c92c605755367ee7bbdd4f31e6dd1

Added some comments.

> Rust's ownership checks seem like they would be extra useful for an "unsafe" memory manager.

Yeah, that reminds me I'd had an idea to make this safer. I'd written some TODOs that dropped somewhere along the way.

Ultimately I'd like to keep track of the memory ranges of borrowed references using another IntervalMap, and enforce that mutable references are exclusive from any other borrowed references. I think to do that these'll need to return a Ref-like object that updates the MemoryManager's bookkeeping when they're dropped.

I'd rather defer that change for now, but I added some comments about it, and for now made the methods that return mutable references take `&mut self`.

sporksmith:
P.S. @stevenengler thanks for the careful review!

src/main/host/memory_manager.rs
sporksmith: Thinking about this some more - maybe only allowing one outstanding ref at a time isn't actually that bad. Small values (e.g. integers) can just be copied out, and even if we end up doing multiple lookups of the same ref in some scenarios, that's probably cheaper than the bookkeeping I was thinking of where we have to mutate another intervalmap everytime a ref is borrowed or dropped.

Going to take a quick stab at removing internal mutability and just requiring a mutable ref in most (all?) of MemoryManager's methods, and see how it looks

sporksmith: I tried it, and was almost ready to go with that approach, but realized that if we end up having to do multiple accesses of pointers that we're *not* mapping; i.e. that we have to use the thread->get[Readable|Writeable]Pointer APIs, it could be a significant performance hit. I think I'd rather leave it as-is for now.

I updated the comments again though to note that while we might consider tracking borrowed memory ranges, it might be better to just see if we can avoid needing to hold a mutable ref simultaneously with other refs. I suspect we largely can.

Anyway, the net result of this comment thread (so far) are some more detailed comments, and changing the methods that return a mutable ref to take `&mut self`, which I think are both improvements at least :)

stevenengler: > If the methods that return immutable references took a mutable reference to the MemoryManager, then we'd never be able to have multiple outstanding immutable refs. https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5b2c92c605755367ee7bbdd4f31e6dd1

Ah, makes sense.

Regarding the comment "Perhaps we could require the caller to explicitly call another hook on the first syscall after the `execve` is done", it seems like this would allow you to get rid of a few RefCells (`shm_file` and `regions`), so I don't think making the call explicit is a bad thing. But I'm okay with how it is now, thanks for adding the comments.

> Ultimately I'd like to keep track of the memory ranges of borrowed references using another IntervalMap, and enforce that mutable references are exclusive from any other borrowed references. I think to do that these'll need to return a Ref-like object that updates the MemoryManager's bookkeeping when they're dropped.
>
> I'd rather defer that change for now, but I added some comments about it, and for now made the methods that return mutable references take &mut self.

I think this is a good idea (but for a possible future PR). Another option (if willing to change the interface) could be a function or macro that allows you to take multiple mutable references at once so that the bookkeeping overhead isn't needed. Something like `(ref1, ref2, ref3) = manager_get_mut!(mem_manager, plugin_ptr_1, plugin_ptr_2, plugin_ptr_3)`, which checks that the memory doesn't overlap, but also keeps a mutable reference to the memory manager so that additional mutable references can't be taken.

> I tried it, and was almost ready to go with that approach, but realized that if we end up having to do multiple accesses of pointers that we're not mapping; i.e. that we have to use the thread->get[Readable|Writeable]Pointer APIs, it could be a significant performance hit. I think I'd rather leave it as-is for now.

Sounds good to me.

src/main/host/memory_manager.rs
stevenengler: > Rust won't allow you to call a mutable method of an object from which there are any borrowed lifetimes. e.g.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=368f25542a78ace3935d00214dd260eb

I'm still not convinced here. The `get_mapped_ptr()` returns a `*const c_void` which in `get_ref()` is dereferenced and converted to a Rust reference with `unsafe { &*(raw as *const T) }`. I think this operation will cause the `Ref` to go out of scope, so `RefCell` will think it's no longer borrowed? So I think Rust will allow you to call that method with a mutable reference to the memory manager, even if there are outstanding mutable or immutable references. https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cbb4bf1fef3e098a01b236df26eb15fd

src/main/host/memory_manager.rs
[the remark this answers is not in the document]
stevenengler: Oops sorry, I shouldn't suggest changes without checking that they work first :/

src/main/host/memory_manager.rs
stevenengler: Okay this makes sense, feel free to leave it as-is.

src/main/host/memory_manager.rs
```diff
@@ -0,0 +1,769 @@
...
+    // the memory file. e.g. called after execve and in Drop.
+    fn reset_and_unmap_all(&self) {
+        // Mappings are no longer valid. Clear out our map, and unmap those regions.
+        let mutations = self
+            .regions
+            .borrow_mut()
+            .clear(std::usize::MIN..std::usize::MAX);
+        for m in mutations {
+            match m {
+                Mutation::Removed(interval, region) => {
```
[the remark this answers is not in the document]
stevenengler: I think I've written too much Python lately :)

src/main/host/memory_manager.rs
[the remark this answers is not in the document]
sporksmith: No worries :)

src/main/host/memory_manager.rs
sporksmith: > Something like (ref1, ref2, ref3) = manager_get_mut!(mem_manager, plugin_ptr_1, plugin_ptr_2, plugin_ptr_3), which checks that the memory doesn't overlap, but also keeps a mutable reference to the memory manager so that additional mutable references can't be taken.

Yeah, something like that might end up working.

Thinking about this some more, I've flip-flopped again and removed the internal mutability. I think preferring simplicity and static checking makes sense until we know for sure we need something else.

src/main/host/memory_manager.rs
sporksmith: Yeah, we're counting on the static safety in this case, not the Ref (even before removing it). The reference returned e.g. by MemoryManager::get_ref is statically borrowed from the MemoryManager, so calling any other mutable methods on MemoryManager while that borrowed reference is alive would be a compile-time error.

As you show in your example, that static check doesn't work for pointers, only for refs. It works when calling e.g. MemoryManager::get_ref, but the signature is implicitly `pub unsafe fn get_ref<'a, T>(&'a mut self, thread: &'a impl Thread, src: PluginPtr) -> Result<&'a T, i32>` - from the caller's point of view the ref is borrowed from the memorymanager and the thread, even though internally the ref is constructed from a pointer.

If the caller uses `MemoryManager::get_readable_ptr` to get a pointer and converts it to a reference itself, as in your example, all bets are off again, since the reference won't be viewed as borrowed from the MemoryManager.

stevenengler: Ah okay this clears it up for me, thanks!

src/main/host/memory_manager.rs
stevenengler: Sorry for making you second-guess yourself :) Either approach is good with me.

PR #918, opened 2020-08-12
Force frame pointers in Rust. Useful to get sensible prof reports

sporksmith:


sporksmith:
> Would it be better to enable this only for debug builds? But I'm okay with it in any case.

Yeah, I wondered that a bit too, but opted to follow the existing precedent of setting -fno-omit-frame-pointer for all builds in the C code.

Presumably we want it for being able to profile release builds. I suppose we could make it an orthogonal cmake build option, but my understanding is that the performance cost is generally negligible. To double-check, I just did some quick ad-hoc timing of phold (with the simulation time changed from 3 to 30) and wasn't able to measure a significant difference.

Issue #888, opened 2020-07-15
Make thread_getWriteablePointer and thread_getReadablePointer zero-copy


sporksmith:
The `thread_get(Readable|Writeable)Pointer` interfaces were designed so that the implementations could potentially be zero-copy, but right now they both copy data from the plugin into a Shadow-allocated buffer. Worse, this copy involves syscalls, and for the `threadpreload` involves context switches between Shadow and the plugin.

In most or all cases, we should be able to arrange for plugin memory that Shadow needs to access to be directly mmap'd in shadow, such that these interfaces should only need to do some arithmetic to translate a plugin-pointer to a shadow-pointer.

sporksmith:
Initial proof of concept: #870. This currently naively maps maps individual pages on-demand and assumes they're valid forever. I had to only turn the optimization on after the first `sendto` syscall; I think because handling invalidations due to earlier execs, mmaps, munmaps, etc. is currently unhandled. It also maps all such pages rw inside the plugin. Since it only does each mmap at a single-page granularity, it has to fall back to the legacy implementations for accesses that cross page boundaries (though this shouldn't happen in the phold benchmark).

It substantially improves performance in the phold benchmark: about 2.2s -> 1.3s for preload, 1.1s -> 0.9s for ptrace.

Now that we know it works and gives a substantial benefit, I'm working on implementing it "for real". The big pieces are:

* [] Implement a data structure to track variable-sized, mmap'd memory regions for each plugin.
* [] Track syscalls that change address-space mappings in the plugin (e.g. mmap, munmap, mremap, brk). Minimally we need to invalidate any of our previous mappings that overlap the affected areas. We could also potentially use these hooks to proactively mmap areas we may need to access. If we don't do that, we probably will at least want to keep track of metadata about the mapping (such as permissions), so that we can map it later if needed.
* [] Maybe parse /proc/*/maps to get the ground-truth current state of the plugin's address space when we try to do a mapping. (e.g. find the bounds and permissions of a region we want to map). We might not need this if we successfully hook into all the syscalls above that would change such mappings, but there could be some gaps there. e.g. with preload interposition we don't see syscalls before our shim constructor runs, and so don't see some of the initial address-space setup including loading in code and relevant static data areas.

PR #919, opened 2020-08-12
Split Rust Thread interface (trait) from implementation

sporksmith:
This will make it easier to later have Rust-native Thread implementations rather than wrapping the C interface.

Wraps some additional Thread functionality, and provides default-implementation in the Thread trait, e.g. to provide type-safe wrappers for specific syscalls.


sporksmith:
Progress on #888

PR #920, opened 2020-08-12
IntervalMap: add iter_from method

sporksmith:


src/main/utility/interval_map.rs
```diff
@@ -99,6 +99,15 @@ impl<V: Clone> IntervalMap<V> {
         ItemIter { map: self, i: 0 }
     }
 
+    pub fn iter_from(&self, begin: usize) -> ItemIter<V> {
```
stevenengler, line 102: Maybe add a comment for this function since all of the other functions are well documented?

sporksmith: Done

PR #922, opened 2020-08-13
Don't use heap in base and shim logger

### LINKED ISSUE #917: Avoid using the heap in shim logger
The shim logger currently uses some helper functions that allocate memory on the heap. This can lead to corruption during the process of manipulating the heap mapping, and/or if it ends up triggering brk syscalls.

sporksmith:
Fixes #917

### DIFF
--- a/src/shim/shim_logger.c
+++ b/src/shim/shim_logger.c
@@ -3,6 +3,9 @@
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <sys/param.h>
+#include <sys/time.h>
+#include <time.h>
 
 #include "shim/shim.h"
 #include "support/logger/log_level.h"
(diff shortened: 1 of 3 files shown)


src/shim/shim_logger.c
```diff
@@ -41,18 +46,20 @@ void shimlogger_log(Logger* base, LogLevel level, const gchar* fileName,
...
-            time_string, simulation_nanos_string, loglevel_toStr(level), fileName,
-            lineNumber, functionName, message);
+    // Keep appending to string. These functions all ensure NULL-byte termination.
+
+    offset += logger_elapsed_string(&buf[offset], sizeof(buf) - offset);
+    offset += snprintf(&buf[offset], sizeof(buf) - offset, " [");
+    offset += _simulation_nanos_string(&buf[offset], sizeof(buf) - offset);
+    offset += snprintf(&buf[offset], sizeof(buf) - offset, "] [shd-shim] [%s] [%s:%i] [%s] ",
+                       loglevel_toStr(level), logger_base_name(fileName), lineNumber, functionName);
+    offset += vsnprintf(&buf[offset], sizeof(buf) - offset, format, vargs);
```
stevenengler, line 56: Although it's unlikely since the buffer is large, there is a possible buffer overflow if offset is ever >200 any time before the last `vsnprintf` line since `snprintf` returns the number of bytes it would have written (and not how many it actually wrote), and `sizeof(buf) - offset` will wrap around if less than 0. Maybe add just an `assert(offset <= 200)` before the last `vsnprintf` just to be safe?

src/support/logger/logger.c
```diff
@@ -53,31 +54,48 @@ int64_t logger_elapsed_micros() {
...
+                                const int lineNumber, const char* format, va_list vargs) {
+    // Stack-allocated to avoid dynamic allocation.
+    char buf[200];
+    size_t offset = 0;
+
+    // Keep appending to string. These functions all ensure NULL-byte termination.
+    offset += logger_elapsed_string(&buf[offset], sizeof(buf) - offset);
+    offset += snprintf(&buf[offset], sizeof(buf) - offset, "%s [%s:%i] [%s] ",
+                       loglevel_toStr(level), logger_base_name(fileName), lineNumber, functionName);
+    offset += vsnprintf(&buf[offset], sizeof(buf) - offset, format, vargs);
```
stevenengler, line 95: Same as above comment.

src/shim/shim_logger.c
sporksmith: > Although it's unlikely since the buffer is large, there is a possible buffer overflow if offset is ever >200 any time before the last vsnprintf line

Oof, thanks for catching that

> since snprintf returns the number of bytes it would have written (and not how many it actually wrote), and sizeof(buf) - offset will wrap around if less than 0.

Sigh. Yeah I was thinking it returned the number of bytes *actually* written, like `write` does, and how I wish it worked :).

> Maybe add just an assert(offset <= 200) before the last vsnprintf just to be safe?

I'd rather we truncate than crash. Unfortunately I couldn't come up with a very elegant way of doing so, but done now in an ugly way. Maybe as you said the guards aren't really necessary for the fixed-ish size strings above, but I'd rather have them in place for uniformity, and in case anyone tweaks the buffer size or adds/modifies those strings later.

This time tested with a smaller buffer size (20) to validate that it truncated and didn't crash.

sporksmith: Oh! I realized what you meant now about asserting before the last vsnprintf; since all the strings before there are fixed-ish size, an assert there wouldn't be too bad, and we can let the last vsnprintf truncate.

I suppose that would be *pretty* safe, but I think I'd still rather truncate than crash there, since e.g. the file-name-length is a bit difficult to bound for sure.

PR #924, opened 2020-08-13
Use shared memory + spinlock & semaphore to perform IPC message passing in preload mode.

RWails:


src/shim/gate.cc
```diff
@@ -0,0 +1,45 @@
...
+
+#include <sched.h>
+
+// (rwails) The number times we should increment the counter and 
+// check the atomic bool before we fall back to the semaphore.
+// TODO: Move to an environment variable?
+#define SHD_GATE_SPIN_MAX 8096
+
+void gate_init(Gate *gate) {
+  std::memset(gate, 0, sizeof(Gate));
```
sporksmith, line 14: IIRC memset'ing an object to zero is a bit of a gray area for type safety in the standard. It's *probably* fine in practice, but better to either use struct initialization to default-init the struct in a type-aware way (`*gate = {0}`), and/or explicitly init all members (`gate->x` is currently not initialized other than in this memset).

src/shim/gate.h
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+struct Gate {
+  std::atomic<bool> x;
+  sem_t semaphore;
+  std::size_t spin_ctr;
+  pthread_spinlock_t lock;
+};
+
+void gate_init(Gate *gate);
```
sporksmith, line 21: I'm struggling a little bit with the "gate" metaphor. Could perhaps use a brief explanation (and/or a reference if it's a standard term, though I didn't have much luck searching)

src/main/host/thread_preload.c
```diff
@@ -26,7 +27,8 @@ struct _ThreadPreload {
 
     SysCallHandler* sys;
 
-    int eventFD;
+    pid_t childPID;
```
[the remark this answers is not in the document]
RWails: Fixed.

src/main/host/thread_preload.c
```diff
@@ -180,15 +162,18 @@ pid_t threadpreload_run(Thread* base, gchar** argv, gchar** envv) {
...
+        shmemallocator_globalBlockSerialize(&thread->ipc_blk);
 
-    char buf[64];
-    snprintf(buf, 64, "%d", child_fd);
+    char ipc_blk_buf[SHD_SHMEM_BLOCK_SERIALIZED_MAX_STRLEN] = {0};
+    shmemblockserialized_toString(&ipc_blk_serial, ipc_blk_buf);
 
     /* append to the env */
-    myenvv = g_environ_setenv(myenvv, "_SHD_IPC_SOCKET", buf, TRUE);
+    myenvv = g_environ_setenv(myenvv, "_SHD_IPC_BLK", ipc_blk_buf, TRUE);
```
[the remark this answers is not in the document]
RWails: The env should use the character buffer (ipc_blk_buf).

(The character buffer is a string describing the serialized block.)

src/main/host/thread_preload.c
```diff
@@ -278,7 +261,13 @@ SysCallCondition* threadpreload_resume(Thread* base) {
                     .event_id = SHD_SHIM_EVENT_SHMEM_COMPLETE,
                 };
 
-                shimevent_sendEvent(thread->eventFD, &ipc_complete_ev);
+                ShimEvent resp = {0};
```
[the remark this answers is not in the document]
RWails: Fixed.

src/main/shmem/buddy.c
```diff
@@ -159,13 +159,15 @@ static void _buddy_alloc_split_blocks(BuddyControlBlock* bcb, uint32_t k,
 #ifndef NDEBUG
 static void _print(void* pool, size_t pool_nbytes, void* meta) {
 
+    fprintf(stderr, "-----------------------\n");
```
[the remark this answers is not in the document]
RWails: This is old debugging code. Removed for now.

src/shim/CMakeLists.txt
```diff
@@ -16,11 +16,14 @@ add_cflags(-fPIC)
...
 set(SHIM_HELPER_FILES
+  gate.cc
+  ipc.cc
   shim_event.c
 )
 add_library(${SHIM_HELPER_LIB} STATIC ${SHIM_HELPER_FILES})
 # -D_GNU_SOURCE enables some additional features in libc, such at RTLD_NEXT.
-target_compile_options(${SHIM_HELPER_LIB} PRIVATE -D_GNU_SOURCE)
+target_compile_options(${SHIM_HELPER_LIB} PRIVATE -D_GNU_SOURCE -fPIC)
+target_link_libraries(${SHIM_HELPER_LIB} PRIVATE shadow-shmem)
```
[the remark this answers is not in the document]
RWails: Changed to interface.

src/shim/gate.cc
RWails: Fixed. n.b. C++ doesn't support the lovely `= {0}` syntax to zero all struct members.

src/shim/gate.h
RWails: Renamed to BinarySpinningSem

src/shim/ipc.cc
```diff
@@ -5,20 +5,20 @@
...
 struct IPCData {
     ShimEvent plugin_to_shadow, shadow_to_plugin;
-    Gate xfer_ctrl_to_plugin, xfer_ctrl_to_shadow;
+    BinarySpinningSem xfer_ctrl_to_plugin, xfer_ctrl_to_shadow;
 };
 
 extern "C" {
 
 void ipcData_init(IPCData *ipc_data) {
     memset(ipc_data, 0, sizeof(IPCData));
```
sporksmith, line 18: As before, memset is a bit fragile here. Now I remember that this is especially the case for a C++ object, since a memset won't run the constructors of any member objects, if applicable.

As far as I can tell struct init *does* technically work here, but actually is also fragile due to c++ *destuctors*: if IPCData or any of its members have destructors, they'd run on the garbage data at `ipc_data` before calling IPCData's constructor (you haven't defined one, but IIRC there's implicitly one that default-initializes all members).

I think the idiomatic, safe way to do this is to replace `IPCData::init` with a constructor `IPDCata::IPCData`, and here use a placement new.

Demo: https://godbolt.org/z/df9Gjc

sporksmith: Alternatively, we could add a static assertion that IPCData (and therefore its members) doesn't have any nontrivial constructor or destructors, in which case the struct-init, or even the memset, should be safe: https://godbolt.org/z/c6nsTa

```
    void ipcData_init(IPCData *ipc_data) {
        static_assert(std::is_trivial<IPCData>());
        static_assert(std::is_trivially_destructible<IPCData>());
        *ipc_data = {};
    }
```

RWails: Good catch! I think memsetting a class has bitten me before. I had heard of placement `new`, but had forgotten its use since. Fixed!

src/main/shmem/buddy.c
```diff
@@ -156,50 +156,6 @@ static void _buddy_alloc_split_blocks(BuddyControlBlock* bcb, uint32_t k,
     }
 }
 
-#ifndef NDEBUG
-static void _print(void* pool, size_t pool_nbytes, void* meta) {
```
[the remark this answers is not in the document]
RWails: I think let's leave it out. The block is pretty old and probably only makes sense to me. I can dig it out of the history if needed.

sporksmith (approved):
Looks good! The coverage test is currently broken, so you can ignore that failure. @stevenengler is about to disable it.

You should be able to fix the bindings-lint check by running `cd build; make bindings`. (You may need to install bindgen and cbindgen first if you haven't already)

sporksmith:
> You should be able to fix the bindings-lint check by running `cd build; make bindings`. (You may need to install bindgen and cbindgen first if you haven't already)

Actually the binding failure is spurious too; will be fixed by #940

stevenengler:
All of the CI issues should be resolved now.

PR #925, opened 2020-08-14
Move libc-dbg to preceeding apt-get command

pastly:
Note: I don't know if there is a reason that matters for having it be its own apt-get command. I suspect there isn't one.

I neglected to install libc-dbg because I didn't want the python stuff, but didn't see the libc-dbg line just before it.

### DIFF
--- a/docs/1.1-Shadow.md
+++ b/docs/1.1-Shadow.md
@@ -62,6 +62,7 @@ sudo apt-get install -y \
     cmake \
     g++ \
     gcc \
+    libc-dbg \
     libglib2.0-0 \
     libglib2.0-dev \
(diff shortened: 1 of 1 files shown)


jtracey:
IIRC it's because in many apt-based distros, debug symbols are in different repos that aren't enabled by default, so moving is out into its own line makes those kinds of errors clearer. Maybe just add a specific note of that?

sporksmith:
This change lgtm. I suspect it's on a separate line here due to copy-paste-modifying the yum commands above; there it's a different command rather than just another package.

robgjansen:
Blocked on `lint_python` again... I'll just override and merge.

Issue #909, opened 2020-08-05
Support multi-threaded Shadow


robgjansen:
We want to support our work-stealing scheduling policy in multi-process Shadow.

robgjansen:
This has been transitioned to @sporksmith now.

PR #926, opened 2020-08-14
Support multiple worker threads with work stealing

robgjansen:
When using the work stealing scheduling policy, it is possible
that hosts are migrated between worker threads depending on the
plugin workloads. When migrating a host to a new worker thread,
the previous worker thread that was running it will need to detach
its ptrace attachments so that the new thread can attach and
continue tracing the plugin.

This commit adds a hook that detaches all ptrace attachments at
the end of each round of execution; in the following round, it
becomes possible again that the host could be stolen by another
worker. Also, before resuming a a ptrace session, the worker
checks and does a ptrace attach first if needed.

refs shadow/shadow#909


robgjansen:
Obsoleted by #927 .

PR #927, opened 2020-08-14
Support multiple worker threads with work stealing

sporksmith:
When using the work stealing scheduling policy, it is possible
that hosts are migrated between worker threads depending on the
plugin workloads. When migrating a host to a new worker thread,
the previous worker thread that was running it will need to detach
its ptrace attachments so that the new thread can attach and
continue tracing the plugin.

This commit adds a hook that detaches all ptrace attachments at
the end of each round of execution; in the following round, it
becomes possible again that the host could be stolen by another
worker. Also, before resuming a a ptrace session, the worker
checks and does a ptrace attach first if needed.

refs shadow/shadow#909

### DIFF
--- a/src/main/core/scheduler/scheduler_policy_host_steal.c
+++ b/src/main/core/scheduler/scheduler_policy_host_steal.c
@@ -10,6 +10,7 @@
 #include "main/core/scheduler/scheduler_policy.h"
 #include "main/core/support/definitions.h"
 #include "main/core/work/event.h"
+#include "main/core/worker.h"
 #include "main/host/host.h"
 #include "main/utility/priority_queue.h"
 #include "main/utility/utility.h"
@@ -191,6 +192,8 @@ static void _schedulerpolicyhoststeal_migrateHost(SchedulerPolicy* policy, Host*
         utility_assert(tdata->runningHost != tdataNew->runningHost);
         /* migrate the TLS of all objects associated with this host */
 //        host_migrate(host, &oldThread, &newThread);
+        debug("Migrating host %s from thread %u to thread %u", host_getName(host), tdata->tnumber,
+              tdataNew->tnumber);
     }
     _schedulerpolicyhoststeal_addHost(policy, host, newThread);
 }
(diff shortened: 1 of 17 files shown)


robgjansen:
Should we enable the ptrace test case that uses multiple worker threads at the bottom of `src/test/phold/CMakeLists.txt` to show that the simulation with multiple worker threads is working?

src/main/core/scheduler/scheduler_policy_host_steal.c
```diff
@@ -308,6 +310,8 @@ static Event* _schedulerpolicyhoststeal_popFromThread(SchedulerPolicy* policy, H
         if(nextEvent == NULL) {
             /* no more events on the runningHost, mark it as NULL so we get a new one */
             g_queue_push_tail(tdata->processedHosts, host);
+            /* detach all ptrace attachments for this host so it can be stolen next round */
+            host_detachAllPlugins(host);
```
robgjansen, line 314: I realized that this should become:
```
worker_setActiveHost(host);
host_detachAllPlugins(host);
worker_setActiveHost(NULL);
```

src/main/host/process.c
```diff
@@ -400,6 +400,17 @@ void process_schedule(Process* proc, gpointer nothing) {
     }
 }
 
+void process_detachPlugin(gpointer procptr, gpointer nothing) {
+    Process* proc = procptr;
+    MAGIC_ASSERT(proc);
+    if (proc->interposeMethod == INTERPOSE_PTRACE) {
+        // Detach all ptrace attachements
+        if (proc->mainThread) {
+            threadptrace_detach(proc->mainThread);
```
robgjansen, line 409: I realized that this should become:
```
worker_setActiveProcess(proc);
threadptrace_detach(proc->mainThread);
worker_setActiveProcess(NULL);
```

src/main/core/scheduler/scheduler_policy_host_steal.c
sporksmith: done

src/main/host/process.c
sporksmith: done

src/main/host/thread_ptrace.c
```diff
@@ -406,9 +438,136 @@ static void threadptrace_flushPtrs(Thread* base) {
...
+        }
+        reason = _getStopReason(wstatus);
+    }
+    if (reason.type != STOPREASON_SYSCALL) {
+        error("unexpected stop reason: %d", reason.type);
+        utility_assert(reason.type == STOPREASON_SYSCALL);
+    }
+
+    // Restore our saved instruction pointer to where it was before we detached - just after the
+    // syscall insn.
```
robgjansen, line 490: What is "insn"? Instruction?

sporksmith: Yup; expanded to 'instruction' for clarity

PR #928, opened 2020-08-16
syscall handler for getrandom

robgjansen:
Extended random tests and `getrandom()` syscall handler.


src/shim/preload_syscalls.c
```diff
@@ -237,6 +237,10 @@ NOREMAP(int, futimesat, (int a, const char* b, const struct timeval c[2]), a, b,
 NOREMAP(ssize_t, getdents, (int a, void* b, size_t c), a, b, c);
 NOREMAP(ssize_t, getdents64, (int a, void* b, size_t c), a, b, c);
 NOREMAP(int, getpeername, (int a, struct sockaddr* b, socklen_t* c), a, b, c);
+// TODO the following (getrandom) causes all shadow tests to fail with timeout in centos 8
+// I believe centos 8 contains a version of libc that contains a getrandom wrapper, whereas
+// it was previously only available through the syscall() function.
+//NOREMAP(ssize_t, getrandom, (void* a, size_t b, unsigned int c), a, b, c);
```
sporksmith, line 243: Hmm. Yeah, it's simple to use the `REMAP` wrapper to use a wrapper function instead, and I guess that function could open and read `/dev/u?random`. It'd be nice if we could do that only on platforms where it's needed, though. I guess `SYS_random` must still be defined on centos 8 since the code compiles there and just fails at runtime...?

src/test/random/test_random.c
```diff
@@ -3,73 +3,134 @@
...
-#define HIGH_THRESH 0.9f
+// The number of random values to generate with each method.
+#define RGENLEN 100
+// The number of buckets to use when checking random value distribution.
+#define BUCKETLEN 10
 
-static int _test_dev_urandom() {
-    const char* path = "/dev/urandom";
-    uint numLow = 0, numHigh = 0;
+static int _check_randomness(double* fracs, int len) {
```
sporksmith, line 21: Seems worth briefly commenting on the methodology here. It looks like the idea is that this is basically a fast "plausibly random" test suitable for testing that the plumbing isn't totally broken, but isn't trying to really evaluate the underlying RNG?

src/shim/preload_syscalls.c
robgjansen: I don't think we need (or want) a remapping. Preloading `getrandom()` and asking shadow to complete a getrandom syscall _should_ work fine. It doesn't seem to be any different than any other libc function we are preloading. And I'm really not sure how preloading `getrandom()` is causing timeouts when the `syscall(SYS_getrandom)` version works fine.

I would expect the opposite to happen: without preloading `getrandom()`, then the libc version gets called on centos 8, which could possibly block trying to get random bytes from the kernel (due to low available entropy). Our version is guaranteed not to block in any case.

And now I'm wondering why `getrandom()` is being preloaded in the ptrace mode tests anyway. I thought we do not preload things when running in ptrace mode?

sporksmith: > I don't think we need (or want) a remapping. Preloading getrandom() and asking shadow to complete a getrandom syscall should work fine. It doesn't seem to be any different than any other libc function we are preloading. And I'm really not sure how preloading getrandom() is causing timeouts when the syscall(SYS_getrandom) version works fine.

> I would expect the opposite to happen: without preloading getrandom(), then the libc version gets called on centos 8, which could possibly block trying to get random bytes from the kernel (due to low available entropy). Our version is guaranteed not to block in any case.

Ah right. Hmm. I wonder if something could be trying to use getrandom before main is called; even that should only affect preload-mode though (if it gets used before the shim constructor runs).

> And now I'm wondering why getrandom() is being preloaded in the ptrace mode tests anyway. I thought we do not preload things when running in ptrace mode?

Right now the shim gets loaded for ptrace mode as well; when these functions aren't running under shadow's thread-preload though, they make the real syscall instead of trying to talk to the shim pipe. i.e. most of the wrappers become ~equivalent to their libc counterparts.

We still need some parts of the shim in ptrace-mode; e.g. the wrapper around `getaddrinfo` that exclusively uses the hosts file instead of trying to make DNS requests.

I'd originally thought about splitting the shim so that ptrace-mode only gets any libc wrappers it needs, and preload-mode gets those + everything else, but doing so adds a bit of complexity. We could certainly do it if there's a reason to, though.

src/test/random/test_random.c
robgjansen: Comment added.

(You're right, we're just testing the "plumbing" and not the quality of the underlying RNGs. Another useful test here is for determinism, but that is currently tested as part of the determinism test case (though we may decide to move it here later).)

src/shim/preload_syscalls.c
robgjansen: OK, right.

IDK, might be nice to not intercept things we don't need to, and although it would add some complexity, I think it could make it easier to understand what is happening in ptrace mode because we could confine the worst of the complexity to preload mode. No strong opinion though, especially not in the short term.

PR #931, opened 2020-08-18
Memory manager mmap

sporksmith:
* Updates our understanding of the plugin's memory space; all accesses
should now be in a region we know about.
* Moves anonymous mappings into our shared memory file for fast access.
Notably allocators typically use such mappings for large allocation
requests (e.g. 100 MB).


robgjansen:
Reminder: maybe add `mmap2` to this PR.

sporksmith:
When adding the requested documentation for handle_mutations I realized there was a bug in the mremap handling. I extended the tests to more thoroughly cover mremap and ended up changing it a fair bit.

I also added support support and testing for mmap2.

Could you take another look? Normally GH has an option to view the diff since your last review, but I might have broken that by rebasing on the latest dev; if so, you should still be able to manually select the range of commits to look at (you want "Document handle_mutations" onward)

PR #933, opened 2020-08-19
Support logging in rust

stevenengler:
This PR adds support for using Shadow's logger from Rust. It wraps the `support/logger` utility with a Rust logger following the `Log` trait. This Rust logger requires memory allocations to convert to null-terminated strings. In the future it would be good to use a Rust logger for shadow, but this isn't super straightforward due to the required conversion between string types (and the required memory allocations), and missing variadic functions in Rust.

Note: Like standard rust, log messages during unit tests will not be displayed.


.github/workflows/lint.yml
```diff
@@ -35,44 +35,31 @@ jobs:
...
-        run: echo "::add-path::$HOME/.cargo/bin"
-      - name: Install bindgen
-        run: cargo install --force bindgen
-      - name: bindgen check
+      - name: Install Rust packages
+        run: cargo install --force cbindgen bindgen
+      - name: Check bindings
         run: |
-          cp src/main/cbindings/wrapper.rs orig.rs
+          mv src/main/bindings/rust/wrapper.rs /tmp/main-bindings-rust.rs
```
sporksmith, line 58: I wonder if we can replace the mv's and explicit diff's with just overwriting in place and doing a `git diff`? It'd be nice not to have to maintain the list of binding files here.

stevenengler: Good idea! Added.

PR #934, opened 2020-08-20
Ensure that shim is initialized before first wrapped syscall is made.

sporksmith:
Previously, we couldn't guarantee that some global constructor might not
call some of our wrapped syscalls before the shim was initialized by its
global constructor.

Our syscall wrapper now ensures that the shim is initialized, using a
pthread_once construct. We still use a global constructor to ensure that
the shim is initialized before calling main.

### DIFF
--- a/src/shim/shim.c
+++ b/src/shim/shim.c
@@ -75,8 +75,7 @@ static TIDBlkPair _shim_tidBlkTreeGet(pthread_t tid) {
     return tid_fd;
 }
 
-__attribute__((constructor(SHIM_CONSTRUCTOR_PRIORITY))) static void
-_shim_load() {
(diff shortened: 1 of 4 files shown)


src/shim/shim.c
```diff
@@ -145,6 +144,22 @@ _shim_load() {
     shim_enableInterposition();
 }
 
+// This function should be called before any wrapped syscall. We also use the
+// constructor attribute to be completely sure that it's called before main.
+__attribute__((constructor)) void shim_ensure_init() {
+    static __thread bool in_init = false;
```
RWails, line 150: We can't use `thread_local` here, can we? I think that's more portable if all of our compilers support it.

Issue #941, opened 2020-08-24
Coverage tests failing on rust nightly (incompatible LLVM versions)


stevenengler:
The coverage tests are failing on the most recent rust nightly, but were working on `nightly-2020-08-20`.

sporksmith:
It looks like this is due to the rust-nightly switching over to llvm-11.

When `--coverage` is enabled, a global destructor hook is added that appends profiling data to `gcda` files. It looks like the destructor hooks from llvm-10 (which end up in our C test binaries), and those from llvm-11 (which end up in our Rust binaries), are incompatible.

Unfortunately it looks like compatibility of the raw profile data between llvm versions is explicitly not supported: https://clang.llvm.org/docs/SourceBasedCodeCoverage.html#format-compatibility-guarantees

We'll either need to use the same (or "close enough") versions of llvm to compile our C and Rust code, or arrange that they don't both touch the same profile data files.

stevenengler:
Our current fix is to use `clang-11` in Ubuntu 20.10 (see #945), but leaving this issue open since it would be nice to have a better solution where we don't need to worry about Rust changing its LLVM version and breaking our CI tests.

stevenengler:
Closing since the shadow repo doesn't do coverage builds anymore.

PR #942, opened 2020-08-24
Using older Rust nightly version for coverage tests

stevenengler:
Until #941 is resolved, this PR uses `nightly-2020-08-20` for coverage tests.


stevenengler:
@sporksmith This might be better than disabling the test entirely (let me know if you disagree). And you can ignore the failing lint test, it will be fixed in #940.

PR #944, opened 2020-08-24
Fix clippy warnings in byte queue code

stevenengler:
Looked over the code again and it seems okay, but if there are any bugs, they should be caught fairly quickly in debug mode.

### DIFF
--- a/src/main/utility/byte_queue.rs
+++ b/src/main/utility/byte_queue.rs
@@ -57,10 +57,10 @@ impl ByteQueue {
 
         // while there are bytes to copy
         while bytes_copied < src_len {
-            let head_space = self.chunks.front().unwrap().buf.len() - self.head_write_offset;
+            let head_space = self.chunk_capacity - self.head_write_offset;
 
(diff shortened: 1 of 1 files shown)


sporksmith:
> Looked over the code again and it seems okay, but if there are any bugs, they should be caught fairly quickly in debug mode.

I was initially concerned about silent underflow in some of these places, but yes it looks like it'll be checked and panic in debug mode. https://github.com/rust-lang/rfcs/pull/560/files

stevenengler:
> > Looked over the code again and it seems okay, but if there are any bugs, they should be caught fairly quickly in debug mode.
> 
> I was initially concerned about silent underflow in some of these places, but yes it looks like it'll be checked and panic in debug mode. https://github.com/rust-lang/rfcs/pull/560/files

Yeah I considered using `checked_sub()` here, but after looking over the code more, I think checking in only debug mode should be okay since it would crash soon after due to slice indexing errors, which will get caught in release mode as well.

PR #945, opened 2020-08-24
Changed coverage CI tests to use clang 11

stevenengler:
Ubuntu 20.10 is apparently planning on making LLVM 11 the default in September, just after the official LLVM 11 release date. So we should be able to simplify these changes when this happens.

https://discourse.ubuntu.com/t/groovy-gorilla-release-schedule/15531


stevenengler:
@sporksmith Setting you as reviewer, but feel free to leave this for now if you want to consider alternatives.

sporksmith (approved):

PR #946, opened 2020-08-25
Save process exit codes

stevenengler:
Uses a SIGKILL rather than SIGTERM since we don't want shadow to block waiting for the process, and we do not yet allow the process to shutdown properly.

This PR creates a new file `*.exitcode`, which should contain the exit status of the process.

### DIFF
--- a/src/main/host/process.c
+++ b/src/main/host/process.c
@@ -59,6 +59,8 @@
 
 #include "main/host/thread_ptrace.h"
 
+static gchar* _process_outputFileName(Process* proc, const char* type);
+
 struct _Process {
     /* Host owning this process */
     Host* host;
@@ -152,7 +154,21 @@ static void _process_logReturnCode(Process* proc, gint code) {
                         ((code == 0) ? "success" : "error"), code,
                         process_getName(proc));
 
-        if(code == 0) {
+        gchar* fileName = _process_outputFileName(proc, "exitcode");
+        FILE *exitcodeFile = fopen(fileName, "we");
+        g_free(fileName);
+
+        if (exitcodeFile != NULL) {
+            fprintf(exitcodeFile, "%d", code);
+            fclose(exitcodeFile);
+        } else {
+            warning("Could not open '%s' for writing: %s", mainResultString->str, strerror(errno));
+        }
+
+        // if there was no error or was intentionally killed
+        // TODO: once we've implemented clean shutdown via SIGTERM,
+        //       treat death by SIGKILL as a plugin error
+        if (code == 0 || code == return_code_for_signal(SIGKILL)) {
             message("%s", mainResultString->str);
         } else {
             warning("%s", mainResultString->str);
@@ -348,6 +364,8 @@ void process_stop(Process* proc) {
         proc->plugin.isExecuting = TRUE;
         thread_terminate(proc->mainThread);
         proc->plugin.isExecuting = FALSE;
+        _process_logReturnCode(proc, thread_getReturnCode(proc->mainThread));
+
         debug("unreffing main thread %p", proc->mainThread);
         thread_unref(proc->mainThread);
         proc->mainThread = NULL;
--- a/src/main/host/thread_ptrace.c
+++ b/src/main/host/thread_ptrace.c
@@ -341,7 +341,7 @@ static void _threadptrace_updateChildState(ThreadPtrace* thread, StopReason reas
             debug("child %d terminated by signal %d", thread->base.nativePid,
                   reason.exited_signal.signal);
             thread->childState = THREAD_PTRACE_CHILD_STATE_EXITED;
-            thread->returnCode = -1;
+            thread->returnCode = return_code_for_signal(reason.exited_signal.signal);
             return;
         case STOPREASON_EXITED_NORMAL:
             thread->childState = THREAD_PTRACE_CHILD_STATE_EXITED;
@@ -372,11 +372,13 @@ static void _threadptrace_updateChildState(ThreadPtrace* thread, StopReason reas
 
 static void _threadptrace_nextChildState(ThreadPtrace* thread) {
     // Wait for child to stop.
+
     int wstatus;
     if (waitpid(thread->base.nativePid, &wstatus, 0) < 0) {
         error("waitpid: %s", g_strerror(errno));
         return;
     }
+
     StopReason reason = _getStopReason(wstatus);
     _threadptrace_updateChildState(thread, reason);
 }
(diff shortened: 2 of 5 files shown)


src/main/host/process.c
```diff
@@ -150,7 +152,17 @@ static void _process_logReturnCode(Process* proc, gint code) {
                         ((code == 0) ? "success" : "error"), code,
                         process_getName(proc));
 
-        if(code == 0) {
+        gchar* fileName = _process_outputFileName(proc, "exitcode");
+        FILE *exitcodeFile = fopen(fileName, "we");
```
sporksmith, line 156: Did you mean to use "x" for exclusive rather than "e"?

src/main/host/process.c
```diff
@@ -150,7 +152,17 @@ static void _process_logReturnCode(Process* proc, gint code) {
...
+        g_free(fileName);
+
+        if (exitcodeFile != NULL) {
+            fprintf(exitcodeFile, "%d", code);
+            fclose(exitcodeFile);
+        }
```
sporksmith, line 162: perhaps log a warning otherwise?

src/main/host/thread_ptrace.c
```diff
@@ -372,13 +370,19 @@ static void _threadptrace_updateChildState(ThreadPtrace* thread, StopReason reas
...
     if (waitpid(thread->base.nativePid, &wstatus, 0) < 0) {
         error("waitpid: %s", g_strerror(errno));
         return;
     }
+
     StopReason reason = _getStopReason(wstatus);
     _threadptrace_updateChildState(thread, reason);
+
+    if (thread->childState == THREAD_PTRACE_CHILD_STATE_EXITED) {
+        thread->returnCode = wstatus;
```
sporksmith, line 384: `WEXITSTATUS(wstatus)` (see `man waitpid`)

src/main/host/thread_ptrace.c
```diff
@@ -659,9 +663,12 @@ void threadptrace_terminate(Thread* base) {
...
     }
 
-    if (ptrace(PTRACE_CONT, thread->base.nativePid, 0, SIGTERM) < 0) {
-        warning("ptrace %d: %s", thread->base.nativePid, g_strerror(errno));
+    // need to kill() and not ptrace() since the process may not be stopped
+    if (kill(thread->base.nativePid, SIGKILL) < 0) {
+        warning("kill %d: %s", thread->base.nativePid, g_strerror(errno));
     }
+
+    _threadptrace_nextChildState(thread);
```
sporksmith, line 671: I'm not sure this is guaranteed to process the "killed" stop; there could be other stops before it. We could call `threadptrace_continue` to process all remaining stops, but perhaps if we get to this point better to just set thread->returnCode right here and return.

sporksmith (changes requested):

src/main/host/process.c
stevenengler: I used `e` for the `O_CLOEXEC` flag. It's unlikely that another thread will fork+exec while we're cleaning up this process, but I don't think it will hurt.

src/main/host/thread_ptrace.c
stevenengler: So I'm not sure if the `WEXITSTATUS` macro will return the signal that terminated the process, or if you need to call `WTERMSIG` for that. Originally I just wanted to return all of the bits so I wasn't missing anything. I'll do a test to check, but if you see this in the meantime and know whether it does, please let me know.

sporksmith: In general the full contents of `wstatus` is a bit opaque afaik. The documented API in waitpid says to use the macros to make any sense out of it.

Yeah, the exit code might not necessarily be the signal number when killed by an unhandled signal, but it should be non-zero, which should be good enough for our purposes.

OTOH on normal exit `wstatus` might still be non-zero, and we wouldn't want to say there was a non-zero exit status.

src/main/host/thread_ptrace.c
stevenengler: I'm not sure what you mean here, can't there only be a single stop at a time? If the process is in the stopped state when it receives a sigkill, I would expect a future waitpid() to return the status from the sigkill and not the stop.

src/main/host/thread_ptrace.c
stevenengler: I ran a test, and if the process is killed by a signal, `WIFEXITED(wstatus)` and `WEXITSTATUS(wstatus)` will both return 0. So if we aim to model something like bash, we should return `WEXITSTATUS(wstatus)` if the process exited normally, or `128+WTERMSIG(wstatus)` if the process exited by a signal.

sporksmith: SGTM

src/main/host/thread_ptrace.c
sporksmith: The child process might have already stopped for another reason before the kill signal was received. I *think* in that case this would see that other stop. The kill signal would still be pending, and would generate another stop after the process is resumed.

src/main/host/process.c
sporksmith: Ah, makes sense.

src/main/host/process.c
stevenengler: Added.

src/main/host/thread_ptrace.c
stevenengler: I think sigkill would be handled differently and wouldn't be queued, since ptrace cannot intercept sigkill (and therefore won't cause a signal-delivery-stop), and the process should exit immediately (regardless of if it's already stopped).

What do you think about the latest commit? I'm fairly confident that the waitpid will return the result from the sigkill, but I added an error() message to check if that's not the case, so hopefully we can tell if this happens in practice while we're still in the testing stages of phantom. I don't think we're any worse off than before.

src/main/host/process.c
```diff
@@ -150,7 +152,19 @@ static void _process_logReturnCode(Process* proc, gint code) {
...
+
+        if (exitcodeFile != NULL) {
+            fprintf(exitcodeFile, "%d", code);
+            fclose(exitcodeFile);
+        } else {
+            warning("Could not open '%s' for writing: %s", mainResultString->str, strerror(errno));
+        }
+
+        // if there was no error or was intentionally killed
+        if(code == 0 || code == 128+SIGKILL) {
```
sporksmith, line 167: Maybe a TODO here to treat death by SIGKILL as an error (or at least a warning) once we've implemented clean shutdown via SIGTERM?

src/main/host/thread_ptrace.c
```diff
@@ -341,7 +341,7 @@ static void _threadptrace_updateChildState(ThreadPtrace* thread, StopReason reas
             debug("child %d terminated by signal %d", thread->base.nativePid,
                   reason.exited_signal.signal);
             thread->childState = THREAD_PTRACE_CHILD_STATE_EXITED;
-            thread->returnCode = -1;
+            thread->returnCode = reason.exited_signal.signal + 128;
```
sporksmith, line 344: Could you add a helper like `int return_code_for_signal(int signal)`? That'd consolidate the magic `+ 128` in one place, where we can also document that e.g. this is what bash does.

sporksmith (approved):

src/main/host/thread_ptrace.c
stevenengler: Added. (And also now using this for preload return codes as well, previously it used a return code of -1.)

src/main/host/process.c
stevenengler: Added.

sporksmith (approved):
Not sure if you're waiting on me, but reapproving to be sure :)

Issue #826, opened 2020-06-09
Support multi-threaded plugins


robgjansen:
In order to support the Tor use-case in the multi-process architecture, we'll need to be able to handle plugins that run multiple threads (because Tor uses multiple threads for its onion-skin CPU workers).

This requires us to support the `clone` syscall, and may also require some work to enable the `pthread` library to function correctly, at least the subset of the API that Tor uses.

sporksmith:
Some subtasks:

- [ ] clone syscall (create new threads). #949 and #952
- [ ] futex syscall (synchronize across threads)
- [ ] Once futex is implemented, honor the `CLONE_CHILD_CLEARTID` flag in clone by doing a futex wakeup on the specified address when the child exits. e.g. this is used by pthread to join a child thread.
- [ ] set_robust_list syscall. This gives the OS (or Shadow, in this case) a list of futexes that should be woken up when the calling thread dies.- [ ] exit_group syscall. Right now it's unhandled, meaning we let it execute natively and don't do anything special. This is fine when there are no other threads. If there are other threads still alive, though, the native execution of exit_group will kill the native threads, but Shadow threads still need to be cleaned up.
- [ ] gettid syscall

PR #949, opened 2020-08-28
Move MemoryManager to Process

sporksmith:
Progress on #826 

Move MemoryManager from SyscallHandler to Process so that it can be shared amongst threads.

Also:
* Instead of setting a flag in MemoryManager to re-initialize itself after an exec, just destroy the whole thing. The SysCallHandler will re-create it lazily on the next syscall.
* Unlink shared memory file as soon as we're done opening it, reducing the risk of leaving stale files around after a crash.

### DIFF
--- a/src/main/host/syscall/epoll.c
+++ b/src/main/host/syscall/epoll.c
@@ -105,7 +105,7 @@ SysCallReturn syscallhandler_epoll_ctl(SysCallHandler* sys,
     }
 
     const struct epoll_event* event =
-        process_getReadablePtr(sys->process,sys->thread, eventPtr, sizeof(*event));
+        process_getReadablePtr(sys->process, sys->thread, eventPtr, sizeof(*event));
 
     debug("Calling epoll_control on epoll %i with child %i", epfd, fd);
     errorCode = epoll_control(epoll, op, descriptor, event);
@@ -186,7 +186,7 @@ SysCallReturn syscallhandler_epoll_wait(SysCallHandler* sys,
     guint numEventsNeeded = MIN((guint)maxevents, numReadyEvents);
     size_t sizeNeeded = sizeof(struct epoll_event) * numEventsNeeded;
     struct epoll_event* events =
-        process_getWriteablePtr(sys->process,sys->thread, eventsPtr, sizeNeeded);
+        process_getWriteablePtr(sys->process, sys->thread, eventsPtr, sizeNeeded);
 
     /* Retrieve the events. */
     gint nEvents = 0;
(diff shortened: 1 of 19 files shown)


src/main/host/syscall/epoll.c
```diff
@@ -105,7 +105,7 @@ SysCallReturn syscallhandler_epoll_ctl(SysCallHandler* sys,
     }
 
     const struct epoll_event* event =
-        memorymanager_getReadablePtr(sys->memoryManager,sys->thread, eventPtr, sizeof(*event));
+        process_getReadablePtr(sys->process,sys->thread, eventPtr, sizeof(*event));
```
robgjansen, line 108: Looks like you didn't introduce this here, but need a space after `sys->process,`. I'm surprised clang-format didn't catch this.

This same issue occurs many times throughout the diff below. I'm only noting it here, but please fix everywhere.

src/main/host/syscall/mman.c
```diff
@@ -165,8 +165,11 @@ static SysCallReturn _syscallhandler_mmap(SysCallHandler* sys, PluginPtr addrPtr
     }
 
     // Delegate execution of the mmap itself to the memorymanager.
```
robgjansen, line 167: Could we update these delegation comments to explain why/when the memory manager could be NULL and why NATIVE is OK in that case?

src/main/host/syscall_handler.c
```diff
@@ -179,8 +173,8 @@ SysCallReturn syscallhandler_make_syscall(SysCallHandler* sys,
 
     // Lazily initialize memoryManager on the first syscall. It needs a thread
     // in the syscall state for its initialization.
-    if (!sys->memoryManager) {
-        sys->memoryManager = memorymanager_new(sys->thread);
+    if (!process_getMemoryManager(sys->process)) {
+        process_setMemoryManager(sys->process, memorymanager_new(sys->thread));
```
robgjansen, line 177: So, iiuc, in the syscall handling code, `process_getMemoryManager()` should always return non-NULL. So should the NULL checks above (with fallback to NATIVE) be an assertion error instead?

sporksmith: Ah, at the last minute I was trying to make it easy to disable the MM completely for experimentation by just commenting out the line where it gets initialized. Since we don't have immediate plans for that I'll go ahead and make these asserts rather than falling back.

sporksmith: Ah, I mean e.g. in mman.c we can drop the fallbacks to NATIVE. Here we're lazily initializing it; I'll add better comments.

src/main/host/syscall/mman.c
sporksmith: Done. (Changed to assert)

PR #956, opened 2020-09-01
Added tests for the connect() call

### LINKED ISSUE #836: Thoroughly test the connect() syscall
See #832 for an overview of our approach.

stevenengler:
Closes #836.


src/test/socket/connect/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+add_test(NAME connect COMMAND ../../target/debug/test_connect)
+foreach(Method ptrace preload)
+    # runs the shadow tests with only the passing tests, not all tests
+    add_test(NAME connect-shadow-${Method} COMMAND ${CMAKE_BINARY_DIR}/src/main/shadow --interpose-method=${Method}
+	    -l debug -d connect.shadow-${Method}.data ${CMAKE_CURRENT_SOURCE_DIR}/connect.shadow.config.xml)
+endforeach()
+
+set(TestNames connect connect-shadow-ptrace connect-shadow-preload)
```
sporksmith, line 8: Consider appending to TestNames next to corresponding add_test commands, to make it easier to keep it consistent with the actual set of tests.

src/test/socket/connect/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
...
+set_property(TEST ${TestNames} PROPERTY ENVIRONMENT "RUST_BACKTRACE=1")
+set_property(TEST ${TestNames} PROPERTY FAIL_REGULAR_EXPRESSION "main error code '.*' for process")
```
sporksmith, line 10: Is this to catch the case where Shadow killed the process but exited with a 0 status? Probably worth a brief explanation and reference to the issue.

src/test/socket/connect/test_connect.rs
```diff
@@ -0,0 +1,526 @@
...
+    let num_tests = tests.len();
+    let tests: std::collections::BTreeMap<_, _> = tests.into_iter().collect();
+
+    // make sure we didn't have any duplicate tests
+    assert_eq!(num_tests, tests.len());
+
+    tests
+}
+
+fn get_all_tests() -> std::collections::BTreeMap<String, test_utils::TestFn> {
```
sporksmith, line 84: OK for this PR since it's the established pattern so far, but I wonder if it'd be a bit easier to read & maintain by adding a third parameter to each test, rather than having two lists. e.g.

```
let tests: Vec<(String, test_utils::TestFn, /* passing: */)bool> = vec![
    (append_args("test_after_close"),
      Box::new(move || test_after_close(sock_type, flag)),
      true),
    ...
     (append_args("test_non_existent_server"),
      Box::new(move || test_non_existent_server(sock_type, flag),
       false))
),
```

We could then construct the list of tests to actually run with a simple filter. That could be done from out here, though since all of our tests follow this pattern it could perhaps alternatively be moved into the test utility.

src/test/socket/connect/test_connect.rs
```diff
@@ -0,0 +1,526 @@
...
+                (append_args("test_interface_loopback"),
+                    Box::new(move || test_interface(sock_type, flag, libc::INADDR_LOOPBACK, None))),
+                (append_args("test_interface_loopback_any"),
+                    Box::new(move || test_interface(sock_type, flag, libc::INADDR_LOOPBACK, Some(libc::INADDR_ANY)))),
+                (append_args("test_interface_any"),
+                    Box::new(move || test_interface(sock_type, flag, libc::INADDR_ANY, None))),
+                (append_args("test_interface_any_any"),
+                    Box::new(move || test_interface(sock_type, flag, libc::INADDR_ANY, Some(libc::INADDR_ANY)))),
+                (append_args("test_double_connect_same_addr"),
+                    Box::new(move || test_double_connect(sock_type, flag, false))),
```
sporksmith, line 66: The reader has to look pretty far away to understand what 'true' and 'false' means in this context. Consider annotating what bool parameters are at call sites (or avoiding boolean params in favor of an enum).

```
test_double_connect(sock_type, flag, /* change_address= */false)
```

or

```
enum ChangeAddress{ True, False };
...
test_double_connect(sock_type, flag, ChangeAddress::False)
```

Nice writeup about this issue here: https://abseil.io/tips/94

src/test/socket/connect/test_connect.rs
```diff
@@ -0,0 +1,526 @@
...
+        sin_family: libc::AF_INET as u16,
+        sin_port: 0u16.to_be(),
+        sin_addr: libc::in_addr {
+            s_addr: server_interface.to_be(),
+        },
+        sin_zero: [0; 8],
+    };
+
+    // bind on the server address
+    let rv = unsafe {
```
sporksmith, line 327: Optional: in places like this the `assert` can be inlined, letting us get rid of the named variable. This is helpful to someone reading the code since they don't have to try to remember what `rv` is (not knowing whether or not it's used later). Unlike C/C++ `assert`s are never compiled out (only `debug_assert`s)

src/test/socket/connect/CMakeLists.txt
stevenengler: Added.

src/test/socket/connect/CMakeLists.txt
stevenengler: Added.

src/test/socket/connect/test_connect.rs
stevenengler: Yeah I haven't really been happy with how the testing code is structured, so I'll try something like that outside of this PR.

src/test/socket/connect/test_connect.rs
stevenengler: Good idea, I went with the first since it's just a test, but for public interfaces the second approach seems like a good idea.

src/test/socket/connect/test_connect.rs
stevenengler: I like to keep things that change state outside of assertions since I find them easy to miss when scanning through code (and rustfmt can make them confusing when they use several lines), but I definitely agree with not polluting the function with extra variables, especially these test functions that can get fairly long.

Instead I'm now using explicit scopes so that it's easier to see that `rv` won't be used again later:

```
{
    let rv = unsafe { libc::close(fd) };
    assert_eq!(rv, 0);
}
```

PR #960, opened 2020-09-03
Restructured part of the socket API tests

stevenengler:
An attempt to simplify and standardize the tests. The functionality should not have changed in this PR, only the way that the tests are structured.

### DIFF
--- a/src/test/socket/accept/test_accept.rs
+++ b/src/test/socket/accept/test_accept.rs
@@ -24,11 +24,13 @@ fn main() -> Result<(), String> {
     // should we summarize the results rather than exit on a failed test
     let summarize = std::env::args().any(|x| x == "--summarize");
 
-    let tests = if run_only_passing_tests {
-        get_passing_tests()
-    } else {
-        get_all_tests()
-    };
+    let mut tests = get_tests();
+    if run_only_passing_tests {
+        tests = tests
+            .into_iter()
+            .filter(|x| x.shadow_passing() == test_utils::ShadowPassing::Yes)
+            .collect()
+    }
 
     test_utils::run_tests(&tests, summarize)?;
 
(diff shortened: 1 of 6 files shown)


src/test/socket/accept/test_accept.rs
```diff
@@ -24,32 +24,53 @@ fn main() -> Result<(), String> {
...
-    };
+    let mut tests = get_tests();
+    if run_only_passing_tests {
+        tests = tests
+            .into_iter()
+            .filter(|x| x.shadow_passing() == test_utils::ShadowPassing::Yes)
+            .collect()
+    }
 
-    test_utils::run_tests(&tests, summarize)?;
```
sporksmith, line 33: Why pull this conditional out of test_utils::run_tests?

src/test/test_utils.rs
```diff
@@ -5,6 +5,133 @@
 
 //! Utilities helpful for writing Rust integration tests.
 
+use std::error;
+use std::fmt;
+
+#[derive(Debug)]
```
sporksmith, line 11: Is it worth having a custom error type here, rather than using `Box<dyn Error>` and letting the test functions return whatever they want? (e.g. see proc_maps.rs for example)

sporksmith (approved):
Some minor questions/comments, but approving since I think this is a net-improvement as-is and don't want to leave you blocked until Tuesday

src/test/test_utils.rs
stevenengler: I chose a custom error type rather than a `dyn Error` since there's no nice way to pass custom information in a trait object if we want that in the future. I have changed it to use generics instead of a custom type or trait object, so it's more flexible than before, but also allows tests to pass custom information in errors since it uses a concrete type.

src/test/socket/accept/test_accept.rs
stevenengler: I did that since when `summarize == true`, it will never return an `Err` even if a test failed. I didn't want the caller to think that the tests passed just because it returned `Ok`. I have changed it back to how it was before though since it was overly complex.

PR #961, opened 2020-09-03
Clear the signal mask at startup

stevenengler:
We don't technically need this at the moment now that we've switched to SIGKILL for child processes, but since the current behaviour might be unexpected in the future, probably best to add this now before I forget about it.

### DIFF
--- a/src/main/core/main.c
+++ b/src/main/core/main.c
@@ -189,6 +189,12 @@ gint main_runShadow(gint argc, gchar* argv[]) {
         return EXIT_FAILURE;
     }
 
+    /* unblock all signals in shadow and child processes since cmake's ctest blocks SIGTERM (and
+     * maybe others) */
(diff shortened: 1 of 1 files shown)


stevenengler:
@sporksmith I'm not sure if you missed this one, or maybe my wording above sounded like I didn't want to merge this. Anyways, just pinging you :) I think it would be good to add this to reduce the possibility of unexpected signal-based problems in the future.

sporksmith:
> @sporksmith I'm not sure if you missed this one, or maybe my wording above sounded like I didn't want to merge this. Anyways, just pinging you :) I think it would be good to add this to reduce the possibility of unexpected signal-based problems in the future.

Oops, yeah I think I'd missed that you'd requested a review. Thanks for pinging!

sporksmith (approved):

PR #967, opened 2020-09-04
Use valid sockaddr when testing invalid file descriptors

stevenengler:
Should improve the reliability of socket API integration tests on different kernels.


sporksmith (approved):

stevenengler:
@robgjansen This should fix the issue you were having with the tests in your VM. If it doesn't, please let me know.

PR #968, opened 2020-09-08
Run additional tests using patched interposable libc.

sporksmith:
Right now some preload tests require a libc that's been patched to make interposable syscalls. This enables those tests in the "ilibc" ctest configuration, and updates our CI to install the patched libc and run those tests on centos 7.


robgjansen (approved):
LGTM!

Are there additional preload test cases that we want to enable in the `ilibc` configuration, other than just `phold` and `file`?

sporksmith:
Enabled epoll as well. I *think* that's the only other ones that were enabled for ptrace but not for preload. (grepping for `FIXME` in the test build files)

I also stumbled across the tcp tests - while src/test/tcp/CMakeLists.txt has specs for shadow-ptrace tests, the whole tcp directory is disabled in src/test/CMakeLists.txt.

PR #972, opened 2020-09-10
Convert existing tests from xml to yaml, and use the yaml2xml converter in the test cases

florian-vuillemot:
https://github.com/shadow/shadow/issues/773


robgjansen:
FYI, @stevenengler from our dev team will be helping move this along :)

stevenengler:
Once you're ready for me to take a look, just mark me as a reviewer (I'm not sure if you're still making changes).

florian-vuillemot:
@stevenengler yes please, you could. Don't hesitate to check the recent comment of @robgjansen on the https://github.com/shadow/shadow/issues/773 to be sure we all follow the same idea.
I let the "WIP", because I will work on other tests once you will have validated this first step.

src/test/bind/CMakeLists.txt
```diff
@@ -5,5 +5,11 @@ add_executable(test-bind test_bind.c ../test_common.c)
 
 ## register the tests
 add_test(NAME bind COMMAND test-bind)
-add_test(NAME bind-shadow-ptrace COMMAND ${CMAKE_BINARY_DIR}/src/main/shadow --interpose-method=ptrace -l debug -d bind.shadow-ptrace.data ${CMAKE_CURRENT_SOURCE_DIR}/bind.test.shadow.config.xml)
-add_test(NAME bind-shadow-preload COMMAND ${CMAKE_BINARY_DIR}/src/main/shadow --interpose-method=preload -l debug -d bind.shadow-preload.data ${CMAKE_CURRENT_SOURCE_DIR}/bind.test.shadow.config.xml)
+add_test(
+        NAME bind-shadow-ptrace
+        COMMAND sh -c "python3 ${CMAKE_SOURCE_DIR}/src/tools/convert.py yaml2xml ${CMAKE_CURRENT_SOURCE_DIR}/bind.test.shadow.config.yaml --output - | ${CMAKE_BINARY_DIR}/src/main/shadow --interpose-method=ptrace -l debug -d bind.shadow-ptrace.data -"
```
stevenengler, line 10: I think breaking the command (and the one below) over two lines would help readability here. (Most of our tests currently use only a single line because we just recently upgraded to a CMake version that supports splitting them across lines.)

stevenengler (changes requested):
Looks good, just added a few changes, and after those feel free to convert the other test cases.

src/tools/convert.py
```diff
@@ -110,32 +112,43 @@ def xml_nodes_to_dict(xml_nodes):
...
+    stream = None
     if args.output:
-        return args.output
-    return args.filename.replace(original_extension, target_extension)
+        filename = '/dev/stdout' if '-' == args.output else args.output
+        stream = open(filename, 'w', encoding='utf8')
+    else:
+        stream = open(args.filename.replace(original_extension, target_extension), 'w', encoding='utf8')
+    yield stream
+    stream.close()
```
stevenengler, line 136: What's the reason for the context manager here? The returned file object should already implement `__enter__()` and `__exit__()`, so I think the `with get_output_stream(...) as stream` below should work fine without this extra contextmanager decorator. For example:

```python
def f():
   return open('/tmp/abc', 'w')
with f() as stream:
   print(stream.closed)
print(stream.closed)
```

Returns:

```
False
True
```

stevenengler (changes requested):

src/tools/convert.py
florian-vuillemot: ahah ! Of course ! I didn't think about it but you are totally right ! Thanks :)

florian-vuillemot:
@stevenengler everything looks good to you ? Can I spread this method on others tests ?

stevenengler:
> @stevenengler everything looks good to you ? Can I spread this method on others tests ?

Looks good to me, thanks for making the changes! And yep, adding this to the other tests sounds good.

stevenengler:
If you haven't started on the `src/test/socket/*` tests, you can leave those and I can do them later since I'm currently modifying them. (If you've already started them that's fine too.)

florian-vuillemot:
@stevenengler ok I will ignore them

florian-vuillemot:
@sporksmith I let `src/test/phold/phold.test.shadow.config.xml` and `src/test/file/file.test.shadow.config.xml` because there is a commentary in them. Should I update the script to take care of the comment or should I remove them or put them in another place ?

florian-vuillemot:
@stevenengler , could you give me a hand with the centos7 error ?

stevenengler:
@florian-vuillemot: it looks like the xml2yaml script isn't converting the `environment` attribute on the shadow element.

https://github.com/shadow/shadow/blob/d7f05166774ae196f986070e6116386e754218db/src/test/file/file.test.shadow.config.xml#L1-L6

florian-vuillemot:
Well, I will pay more attention the next time I will check the conversion.. Thank you !

stevenengler:
No problem, is this a bug in the script that we should fix? I took a quick look but I don't see an obvious reason why it would skip that attribute.

sporksmith:
> @sporksmith I let `src/test/phold/phold.test.shadow.config.xml` and `src/test/file/file.test.shadow.config.xml` because there is a commentary in them. Should I update the script to take care of the comment or should I remove them or put them in another place ?

Updating the script to preserve comments would be a nice feature, though manually recreating the comments would also be ok.

florian-vuillemot:
@stevenengler after having tested the `convert.py` script I think I might have forgotten to run the script on the new (merged) XML file with the `environment` option  ^^

florian-vuillemot:
> > @sporksmith I let `src/test/phold/phold.test.shadow.config.xml` and `src/test/file/file.test.shadow.config.xml` because there is a commentary in them. Should I update the script to take care of the comment or should I remove them or put them in another place ?
> 
> Updating the script to preserve comments would be a nice feature, though manually recreating the comments would also be ok.

I just copy/past commentaries because it is quick, simple and I don't think we will have a huge amount of files with commentaries, so automation does not look mandatory. If I'm wrong, don't hesitate to open an issue and assign it to me.

src/test/clone/CMakeLists.txt
```diff
@@ -12,6 +12,10 @@ target_link_libraries(test_clone ${GLIB_LIBRARIES}
...
-add_test(NAME clone-shadow-ptrace COMMAND
-    ${CMAKE_BINARY_DIR}/src/main/shadow --interpose-method=ptrace -l debug -d
-    clone.shadow_ptrace.data ${CMAKE_CURRENT_SOURCE_DIR}/clone_test_config.xml)
+add_test(
+	NAME clone-shadow-ptrace
+	COMMAND sh -c "\
+        ${yaml2xml} ${CMAKE_CURRENT_SOURCE_DIR}/clone_test_config.yaml --output - \
+	| ${CMAKE_BINARY_DIR}/src/main/shadow --interpose-method=ptrace -l debug -d clone.shadow_ptrace.data - \
+	"
+)
```
stevenengler, line 21: A lot of the tests are mixing spaces and tabs for indentation, could you modify the indentation to be consistent within each test, and preferably between tests? (Ideally 4 spaces for indentation, but being consistent is the most important.)

For example, this test might be formatted like:

```
add_test(
    NAME clone-shadow-ptrace
    COMMAND sh -c "\
    ${yaml2xml} ${CMAKE_CURRENT_SOURCE_DIR}/clone_test_config.yaml --output - \
    | ${CMAKE_BINARY_DIR}/src/main/shadow --interpose-method=ptrace -l debug -d clone.shadow_ptrace.data - \
    "
)
```

stevenengler (changes requested):
Other than the indentation, looks good to me. I'll take another look for typos in a final review, but I didn't see any on this pass.

robgjansen:
FYI, yesterday I merged another test in `src/test/futex` that should probably also be converted. Sorry!

src/test/clone/CMakeLists.txt
florian-vuillemot: arff good catch !

florian-vuillemot: A configuration error with my IDE. I updated and its looks good now

src/test/socket/socketpair/socketpair.shadow.config.yaml
```diff
@@ -0,0 +1,32 @@
+topology:
```
stevenengler, line 1: Need to delete the `src/test/socket/socketpair/socketpair.shadow.config.xml` file.

stevenengler (changes requested):
Almost there!

stevenengler:
I also just noticed that there are two other directories that need to be converted, `src/test/tcp` and `src/test/poll`. Some of these are currently commented out, but I hope to re-enable them soon, so they would be good to convert.

stevenengler (approved):

stevenengler:
Thanks @florian-vuillemot for the changes! It's great to be another step closer to full yaml support. I've checked the box at https://github.com/shadow/shadow/issues/773#issue-605046433 since I'm not sure which accounts are allowed to edit comments by other accounts.

robgjansen:
Woo! Thanks for all the work to get this merged! 🥇

PR #973, opened 2020-09-10
Disabled socketpair() tests in shadow

stevenengler:
When I enabled the shadow tests for `socketpair()`, I forgot to check that Shadow actually supported remapping the `socketpair()` system call rather than just calling the native system call. This PR disables the tests in shadow.


robgjansen (approved):
This is exactly what I had in mind. Happened to look at it, so thought I might as well submit a review :)

stevenengler:
Thanks! Just rebasing and then I'll merge.

Issue #842, opened 2020-06-11
Thoroughly test the shutdown() syscall


robgjansen:
See #832 for an overview of our approach.

The `src/test/shutdown` test is fairly extensive and a good place to start.

PR #975, opened 2020-09-11
Added tests for the shutdown() call

stevenengler:
Closes shadow/shadow#842.


src/test/test_utils.rs
```diff
@@ -163,3 +163,58 @@ pub fn get_errno_message(errno: i32) -> String {
...
+    expected_errnos: &[libc::c_int],
+    line: u32,
+) -> Result<libc::c_int, String>
+where
+    F: FnOnce() -> libc::c_int,
+{
+    let rv = f();
+    let errno = get_errno();
+
+    match expected_errnos.is_empty() {
```
sporksmith, line 187: `if` rather than `match`?

src/test/test_utils.rs
```diff
@@ -163,3 +163,58 @@ pub fn get_errno_message(errno: i32) -> String {
     }
     cstr.to_string_lossy().into_owned()
 }
+
+/// Calls check_system_call(), but automatically passes the current line number.
+#[macro_export]
+macro_rules! check_system_call {
```
sporksmith, line 169: Not sure I understand the motivation for passing the line number to `check_system_call` and wrapping it in a macro to do so. If the caller were to do `check_system_call(f, expected_errnos).unwrap()`, wouldn't that achieve the same?

src/test/test_utils.rs
```diff
@@ -8,22 +8,22 @@
...
+pub enum TestEnvironment {
+    Shadow,
+    Libc,
 }
 
 pub struct ShadowTest<T, E> {
     name: String,
     func: Box<dyn Fn() -> Result<T, E>>,
-    shadow_passing: ShadowPassing,
+    passing: Vec<TestEnvironment>,
```
sporksmith, line 19: Optional - it's a *little* weird to use a `Vec` when conceptually it's representing a set. Is there a reason not to use `HashSet`?

src/test/socket/shutdown/test_shutdown.rs
```diff
@@ -0,0 +1,974 @@
...
+}
+
+/// Generate a pair of connected TCP sockets.
+fn setup_stream_sockets(domain: libc::c_int, flag: libc::c_int) -> (libc::c_int, libc::c_int) {
+    let fd_client = unsafe { libc::socket(domain, libc::SOCK_STREAM | flag, 0) };
+    let fd_listener = unsafe { libc::socket(domain, libc::SOCK_STREAM | flag, 0) };
+    assert!(fd_client >= 0);
+    assert!(fd_listener >= 0);
+
+    {
```
sporksmith, line 255: Optional throughout, but I still slightly prefer keeping things a bit more contact by inlining the function call rather than introducing a local variable and limiting its lifetime with a new scope. e.g.

```
assert_eq!(unsafe { libc::listen(fd_listener, 10) }, 0);
```

sporksmith (approved):

src/main/host/descriptor/udp.c
```diff
@@ -24,12 +24,36 @@
 #include "main/utility/utility.h"
 #include "support/logger/logger.h"
 
+enum UDPState {
```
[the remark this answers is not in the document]
stevenengler: I used this format in order to keep the same structure as `tcp.c`. Should I update the enums and prefixes for file-scoped symbols in `tcp.c` as well?

src/main/host/descriptor/udp.c
```diff
@@ -24,12 +24,36 @@
...
+    UDPS_CLOSED, UDPS_ESTABLISHED,
+};
+
+static const gchar* udpStateStrings[] = {
```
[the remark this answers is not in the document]
stevenengler: Same question as above.

src/main/host/descriptor/udp.c
```diff
@@ -24,12 +24,36 @@
...
+    "UDPS_CLOSED", "UDPS_ESTABLISHED",
+};
+
+static const gchar* udp_stateToAscii(enum UDPState state) {
```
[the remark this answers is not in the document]
stevenengler: Same question as above.

src/main/host/syscall/socket.c
```diff
@@ -580,7 +598,12 @@ SysCallReturn _syscallhandler_sendtoHelper(SysCallHandler* sys, int sockfd,
         /* TODO: Dynamically compute size based on how much data is actually
          * available in the descriptor. */
         size_t sizeNeeded = MIN(bufSize, SYSCALL_IO_BUFSIZE);
-        const void* buf = process_getReadablePtr(sys->process, sys->thread, bufPtr, sizeNeeded);
+
+        const void* buf = NULL;
+        /* If the size is 0, nothing should be trying to read the memory so NULL should be okay. */
+        if (sizeNeeded > 0) {
```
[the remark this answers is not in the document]
stevenengler: It worked for ptrace but not preload. I've fixed this for preload, and added additional checks for ptrace so that we don't keep record of shared memory blocks of size 0.

src/main/host/syscall/socket.c
```diff
@@ -1042,19 +1070,27 @@ SysCallReturn syscallhandler_shutdown(SysCallHandler* sys,
...
     }
 
-    return (SysCallReturn){
-        .state = SYSCALL_DONE,
-        .retval.as_i64 = (int64_t)tcp_shutdown(tcp_desc, how)};
+    TCP* tcp_desc = NULL;
+    errcode = _syscallhandler_validateTCPSocketHelper(sys, sockfd, &tcp_desc);
+    if (errcode >= 0) {
+        return (SysCallReturn){
+            .state = SYSCALL_DONE, .retval.as_i64 = (int64_t)tcp_shutdown(tcp_desc, how)};
```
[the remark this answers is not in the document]
stevenengler: Changed this.

src/test/test_utils.rs
stevenengler: Good catch, thanks.

src/test/test_utils.rs
stevenengler: All of the socket tests have the `summarize` option which should run all of the tests, even if some fail. So you can't unwrap results, and instead need to propagate them which means you lose the location of the error. The `--summarize` option is useful for debugging, but maybe it's not worth the added code complexity.

src/test/test_utils.rs
stevenengler: I used `Vec` to reduce the verbosity since it makes the test cases easier to read (`vec![A, B]` vs `[A, B].iter().cloned().collect()` for each test case), but I've changed it to a `HashSet` since it is the right data type.

src/test/socket/shutdown/test_shutdown.rs
stevenengler: I'm still not convinced that this is more readable, especially with larger function calls and if you have assert messages, but I converted the ones that will fit on a single line.

stevenengler:
@sporksmith Can you take a look at commit 2de33f236 and make sure that change looks reasonable to you?

src/main/host/descriptor/udp.c
```diff
@@ -212,9 +238,20 @@ static gboolean _udp_close(Descriptor* descriptor) {
     UDP* udp = _udp_fromDescriptor(descriptor);
     MAGIC_ASSERT(udp);
     /* Deregister us from the process upon return. */
+    _udp_setState(udp, UDPS_CLOSED);
     return TRUE;
 }
 
+gint udp_shutdown(UDP* udp, gint how) {
```
florian-vuillemot, line 245: Why passing the `how` argument here ? This argument looks to indicate how to close the file descriptor (`how != SHUT_RD && how != SHUT_WR && how != SHUT_RDWR`) but I have never heard about a way to close file descriptor depending on the way it was opened... Could you please explain me how it works?

stevenengler: For UDP sockets, we aren't currently using the `how` argument, but we should be if we wanted our UDP tests to pass in Shadow. This `udp_shutdown()` function should imitate the `int shutdown(int sockfd, int how)` [system call](https://man7.org/linux/man-pages/man2/shutdown.2.html), which allows you to separately shut down the read or write ends of a socket. For TCP, calling `shutdown(fd, SHUT_WR)` would send a `FIN` packet and prevent you from writing any more to the socket, but still allowing you to continue reading on the socket. For UDP this `shutdown()` call is much less meaningful and I think should just set a flag to prevent the program from either reading or writing new messages.

src/main/host/descriptor/udp.c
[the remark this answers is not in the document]
sporksmith: Oh, that's surprising. It looks like Rob wrote it that way, and that's who I thought established/preferred the other convention. I'm fine with leaving as-is (though maybe @robgjansen will want to chime in?)

src/main/host/descriptor/udp.c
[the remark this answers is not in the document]
sporksmith: I'd lean towards using the prefix here, and perhaps adding it in tcp.c.

src/test/test_utils.rs
sporksmith: Ahhh ok.

I'm trying to think how we can avoid writing macros for every such helper function. Given that this is the first time it's come up, maybe it's fine to proceed as-is.

Another option might be have a utility function that takes the list of tests, config, and runs them (this might be a good idea anyway), and just have it catch panics in individual tests and stuff them into an error for the summary. This would let us use built in assert helpers, unwrap, etc.

src/test/socket/shutdown/test_shutdown.rs
sporksmith: There's definitely a bit of a tradeoff here. The actual assertion might be a little easier to understand with the temp variable and new scope, but adding so many more lines has a cost of making the code more sparse, making it difficult to look at as much at once and see the overall structure, etc.

I like how you have it now; I agree that it becomes more worth it to break it up with longer / more-complex calls and assertions.

src/test/test_utils.rs
sporksmith: Looking at it again, just using two separate fields might make the most sense since we don't expect the number of possible environments to grow beyond 2.

I'm fine with leaving as you have it now, and agree the extra verbosity is annoying but not enough to prefer a `Vec`. I wouldn't be opposed to writing a `set!` macro analogous to the `vec!` macro :)

sporksmith:
Done; sorry for all the individual comments instead of batching into a review. I couldn't get the 'files changed' tab to show me all the open comment threads, and the 'conversation' tab doesn't have the review option. (It might have been a little easier without the rebase, since that "orphan"d the comments)

src/main/host/descriptor/udp.c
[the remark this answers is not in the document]
robgjansen: Looks like I wrote most of those non-typedef'd TCP enums 9 years ago. My guess is that I didn't realize you could typedef enums at the time :-)

I don't like writing out the `struct` and `enum` prefixes all the time. Also, typedef'ing structs is useful for modularization, because you can hide the struct definition inside the implementation c file. Obviously that's not relevant to objects that are only ever defined and used at the file level, though, but being consistent in both cases is nice.

Maybe a compromise would be to always typedef all structs, but don't typedef enums? That way if you don't see `enum` you know it's a `struct`.

In any case though, this is one of those bike shed issues and I don't think it's really worth going through and fixing this everywhere in this PR. We can discuss further if we think the issue is causing enough confusion/annoyance to warrant effort on refactoring rather than just replacing with Rust code ;)

src/test/test_utils.rs
stevenengler: I thought about this a long time ago, but my reason at the time was that you probably also want to suppress the panic message (including the backtrace if the environment variable is set), so you need to also set the global panic hook which can cause issues in rust's standard test runner which uses multi-threading. But since we aren't using rust's test runner and are not using multi-threading, this may be fine to do.

src/main/host/descriptor/udp.c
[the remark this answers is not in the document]
stevenengler: Added the prefix for these file-scoped symbols (in `tcp.c` as well).

src/test/test_utils.rs
stevenengler: I'm a little worried that in the future we might also want "Shadow-ptrace" and "Shadow-preload" options as well if a test only works under one of the runtimes. So I think the flexibility of using a set is good for now, but we can revisit it later.

src/main/host/descriptor/udp.c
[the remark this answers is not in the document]
stevenengler: Okay thanks, leaving for now, but I did add the `_` prefix as discussed below.

stevenengler:
@sporksmith Sorry about the rebase, it wasn't until I pushed it that I realized it was probably a mistake. And I didn't know that you could reply to comments as part of the "review" process to avoid sending a bunch of emails. I'll try that next time.

Issue #951, opened 2020-08-28
phold off-by-one error


rwails:
**Describe the issue**

The phold test has an off-by-one error in the way hosts talk to each other / are weighted. I think the weights file should contain `n` lines for `n` hosts, but seems to only require `n-1` lines. One host during the experiment doesn't receive traffic from the other hosts.

(rwails to fill in more.)

**To Reproduce*

Run phold :-)

RWails:
Fixed from #976.

PR #976, opened 2020-09-11
Fix ThreadPtrace migration

sporksmith:
When running phold-threaded-shadow-ptrace with more hosts (e.g. 20
instead of the default 10), we'd get errors in threadptrace_doDetach. It
turns out that a host can end up calling detach on a plugin that's
already detached.

Since threadptrace_doDetach was assuming it was starting from an
attached state, things didn't go well.

We now simply silently return if already detached.

### DIFF
--- a/src/main/host/thread_ptrace.c
+++ b/src/main/host/thread_ptrace.c
@@ -504,6 +504,10 @@ static void _threadptrace_doAttach(ThreadPtrace* thread) {
 
 static void _threadptrace_doDetach(ThreadPtrace* thread) {
     utility_assert(thread->childState == THREAD_PTRACE_CHILD_STATE_SYSCALL);
+    if (thread->needAttachment) {
+        // We're already detached.
+        return;
+    }
 
     // First rewind the instruction pointer so that the current syscall will be
     // retried when we resume. This isn't *strictly* necessary, but it's not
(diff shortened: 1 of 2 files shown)


robgjansen (approved):
👍 

It looks like e40e9c0 fixes #951, and I didn't find an issue for the ThreadPTrace Migration issue.

sporksmith:
> It looks like [e40e9c0](https://github.com/shadow/shadow/commit/e40e9c0897379e9e2b5a994dbe5c63a58160b189) fixes #951

Not that I'm aware of. This fixes an issue when there's a newline at the end of the weights file. The checked in weights file doesn't have one, but when editing locally vim adds one whether I want it to or not :-D.

> , and I didn't find an issue for the ThreadPTrace Migration issue.

Yeah, I didn't bother creating a separate issue for that one. I can if it's useful for bookkeeping, but I figured explaining the issue in the PR + git commit was probably enough.

robgjansen:
> Not that I'm aware of

Oh, sorry, invalid assumption on my part!

> Yeah, I didn't bother creating a separate issue for that one. I can if it's useful for bookkeeping, but I figured explaining the issue in the PR + git commit was probably enough.

PR + git commit is fine, since the PR is linked to the milestone too :)

PR #979, opened 2020-09-14
Avoid performing an atomic write in spinlock loop.

### LINKED ISSUE #977: Avoid writes on every spinlock loop
IIUC, this is an optimization that will invalidate cache less often.

sporksmith:
In my local tests this is performance-neutral. I suspect this is because the spinlock isn't under contention, so it doesn't matter if we're invalidating the cache on other cores. It could end up mattering if there happen to be other shared data in the same cache lines/pages, or we change things s.t. the lock could be under contention, though.

Fixes #977


robgjansen:
Is this no longer a WIP draft, since a review was requested?

sporksmith:
> Is this no longer a WIP draft, since a review was requested?

Oops, right

rwails (approved):
I ran this change through the phold benchmarks and noticed a very slight (perhaps non-significant) improvement in performance. I think worth it to merge!

PR #981, opened 2020-09-14
Improve rust build process

stevenengler:
Set Cargo's target directory to the 'build/' directory (so now 'cargo doc' and others will put files in the 'build/' directory), and stop CMake from adding unnecessary files into the 'src/' directory.

The Cargo workspace was removed since it was not working correctly, and we cannot share dependencies between packages because of different `RUSTFLAGS`.


robgjansen (approved):
Thank you!

PR #982, opened 2020-09-14
Syscall handler for futex

robgjansen:
This add a simple futex test case and a minimal implementation of futex in Shadow. Support was only implemented for basic `FUTEX_WAIT` (with or without timeout) and `FUTEX_WAKE` operations.

Rather than subtyping descriptor, I ended up doing some refactoring on the "Status" and "Listener" bits that help us implement blocking syscalls on descriptors in order to extend the `SyscallCondition` to support a futex as well. This design is more extensible in case we ever want to add new types of blocking objects in the future (maybe signals?).


robgjansen:
The thing that is missing here is what to do when a cloned thread exits and the main thread calls `pthread_join`. Since I didn't implement support for that yet, what currently happens is the main thread blocks indefinitely on a futex. So we'll need to handle this case.

src/main/host/syscall/futex.c
```diff
@@ -0,0 +1,164 @@
...
+
+            // TODO: does this load/compare need to be done atomically?
+            // `man 2 futex`: blocking via a futex is an atomic compare-and-block operation
+            const uint32_t* futexVal =
+                process_getReadablePtr(sys->process, sys->thread, uaddrptr, sizeof(uint32_t));
+
+            debug("Futex value is %" PRIu32 ", expected value is %" PRIu32, *futexVal, expectedVal);
+            if (*futexVal != expectedVal) {
+                debug("Futex values don't match, try again later");
+                result = -EAGAIN;
```
sporksmith, line 82: Should this only be if `!_syscallhandler_wasBlocked(sys)`? e.g.:

* futex F has value X
* thread 1 waits on futex F with expected value X, causing it to get blocked
* thread 2 assigns Y to futex F and does a FUTEX_WAKE on it.

I believe in the current code thread 1 would be woken up but get -EAGAIN, whereas it should get woken up and get 0.

src/main/host/syscall/futex.c
```diff
@@ -0,0 +1,164 @@
...
+            FutexTable* ftable = host_getFutexTable(sys->host);
+            Futex* futex = futextable_get(ftable, faddr);
+
+            if (_syscallhandler_wasBlocked(sys)) {
+                utility_assert(futex != NULL);
+
+                // We already blocked on wait, so this is either a timeout or wakeup
+                if (timeout && _syscallhandler_didListenTimeoutExpire(sys)) {
+                    // Timeout while waiting for a wakeup
+                    debug("Futex %p timeout out while waiting", faddr);
```
sporksmith, line 96: Is there a possible race condition here if thread A does a WAKE-1 operation, waking thread B, but when it gets here finds the timeout *just* expired? Or does Shadow's scheduling guarantee that can't happen?

src/main/host/syscall/futex.c
```diff
@@ -0,0 +1,164 @@
...
+                }
+
+                // Dynamically clean up the futex if needed
+                if (futex_getListenerCount(futex) == 0) {
+                    debug("Dynamically freed a futex object for futex addr %p", faddr);
+                    bool success = futextable_remove(ftable, futex);
+                    utility_assert(success);
+                }
+                break;
+            }
```
sporksmith, line 111: It looks like when we intend to wake up a thread, we still fall through and end up returning SYSCALL_BLOCK? (and possibly using a now-freed futex pointer?)

It looks like there should be a `return` here?

src/main/host/syscall/futex.c
```diff
@@ -0,0 +1,164 @@
...
+
+            FutexTable* ftable = host_getFutexTable(sys->host);
+            Futex* futex = futextable_get(ftable, faddr);
+
+            debug("Found futex %p at futex addr %p", futex, faddr);
+
+            if (futex && val > 0) {
+                debug("Futex waking %i waiters", val);
+                result = futex_wake(futex, (unsigned int)val);
+            }
```
sporksmith, line 141: Maybe worth a `debug` message if there were no waiters?

src/main/host/syscall_condition.c
```diff
@@ -39,22 +38,32 @@ struct _SysCallCondition {
...
+            case TRIGGER_DESCRIPTOR: {
+                descriptor_ref(cond->trigger.object.as_descriptor);
+                break;
+            }
+            case TRIGGER_FUTEX: {
+                futex_ref(cond->trigger.object.as_futex);
+                break;
+            }
+            case TRIGGER_NONE:
+            default: {
```
sporksmith, line 63: warning in default case?

src/main/host/syscall_condition.c
```diff
@@ -112,8 +134,21 @@ static void _syscallcondition_free(SysCallCondition* cond) {
...
+            case TRIGGER_DESCRIPTOR: {
+                descriptor_unref(cond->trigger.object.as_descriptor);
+                break;
+            }
+            case TRIGGER_FUTEX: {
+                futex_unref(cond->trigger.object.as_futex);
+                break;
+            }
+            case TRIGGER_NONE:
+            default: {
```
sporksmith, line 148: Make exhaustive and/or warning for unhandled enumerator?

src/main/host/syscall_condition.c
sporksmith: alternatively no `default` case at all to get a compiler warning if another enumerator gets added later without updating this switch; `return` statements in each case; and a `warning` or `error` after the whole statement to detect at runtime if we got a bad enumerator here.

Blatant self-promotion (a bit Google-centric though): https://abseil.io/tips/147

e.g.
```
        switch (cond->trigger.type) {
            case TRIGGER_DESCRIPTOR: {
                descriptor_ref(cond->trigger.object.as_descriptor);
                return;
            }
            case TRIGGER_FUTEX: {
                futex_ref(cond->trigger.object.as_futex);
                return;
            }
            case TRIGGER_NONE: {
                 return;
            }
            // no default, forcing a compiler warning if not kept up to date
        }
        // Complain loudly if we get a non-enumerator at run-time.
        error("Unhandled enumerator %d", cond->trigger.type);
```

src/main/host/syscall_condition.c
```diff
@@ -68,24 +77,37 @@ static void _syscallcondition_cleanupListeners(SysCallCondition* cond) {
...
+                descriptor_removeListener(
+                    cond->trigger.object.as_descriptor, cond->triggerListener);
+                break;
+            }
+            case TRIGGER_FUTEX: {
+                futex_removeListener(cond->trigger.object.as_futex, cond->triggerListener);
+                break;
+            }
+            case TRIGGER_NONE:
+            default: {
```
sporksmith, line 100: Make exhaustive and/or warning for unhandled enumerator?

src/main/host/syscall_condition.c
```diff
@@ -148,11 +183,27 @@ static void _syscallcondition_logListeningState(SysCallCondition* cond,
...
+                break;
+            }
+            case TRIGGER_FUTEX: {
+                g_string_append_printf(string, "status on futex %p%s",
+                                       futex_getAddress(cond->trigger.object.as_futex),
+                                       cond->timeout ? " and " : "");
+                break;
+            }
+            case TRIGGER_NONE:
+            default: {
```
sporksmith, line 201: Make exhaustive and/or warning for unhandled enumerator?

src/main/host/syscall_condition.c
```diff
@@ -167,6 +218,29 @@ static void _syscallcondition_logListeningState(SysCallCondition* cond,
...
+                return true;
+            }
+            break;
+        }
+        case TRIGGER_FUTEX: {
+            // Futex status doesn't change
+            return true;
+        }
+        case TRIGGER_NONE:
+        default: {
```
sporksmith, line 236: Make exhaustive and/or warning for unhandled enumerator?

src/main/host/syscall_condition.c
```diff
@@ -253,37 +327,53 @@ void syscallcondition_waitNonblock(SysCallCondition* cond, Process* proc,
...
+                /* Monitor the requested status an every status change. */
+                statuslistener_setMonitorStatus(
+                    cond->triggerListener, cond->trigger.status, SLF_ALWAYS);
+
+                /* Attach the listener to the descriptor. */
+                futex_addListener(cond->trigger.object.as_futex, cond->triggerListener);
+                break;
+            }
+            case TRIGGER_NONE:
+            default: {
```
sporksmith, line 373: Make exhaustive and/or warning for unhandled enumerator?

src/test/futex/test_futex.c
```diff
@@ -0,0 +1,137 @@
...
+#include <unistd.h>
+
+// Note: this test roughly follows the example at the end of `man 2 futex`
+
+#define PTR_TO_INT(p) ((int)(long)(p))
+#define INT_TO_PTR(i) ((void*)(long)(i))
+#define NUM_LOOPS 100
+
+// The futex word used to synchronize threads
+volatile int futex_word1 = 0; // initial state is unavailable
```
sporksmith, line 25: Might be a little easier to follow below with
```
#define UNAVAILABLE 0
#define AVAILABLE 1
```

src/test/futex/test_futex.c
```diff
@@ -0,0 +1,137 @@
...
+static int _futex_wait(volatile int* word) {
+    while (1) {
+        // Args are: ptr, expected old val, desired new val
+        bool is_available = 1 == __sync_val_compare_and_swap(word, 1, 0);
+
+        if (is_available) {
+            break;
+        } else {
+            int res = syscall(SYS_futex, word, FUTEX_WAIT, 0, NULL, NULL, 0);
+            if (res == -1 && errno != EAGAIN) {
```
sporksmith, line 38: If I'm right about the bug in the `futex` handler, then after woken we'll have EAGAIN and continue to loop; presumably forever. Was thinking maybe this is the "shadow exits without error even if we had to kill the plugin violently" problem, but since you pthread_join the child threads shouldn't that hang and cause test to timeout? Maybe pthread_join is returning before it should due to some other bug or missing functionality...?

Possible alternative to relying on pthread_join is to busy wait on some volatile "finished" flag as in the clone and (detached) pthread test. Seems like that'd be reasonable to do here since pthread_join itself depends on futex, so could cause confusing behavior here if futex isn't working correctly.

sporksmith (changes requested):
I know you don't have a lot more cycles to work on this - feel free to drop in TODOs/FIXMEs and have me follow up.

src/main/host/syscall/futex.c
```diff
@@ -0,0 +1,164 @@
...
+                if (!(timeout->tv_nsec >= 0 && timeout->tv_nsec <= 999999999)) {
+                    debug("A futex timeout was given, but the nanos value is out of range");
+                    result = -EINVAL;
+                    break;
+                }
+            } else {
+                timeout = NULL;
+            }
+
+            // TODO: does this load/compare need to be done atomically?
```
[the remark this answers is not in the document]
robgjansen: Yeah. Wasn't sure if atomic ops are necessary on volatile variables, but it doesn't appear so. :)

src/main/host/futex.c
```diff
@@ -0,0 +1,107 @@
...
+#include <glib.h>
+
+#include "main/core/support/definitions.h"
+#include "main/core/support/object_counter.h"
+#include "main/core/worker.h"
+#include "main/utility/utility.h"
+#include "support/logger/logger.h"
+
+struct _Futex {
+    // The unique address that is used to refer to this futex
```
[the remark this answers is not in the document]
robgjansen: Agree, we should do this. For now I'm stubbing out a function in process.c, and leaving the physical ptr conversion part as a TODO.

src/main/host/syscall/futex.c
robgjansen: I think you're right, good catch!

src/main/host/syscall/futex.c
robgjansen: The break should cause us to break out of the switch statement, and then we return done immediately after that. (I'm going to add `// return SYSCALL_DONE` after the breaks to make that more clear.)

src/main/host/syscall/futex.c
robgjansen: I think we can consider whichever callback gets executed first is the thing that happened first. In any case, the futex man page states that spurious wakeups can occur, so I wouldn't call it a race condition even if I'm wrong.

src/main/host/syscall/futex.c
sporksmith: Ah, I must have missed the `break` or incorrectly matched it up with the next brace.

sporksmith: Optional - maybe it'd be worth breaking the (implemented) cases out into helper functions to make it easier to see the overall structure?

src/test/futex/test_futex.c
robgjansen: Hmm... I'm not exactly sure what was going on there. I thought the atomics ops were preventing any infinite loops. In any case, I stopped using `pthread_join` in favor of the busy loop (as is done in the clone test) as suggested.

robgjansen:
I don't understand the lint warning or how to fix it. I changed up `typedef struct {uint64_t val} PluginPtr;` in `main/host/syscall_types.c` and the warnings seem to indicate it has something to do with that. I didn't touch `Thread`, though.

I played around with defining the Virtual and PhysicalPtr structs separately rather than creating aliases, but that didn't seem to help on my local machine so I didn't commit that yet.

sporksmith:
> I don't understand the lint warning or how to fix it. I changed up `typedef struct {uint64_t val} PluginPtr;` in `main/host/syscall_types.c` and the warnings seem to indicate it has something to do with that. I didn't touch `Thread`, though.
> 
> I played around with defining the Virtual and PhysicalPtr structs separately rather than creating aliases, but that didn't seem to help on my local machine so I didn't commit that yet.

Ah, you probably need to merge (or rebase onto) dev to get https://github.com/shadow/shadow/pull/993. (And maybe locally update and rebuild bindings?)

sporksmith (approved):

PR #985, opened 2020-09-15
Handle libpthread's way of allocating stacks

sporksmith:
libpthread mmap's a region to use as a new thread stack, initially with PROT_NONE. It then updates the protections using `mprotect`.

* Dropped a sanity check in our mmap-handling, that was broken because PROT_NONE is zero.
* Added handling for mprotect to update Shadow's bookkeeping and prot settings.


sporksmith:
Folded into #986.

PR #986, opened 2020-09-15
Support pthread_create

sporksmith:
* Handle PROT_NONE in mmap
* Handle mprotect syscall
* Handle gettid syscall
* Add and enable pthread_create detached thread test


src/main/host/syscall/clone.c
```diff
@@ -72,3 +72,8 @@ SysCallReturn syscallhandler_clone(SysCallHandler* sys, const SysCallArgs* args)
 
     return (SysCallReturn){.state = SYSCALL_DONE, .retval.as_i64 = thread_getID(child)};
 }
+
+SysCallReturn syscallhandler_gettid(SysCallHandler* sys, const SysCallArgs* args) {
+    utility_assert(sys && args);
+    return (SysCallReturn){.state = SYSCALL_DONE, .retval.as_i64 = thread_getID(sys->thread)};
+}
```
robgjansen, line 79: You beat me to it! I also [implemented `gettid`](https://github.com/shadow/shadow/pull/982/commits/dc6e586333523ff092290555d8f38d6d0bc1918c), but I put mine in `unistd.c` (alongside `getpid`).

robgjansen (approved):
LGTM! I'm really falling behind w.r.t. Rust though, so you might wanna get @stevenengler to look over the Rust bits.

src/main/host/syscall/clone.c
sporksmith: I considered putting it there, but since it isn't actually in unistd.h thought it made more sense to put it with "thread stuff" than "get an ID stuff". Happy to move it if you think unistd.c makes more sense though.

sporksmith:
@stevenengler could you take a look at the rust?

src/main/host/memory_manager.rs
```diff
@@ -1159,4 +1272,23 @@ mod export {
...
+
+    #[no_mangle]
+    pub unsafe extern "C" fn memorymanager_handleMprotect(
+        memory_manager: *mut MemoryManager,
+        thread: *mut c::Thread,
+        addr: c::PluginPtr,
+        size: usize,
+        prot: i32,
+    ) -> c::SysCallReg {
+        let memory_manager = &mut *memory_manager;
```
stevenengler, line 1284: Check for a null pointer here?

stevenengler (approved):
I don't really follow what's happening in `handle_mprotect()`, but the rust code generally looks good to me.

src/main/host/syscall/clone.c
robgjansen: It's fine. I'll remove mine from unistd.

robgjansen:
Oh yeah, I should have included the general comment that I would have liked some additional comments to explain what was going on in some of the Rust code. Seems like that could be helpful for Steven too.

src/main/host/syscall/clone.c
robgjansen: `set_tid_address` is a thread thing and is in unistd though, so who knows.

stevenengler:
Maybe for another PR, but I think the memory manager could use more top-level comments. For example, what the memory manager does and what it's responsible for, what do functions like `handle_mprotect` and `handle_mremap` actually do in the context of the memory manager, etc? The regions and mutations seem like a nice abstraction, but I find the code can be hard to follow before I understand what the general goal is.

src/main/host/memory_manager.rs
sporksmith: Good point. Done throughout.

sporksmith:
I have PR #987 q'd up that builds on this one, switching from println to the logging macros. I'll make a pass on the comments in that PR is as well.

PR #987, opened 2020-09-15
Memory manager clarity improvements

sporksmith:
* Switch from `println!` to logging macros.
* Use nix function wrappers instead of libc wrappers.
* Use nix Result and Error types, which use an Errno type, rather than just an i32 to represent errno errors.
* Simplify some pointer arithmetic using std::ptr::add, and util::ops::Range::len.
* Flesh out some comments.


src/main/host/memory_manager.rs
```diff
@@ -953,14 +1002,19 @@ impl MemoryManager {
...
-    // Extend the portion of the stack that we've mapped downward to include `src`.  This is
-    // carefuly designed *not* to invalidate any outstanding borrowed references or pointers, since
-    // otherwise a caller trying to marshall multiple syscall arguments might invalidate the first
-    // argument when marshalling the second.
+    // Extend the portion of the stack that we've mapped downward to include `src`.
+    //
+    // This is carefuly designed *not* to invalidate any outstanding borrowed references or
+    // pointers, since otherwise a caller trying to marshall multiple syscall arguments might
+    // invalidate the first argument when marshalling the second. (While the Rust API currently
+    // prevents there from being any outstanding references, the C API does not).
```
stevenengler, line 1010: Make these `///` so that they show up in the docs? (And for other functions in this file.)

stevenengler (commented):
Looks good!

src/main/host/memory_manager.rs
sporksmith: Does it make sense to use `///` for non-public methods? I suppose it doesn't hurt...

stevenengler: I think it makes sense, since if you're developing the library itself and not just using it in another project, you may run `cargo doc --document-private-items` which will include those functions. And rust seems to use `///` for non-public methods as well: https://doc.rust-lang.org/src/alloc/vec.rs.html#2978

src/main/host/memory_manager.rs
```diff
@@ -814,15 +847,17 @@ impl MemoryManager {
...
-                /* new_len: */ new_heap.end - new_heap.start,
-                /* flags: */ 0,
-                /* new_addr: */ PluginPtr::from(0usize),
-            );
-            assert!(res.is_ok());
+            thread
+                .native_mremap(
+                    /* old_addr: */ PluginPtr::from(self.heap.start),
+                    /* old_len: */ self.heap.end - self.heap.start,
+                    /* new_len: */ new_heap.end - new_heap.start,
```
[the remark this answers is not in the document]
sporksmith: Yup, missed these ones. Thanks!

src/main/host/memory_manager.rs
sporksmith: SG. Thanks!

robgjansen (commented):
Thanks for the extra comments! I'm happy if Steven's happy :)

stevenengler:
> Thanks for the extra comments! I'm happy if Steven's happy :)

I'm always happy so that's not a good basis for decision making :)

(I thought I approved this PR when I left a review, but I guess not, so approving now.)

stevenengler (approved):

PR #993, opened 2020-09-22
Updated the (c)bindgen version numbers for the CI

stevenengler:


sporksmith (approved):
Oof; I imagine a bit of a headache preceded this. At some point maybe we should revisit whether having checked-in bindings is worth it...

PR #994, opened 2020-09-23
Added `set!` macro for tests

stevenengler:
Based on the idea by @sporksmith in shadow/shadow#975.

### DIFF
--- a/src/test/socket/accept/test_accept.rs
+++ b/src/test/socket/accept/test_accept.rs
@@ -3,8 +3,8 @@
  * See LICENSE for licensing information
  */
 
-use test_utils::AsMutPtr;
 use test_utils::TestEnvironment as TestEnv;
+use test_utils::{set, AsMutPtr};
 
 struct AcceptArguments {
     fd: libc::c_int,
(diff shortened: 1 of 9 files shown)


sporksmith (approved):
Semi-orthogonal to this change, but maybe the TestCase ought to have the set of environments where we *don't* want it to run. The usual default we'd want is to run it in all environments, and we only want to explicitly exclude ones we know won't work.

stevenengler:
> Semi-orthogonal to this change, but maybe the TestCase ought to have the set of environments where we _don't_ want it to run. The usual default we'd want is to run it in all environments, and we only want to explicitly exclude ones we know won't work.

I tried this change, but then it still looks like the test is designed to run in the listed environments (for example the "shadow test environment" in the below example). We could improve this with something like `/* exclude_envs= */ set![TestEnv::Shadow]`, but if you have any suggestions let me know.

```rust
test_utils::ShadowTest::new(
    &append_args("test_write_after_peer_shutdown"),
    move || test_write_after_peer_shutdown(domain, sock_type, flag, how),
    set![TestEnv::Shadow],
),
```

stevenengler:
I'll merge this PR, but we can continue the discussion here anyways for a potential new PR.

PR #1000, opened 2020-09-30
Support TCP tests in preload mode

stevenengler:
These changes are enough to make the tcp-shadow-preload tests work.

### DIFF
--- a/src/shim/preload_libraries.c
+++ b/src/shim/preload_libraries.c
@@ -6,12 +6,15 @@
 #include <assert.h>
 #include <dlfcn.h>
 #include <errno.h>
+#include <fcntl.h>
 #include <glib.h>
 #include <netdb.h>
 #include <stdarg.h>
 #include <stdlib.h>
 #include <string.h>
 #include <sys/socket.h>
+#include <sys/stat.h>
+#include <sys/statfs.h>
 #include <sys/syscall.h>
 #include <sys/types.h>
 #include <sys/utsname.h>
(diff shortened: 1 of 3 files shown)


src/shim/preload_libraries.c
```diff
@@ -477,3 +480,70 @@ FILE *tmpfile(void) {
         free(name);
     return rv;
 }
+
+static void _convert_stat_to_stat64(struct stat* s, struct stat64* s64) {
+    memset(s64, 0, sizeof(*s64));
+
```
sporksmith, line 486: Optional, *might* be worth using a macro to avoid possibility of copying the wrong field. e.g.

```
#define COPY_X(x) s64->##x = s->##x
COPY_X(st_dev);
COPY_X(st_ino);
...
#undef COPY_X
```

src/shim/preload_libraries.c
```diff
@@ -477,3 +480,70 @@ FILE *tmpfile(void) {
...
+    s64->st_blocks = s->st_blocks;
+    s64->st_atim = s->st_atim;
+    s64->st_mtim = s->st_mtim;
+    s64->st_ctim = s->st_ctim;
+}
+
+static void _convert_statfs_to_statfs64(struct statfs* s, struct statfs64* s64) {
+    memset(s64, 0, sizeof(*s64));
+
+    s64->f_type = s->f_type;
```
sporksmith, line 505: Optional - might be worth a macro; as above.

src/shim/preload_libraries.c
```diff
@@ -477,3 +480,70 @@ FILE *tmpfile(void) {
...
+
+int fstatfs64(int a, struct statfs64* b) {
+    struct statfs s;
+    int rv = syscall(SYS_fstatfs, a, &s);
+    _convert_statfs_to_statfs64(&s, b);
+    return rv;
+}
+
+int __fxstat(int ver, int a, struct stat* b) {
+    // the version doesn't matter because we never make the syscall directly,
```
sporksmith, line 537: I don't quite parse this comment. What "version" are you referring to?

src/shim/preload_libraries.c
```diff
@@ -477,3 +480,70 @@ FILE *tmpfile(void) {
...
+    // and will later pass it back to glibc's fstat
+    return syscall(SYS_fstat, a, b);
+}
+
+int __fxstat64(int ver, int a, struct stat64* b) {
+    // the version doesn't matter because we never make the syscall directly,
```
sporksmith, line 543: Same here

src/shim/preload_libraries.c
```diff
@@ -477,3 +480,70 @@ FILE *tmpfile(void) {
...
+int fstatfs64(int a, struct statfs64* b) {
+    struct statfs s;
+    int rv = syscall(SYS_fstatfs, a, &s);
+    _convert_statfs_to_statfs64(&s, b);
+    return rv;
+}
+
+int __fxstat(int ver, int a, struct stat* b) {
+    // the version doesn't matter because we never make the syscall directly,
+    // and will later pass it back to glibc's fstat
```
sporksmith, line 538: By 'pass it back to', do you mean we expect this to only ever be called by glibc's fstat?

sporksmith (approved):

src/shim/preload_libraries.c
stevenengler: Good idea, thanks!

src/shim/preload_libraries.c
stevenengler: Done.

src/shim/preload_libraries.c
stevenengler: I updated the comment and added a check that the `stat` struct is the expected version. On x86_64 (where `STAT_IS_KERNEL_STAT` is true) I think we should be able to ignore this argument altogether and just pass the stat struct through to shadow (and eventually glibc), but I don't think checking the value will hurt (glibc would also error if this argument were incorrect).

src/shim/preload_libraries.c
stevenengler: I worded this badly. See above.

sporksmith (approved):

PR #1004, opened 2020-10-02
Preload/ptrace hybrid mode

### LINKED ISSUE #991: Add a fallback mechanism for interposing syscalls in ThreadPreload
# Background/Motivation

Right now ThreadPreload can miss syscalls when the plugin calls a libc interface that we haven't reimplemented in the shim. This is because inside glibc, syscall wrapper functions are called via private symbols. Calls to such symbols are compiled as direct calls that don't go through the PLT. As a result, they can't be interposed via `LD_PRELOAD`. For more about this issue, see https://www.jimnewsome.net/posts/interposing-internal-libc-calls/.

Our plan has been to reimplement all of the libc interfaces that make syscalls that we want to be able to interpose. For some interfaces, this is fairly trivial (e.g. `sleep(3)`, which basically just scales its argument and calls `usleep(2)`). For others though, it's more complex. `fwrite(3)` doesn't just call `write(2)` - it also does buffering in user-space. `pthread_create(3)` doesn't just call `clone` - it does a decent amount of work in user-space to set things up properly, including setting up thread-local-storage.

It'd be preferable if we didn't have to write and maintain all of this functionality that would be mostly duplicative of glibc.

In the short term we're doing initial development and benchmarking using a [patched libc](https://www.jimnewsome.net/posts/patching-glibc-to-make-syscalls-interposable/) that makes its syscalls by calling the `syscall(2)` *function* via the PLT; this allows us to interpose everything with our own `syscall` function in the `LD_PRELOAD`ed shim. However, we don't want to write and maintain such glibc patches for every supported platform.

# Alternatives under consideration

1. `ptrace`-attach in ThreadPreload as well. On a syscall-stop, service the syscall as ThreadPtrace does now. In this model, parts of the libc interface that we *do* reimplement would go through the ThreadPreload fast path, and parts that we don't would go through the slower ptrace-path, but would at least be reliably interposed.
2. `ptrace`-attach in ThreadPreload as well. On a syscall-stop, hot-patch the executable in-memory to call the shim's syscall function instead of doing a direct syscall. This is more complex, but potentially means good performance without reimplementing any of the libc interfaces at all.
3. Patch syscall instructions at load-time, using a tool such as https://github.com/srg-imperial/SaBRe.

We're currently leaning towards first implementing 1. (reimplementing hot paths of the libc interface, and using ptrace to handle other paths) since it's relatively simple, and we believe it shouldn't be too difficult to reimplement the "hot path interfaces" for tor.

# Rejected alternatives

1. Dynamically rewrite the entire binary with a tool like [pin](https://software.intel.com/content/www/us/en/develop/articles/pin-a-dynamic-binary-instrumentation-tool.html), [valgrind](https://valgrind.org/), or [dynamoRIO](https://dynamorio.org/). These would introduce substantial runtime overhead and possibly unacceptably large memory overhead (particularly if every page of every program is rewritten, and those pages don't end up getting shared across multiple instances of the same process). These are also fairly complex dependencies to bring in.
2. Tools such as https://dyninst.org/dyninst enable more-selective binary rewriting. This could address the performance and memory usage concerns vs rewriting the whole binary, but is still a complex dependency to bring in.

sporksmith:
This extends ptrace-mode to process syscalls via the same shim IPC mechanism as preload-mode. Syscalls that aren't handled by that mechanism (e.g. because there is no corresponding shim wrapper function) are still handled via ptrace stops.

This also adds a ptrace-noipc mode that preserves the old behavior of handling everything through ptrace only. This mode is intended primarily for measurement / comparison purposes, and can probably be removed before release.

Fixes #991


sporksmith:
Fixes #991

src/main/host/thread_ptrace.c
```diff
@@ -172,6 +207,19 @@ typedef struct _ThreadPtrace {
     // True if we have detached ptrace from the plugin and should attach before
     // executing another ptrace operation.
     bool needAttachment;
+
+    // Shared memory block for fast communication with plugin process.
+    // Intentionally opaque; details are implementation details of ShimEvent.
```
RWails, line 212: Wouldn't opaqueness here imply that ipcBlk is a pointer to struct?

src/main/host/thread_ptrace.c
```diff
@@ -386,42 +471,164 @@ static pid_t _waitpid_spin(pid_t pid, int *wstatus, int options) {
...
+// Waits for a ptrace or shim event.
+static StopReason _threadptrace_hybridSpin(ThreadPtrace* thread) {
+    // Only used for an shim-event-stop; lifted out of the loop so we don't
+    // re-initialize on every iteration.
+    StopReason event_stop = {
+        .type = STOPREASON_SHIM_EVENT,
+    };
+    while (1) {
+        if (thread->enableIpc &&
+            shimevent_tryRecvEventFromPlugin(thread->ipcBlk.p, &event_stop.shim_event) == 0) {
```
RWails, line 483: I'm confused here. `shimevent_tryRecvEventFromPlugin` takes a `struct IPCData *data` as its first argument. Why is a pointer to a shared-memory block being used?

sporksmith: It compiles because thread->ipcBlk.p is declared type `void *`, and works since it is in this case actually type `struct IPCData*`.

Added a method to get the typed pointer for clarity.

src/main/host/thread_ptrace.c
```diff
@@ -386,42 +479,165 @@ static pid_t _waitpid_spin(pid_t pid, int *wstatus, int options) {
...
 
 pid_t threadptrace_run(Thread* base, gchar** argv, gchar** envv) {
     ThreadPtrace* thread = _threadToThreadPtrace(base);
 
-    thread->base.nativeTid = _threadptrace_fork_exec(argv[0], argv, envv);
+    /* set the env for the child */
+    gchar** myenvv = g_strdupv(envv);
+
+    if (thread->enableIpc) {
+        ShMemBlockSerialized ipcBlkSerial = shmemallocator_globalBlockSerialize(&thread->ipcBlk);
```
RWails, line 574: Does it make any sense to (optionally) set up this block in the parent class? Both subclasses currently implement this behavior.

src/main/host/thread_ptrace.c
```diff
@@ -299,9 +374,12 @@ static void _threadptrace_enterStateSignalled(ThreadPtrace* thread,
             error("ptrace: %s", g_strerror(errno));
             return;
         }
+        debug("threadptrace_enterStateSignalled regs: rip=%llx rax=%llx arg0=%llx arg1=%llx "
+              "arg2=%llx arg3=%llx arg4=%llx arg5=%llx",
+              regs.rip, regs.rax, regs.rdi, regs.rsi, regs.rdx, regs.r10, regs.r8, regs.r9);
         uint64_t eip = regs.rip;
-        const uint8_t* buf = thread_getReadablePtr(
-            _threadPtraceToThread(thread), (PluginPtr){eip}, 16);
+        const uint8_t* buf = process_getReadablePtr(
```
RWails, line 381: Why `16` bytes here?

src/main/host/thread_ptrace.c
```diff
@@ -457,8 +673,24 @@ static void threadptrace_flushPtrs(Thread* base) {
     _threadptrace_flushPtrs(thread);
 }
 
+static const char* _regs_to_str(const struct user_regs_struct* regs) {
```
RWails, line 676: I wonder if there's some C preprocessor magic that would compress this function. I'm not sure that it would be better, though...

src/main/host/thread_ptrace.c
```diff
@@ -299,9 +374,12 @@ static void _threadptrace_enterStateSignalled(ThreadPtrace* thread,
             error("ptrace: %s", g_strerror(errno));
             return;
         }
+        debug("threadptrace_enterStateSignalled regs: rip=%llx rax=%llx arg0=%llx arg1=%llx "
+              "arg2=%llx arg3=%llx arg4=%llx arg5=%llx",
+              regs.rip, regs.rax, regs.rdi, regs.rsi, regs.rdx, regs.r10, regs.r8, regs.r9);
```
RWails, line 379: A couple of times thru out this code and elsewhere we maintain the mapping between registers and syscalls args. Does it make sense to refactor with something like a `regs_to_syscall_args()` function?

src/shim/binary_spinning_sem.cc
```diff
@@ -42,3 +44,18 @@ void BinarySpinningSem::wait(bool spin) {
     _x.store(false, std::memory_order_release);
     pthread_spin_unlock(&_lock);
 }
+
+int BinarySpinningSem::trywait() {
+    int rv = sem_trywait(&_semaphore);
+    if (rv != 0) {
+        assert(rv == -1 && errno == EAGAIN);
```
RWails, line 51: Any chance for us to receive an interrupt/signal here? Maybe we should allow for `EINTR`

src/shim/binary_spinning_sem.cc
```diff
@@ -42,3 +44,18 @@ void BinarySpinningSem::wait(bool spin) {
...
+        return rv;
+    }
+
+    _spin_ctr = 0;
```
RWails, line 55: Hm, maybe we should not make spin_ctr a member variable of `BinarySpinningSem`. We don't need the state outside of a `wait` call and it's just creating more, unnecessary work for us here.

src/shim/binary_spinning_sem.h
```diff
@@ -61,6 +61,13 @@ class BinarySpinningSem {
      */
     void wait(bool spin = true);
 
+    /*
+     * Atomically check if the semaphore is available (has value one). If
+     * so takes the semaphore (sets it back to zero) and returns 0. Otherwise
+     * returns -1 and sets errno to EAGAIN.
+     */
+    int trywait();
```
RWails, line 69: n.b. I haven't worked it out, but this might violate thread-safety if called in parallel with wait(). I think that's OK because that's not a use case we're really using, but maybe worth noting.

rwails (commented):
OK -- no glaring issues after a first pass through :-) I think I need to digest the changes to `thread_ptrace` for a little longer, but I'll try to have that wrapped up when Rob gets his review in.

src/shim/binary_spinning_sem.h
sporksmith: Hmm... I'd rather make this safe than add a potential footgun of `trywait` and `wait` not being safe to use together...

The way I look at it is that `_semaphore` is the authority on whether the BinarySpinningSem is available, and `_x` is only a hint. TBH I don't think I understand what `_lock` is doing.

I changed it to try grabbing `_x` first to avoid a thread calling `_wait`, grabbing `_x`, breaking out of its spin loop, only to have a concurrent thread grab the semaphore out from under it via `trywait`. Does this address your concern?

FWIW a spurious break in `wait` is already (and still) possible when two threads call `wait`: thread A can hit its spin max (or not spin at all) and wait on `_semaphore` without having grabbed `_x` first. After the semaphore is released, another thread calling `wait` could grab `_x` but fail to grab the semaphore, which the first waiter grabbed out from under it. I don't see a way to avoid that problem; perhaps another reason to see if we should reduce to just spinning directly on the semaphore with `sem_trywait`. I'll take a closer look at that, but would rather save that for another PR.

src/main/host/thread_ptrace.c
```diff
@@ -190,6 +247,55 @@ static Thread* _threadPtraceToThread(ThreadPtrace* thread) {
...
+    static char buf[1000];
+    ssize_t offset = 0;
+    sprintf(&buf[offset], "arg0:%lld ", regs->rdi);
+    sprintf(&buf[offset], "arg1:%lld ", regs->rsi);
+    sprintf(&buf[offset], "arg2:%lld ", regs->rdx);
+    sprintf(&buf[offset], "arg3:%lld ", regs->r10);
+    sprintf(&buf[offset], "arg4:%lld ", regs->r8);
+    sprintf(&buf[offset], "arg5:%lld", regs->r9);
+    return buf;
+}
```
robgjansen, line 297: Should this be in a `#ifdef DEBUG` block, since it's only ever called from a `debug()` log macro (which is also wrapped in an `#ifdef DEBUG` block)?

src/main/host/thread_ptrace.c
```diff
@@ -386,42 +524,165 @@ static pid_t _waitpid_spin(pid_t pid, int *wstatus, int options) {
...
+    gchar** myenvv = g_strdupv(envv);
+
+    if (thread->enableIpc) {
+        ShMemBlockSerialized ipcBlkSerial = shmemallocator_globalBlockSerialize(&thread->ipcBlk);
+
+        char ipcBlkBuf[SHD_SHMEM_BLOCK_SERIALIZED_MAX_STRLEN] = {0};
+        shmemblockserialized_toString(&ipcBlkSerial, ipcBlkBuf);
+
+        /* append to the env */
+        myenvv = g_environ_setenv(myenvv, "_SHD_IPC_BLK", ipcBlkBuf, TRUE);
```
robgjansen, line 625: Should we be consistent about name prefixes for env variables we are setting? I think others are prefixed with `SHADOW` rather than `_SHD`.

src/main/host/thread_ptrace.c
```diff
@@ -600,29 +796,100 @@ SysCallCondition* threadptrace_resume(Thread* base) {
...
+                        .event_id = SHD_SHIM_EVENT_SYSCALL_COMPLETE,
+                        .event_data = {
+                            .syscall_complete = {.retval = ret.retval,
+                                                 .simulation_nanos =
+                                                     worker_getEmulatedTime()},
+
+                        }};
+                        shimevent_sendEventToPlugin(_threadptrace_ipcData(thread), &shim_result);
+                        break;
+                                       }
```
robgjansen, line 820: Is this bracket is in a weird location? I expected it to be one tab back from the tab-stop on the previous line, but maybe this is where clang-format placed it?

src/main/host/thread_ptrace.c
```diff
@@ -600,29 +796,100 @@ SysCallCondition* threadptrace_resume(Thread* base) {
             case THREAD_PTRACE_CHILD_STATE_TRACE_ME:
                 debug("THREAD_PTRACE_CHILD_STATE_TRACE_ME");
                 break;
+            case THREAD_PTRACE_CHILD_STATE_IPC_SYSCALL: {
+                debug("THREAD_PTRACE_CHILD_STATE_IPC_SYSCALL");
+                SysCallReturn ret = syscallhandler_make_syscall(thread->sys, &thread->syscall_args);
+                switch (ret.state) {
```
robgjansen, line 802: Nested switch statements? Could we break this out into its own function for readability? I would also prefer to break out the longer switch cases too when that is feasible.

src/main/host/thread_ptrace.c
```diff
@@ -600,29 +796,100 @@ SysCallCondition* threadptrace_resume(Thread* base) {
...
+                                                        }
             case THREAD_PTRACE_CHILD_STATE_SYSCALL:
                 debug("THREAD_PTRACE_CHILD_STATE_SYSCALL");
 
-                // Ask the syscall handler to handle it.
-                _threadptrace_handleSyscall(thread);
+                SysCallReturn ret = _threadptrace_handleSyscall(thread, &thread->syscall_args);
 
-                switch (thread->syscall.sysCallReturn.state) {
+                switch (ret.state) {
```
robgjansen, line 872: Another nested switch. It would be nice to break this out for readability.

robgjansen (commented):
Mostly minor comments and renaming; you know I love my bikeshedding ;)

I'm going to let Ryan give final approval here.

src/shim/shim.c
```diff
@@ -15,43 +16,101 @@
...
 #include "support/logger/logger.h"
 
-// Whether Shadow is using INTERPOSE_PRELOAD
-static bool _using_interpose_preload;
+typedef enum {
+    INTERPOSE_NONE,
+    INTERPOSE_PRELOAD,
+    INTERPOSE_PTRACE,
+    INTERPOSE_PTRACE_NOIPC,
+} InterposeType;
```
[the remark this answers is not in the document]
sporksmith: Done-ish.

Here in shim.c, with the benefit of hindsight I think it's easier to have two independent variables tracking whether we're using ptrace and whether we're using preload ipc.

I changed some other places based on your suggestion here.

I left the command-line option as "ptrace" and class ThreadPtrace to get "preload+ptrace", mostly so that I could flip the default to use preload+ptrace in this PR (thus validating it in the tests) without also bringing in a new big rename. If we do want to change those, I suggest we do it in a follow-up PR as a pure refactor.

src/shim/shim.c
```diff
@@ -15,43 +16,101 @@
...
+        }
+    }
 }
 
 void shim_enableInterposition() {
     assert(_shim_disable_interposition > 0);
-    --_shim_disable_interposition;
+    if (--_shim_disable_interposition == 0 && _interpose_type == INTERPOSE_PTRACE) {
+        if (_shim_shared_mem) {
+            debug("disabling native-syscalls via shmem %p %p", &_shim_shared_mem, _shim_shared_mem);
```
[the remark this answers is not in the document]
sporksmith: Dropped `&_shim_shared_mem`; I think I'd thrown that in there when debugging at some point to validate that the thread-local-storage was WAI

src/shim/shim.c
```diff
@@ -15,43 +16,101 @@
...
+        } else {
+            debug("disabling native-syscalls via syscall");
+            shadow_set_ptrace_allow_native_syscalls(false);
+        }
+    }
 }
 
 bool shim_interpositionEnabled() {
-    return _using_interpose_preload && !_shim_disable_interposition;
+    return (_interpose_type == INTERPOSE_PRELOAD || _interpose_type == INTERPOSE_PTRACE) &&
```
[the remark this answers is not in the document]
sporksmith: Done(ish)

src/shim/shim.c
```diff
@@ -15,43 +16,101 @@
...
 
+    const char* interpose_method = getenv("SHADOW_INTERPOSE_METHOD");
+    assert(interpose_method);
+    if (!strcmp(interpose_method, "PRELOAD")) {
+        return INTERPOSE_PRELOAD;
+    }
+    if (!strcmp(interpose_method, "PTRACE")) {
+        return INTERPOSE_PTRACE;
+    }
+    if (!strcmp(interpose_method, "PTRACE_NOIPC")) {
```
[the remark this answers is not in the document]
sporksmith: Done

src/main/host/thread_ptrace.c
sporksmith: Done

src/shim/shim.c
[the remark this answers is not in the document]
robgjansen: Makes sense!

src/main/host/thread_ptrace.c
robgjansen: OK, cool :)

sporksmith:
@RWails per offline conversation, can I get an approval here so I can merge? :)

rwails (approved):

PR #1008, opened 2020-10-14
Update cbindgen to 0.15.0

sporksmith:
This fixes the bindings check in our CI. The previous version we'd pinned to has been [yanked](https://crates.io/crates/cbindgen/0.14.6) from crates.io. (Not sure why, but presumably because of some serious bug)


robgjansen (approved):
👍

PR #1009, opened 2020-10-16
Move --output-on-failure from setup to ci

sporksmith:
When debugging locally it's generally easier to look at build/Testing/Temporary/LastTest.log,
instead of trying to scroll through direct terminal output.

--output-on-failure is more useful in the ci, where it would otherwise
be an extra step to download LastTest.log from a GitHub artifact or copy
out of a local Docker container.


stevenengler (approved):
I didn't know about the `build/Testing/Temporary/LastTest.log` file, I used to always just redirect the output of `./setup test` to a file. So looks good to me.

PR #1010, opened 2020-10-16
Expand futex tests

sporksmith:
Added some more granular tests for ease of debugging.

Also added tests for FUTEX_WAIT_BITSET and FUTEX_WAKE_BITSET. These aren't implemented in Shadow yet, but it looks like these are used in `sem_wait` and `sem_post`. I'd started to work on these to support handling the resulting futex syscalls in ptrace/preload "hybrid mode", but for now ended up handling it a different way for now (by setting a flag to let Shadow know that the plugin is executing in the shim).


robgjansen (approved):
Looks plausible :)

PR #1011, opened 2020-10-16
Don't destroy global shmem objects

sporksmith:
Explicitly destroying them doesn't have a clear benefit, and introduces
a danger of accidentally using them (e.g. in another global destructor)
after they've been destroyed.

In particular I stumbled across this while developing ptrace/preload
hybrid mode, where other syscalls can continue to be made and handled in
other global destructors.

For more on this topic see
https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

### DIFF
--- a/src/main/shmem/shmem_allocator.c
+++ b/src/main/shmem/shmem_allocator.c
@@ -115,22 +115,6 @@ struct _ShMemSerializer {
 static ShMemAllocator* _global_allocator = NULL;
 static ShMemSerializer* _global_serializer = NULL;
 
-/*
- * hook used to cleanup at exit.
- */
-static void _shmemallocator_destroyGlobal() {
-    assert(_global_allocator);
-    shmemallocator_destroy(_global_allocator);
-}
-
-/*
- * hook used to cleanup at exit.
- */
-static void _shmemserializer_destroyGlobal() {
-    assert(_global_serializer);
-    shmemserializer_destroy(_global_serializer);
-}
-
 ShMemAllocator* shmemallocator_getGlobal() {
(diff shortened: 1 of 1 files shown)


RWails:
I think the main reason for the destructor is to prevent spurious memory-leak warnings from tools like Valgrind, although this isn't good reason to keep code around that's crashing the shutdown process. Does it make sense to change the allocator and serializer to be statically allocated variables inside of shmem_allocator.c? I don't think there's any reason why these objects have to be malloc'd.

sporksmith:
> I think the main reason for the destructor is to prevent spurious memory-leak warnings from tools like Valgrind,

A global is still reachable at program exit, and thus generally not classified as a leak. I verified that Valgrind correctly distinguishes this: https://github.com/sporksmith/dev-journal/blob/master/valgrind-global-leaks/README.ipynb

> Does it make sense to change the allocator and serializer to be statically allocated variables inside of shmem_allocator.c

Hmm, that could work, though I'd lean towards keeping them heap allocated vs. adding a state of "statically allocated but not yet initialized". Since Valgrind can tell this isn't a leak I'm inclined to leave it; wdyt?

Btw I realized that if we don't destroy the allocator then the /dev/shm files don't get cleaned up until the next Shadow run. In the shim we only need a serializer since Shadow does all the allocation, so I restored the allocator destructor.

RWails:
I didn't realize that valgrind does not flag global variables! TMYK.

I think the changes you made make sense. One other option is to create a routine that cleans up all shared-memory pages, implicitly invalidating all allocated shared memory blocks. That could be called instead of destroy during exit --- then, the only blocks that leak would be ones that are allocated during exit.

RWails (approved):

PR #1013, opened 2020-10-23
Enable block msg and add command-line options

### LINKED ISSUE #965: Signal plugin when Shadow will block on a syscall
Currently in preload mode, the plugin makes a syscall and then spins until shadow gives it an answer. But sometimes Shadow needs to block before the ssycall can be completed, for example when it is waiting for a packet to read data. In this case, Shadow should send a message back to the plugin to inform it that we will be blocking it. This can be used as a signal so the plugin can stop spinning and instead do a `semaphore_wait()` call.

This strategy could allow us to avoid implementing an algorithm to dynamically compute the "optimal" spin time (as requested in #894), because we could just always spin until we get a response from Shadow.

sporksmith:
Main functional change here is enabling the explicit message from shadow to shim to block.

I also changed that from being controlled by an env variable, and the max-spin threshold from being controlled by a hard-coded constant, to both be controlled by command-line flags.

This also adds a mechanism for code modules to declare command-line flags locally, so that they don't need to be plumbed through intermediate layers of code that shouldn't have to know about them.

Fixes #965.


sporksmith:
Btw I also verified locally that with the explicit block message, setting the spin-max to unlimited seems to be neutral or slightly faster. Probably ought to make changing that default a separate PR, though (and run the full bench suite)

sporksmith:
+ @RWails for semaphore bits

src/main/host/thread_preload.c
```diff
@@ -21,6 +22,19 @@
...
+static gint _sendExplicitBlockMessage = true;
+OPTION_EXPERIMENTAL_ENTRY(
+    "send-explicit-block-message", 0, 0, G_OPTION_ARG_INT, &_sendExplicitBlockMessage,
+    "Send message to plugin telling it to stop spinning when a syscall blocks", "[0|1]")
+
+static gint _spinMax = 8096;
+OPTION_EXPERIMENTAL_ENTRY(
+    "preload-spin-max", 0, 0, G_OPTION_ARG_INT, &_spinMax,
+    "Max number of iterations to busy-wait on ICP sempahore before blocking. -1 for unlimited.",
+    "[n]")
```
robgjansen, line 36: I think with our current set of options, we specify the default value of the option in brackets at the end of the description string (line 35), and I don't think we use brackets for the part in line 36. We should make our use of brackets consistent, and also I think we want to specify the default value.

robgjansen (approved):

src/shim/binary_spinning_sem.cc
```diff
@@ -24,10 +17,11 @@ void BinarySpinningSem::post() {
 }
 
 void BinarySpinningSem::wait(bool spin) {
+    ssize_t spin_ctr = 0;
     // Based loosely on
     // https://probablydance.com/2019/12/30/measuring-mutexes-spinlocks-and-how-bad-the-linux-scheduler-really-is/.
     if (spin) {
-        while (_spin_ctr++ < _thresh) {
+        while (_thresh < 0 || spin_ctr++ < _thresh) {
```
RWails, line 24: Probably not worth addressing in this PR, but I bet some use of `__builtin_expect` might give a good hint to the platform -- both these conditions are low entropy

RWails (approved):
Looks good!

src/main/host/thread_preload.c
sporksmith: Done

src/shim/binary_spinning_sem.cc
sporksmith: Hmm, yeah I'm a bit wary of such hints in general. Ironically it's not clear we have a reason to highly optimize the straight-line execution cost of this loop, since making the loop tighter just means we'll iterate more times.

Making it faster or slower does interact with `SHD_GATE_SPIN_MAX`, though. I've been wondering whether we'd be better off using RDTSC inside the loop and basing the max off of cycle count instead of loop count so that we can decouple "how long to spin" from "how many instructions ended up inside the loop in this compilation".

Either way, yeah let's save for another PR :)

PR #1014, opened 2020-10-26
Enable epoll preload tests on all platforms

stevenengler:
The epoll preload tests were set to only run under the "ilibc" configuration, but the patched libc wasn't actually being used (`LD_LIBRARY_PATH` wasn't set in the shadow config file). The epoll preload tests seem to work on all CI platforms without the patched libc, so the "ilibc" requirement has been removed.


sporksmith (approved):
Cool!

PR #1017, opened 2020-11-03
Use sem_trywait instead of separate atomic

sporksmith:
sem_trywait already provides an efficient way of trying to check a
semaphore without blocking. Using it instead of a separate atomic
simplifies things quite a bit.

Performance seems to be the same for a 30s phold simulation.

This also eliminates some subtle interactions between the atomic and
the semaphore.

For example:
 * Thread A: calls `wait`, hits spin threshold, blocks on `sem_wait`
 * Thread B: calls `post`
 * Thread C: calls `wait`, sees that the atomic is available and breaks
out of the loop
 * Thread A: wakes up from `sem_wait` and grabs the semaphore
 * Thread C: calls `sem_wait` expecting it not to block, but it does
block since the semaphore was taken out from under it.


sporksmith:
It looks like glibc uses atomics in a similar way as we were, except directly on the semaphore itself. e.g. https://github.com/bminor/glibc/blob/b8d3e8fbaac154979078e6665aa983742dec9539/nptl/sem_waitcommon.c#L126

RWails (commented):
Yeah, this is a great catch Jim. I wasn't convinced that the concurrency logic was 100% correct in this module and this simplifies a lot. Indeed, a lot of the logic we had implemented is similar to: <https://code.woboq.org/userspace/glibc/nptl/sem_waitcommon.c.html#124>

Let me benchmark the change, but otherwise good to go

sporksmith:
> Let me benchmark the change, but otherwise good to go

@rwails Friendly ping - would be good to get this in so I can bring the hybrid mode PR up to date

RWails:
> > Let me benchmark the change, but otherwise good to go
> 
> @RWails Friendly ping - would be good to get this in so I can bring the hybrid mode PR up to date

Have a job running now, should be ready in a few hours.

RWails:
https://github.com/shadow/benchmark-results-phantom/blob/master/result_dir/phold_benchmark_2020-11-10T21:37:05.244207_7d4ad27f/perf.csv

Looks like a slight regression introduced by this change, but maybe worth it to reduce complexity. WDYT?

sporksmith:
Hmm, in a couple of the preload data points I spot checked, it was a 5% and 8% regression. Maybe we should hold off on this for now, pending further investigation.

sporksmith:
I tried to repro the regression locally but so far haven't been able to.

My machine only has 4 cores, so running a large number of work_threads isn't very meaningful. Looking at the lowest # of threads in your data (4), I get that trywait is 1% slower @ 32-hosts, 9% slower @ 64 hosts, and 6% slower @ 128 hosts.

Locally I tried running with 2 threads, 64 hosts, 30 seconds of sim time, but I get basically neutral results. There's a fair bit of variance, so *maybe* we could've just gotten 'unlucky' in your data, but I haven't spotted anywhere where it doesn't look like a regression. (Though maybe I should more carefully analyze/graph).

dev mean 16.2s, 95% [confidence interval](https://stackoverflow.com/questions/15033511/compute-a-confidence-interval-from-sample-data) [15.0, 17.4]

```
2/2 Test #37: phold-threaded-shadow-preload ....   Passed   16.08 sec
2/2 Test #37: phold-threaded-shadow-preload ....   Passed   15.19 sec
2/2 Test #37: phold-threaded-shadow-preload ....   Passed   17.77 sec
2/2 Test #37: phold-threaded-shadow-preload ....   Passed   15.64 sec
2/2 Test #37: phold-threaded-shadow-preload ....   Passed   16.17 sec
```

trywait mean 16.0s; 95% confidence interval [14.9s, 17.2s]

```
2/2 Test #37: phold-threaded-shadow-preload ....   Passed   16.67 sec
2/2 Test #37: phold-threaded-shadow-preload ....   Passed   16.18 sec
2/2 Test #37: phold-threaded-shadow-preload ....   Passed   15.05 sec
2/2 Test #37: phold-threaded-shadow-preload ....   Passed   15.10 sec
2/2 Test #37: phold-threaded-shadow-preload ....   Passed   17.15 sec
```

Any ideas? Maybe worth just running the benchmark script again?

RWails (approved):

Issue #841, opened 2020-06-11
Thoroughly test the sendto() syscall


robgjansen:
See #832 for an overview of our approach.

(The libc `send` function redirects to the `sendto` syscall.)

Issue #840, opened 2020-06-11
Thoroughly test the recvfrom() syscall


robgjansen:
See #832 for an overview of our approach.

(The libc `recv` function redirects to `recvfrom`.)

PR #1019, opened 2020-11-05
Improved sendto() and recvfrom() support and added tests

stevenengler:
Various socket improvements, mostly for UDP, and tests to check the behaviour of `sendto()` and `recvfrom()`.

Closes shadow/shadow#841, shadow/shadow#840.


src/main/host/descriptor/socket.h
```diff
@@ -84,7 +84,8 @@ void socket_init(Socket* socket, SocketFunctionTable* vtable,
 void socket_pushInPacket(Socket* socket, Packet* packet);
 void socket_dropPacket(Socket* socket, Packet* packet);
 Packet* socket_pullOutPacket(Socket* socket);
-Packet* socket_peekNextPacket(const Socket* socket);
+Packet* socket_peekNextOutwardPacket(const Socket* socket);
```
sporksmith, line 87: Nit - maybe just `socket_peekNextOutPacket` and `socket_peekNextInPacket` for consistency with `socket_pullOutPacket`?

Double nit - if we did want to spell it out more, slight preference for `(in|out)Bound` instead of `(in|out)Ward`.

src/main/host/descriptor/tcp.c
```diff
@@ -2280,6 +2280,17 @@ static gssize _tcp_receiveUserData(Transport* transport, gpointer buffer,
     gsize offset = 0;
     gsize copyLength = 0;
 
+    if ((socket_getInputBufferLength(&(tcp->super)) == 0) && (tcp->partialUserDataPacket == NULL) &&
```
sporksmith, line 2283: Can drop extra parens - `&tcp->super`

src/test/socket/sendto_recvfrom/test_sendto_recvfrom.rs
```diff
@@ -0,0 +1,1379 @@
...
+    test_utils::run_tests(&tests, summarize)?;
+
+    println!("Success.");
+    Ok(())
+}
+
+fn get_tests() -> Vec<test_utils::ShadowTest<(), String>> {
+    let mut tests: Vec<test_utils::ShadowTest<_, _>> = vec![];
+
+    tests.extend(vec![
```
sporksmith, line 83: Nit - why not use this `vec!` above to construct the original `Vec`, rather than constructing an empty one and then extending it?

src/test/socket/sendto_recvfrom/test_sendto_recvfrom.rs
```diff
@@ -0,0 +1,1379 @@
...
+            set![TestEnv::Libc, TestEnv::Shadow],
+        ),
+        test_utils::ShadowTest::new(
+            "test_zero_len_buf <udp>",
+            || test_zero_len_buf(libc::SOCK_DGRAM),
+            set![TestEnv::Libc, TestEnv::Shadow],
+        ),
+        // This test fails in shadow (debug mode) on centos 7, but only in the
+        // GitHub CI, so disabling for now.
+        // The cause of failure seems to be that centos 7 is about twice as slow
```
sporksmith, line 131: We could just increase the timeout for this test, as we do for phold.

sporksmith (approved):
Looks good! Just a couple nits

src/main/host/descriptor/socket.h
stevenengler: I used the shortened form for consistency.

src/main/host/descriptor/tcp.c
stevenengler: Done.

src/test/socket/sendto_recvfrom/test_sendto_recvfrom.rs
stevenengler: Done. Not sure why I wrote it this way.

src/test/socket/sendto_recvfrom/test_sendto_recvfrom.rs
stevenengler: I increased it to 60 seconds. It seems to take about 40 seconds in the CI on CentOS 7 (and 3 seconds on my laptop).

Issue #1018, opened 2020-11-05
Stop using "add-path" in CI workflow


stevenengler:
We will need to update our CI workflows to stop using the `add-path` option, otherwise it will stop working some time in the near future: https://github.com/actions/toolkit/security/advisories/GHSA-mfwh-5m23-j46w

> The runner will release an update that disables the `set-env` and `add-path` workflow commands in the near future.

PR #1022, opened 2020-11-06
Fix deprecated 'add-path' workflow command

stevenengler:
Closes shadow/shadow#1018.


sporksmith (approved):
Thanks for taking care of this!

PR #1025, opened 2020-11-09
Tests: add Rust tests for poll

vrask:


stevenengler:
Sorry this is taking awhile, the test is failing when I run it locally, so I'm trying to figure out why.

vrask:
> Edit: Figured out the issues, the code has a buffer overflow and a file permission error. I'll do a full review.

Look forward to the full review.

Where is the buffer overflow? I copied the interactions with raw buffers essentially verbatim from the C tests, so if they exist in this PR, they likely exist in the original code (unless I have made some idiotic mistake).

Perhaps I should be more suspicious of the code I am porting to Rust?

src/test/poll/test_poll.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+    return Ok(());
+}
+
+fn test_creat() -> Result<(), String> {
+    let test_file = "testpoll.txt";
+    let mut fd = unsafe { libc::creat(test_file.as_ptr() as *const libc::c_char, 0664) };
+    if fd < 0 {
+        return Err(format!("error: could not create {}", test_file));
+    }
```
stevenengler, line 120: There are two issues here:

1. In C, the leading `0` in the literal `0664` indicates that it is an octal. In Rust you must use `0o644` instead, otherwise the file is created with incorrect permissions.
2. There is a buffer overflow here since strings in Rust do not have null terminators. Instead you need to convert it to a null-terminated `CString` with something like:

```rust
let test_file = b"testpoll.txt";
let test_file = std::ffi::CString::new(*test_file).unwrap();
let fd = unsafe { libc::creat(test_file.as_bytes_with_nul() as *const _ as *const libc::c_char, 0o664) };
if fd < 0 {
    return Err(format!("error: could not create {:?}", test_file));
}
```

stevenengler (commented):
I didn't finish this review today, but here are some in-progress comments so you can see what I meant in my earlier comment.

src/test/poll/test_poll.rs
```diff
@@ -0,0 +1,224 @@
...
+            buf.as_ptr() as *const libc::c_char,
+            "test".as_ptr() as *const libc::c_char,
+            4,
+        )
+    };
+    if rv != 0 {
+        return Err("error: did not read 'test' from pipe.".to_string());
+    } else {
+        return Ok(());
+    }
```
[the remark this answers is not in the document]
stevenengler: Normally we try to avoid adding rust dependencies, but we're already using the nix crate in shadow, so we want to start using it tests as well for syscalls that we're not explicitly testing to make the code easier to write and read.

src/test/poll/test_poll.rs
stevenengler: [As an aside, the old code created the following file: `src/test/poll/testpoll.txterror: could not create error: poll on empty file failederror: expected EOF to be readable from empty file. revents=error: could not write to test_pipe?`, which I can't figure out how to delete since the filename also contains non-ascii bytes :) ]

src/test/poll/test_poll.rs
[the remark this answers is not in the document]
vrask: > Normally we try to avoid adding rust dependencies, but we're already using the nix crate in shadow, so we want to start using it tests as well for syscalls that we're not explicitly testing to make the code easier to write and read.

Saw use of `nix` when reviewing code, and agree using it is a great idea. Will add it with the fixups for this PR.

src/test/poll/test_poll.rs
vrask: > There are two issues here:

    In C, the leading 0 in the literal 0664 indicates that it is an octal. In Rust you must use 0o644 instead, otherwise the file is created with incorrect permissions.
    There is a buffer overflow here since strings in Rust do not have nul terminators. Instead you need to manually add the nul byte, or convert it to a nul-terminated CString with something like:

Ah, and here is the idiotic mistake... Thanks for the catch, will fix.

vrask: While I'm changing file permissions, do the test files need to be group/world readable? Would a more restrictive `0o600` still work?

stevenengler: I think I would still keep them group/world readable just because there's no reason not to, and it often makes things less difficult. While it would be fine to make it more restrictive, it can make things more painful when sharing the files between different systems where the uid might change (for example between a host and container which we do a lot, or on a USB drive). If someone is worried about the world-read permission, they'll probably have set a umask already, or set more restrictive permissions on a parent directory.

vrask:
Added the suggested fixes. Will clean up anything else that needs it, after the full review.

src/test/poll/test_poll.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    /* poll will check when pipe has info to read */
+    let mut read_poll = libc::pollfd {
+        fd: pfds[0],
+        events: libc::POLLIN,
+        revents: 0,
+    };
+
+    /* First make sure there's nothing there */
+    let mut ready = unsafe { libc::poll(&mut read_poll as *mut libc::pollfd, 1, 100) };
```
stevenengler, line 40: Everything in the function at this point can be wrapped with `test_utils::run_and_close_fds`, which will automatically close the file descriptors when we exit the closure and lets you remove all of the `libc::close()` calls. This helps make the code more readable and prevents us from forgetting to close file descriptors:

```rust
test_utils::run_and_close_fds(&[pfd_read, pfd_write], || {
    let mut read_poll = ...
    let mut ready = ...
    if ready < 0 {
        return Err("error: poll failed".to_string());
    } else if ready > 0 {
        return Err(...);
    }
    fd_write(pfds[1])?;
    ...
})?;
```

For example see: https://github.com/shadow/shadow/blob/5079bb55ecc7ad7574389fadfe0d6eed7f56fcf9/src/test/socket/sendto_recvfrom/test_sendto_recvfrom.rs#L322-L342

src/test/poll/test_poll.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    /* success! */
+    unsafe {
+        libc::close(pfds[0]);
+        libc::close(pfds[1]);
+    }
+    return Ok(());
+}
+
+fn test_creat() -> Result<(), String> {
```
stevenengler, line 108: This function can use the `test_utils::run_and_close_fds` wrapper as well to not require explicit `libc::close()` calls.

stevenengler (changes requested):
Thanks for making those changes. There are a few more I think would help readability.

vrask:
@stevenengler `test_utils::run_and_close_fds` is a really nice wrapper function, thanks for pointing it out. Definitely made the code cleaner.

Applied your suggested fixes, let me know what you think.

src/test/poll/test_poll.rs
```diff
@@ -0,0 +1,186 @@
...
+        /* Check again, should be something to read */
+        fd = nix::fcntl::open(
+            test_file.as_ref(),
+            nix::fcntl::OFlag::O_RDONLY,
+            nix::sys::stat::Mode::empty(),
+        )
+        .map_err(|e| e.to_string())?;
+
+        read_poll.fd = fd;
+        read_poll.events = libc::POLLIN;
```
stevenengler, line 131: You should close the old file descriptor before overwriting it, so this can change to:

```rust
test_utils::run_and_close_fds(&[fd], || {
    ...

    /* write to file */
    fd_write(fd)?;
}

/* Check again, should be something to read */
let fd = nix::fcntl::open(
    test_file.as_ref(),
    nix::fcntl::OFlag::O_RDONLY,
    nix::sys::stat::Mode::empty(),
)
.map_err(|e| e.to_string())?;

test_utils::run_and_close_fds(&[fd], || {
    read_poll.fd = fd;
    read_poll.events = libc::POLLIN;

    ...
}
```

src/test/poll/test_poll.rs
```diff
@@ -0,0 +1,186 @@
...
+            return Err(format!(
+                "error: read_poll has wrong revents: {}",
+                read_poll.revents
+            ));
+        }
+
+        /* Make sure we got what expected back */
+        if let Err(e) = fd_read_cmp(pfd_read) {
+            return Err(e);
+        }
```
stevenengler, line 75: This can be just:

```rust
fd_read_cmp(pfd_read)?;
```

src/test/poll/test_poll.rs
```diff
@@ -0,0 +1,186 @@
...
+            return Err(format!(
+                "error: pipe marked readable. revents={}",
+                read_poll.revents
+            ));
+        }
+
+        /* Now put information in pipe to be read */
+        if fd_write(pfd_write).is_err() {
+            return Err("error: could not write to pipe".to_string());
+        }
```
stevenengler, line 54: This can be just:

```rust
fd_write(pfd_write)?;
```

src/test/poll/test_poll.rs
```diff
@@ -0,0 +1,186 @@
...
+                read_poll.revents
+            ));
+        }
+
+        let test_file = b"testpoll.txt";
+
+        /* write to file */
+        if fd_write(fd).is_err() {
+            return Err(format!("error: could not write to {:?}", test_file));
+        }
```
stevenengler, line 120: This can be just:

```rust
fd_write(pfd_write)?;
```

src/test/poll/test_poll.rs
```diff
@@ -0,0 +1,186 @@
...
+            return Err(format!(
+                "error: read_poll has wrong revents: {}",
+                read_poll.revents
+            ));
+        }
+
+        /* Make sure we got what expected back */
+        if let Err(e) = fd_read_cmp(fd) {
+            return Err(e);
+        }
```
stevenengler, line 148: This can be just:

```rust
fd_read_cmp(fd)?;
```

src/test/poll/test_poll.rs
```diff
@@ -0,0 +1,186 @@
...
+            /* Note: Even though the file is 0 bytes, has no data inside of it, it is still instantly
+             * available for 'reading' the EOF. */
+            return Err(format!(
+                "error: expected EOF to be readable from empty file. revents={}",
+                read_poll.revents
+            ));
+        }
+
+        let test_file = b"testpoll.txt";
+
```
stevenengler, line 116: This was defined earlier in the function, so no need to re-define it here.

stevenengler (changes requested):
Okay, just a few small style-things to change, then I think it will be good to merge after this! After these changes, feel free to rebase and squash.

vrask:
@stevenengler fixes are in, and commits squashed. Thanks for your time reviewing this PR.

stevenengler (approved):

stevenengler:
@vrask Thanks for the PR and changes!

PR #1026, opened 2020-11-17
Syscall handler for fcntl64

robgjansen:
Redirect `fcntl64` to the `fcntl` handler.

### DIFF
--- a/src/main/host/syscall/fcntl.c
+++ b/src/main/host/syscall/fcntl.c
@@ -184,3 +184,7 @@ SysCallReturn syscallhandler_fcntl(SysCallHandler* sys,
 
     return (SysCallReturn){.state = SYSCALL_DONE, .retval.as_i64 = result};
 }
+
+SysCallReturn syscallhandler_fcntl64(SysCallHandler* sys, const SysCallArgs* args) {
+    return syscallhandler_fcntl(sys, args);
+}
--- a/src/main/host/syscall/fcntl.c
+++ b/src/main/host/syscall/fcntl.c
@@ -68,6 +68,15 @@ static int _syscallhandler_fcntlHelper(SysCallHandler* sys, File* file, int fd,
             break;
         }
 
+#if defined(F_GETLK64) && F_GETLK64 != F_GETLK
+        case F_GETLK64: {
+            struct flock64* flk =
+                process_getMutablePtr(sys->process, sys->thread, argReg.as_ptr, sizeof(*flk));
+            result = file_fcntl(file, command, (void*)flk);
+            break;
+        }
+#endif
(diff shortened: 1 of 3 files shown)


src/main/host/syscall/fcntl.c
```diff
@@ -184,3 +184,7 @@ SysCallReturn syscallhandler_fcntl(SysCallHandler* sys,
 
     return (SysCallReturn){.state = SYSCALL_DONE, .retval.as_i64 = result};
 }
+
+SysCallReturn syscallhandler_fcntl64(SysCallHandler* sys, const SysCallArgs* args) {
+    return syscallhandler_fcntl(sys, args);
```
sporksmith, line 189: Took me a bit of figuring out if/why this is safe. From looking at the man page, I *think* it's safe because the only differences are when it uses `*64` variations of the commands... but if that's the case I don't understand why it's a separate syscall at all.

Anyway, could probably use a brief comment here about why we think this is safe.

sporksmith (approved):

src/main/host/syscall/fcntl.c
robgjansen: Hmm, the man page hints at otherwise undocumented `F_GETLK64`, `F_SETLK64`, and `F_SETLKW64` commands for fcntl64 in the NOTES section:

> The original Linux fcntl() system call was not designed to handle
       large file offsets (in the flock structure).  Consequently, an
       fcntl64() system call was added in Linux 2.4.  The newer system call
       employs a different structure for file locking, flock64, and
       corresponding commands, F_GETLK64, F_SETLK64, and F_SETLKW64.
       However, these details can be ignored by applications using glibc,
       whose fcntl() wrapper function transparently employs the more recent
       system call where it is available.

I should implement those I guess.

But I agree, I don't know why this requires a distinct function from `fcntl`, since the `flock64` struct could have already been inferred by the command type.

robgjansen: Unless it's possible to call `fcntl64` with the command, `F_GETLK`, and that would transparently get handled as `F_GETLK64`. But such behavior is not documented anywhere so I would be really annoyed if that were possible.

sporksmith: > Unless it's possible to call fcntl64 with the command, F_GETLK, and that would transparently get handled as F_GETLK64. But such behavior is not documented anywhere so I would be really annoyed if that were possible.

Agreed.

AFAICT the code is probably ok as-is, and we don't necessarily need to implement `F_GETLK64` etc in this PR; it'd just be helpful to have a comment here briefly summarizing why we think this is ok.

PR #1028, opened 2020-11-17
Fix bug in UDP socket stats

stevenengler:
The UDP socket stats were not being set correctly when receiving packets, so fixed that in one commit. Also, the stats were being set twice, once in the socket code (`socket_addToOutputBuffer()` and `socket_removeFromOutputBuffer()`), and then in the UDP code. There's no reason to set it twice, so the UDP socket stat updating was removed in another commit.


sporksmith:
Looks plausible, but I'll let Rob take a look at this one since I haven't touched this code much

robgjansen (approved):
👍

PR #1029, opened 2020-11-18
Pthread join

sporksmith:
Implements clear-and-wake on `set_child_tid`, per set_tid_address(2)

This also appears to be sufficient for `pthread_join` to work, so re-enabled corresponding tests.

This does *not* yet implement robust lists as per `get_robust_list(2)`. One could imagine alternatively implementing `pthread_join` using that mechanism, but it appears that our supported platforms do not.

Progress on #826

### DIFF
--- a/src/main/host/thread.c
+++ b/src/main/host/thread.c
@@ -212,6 +212,11 @@ SysCallCondition* thread_getSysCallCondition(Thread* thread) {
     return thread->cond;
 }
 
+PluginVirtualPtr thread_getTidAddress(Thread* thread) {
+    MAGIC_ASSERT(thread);
+    return thread->tidAddress;
+}
+
 void thread_setTidAddress(Thread* thread, PluginPtr addr) {
     MAGIC_ASSERT(thread);
     thread->tidAddress = addr;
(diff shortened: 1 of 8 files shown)


src/main/host/thread.c
```diff
@@ -212,6 +212,11 @@ SysCallCondition* thread_getSysCallCondition(Thread* thread) {
     return thread->cond;
 }
 
+PluginPhysicalPtr thread_getTidAddress(Thread* thread) {
```
robgjansen, line 215: The call signature does not match the one below in `thread.h` (`PluginVirtualPtr` vs `PluginPhysicalPtr`).

I *think* this one should be `PluginVirtualPtr`?

This would become important if we stop typedef'ing these both to be the same as `PluginPtr`.

robgjansen (approved):
Looks good other than a minor naming issue.

src/main/host/thread.c
sporksmith: Oops, good catch. Yeah, I was thinking of making these separate types to help avoid these mistakes at the same time as improving the virtual -> physical address translation. Maybe I'll see if I can knock that out today

PR #1030, opened 2020-11-20
Improve physical address handling

### LINKED ISSUE #996: Add method to convert plugin virtual to physical addresses
The futex syscall uses memory addresses as an ID that can be used to uniquely identify a futex object. However, normally the kernel converts the virtual memory address passed by the futex caller into a physical address based on page table and index. This guarantees that the address is unique across processes.

We should handle this virtual to physical address conversion. This is currently stubbed out in [`process_getPhysicalAddress`](https://github.com/shadow/shadow/blob/b9797048a196906b15d1435cf24ad4c1194aac7e/src/main/host/process.c#L614) in `process.c`.

sporksmith:
* Make PluginPhysicalPtr a distinct type from PluginVirtualPtr
* Incorporate pid into simulated physical addresses

Fixes #996


robgjansen (approved):
👍

PR #1047, opened 2020-12-09
Prevent idle workers in the work stealing policy

### LINKED ISSUE #1046: Bug in work stealing scheduler policy causing idle workers
The work-stealing scheduling policy is occasionally causing some worker threads to be idle even though there is outstanding work to perform. This is item 1 in #1041, more details below.

There is a bug in the logic that is run at the beginning of each simulation round. When the round changes, each thread moves all of its assigned hosts from its `processed` queue back into the `unprocessed` queue so that the work associated with those hosts in the new round can proceed. There is a race condition here where thread A has nothing to do (no hosts assigned) and tries to start stealing a host from thread B's `unprocessed` queue before thread B has repopulated it for the new round. Thread A then thinks all the work is already done and sits idle, while thread B is stuck with a full queue of hosts that it must process.

Here is a simple program to create a ~5 second cpu workload (on my machine):

```C
#include <stdint.h>
int main(int argc, char* argv[]) {
    volatile uint64_t i = 0;
    while(i < 3000000000)
        i += 1;
}
```

And the shadow config:

```xml
<shadow stoptime="2">
  <topology><![CDATA[<graphml xmlns="http://graphml.graphdrawing.org/xmlns" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://graphml.graphdrawing.org/xmlns http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd">
    <node id="poi-1">
      <data key="d0">US</data>
      <data key="d1">10240</data>
      <data key="d2">10240</data>
    </node>
      <data key="d3">1.0</data>
      <data key="d4">0.0</data>
    </edge>
  </graph>
</graphml>
]]></topology>
  <plugin id="spin" path="/home/rjansen/dev/shadow/build/src/test/spin/libshadow-plugin-test-spin.so"/>
  <host id="testhost" quantity="12">
    <process plugin="spin" starttime="1" arguments=""/>
  </host>
</shadow>
```

If each run of `main` takes 5 seconds, then running 12 hosts and 12 threads should complete in 5 seconds. But sometimes it takes 2 cycles (10 seconds) because some of the workers are idle and others end up processing 2 hosts instead of 1. I was able to observe this behavior consistently when using 12 workers, and about 3 workers on average were idle.

robgjansen:
This fixes two main issues:
- Ensures that the very first round is synchronized properly, by waiting until the workers finish setting up all of the host objects before advancing the master clock to scheduling round 1
- Ensures that during every round, no thread can steal from another thread until the other thread has reset its "unprocessed" hosts queue

Closes #1046

### DIFF
--- a/src/main/core/scheduler/scheduler.c
+++ b/src/main/core/scheduler/scheduler.c
@@ -33,6 +33,7 @@ struct _Scheduler {
 
     /* barrier for worker threads to start and stop running */
     CountDownLatch* startBarrier;
+    CountDownLatch* bootBarrier;
     CountDownLatch* finishBarrier;
     /* barrier to wait for worker threads to finish processing this round */
     CountDownLatch* executeEventsBarrier;
@@ -121,6 +122,7 @@ Scheduler* scheduler_new(SchedulerPolicyType policyType, guint nWorkers, gpointe
     g_mutex_init(&(scheduler->globalLock));
 
     scheduler->startBarrier = countdownlatch_new(nWorkers+1);
+    scheduler->bootBarrier = countdownlatch_new(nWorkers + 1);
     scheduler->finishBarrier = countdownlatch_new(nWorkers+1);
     scheduler->executeEventsBarrier = countdownlatch_new(nWorkers+1);
     scheduler->collectInfoBarrier = countdownlatch_new(nWorkers+1);
@@ -308,6 +310,7 @@ static void _scheduler_free(Scheduler* scheduler) {
     countdownlatch_free(scheduler->collectInfoBarrier);
     countdownlatch_free(scheduler->prepareRoundBarrier);
     countdownlatch_free(scheduler->startBarrier);
+    countdownlatch_free(scheduler->bootBarrier);
     countdownlatch_free(scheduler->finishBarrier);
 
     g_mutex_clear(&(scheduler->globalLock));
@@ -570,6 +573,9 @@ gboolean scheduler_isRunning(Scheduler* scheduler) {
 }
 
 void scheduler_awaitStart(Scheduler* scheduler) {
+    /* Called by worker threads. When we are in single thread mode, all of the barriers have a count
+     * of 1 so the countDownAwait functions always return immediately here. */
+
     /* set up the thread timer map */
     g_mutex_lock(&scheduler->globalLock);
     if(!g_hash_table_lookup(scheduler->threadToWaitTimerMap, GUINT_TO_POINTER(pthread_self()))) {
@@ -585,17 +591,17 @@ void scheduler_awaitStart(Scheduler* scheduler) {
     /* each thread will boot their own hosts */
     _scheduler_startHosts(scheduler);
 
-    if(scheduler->policyType != SP_SERIAL_GLOBAL) {
-        /* when everyone finishes starting the hosts, the main thread can prepare
-         * round 1 */
-        countdownlatch_countDownAwait(scheduler->finishBarrier);
-    }
+    /* when all hosts are started, the main thread can prepare round 1 */
+    countdownlatch_countDownAwait(scheduler->bootBarrier);
 
     /* everyone will wait for the next round to be ready */
     countdownlatch_countDownAwait(scheduler->prepareRoundBarrier);
 }
 
 void scheduler_awaitFinish(Scheduler* scheduler) {
+    /* Called by worker threads. When we are in single thread mode, all of the barriers have a count
+     * of 1 so the countDownAwait functions always return immediately here. */
+
     /* each thread will run cleanup on their own hosts */
     g_mutex_lock(&scheduler->globalLock);
     scheduler->isRunning = FALSE;
@@ -608,6 +614,8 @@ void scheduler_awaitFinish(Scheduler* scheduler) {
 }
 
 void scheduler_start(Scheduler* scheduler) {
+    /* Called by the scheduler thread. */
+
     _scheduler_assignHosts(scheduler);
 
     g_mutex_lock(&scheduler->globalLock);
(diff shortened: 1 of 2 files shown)


robgjansen:
I'm running a before/after Tor simulation on this to understand the performance implications, but the code could be reviewed in the meantime.

src/main/core/scheduler/scheduler.c
```diff
@@ -585,7 +585,13 @@ void scheduler_awaitStart(Scheduler* scheduler) {
     /* each thread will boot their own hosts */
     _scheduler_startHosts(scheduler);
 
-    /* everyone is waiting for the next round to be ready */
+    if (scheduler->policyType != SP_SERIAL_GLOBAL) {
+        /* when everyone finishes starting the hosts, the main thread can prepare
+         * round 1 */
+        countdownlatch_countDownAwait(scheduler->finishBarrier);
```
sporksmith, line 591: I was a bit confused both about reusing `finishBarrier` here, and why wait on `prepareRoundBarrier` below doesn't already accomplish what you want.

I see now that `prepareRoundBarrier` is actually waited at the *end* of a round in `scheduler_continueNextRound`. Would moving it to the *start* of a round solve the bug without having to (re)use another barrier here, as well as being more consistent with its name?

src/main/core/scheduler/scheduler.c
```diff
@@ -609,8 +615,14 @@ void scheduler_start(Scheduler* scheduler) {
...
 
     if(scheduler->policyType != SP_SERIAL_GLOBAL) {
-        /* this will cause a worker to execute the locked initialization in awaitStart */
+        /* this will cause all workers to start their hosts in awaitStart */
         countdownlatch_countDownAwait(scheduler->startBarrier);
+        /* we wait for the workers to finish starting hosts before preparing
+         * round 1 */
+        countdownlatch_countDownAwait(scheduler->finishBarrier);
+        /* reset the barriers so we can use them later */
+        countdownlatch_reset(scheduler->startBarrier);
```
sporksmith, line 624: It doesn't look like `startBarrier` is reused, so this shouldn't be needed? (And same for `finishBarrier` if you end up not reusing it)

src/main/core/scheduler/scheduler_policy_host_steal.c
```diff
@@ -42,6 +43,7 @@ struct _HostStealThreadData {
     /* which worker thread this is */
     guint tnumber;
     GMutex lock;
+    volatile bool isStealable;
```
sporksmith, line 46: I don't think you need `volatile` here; the `__atomic` helpers should already tell the compiler not to optimize away those accesses. (OTOH I suppose it doesn't hurt anything other than being slightly confusing why it's being used here)

src/main/core/scheduler/scheduler_policy_host_steal.c
```diff
@@ -374,6 +379,11 @@ static Event* _schedulerpolicyhoststeal_pop(SchedulerPolicy* policy, SimulationT
         g_rw_lock_reader_lock(&data->lock);
         HostStealThreadData* stolenTdata = g_array_index(data->threadList, HostStealThreadData*, stolenTnumber);
         g_rw_lock_reader_unlock(&data->lock);
+
+        /* make sure the workload has been updated for this round */
+        while (!__atomic_load_n(&stolenTdata->isStealable, __ATOMIC_ACQUIRE)) {
```
sporksmith, line 384: I *think* this looks correct, though could potentially spin more than we'd like, and in general atomic operations outside of implementing higher-level sync primitives is a bit obscure. Maybe a barrier could work instead of a raw atomic?

sporksmith (approved):
Approving since this looks correct AFAICT, though I have some suggestions for clarity/maintainability.

src/main/core/scheduler/scheduler_policy_host_steal.c
sporksmith: (Or if spinning actually is what you want here, maybe a spin lock?)

robgjansen:
In a small 1% Tor network workload, Shadow run time is nearly identical before and after this PR (though maybe it would help more on other workloads like Ryan's cpu-based workloads).

[run_time.pdf](https://github.com/shadow/shadow/files/5670127/run_time.pdf)

src/main/core/scheduler/scheduler.c
robgjansen: I had first added a new barrier, but then decided to re-use the `finishBarrier` since it wasn't needed at the start. But I can see how that could lead to issues later, so I will go back to using a separate one.

The `prepareRoundBarrier` needs to be at the end, because the state that is updated while the workers are waiting at the barrier (i.e., the clock) is what the workers use to break out of their running while loop. If I moved `prepareRoundBarrier` to the beginning, then we would need an extra check and break statement---which is certainly possible but also more lines of code. (Perhaps `transitionRoundBarrier` is a better name.)

Ultimately, I'm not really trying to refactor this admittedly ugly code right now, just trying to minimize the changes needed to get to a correct state. That being said, I do want to rewrite the scheduler in rust (in the dev branch) at some point, because (i) I want to implement a continuous round scheduler to replace the current discrete round scheduler, and (ii) I think we can design a more efficient scheduler/scheduling policy now that we don't have to worry about migrating plugin thread-local storage anymore. But this is all for another month :)

src/main/core/scheduler/scheduler_policy_host_steal.c
robgjansen: I'll remove it. I don't have much experience with atomics, so this was just me being stupid.

src/main/core/scheduler/scheduler_policy_host_steal.c
robgjansen: I don't think a `pthread_barrier` will work here, because I want the bool functionality - I only want the threads waiting for it to turn true to wait, but not the thread that sets it to true. I think `pthread_barrier` requires all threads that decrement the count to wait until it reaches 0.

I was planning to use a `CountDownLatch` but then got excited when I realized I could do it with an atomic with hopefully less synchronization overhead. (The countdown latch uses both a mutex and a condition variable.) I'm using the atomic to prevent a race condition so I don't expect the wait will be very long, so I think spinning is OK.

`pthread_spin_lock` could work, but it's a bit kludgy. If we are using N workers then it's possible that N-1 of them reach this point. I just want the N-1 workers to check/wait for the boolean to become true. If I use a spin lock, I would need to use the unlocked state to represent true, meaning that the N-1 workers would acquire and immediately release the lock every time they want to check if they can steal from another worker. For N workers, this happens N^2 times per round. The atomic seems simpler.

Thoughts?

sporksmith: > I don't think a pthread_barrier will work here
...
> I was planning to use a CountDownLatch

Oops, yeah CountDownLatch was what I meant.

> but then got excited when I realized I could do it with an atomic with hopefully less synchronization overhead. (The countdown latch uses both a mutex and a condition variable.) I'm using the atomic to prevent a race condition so I don't expect the wait will be very long, so I think spinning is OK.

I think the fast path without contention would be the ~same? But yeah, the main practical benefit would be avoiding a potential infinite spin. (re?)reading your comment, it sounds like spinning is probably ok.

> pthread_spin_lock could work, but it's a bit kludgy

Hmm, I see what you mean.

Maybe a semaphore (or even Ryan's spinning semaphore) would be a more appropriate abstraction?

Anway, I'm fine with leaving as-is. It makes the code a little more opaque to directly use these low level atomic functions here, but it's not a big deal.

robgjansen: OK, I'm leaving as-is (Ryan's spinning semaphore is a great idea but only available in dev branch atm).

PR #1056, opened 2020-12-15
Updated dependabot config to use the new GitHub label

stevenengler:


robgjansen (approved):
Sorry for causing this bug when I messed with the labels!

stevenengler:
Don't worry, it would be nice if GitHub let us specify the repository labels in a `.github/` file.

PR #1059, opened 2020-12-17
Prefer vfork to fork

RWails:
Changes thread_preload.c to use `vfork()` instead of `fork()` during
process creation. This change has a significant impact on experiment
performance with many (> 1000) processes to manage.
Initialization runtime is reduced by a large factor.

`prctl(PR_SET_PDEATHSIG, SIGKILL)` and `getppid()` have been moved to
the shim in `_shim_load()`.

n.b. I believe this change does not affect ptrace mode. The `prctl` call
may be duplicated, but shouldn't hurt for now.

### DIFF
--- a/src/main/host/thread_preload.c
+++ b/src/main/host/thread_preload.c
@@ -52,9 +52,7 @@ static ThreadPreload* _threadToThreadPreload(Thread* thread) {
     return (ThreadPreload*)thread;
 }
 
-static Thread* _threadPreloadToThread(ThreadPreload* thread) {
-    return (Thread*)thread;
-}
+static Thread* _threadPreloadToThread(ThreadPreload* thread) { return (Thread*)thread; }
 
 static void _threadpreload_auxFree(void* p, void* _) {
     ShMemBlock* blk = (ShMemBlock*)p;
@@ -80,8 +78,7 @@ static void _threadpreload_auxWrite(void* p, void* t) {
 
     ShimEvent resp = {0};
 
-    req.event_data.shmem_blk.serial =
-        shmemallocator_globalBlockSerialize(&write_blk->blk);
+    req.event_data.shmem_blk.serial = shmemallocator_globalBlockSerialize(&write_blk->blk);
     req.event_data.shmem_blk.plugin_ptr = write_blk->plugin_ptr;
     req.event_data.shmem_blk.n = write_blk->n;
 
@@ -113,7 +110,7 @@ void threadpreload_free(Thread* base) {
     worker_countObject(OBJECT_TYPE_THREAD_PRELOAD, COUNTER_TYPE_FREE);
 }
 
-static gchar** _add_shadow_pid_to_env(gchar **envp) {
+static gchar** _add_shadow_pid_to_env(gchar** envp) {
 
     enum { BUF_NBYTES = 256 };
     char strbuf[BUF_NBYTES] = {0};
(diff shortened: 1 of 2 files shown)


src/main/host/thread_preload.c
```diff
@@ -126,21 +140,6 @@ static pid_t _threadpreload_fork_exec(ThreadPreload* thread, const char* file, c
...
-                error("prctl: %s", g_strerror(errno));
-                return -1;
-            }
-            // Validate that Shadow is still alive (didn't die in between forking and calling
-            // prctl).
-            if (getppid() != shadow_pid) {
-                error("parent (shadow) exited");
-                return -1;
-            }
             int rc = execvpe(file, argv, envp);
```
robgjansen, line 143: I think we want a comment somewhere around here about how the code is designed so that the child process does not execute any code after the `vfork` and before calling `exec`, to guard against future programmers changing this code from accidentally causing undefined behavior.

src/shim/shim.c
```diff
@@ -139,6 +140,34 @@ static void _shim_load() {
...
+    const char* shadow_pid_str = getenv("SHADOW_PID");
+    if (shadow_pid_str) {
+        unsigned long long tmp_pid = 0;
+        pid_t shadow_pid = 0;
+        int rc = sscanf(shadow_pid_str, "%llu", &tmp_pid);
+        if (rc != 1) {
+            error("SHADOW_PID does not contain an unsigned: %s", shadow_pid_str);
+        } else {
+
+            assert((pid_t)tmp_pid == tmp_pid);
```
robgjansen, line 161: Is this vacuously true?

src/main/host/thread_preload.c
```diff
@@ -113,10 +113,24 @@ void threadpreload_free(Thread* base) {
...
+    envp = g_environ_setenv(envp, "SHADOW_PID", strbuf, TRUE);
+
+    return envp;
+}
+
 static pid_t _threadpreload_fork_exec(ThreadPreload* thread, const char* file, char* const argv[],
                                       char* const envp[]) {
     pid_t shadow_pid = getpid();
-    pid_t pid = fork();
+    pid_t pid = vfork();
```
robgjansen, line 133: Maybe add a short comment about why we are using `vfork` since moving away from it isn't a decision that should be taken lightly.

robgjansen (approved):
A couple of comments would be helpful :)

src/shim/shim.c
sporksmith: I think it's to validate it wasn't truncated in the conversion to pid_t? +1 that it could use a comment

It'd be nice if we could parse into a pid_t directly, but I suppose that's tricky since we can't know the format specifier to provide to sscanf (without assuming something about the definition of pid_t).

sporksmith (approved):

src/shim/shim.c
sporksmith: OTOH we could just not do this downcast, and compare directly. C will widen the pid if it's narrower.
```
unsigned long long shadow_pid = 0;
if (sscanf(shadow_pid_str, "%llu", &shadow_pid) != 1) {
...
// Implicitly widen return value of getppid
if (getppid() != shadow_pid) {
```

Btw this control flow would probably be a bit easier to follow by using early exits to avoid excessive nesting.
```
if (/* something failed */) {
  /* log and exit */
}
if (/* something else failed */) {
  /* log and exit */
}
```

src/main/host/thread_preload.c
RWails: Fixed

src/shim/shim.c
RWails: Fixed

src/main/host/thread_preload.c
RWails: Fixed

PR #1062, opened 2020-12-21
Upgrade rand, rand_core, and rand_chacha crates

stevenengler:
This PR replaces #1054, #1055, #1061 since all of these packages need to be upgraded simultaneously.


stevenengler:
(This is ready for review, but I'll wait until January before I assign it to anyone.)

robgjansen (approved):
Pretty simple version bumps :)

PR #1063, opened 2021-01-01
CPU Pinning

RWails:
This PR adds an optional flag that enables CPU pinning for Shadow workers and plugins. Workers are pinned uniformly among logical CPUs available on the machine. Plugin threads are pinned to the same logical CPU that the worker is pinned to. This behavior is enabled with the `--cpu-pin=1` flag.

Pinning can yield a significant performance improvement. On syscall dense workloads, I am measuring a 20-50% improvement in performance (using `--cpu-pin=1 and --preload-spin-max=0`).

### DIFF
--- a/src/main/host/affinity.c
+++ b/src/main/host/affinity.c
@@ -5,65 +5,70 @@
 #endif              // _GNU_SOURCE
 
 #include <assert.h>
+#include <glib.h>
 #include <sched.h>
+#include <stdbool.h>
 #include <stddef.h>
 #include <sys/sysinfo.h>
 
+#include "main/core/support/options.h"
 #include "support/logger/logger.h"
 
 // Compute once and cache this value. Apparently fetching is expensive and
 // requires parsing procfs.
-static int _NCPUS = -1; // Starts uninitialized.
+static int _ncpus = -1; // Starts uninitialized.
 
-__attribute__((constructor)) static void _init_ncpus() { _NCPUS = get_nprocs(); }
+static gint _affinity_enabled = 0;
+
+OPTION_EXPERIMENTAL_ENTRY(
+    "cpu-pin", 0, 0, G_OPTION_ARG_INT, &_affinity_enabled,
+    "Pin workers to fixed CPU cores, pin plugins to the worker's core", "[0|1]")
+
+__attribute__((constructor)) static void _init_ncpus() { _ncpus = get_nprocs(); }
 
 int affinity_getGoodWorkerAffinity(guint worker_thread_id) {
-    assert(_NCPUS > 0);                    // Our host better have at least one CPU.
-    return (int)worker_thread_id % _NCPUS; // Uniform distribution of workers over CPUs
+    assert(_ncpus > 0);                    // Our host better have at least one CPU.
+    return (int)worker_thread_id % _ncpus; // Uniform distribution of workers over CPUs
 }
 
 int affinity_setProcessAffinity(pid_t pid, int old_cpu_num, int new_cpu_num) {
+    assert(_ncpus > 0 && pid >= 0);
 
-    return old_cpu_num;
-
-    if (new_cpu_num == AFFINITY_UNINIT) {
-      return old_cpu_num;
+    // We can short-circuit if there's no work to do.
+    if (!_affinity_enabled || new_cpu_num == AFFINITY_UNINIT || new_cpu_num == old_cpu_num) {
+        return old_cpu_num;
     }
 
     cpu_set_t* cpu_set = NULL;
+    bool set_affinity_suceeded = false;
+    int retval = new_cpu_num;
 
-    if (new_cpu_num >= _NCPUS) {
-        goto FAIL;
+    if (new_cpu_num < _ncpus) { // Good, we are trying to set to a valid CPU
+        cpu_set = CPU_ALLOC(_ncpus);
+
+        if (cpu_set) { // Good, the CPU set allocation succeeded
+
+            size_t cpu_set_size = CPU_ALLOC_SIZE(_ncpus);
+
+            // Clear the CPU set
+            CPU_ZERO_S(cpu_set_size, cpu_set);
+            // And then add the new_cpu_num as the only element of the set
+            CPU_SET_S(new_cpu_num, cpu_set_size, cpu_set);
+
+            int rc = sched_setaffinity(pid, cpu_set_size, cpu_set);
+
+            set_affinity_suceeded = (rc == 0);
+        }
     }
 
-    cpu_set = CPU_ALLOC(_NCPUS);
-
-    if (!cpu_set) {
-        goto FAIL;
-    }
-
-    size_t cpu_set_size = CPU_ALLOC_SIZE(_NCPUS);
-
-    // Clear the CPU set
-    CPU_ZERO_S(cpu_set_size, cpu_set);
-    // And then add the new_cpu_num as the only element of the set.
-    CPU_SET_S(new_cpu_num, cpu_set_size, cpu_set);
-
-    int rc = sched_setaffinity(pid, cpu_set_size, cpu_set);
-    if (rc == -1) {
-        goto FAIL;
+    if (!set_affinity_suceeded) {
+        warning("Could not set CPU affinity for PID %d to %d", (int)pid, new_cpu_num);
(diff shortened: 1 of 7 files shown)


jtracey:
I've not done any experiments with Phantom, but on the current release version of Shadow, I've run some experiments that show pinning to the A sides of CPUs (effectively disabling hyperthreads) is noticeably faster than using all logical CPUs -- specifically, on some large experiments, setting the number of workers to the number of physical cores and pinning to the A sides was significantly faster than pinning to A and B sides, whether setting the number of workers to the number of physical or logical cores (though the details will likely vary with specific workload, and that was pinning the entire process, not each thread, so there were still migrations happening). Pinning can be accomplished with numactl, but since A and B sides are frequently (but not always) interleaved numerically, a user who doesn't know to do that and runs Shadow with fewer workers than logical CPUs will likely end up with the workers bunched up on the same physical cores when distributing via `worker_thread_id % _ncpus`.

It might also be worth trying to keep threads on the same socket when multiple are available (and also bind to the memory for those numa nodes, strictly speaking, but I *think* linux is smart enough to handle that step on its own?). I don't know if there's a standard way to get all this information via glibc/syscalls, but it's available in `/proc/cpuinfo`.

src/main/host/affinity.h
```diff
@@ -0,0 +1,42 @@
...
+ */
+int affinity_setProcessAffinity(pid_t pid, int new_cpu_num, int old_cpu_num);
+
+/*
+ * Helper function. Same semantics as affinity_setProcessAffinity but sets the
+ * affinity of the calling thread/process.
+ */
+static inline int affinity_setThisProcessAffinity(int new_cpu_num, int old_cpu_num) {
+  return affinity_setProcessAffinity(0, new_cpu_num, old_cpu_num);
+}
```
robgjansen, line 40: Is it standard to put static inline code in a header? This seems wrong to me, because it's not really static if it can be called from anyone including this header. Can we just remove static, keep the non-static declaration in the header, and move the definition to the `.c` file?

src/main/host/affinity.c
```diff
@@ -0,0 +1,74 @@
...
+int affinity_getGoodWorkerAffinity(guint worker_thread_id) {
+    assert(_ncpus > 0);                    // Our host better have at least one CPU.
+    return (int)worker_thread_id % _ncpus; // Uniform distribution of workers over CPUs
+}
+
+int affinity_setProcessAffinity(pid_t pid, int new_cpu_num, int old_cpu_num) {
+    assert(_ncpus > 0 && pid >= 0);
+
+    // We can short-circuit if there's no work to do.
+    if (!_affinity_enabled || new_cpu_num == AFFINITY_UNINIT || new_cpu_num == old_cpu_num) {
```
robgjansen, line 38: If `new_cpu_num == AFFINITY_UNINIT` then don't we want to "unpin" the process to allow the scheduler to run it anywhere?

src/main/host/affinity.c
```diff
@@ -0,0 +1,74 @@
...
+            CPU_SET_S(new_cpu_num, cpu_set_size, cpu_set);
+
+            int rc = sched_setaffinity(pid, cpu_set_size, cpu_set);
+
+            set_affinity_suceeded = (rc == 0);
+        }
+    }
+
+    if (!set_affinity_suceeded) {
+        warning("Could not set CPU affinity for PID %d to %d", (int)pid, new_cpu_num);
```
robgjansen, line 65: If we use the `cpu-pin` option and we can't pin, should that be an error or a warning? IIRC, Jim thought error makes more sense, rather than continuing to run the simulation without pinning and potentially tricking the user that pinning is active when it isn't. I'm thinking that logging a warning makes sense but then refusing to run the experiment. That's more work to code because you'll have to propagate the error back, but does it give us the behavior we want?

src/main/host/affinity.h
```diff
@@ -0,0 +1,42 @@
+#ifndef AFFINITY_H_
+#define AFFINITY_H_
+
+#include <glib.h>
+#include <sched.h>
+#include <sys/types.h>
+
+enum { AFFINITY_UNINIT = -1 };
```
robgjansen, line 8: [optional nit] A brief note describing this would be nice.

robgjansen (changes requested):
I think we tested the case that @jtracey found works better for classic shadow when we were testing phantom (i.e., effectively disabling hyperthreading); could you summarize our conclusions there?

src/main/host/affinity.h
RWails: Yeah, this is standard when declaring an inline function. See <https://gcc.gnu.org/onlinedocs/gcc/Inline.html>. This function is just a macro/wrapper that's setting the first argument to 0, so we avoid function call overhead.

RWails:
> I think we tested the case that @jtracey found works better for classic shadow when we were testing phantom (i.e., effectively disabling hyperthreading); could you summarize our conclusions there?

Yeah, so most of the performance improvements due to pinning I believe is due to reducing the cost of context switching (this only applied in the process-oriented `dev` branch). By pinning workers and plugins to the same core we achieve very low cost context switching.

I agree with Justin that there's more benefit that could be gained by intelligently pinning workers to particular CPUs, for example taking NUMA nodes into account. Some experiments we've run so far show that pinning to hyperthread cores pessimizes performance, but for the machines I'm running experiments on the HT cores are all high numbered, so by setting `-w` at `num_cores / 2` we avoid pinning to shared cores. I'd like to leave intelligent pinning to a different PR since it could get complicated if we're parsing `/proc/cpuinfo` or pulling in a dependency to examine the platform.

Thanks for the feedback @jtracey !

src/main/host/affinity.c
RWails: I'm using `AFFINITY_UNINIT` to denote the case when AFFINITY is unknown or has never been set (as opposed to the case when affinity is set to all CPUs). I think it'd be better to introduce a new variable to unpin/"unset" affinity if we want that behavior.

src/main/host/affinity.c
RWails: I promoted to critical so that the user has a higher likelihood of seeing.

Crashing with an error makes me nervous because there are potentially thousands of affinity changes that could occur over the course of an experiment (during work stealing, for example). If a single one of those affinity changes doesn't succeed, I'm not sure the experiment should crash since it could otherwise complete successfully with negligible performance impact.

Regardless, if we want to make this a crash instead of a warning, I think it should be done at the call site and not inside this module.

src/main/host/affinity.h
RWails: OK, I added a comment

src/main/host/affinity.c
```diff
@@ -5,65 +5,70 @@
...
-
-    if (new_cpu_num == AFFINITY_UNINIT) {
-      return old_cpu_num;
+    // We can short-circuit if there's no work to do.
+    if (!_affinity_enabled || new_cpu_num == AFFINITY_UNINIT || new_cpu_num == old_cpu_num) {
+        return old_cpu_num;
     }
 
     cpu_set_t* cpu_set = NULL;
+    bool set_affinity_suceeded = false;
```
armadev, line 43: typo, 'suceeded' (I noticed the typo while reading the discussion :)

robgjansen:
We decided to close this PR and open 2 new ones in order to make sure we have the pinning feature in both master (classic shadow) and dev (phantom) branches. The first PR will add the feature to master, then we'll merge master to dev, and the second PR will make the changes needed to do the pinning in dev.

src/main/host/affinity.c
sporksmith: > Crashing with an error makes me nervous because there are potentially thousands of affinity changes that could occur over the course of an experiment

Haven't looked at the code in this PR yet but when I made that comment I believe the proposed pinning was all done during startup. If something goes wrong during startup I think it makes sense to crash. I agree we want to avoid crashing in the middle of a simulation if we can.

sporksmith:
> We decided to close this PR and open 2 new ones in order to make sure we have the pinning feature in both master (classic shadow) and dev (phantom) branches. The first PR will add the feature to master, then we'll merge master to dev, and the second PR will make the changes needed to do the pinning in dev.

Ok, removing myself as a reviewer on this one, then :)

PR #1075, opened 2021-01-28
Miscellaneous syscalls to support running Tor

robgjansen:


robgjansen:
Oh I see I should enable the `test_getpid_kill` test in `src/test/unistd/test_unistd.rs` now that `kill` is supported.

stevenengler (approved):
Looks good!

During a meeting we discussed a few limitations of the signal support. The plan is just to document these for now.

1. There's a minor issue of causality when using the native Linux signal functionality rather than implementing signals as events/tasks, but this shouldn't be important since only one process/thread per host can be running at once.
2. If the plugin uses the `SIGSTOP` or `SIGCONT` signals, it could cause issues for ptrace mode.
3. In preload mode, this could cause issues for plugins that have syscall handlers. If process A sends a signal to process B, process B which is not currently running (and is waiting on a syscall) will receive the signal and start running the signal handler. The signal handler may make a syscall which gets intercepted by Shadow, and Shadow will try to make a second syscall while it's already processing the first one and will result in an error/crash.
4. In preload mode, even if there is no signal handler registered, if the shim is waiting for a message from a Shadow worker (the shim is waiting on the semaphore), a signal will interrupt the wait and Shadow will continue as if the semaphore wait has finished when it hasn't.

robgjansen:
Another option for dealing with signals is to only allow processes to send signals to themselves, which is the approach taken in the ["Reproducible Containers" paper](https://doi.org/10.1145/3373376.3378519).

Issue #847, opened 2020-06-12
Thoroughly test the getsockopt()/setsockopt() syscalls


robgjansen:
See #832 for an overview of our approach.

PR #1078, opened 2021-01-28
Improve getsockopt() and setsockopt() and add tests

stevenengler:
This PR makes `getsockopt()` and `setsockopt()` behave more similar to Linux, and adds tests for the options that Shadow supports.

Closes shadow/shadow#847.


src/main/host/syscall/socket.c
```diff
@@ -310,40 +300,38 @@ static int _syscallhandler_getTCPOptHelper(SysCallHandler* sys, TCP* tcp,
...
-        info("Unable to store option in %zu bytes.", (size_t)*optlen);
-        return -EINVAL;
-    }
-
-    int* optval = process_getWriteablePtr(sys->process, sys->thread, optvalPtr, sizeof(int));
 
     switch (optname) {
         case SO_SNDBUF: {
-            *optval = socket_getOutputBufferSize(sock);
+            int sndbuf = socket_getOutputBufferSize(sock);
```
sporksmith, line 310: This is a slightly confusing name. `sndbuf_size`?

src/main/host/syscall/socket.c
```diff
@@ -310,40 +300,38 @@ static int _syscallhandler_getTCPOptHelper(SysCallHandler* sys, TCP* tcp,
...
+            int num_bytes = MIN(*optlen, sizeof(sndbuf));
+            memcpy(optval, &sndbuf, num_bytes);
+            *optlen = num_bytes;
             return 0;
         }
         case SO_RCVBUF: {
-            *optval = socket_getInputBufferSize(sock);
+            int rcvbuf = socket_getInputBufferSize(sock);
```
sporksmith, line 317: This is a slightly confusing name. `rcvbuf_size`?

src/main/host/syscall/socket.c
```diff
@@ -310,40 +300,38 @@ static int _syscallhandler_getTCPOptHelper(SysCallHandler* sys, TCP* tcp,
...
-                                              int optname, PluginPtr optvalPtr,
-                                              PluginPtr optlenPtr) {
-    /* Get the len via clone, so we can write to optlenPtr too. */
-    const socklen_t* optlen =
-        process_getReadablePtr(sys->process, sys->thread, optlenPtr, sizeof(*optlen));
-
+static int _syscallhandler_getSocketOptHelper(SysCallHandler* sys, Socket* sock, int optname,
+                                              void* optval, socklen_t* optlen) {
     /* All options we currently support are integers. When adding support for
      * more options, make sure to check length requirements. */
```
sporksmith, line 306: I think this comment needs updating. It looks like this change is meant to do one or both of making it easier to support non-int-sized options in the future, and in the current options handle if the caller passes an integer type smaller than `int`? For the latter, is that actually something we need to support?

src/main/host/syscall/socket.c
```diff
@@ -366,8 +354,15 @@ static int _syscallhandler_setSocketOptHelper(SysCallHandler* sys, Socket* sock,
...
                 process_getReadablePtr(sys->process, sys->thread, optvalPtr, sizeof(int));
-            size_t newsize =
-                (*val) * 2; // Linux kernel doubles this value upon setting
+            size_t newsize = (*val) * 2; // Linux kernel doubles this value upon setting
+
+            // Linux also has limits SOCK_MIN_SNDBUF (slightly greater than 4096) and the sysctl max
+            // limit. We choose some reasonable limits for Shadow. The minimum limit in man 7 socket
+            // is incorrect.
+            newsize = MAX(newsize, 4096);
+            // This upper limit was chosen as an arbitrarily high number.
```
sporksmith, line 363: It's worth mentioning a bit of the reasoning. e.g. that this is effectively a ceiling of how many un-acked packets we can have, since we need to buffer the packets for possible retransmission (or was that a different setting?), and Rob's back-of-the-envelope math to ensure we can saturate a 1 Gbps link

src/main/host/syscall/socket.c
```diff
@@ -377,8 +372,15 @@ static int _syscallhandler_setSocketOptHelper(SysCallHandler* sys, Socket* sock,
...
                 process_getReadablePtr(sys->process, sys->thread, optvalPtr, sizeof(int));
-            size_t newsize =
-                (*val) * 2; // Linux kernel doubles this value upon setting
+            size_t newsize = (*val) * 2; // Linux kernel doubles this value upon setting
+
+            // Linux also has limits SOCK_MIN_RCVBUF (slightly greater than 2048) and the sysctl max
+            // limit. We choose some reasonable limits for Shadow. The minimum limit in man 7 socket
+            // is incorrect.
+            newsize = MAX(newsize, 2048);
+            // This upper limit was chosen as an arbitrarily high number.
```
sporksmith, line 381: Again, a little bit of explanation about context and how we got to this number would be helpful in the future.

src/main/host/syscall/socket.c
```diff
@@ -991,26 +993,41 @@ SysCallReturn syscallhandler_getsockopt(SysCallHandler* sys,
...
     utility_assert(socket_desc);
 
-    /* The pointers must be non-null. */
-    if (!optvalPtr.val || !optlenPtr.val) {
+    /* The optlen pointer must be non-null. */
+    if (!optlenPtr.val) {
         return (SysCallReturn){.state = SYSCALL_DONE, .retval.as_i64 = -EFAULT};
     }
 
+    /* Get the len via clone, so we can write to optlenPtr too. */
```
sporksmith, line 1001: I know you just moved this, but could you update this stale comment? I think we can just drop it.

We used to use a "clone" API here as a workaround for being able to read + write the same pointer, but don't need it since we added `process_getMutablePtr`.

sporksmith (approved):

src/main/host/syscall/socket.c
stevenengler: The goal was just to make it behave more like Linux, where the program can call getsockopt with any length, and it will only copy that many bytes. We don't need to support that, and I can't think of any legitimate reasons a program would do that, but changed it just so that it better follows the Linux system call interface and since we can do it safely. When writing tests, it's useful to test things like passing small buffers to make sure Shadow doesn't segfault. But if Shadow behaves different (even if there isn't a bug/segfault/etc), than we need to test for two different outcomes depending on whether the test is running under Shadow or not, and figured it would just be easier to make Shadow behave the same as Linux. But if there are any downsides to this change let me know and I'm okay with changing it back.

And the sizes for the implemented options in `_syscallhandler_getSocketOptHelper()` are only ints, but `_syscallhandler_getTCPOptHelper()` has an option which returns a `struct tcp_info`, which is much larger, so it's nice to have the `process_getWriteablePtr()` in the parent function instead of duplicating it in both helpers.

src/main/host/syscall/socket.c
stevenengler: So we're not sure what Linux does, but in Shadow this setting doesn't affect the TCP retransmit queue as the socket queue and retransmit queue are separate. So I don't think there's actually a reason to keep it large, but am just doing so anyways to not break anything.

src/main/host/syscall/socket.c
sporksmith: Yup, seems reasonable. Would still be good to update the comment, though, since it still reads as if the size were hard-coded to be an integer. Maybe we can just be dropped entirely?

src/main/host/syscall/socket.c
sporksmith: Should we still have some upper limit to avoid excessive allocation/buffering? 256 MiB seems like a plausible limit to me; I was just looking for a comment with our current understanding of how to choose this limit.

If we do think no upper limit is needed here (e.g. because an upper limit set somewhere else will already have the desired effect), that'd probably also be worth a comment.

sporksmith (approved):

src/main/host/syscall/socket.c
stevenengler: I added it back with hopefully a better comment.

PR #1080, opened 2021-01-29
Add preload and test for getifaddrs

robgjansen:
Using nix here instead of libc to call `getifaddrs` took me way too long :( so I'm leaving the libc version as a TODO.


src/test/ifaddrs/test_ifaddrs.rs
```diff
@@ -0,0 +1,39 @@
...
+                let address_str = address.to_str();
+                let ip = address_str.split(":").collect::<Vec<&str>>()[0];
+                println!(
+                    "found ifaddr interface {} address {}",
+                    ifaddr.interface_name, ip
+                );
+                ip_vec.push(String::from(ip));
+            }
+            None => {}
+        }
```
stevenengler, line 29: I'm not sure if this will work on computers that have an interface without an Inet address. I think it might be better to do:

```rust
if let Some(nix::sys::socket::SockAddr::Inet(address)) = ifaddr.address {
    let address_str = address.to_str();
    ...
}
```

stevenengler (approved):
Looks good, just two small comments.

PR #1084, opened 2021-02-03
Adds CPU pinning to the development branch.

RWails:
Adds CPU pinning to the development branch.
On calls to `thread_run` and`thread_resume`, the plugin's CPU affinity is set to the worker's
current affinity when affinity mode is enabled.

### DIFF
--- a/src/main/host/thread_protected.h
+++ b/src/main/host/thread_protected.h
@@ -52,6 +52,8 @@ struct _Thread {
     // Non-null if blocked by a syscall.
     SysCallCondition* cond;
 
+    int affinity;
+
     MAGIC_DECLARE;
 };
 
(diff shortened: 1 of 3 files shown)


src/main/host/thread_protected.h
```diff
@@ -52,6 +52,8 @@ struct _Thread {
     // Non-null if blocked by a syscall.
     SysCallCondition* cond;
 
+    int affinity;
```
robgjansen, line 55: Nit: comment to explain what this is storing?

robgjansen (approved):

src/main/host/thread_protected.h
RWails: fixed

PR #1086, opened 2021-02-03
Added some Rust wrappers around C types

stevenengler:
This implements some Rust-equivalent types for some C types. There is still more to do in future PRs, but this is a start.

### DIFF
--- a/src/main/host/descriptor/pipe.rs
+++ b/src/main/host/descriptor/pipe.rs
@@ -48,10 +48,7 @@ impl PipeFile {
             return SyscallReturn::Error(nix::errno::Errno::EBADF);
         }
 
-        let num_read = {
-            let mut buffer = self.buffer.borrow_mut();
-            buffer.read(bytes, event_queue)
-        };
+        let num_read = self.buffer.borrow_mut().read(bytes, event_queue);
 
         SyscallReturn::Success(num_read as i32)
     }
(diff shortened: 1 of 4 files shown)


src/main/host/descriptor/pipe.rs
```diff
@@ -159,20 +150,16 @@ impl PipeFile {
         // remove any flags that aren't relevant
         let status = self.filter_status(status);
 
-        if do_set_bits {
-            self.status |= status;
-        } else {
-            self.status &= !status;
-        }
+        self.status.set(status, do_set_bits);
```
sporksmith, line 153: Doesn't this still need to clear the bits if `do_set_bits` is `false`?

src/main/host/descriptor/pipe.rs
```diff
@@ -247,40 +226,36 @@ impl SharedBuf {
...
         event_queue: &mut EventQueue,
     ) {
         let old_status = self.status;
 
-        if do_set_bits {
-            self.status |= status;
-        } else {
-            self.status &= !status;
-        }
+        self.status.set(status, do_set_bits);
```
sporksmith, line 249: Clear if `!do_set_bits`?

sporksmith (approved):

src/main/host/descriptor/pipe.rs
stevenengler: The documentation for [`set(flags, bool)`](https://docs.rs/bitflags/1.2.1/bitflags/example_generated/struct.Flags.html#method.set) says "Inserts or removes the specified flags depending on the passed value", and I double checked the [source](https://docs.rs/bitflags/1.2.1/src/bitflags/lib.rs.html#698) and it calls `remove(flags)` if it's `false`.

sporksmith: Ahh, I missed that `do_set_bits` was getting passed as a parameter there

PR #1090, opened 2021-02-09
Avoid logging in the shim until we have a set a logger

### LINKED ISSUE #1070: Shim in ptrace mode logs to plugin's stdout
When running Shadow in ptrace mode, the shim writes a few log messages to the plugin's stdout file.

For example:

```
$ ./setup build --test --debug --clean
$ ./setup test bind
$ cat build/src/test/socket/bind/bind.shadow-ptrace.data/hosts/testnode/testnode.testbind.1000.stdout
Testing test_invalid_fd... ✓
Testing test_non_existent_fd... ✓
Testing test_short_addr... ✓
Testing test_ipv4 <type=1,flag=0>... ✓
[...]
Testing test_double_bind_loopback_and_any <type=2,flag=524288>... ✓
Testing test_double_bind_loopback_and_any <reversed> <type=2,flag=524288>... ✓
Testing test_unspecified_port <type=2,flag=524288>... ✓
Success.
00:00:00.000000debug [shim.c:45] [shim_disableInterposition] enabling native-syscalls via syscall
00:00:00.000000debug [preload_syscalls.c:188] [syscall] Making real syscall 1000
```

From a git bisect, this was introduced in b980d31b.

robgjansen:
Fixes #1070

### DIFF
--- a/src/main/shmem/shmem_cleanup.c
+++ b/src/main/shmem/shmem_cleanup.c
@@ -96,7 +96,7 @@ static bool _shmemcleanup_unlinkIfShadow(const char* filename,
         strncpy(name_buf + 1, filename, SHD_SHMEM_FILE_NAME_NBYTES - 1);
         int rc = shm_unlink(name_buf);
         if (rc == 0) {
-            info("Removing orphaned shared memory file: %s\n", name_buf);
+            info("Removing orphaned shared memory file: %s", name_buf);
             did_remove = true;
         }
     }
@@ -113,7 +113,7 @@ void shmemcleanup_tryCleanup() {
     }
 
     info(
-        "Num. processes in system's procfs: %u\n", g_hash_table_size(proc_set));
+        "Num. processes in system's procfs: %u", g_hash_table_size(proc_set));
 
     // If we can get a list of running processes on the machine, iterate
     // through the files in shared memory and try to remove them.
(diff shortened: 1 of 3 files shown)


sporksmith (changes requested):
Already discussed offline, but getting this out of my review todos :)

What we discussed:
- make the logger's "no logger" behavior be to drop (< WARNING level) messages instead of writing them to stdout
- Don't install a "stdout logger" in the shim's early init
- Do install a "stdout logger" in shadow's early init

It occurs to me now that an alternative could be to just change shadow's early-init-logger to drop < WARNING level messages. (Or to respect shadow's configured log-level, which would also be good but is a bit more plumbing)

robgjansen:
Rather than installing an "early" shadow logger to stdout, using the default logger will log <= warning messages to stderr following the default logger's (new) behavior. I think this is OK, especially since Shadow right now makes heavy use of `g_printerr` in its "early" code rather than calling log functions.

robgjansen:
Oh crap, maybe I don't want commit dd8a947?

sporksmith:
> Oh crap, maybe I don't want commit [dd8a947](https://github.com/shadow/shadow/commit/dd8a947a29fa645c4fb1ce56c8d4bee9e715cb3b)?

Hmm, it's probably fine to remove the shim's early-logger. In principle the default logger could end up deadlocking on libc file IO, but if we're logging at warning level or higher we're probably in trouble anyway.

I'd also be ok with keeping the early-logger but either hard-coding it to always drop <WARNING messages (I think we rarely look at the shim's debug messages anymore), or to add a log-level parameter to the shim-logger so that we can enable DEBUG log messages after initialization.

sporksmith (approved):

PR #1092, opened 2021-02-10
Realtime scheduler compatibility

### LINKED ISSUE #964: Support optional real-time scheduling mode
In our new process-based architecture, we may get a performance boost if we can reduce unnecessary context switches by utilizing a non-default CPU scheduler to run the shadow and plugin threads. Using a real-time scheduler like `SCHED_FIFO` will give us tighter control over when context switches will occur.

We think it's worth supporting a real-time scheduling mode as an optional run-mode, so that we can test if it provides a performance boost. In order for us to be allowed to use a real-time mode, support needs to be configured on the simulation host via ulimit.

More info:
- https://man7.org/linux/man-pages/man7/sched.7.html
- https://man7.org/linux/man-pages/man2/sched_setscheduler.2.html

sporksmith:
This adds some `sched_yield`s inside shadow to prevent poor performance or deadlocks under CPU contention when using the realtime scheduler. These also make performance less-bad when there's CPU contention, even with the regular scheduler.

An earlier version switched to the realtime scheduler inside Shadow itself. Since we can do it externally with the `chrt` command though, we can keep that functionality decoupled from Shadow itself.

Some analysis at https://github.com/sporksmith/phantom-paper/blob/realtime-experiment/experiments/realtime/analysis.ipynb

Fixes #964


src/main/host/thread_preload.c
```diff
@@ -176,7 +177,11 @@ static void _threadpreload_cleanup(ThreadPreload* thread, int status) {
     thread->isRunning = 0;
 }
 
+double FORK_DURATION_ACC = 0.0;
+
 pid_t threadpreload_run(Thread* base, gchar** argv, gchar** envv) {
+
+
```
robgjansen, line 184: Were these newlines added accidentally?

src/main/utility/timing.h
```diff
@@ -0,0 +1,24 @@
+#ifndef TIMING_H_
+#define TIMING_H_
+#include <stdio.h>
+#include <time.h>
+
+#define RECORD_TIME(TS_NAME) \
+    struct timespec TS_NAME; \
+    clock_gettime(CLOCK_REALTIME, &TS_NAME);
```
robgjansen, line 8: I think we want `CLOCK_MONOTONIC` here instead of `CLOCK_REALTIME` because `CLOCK_REALTIME` can jump around if admin users on the machine adjust the clock between our start and stop stamps.

src/main/core/main.c
```diff
@@ -250,6 +253,8 @@ gint main_runShadow(gint argc, gchar* argv[]) {
         }
     }
 
+    atexit(print_durations);
```
robgjansen, line 256: Instead of logging to stderr atexit, can we log using Shadow's logger (i.e., `message()`) after `_main_helper()` returns (but before we shut down the logger) to make sure the information is stored more persistently?

src/main/host/thread_preload.c
```diff
@@ -203,7 +208,11 @@ pid_t threadpreload_run(Thread* base, gchar** argv, gchar** envv) {
     g_free(envStr);
     g_free(argStr);
 
+    RECORD_TIME(fork_start);
     pid_t child_pid = _threadpreload_fork_exec(thread, argv[0], argv, myenvv);
+    RECORD_TIME(fork_stop);
+
+    FORK_DURATION_ACC += duration_to_seconds(&fork_start, &fork_stop);
```
robgjansen, line 215: Is this addition to a global variable thread safe? Or are we doing all of the forking *before* creating the worker threads?

Instead of the global, I wonder if we might want to store the fork time counter in the (thread-local) `Worker` object, and use a function like `worker_addForkTime(duration_to_seconds(&fork_start, &fork_stop));` to increment it? Then, each worker would log its fork time before destroying itself.

robgjansen (changes requested):

src/main/utility/timing.h
RWails: @robgjansen I think this file and the calls to `RECORD_TIME` were left over from when I was trying to get a tight measure on `fork` performance. I think we can just remove this code.

src/main/core/main.c
```diff
@@ -14,12 +14,15 @@
...
 
+#include <sched.h>
+
 #include "main/core/logger/shadow_logger.h"
 #include "main/core/master.h"
 #include "main/core/support/configuration.h"
 #include "main/core/support/options.h"
 #include "main/host/affinity.h"
 #include "main/utility/utility.h"
+#include "main/utility/timing.h"
```
RWails, line 25: Let's remove this

src/main/host/thread_preload.c
```diff
@@ -16,6 +16,7 @@
 #include "main/host/thread_preload.h"
 #include "main/host/thread_protected.h"
 #include "main/shmem/shmem_allocator.h"
+#include "main/utility/timing.h"
```
RWails, line 19: Let's remove this

src/main/host/thread_preload.c
```diff
@@ -176,7 +177,11 @@ static void _threadpreload_cleanup(ThreadPreload* thread, int status) {
     thread->isRunning = 0;
 }
 
+double FORK_DURATION_ACC = 0.0;
```
RWails, line 180: Let's remove this

src/main/host/thread_preload.c
```diff
@@ -203,7 +208,11 @@ pid_t threadpreload_run(Thread* base, gchar** argv, gchar** envv) {
     g_free(envStr);
     g_free(argStr);
 
+    RECORD_TIME(fork_start);
```
RWails, line 211: Let's remove this

src/main/host/thread_preload.c
```diff
@@ -203,7 +208,11 @@ pid_t threadpreload_run(Thread* base, gchar** argv, gchar** envv) {
...
     pid_t child_pid = _threadpreload_fork_exec(thread, argv[0], argv, myenvv);
+    RECORD_TIME(fork_stop);
```
RWails, line 213: Let's remove this

src/main/host/thread_preload.c
RWails, line 215: Let's remove this

src/main/utility/timing.h
```diff
@@ -0,0 +1,24 @@
+#ifndef TIMING_H_
```
RWails, line 1: Let's remove this whole file

RWails (changes requested):
Changes generally look good. I think there was some cruft that made it into this P/R --- old timing code that I was using to measure fork performance. Let's yank that out, but otherwise this change looks good to me.

sporksmith:
For some reason after cleaning up and rebasing, this PR isn't picking up my force-pushes. Opening a fresh PR #1118

PR #1095, opened 2021-02-12
Test C syscall handlers in CI

stevenengler:
Added tests in the CI environment to test Shadow with the `--use-c-syscalls` build option. Since some pipe tests do not pass using the C syscall handlers, those tests were disabled.

If this PR is merged, I'll make a tracking issue for disabled tests so that we know what tests can be re-enabled once the C syscalls are removed from Shadow.


sporksmith (approved):
Nice

PR #1098, opened 2021-02-16
Simplify preloaded function list

robgjansen:
This makes the code easier to read, less prone to type errors, and easier to extend.

### DIFF
--- a/src/shim/preload_syscall.c
+++ b/src/shim/preload_syscall.c
@@ -1,0 +1,196 @@
+#include <assert.h>
+#include <dlfcn.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <poll.h>
+#include <stdarg.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <sys/epoll.h>
+#include <sys/resource.h>
+#include <sys/select.h>
+#include <sys/stat.h>
+#include <sys/statfs.h>
+#include <sys/syscall.h>
+#include <sys/utsname.h>
+#include <unistd.h>
+
+#include "main/host/syscall/kernel_types.h"
+#include "main/shmem/shmem_allocator.h"
+#include "shim/ipc.h"
+#include "shim/shim.h"
+#include "shim/shim_event.h"
+#include "shim/shim_logger.h"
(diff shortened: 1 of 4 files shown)


robgjansen:
The diff looks more complicated than it should. I basically just cut `preload_syscalls.c` in half: the first half is now in `preload_syscall.c`, and the second half is now in `preload_syscalls.c`. I only really changed the second half significantly.

stevenengler (approved):

src/shim/preload_syscalls.c
```diff
@@ -1,401 +1,130 @@
-// Defines system call wrappers: functions that are
-// documented in man section 2. (See `man man`).
+// Defines system call wrappers: functions that are documented in man section 2. (See `man man`).
+// This file defines ths symbols that will be included in the preload library,
```
sporksmith, line 2: ths -> the

src/shim/preload_syscalls.c
```diff
@@ -1,401 +1,130 @@
...
-#include <stdarg.h>
-#include <stdint.h>
-#include <stdlib.h>
-#include <sys/epoll.h>
-#include <sys/resource.h>
-#include <sys/select.h>
-#include <sys/stat.h>
-#include <sys/statfs.h>
+// Get the SYS_xxx definitions
 #include <sys/syscall.h>
```
sporksmith, line 6: Probably worth a comment here to avoid including any other headers, which could end up conflicting with the definitions below.

sporksmith (approved):
Nice! Yeah, this should make the syscall list a bit easier to maintain.

src/shim/preload_syscall.c
```diff
@@ -0,0 +1,196 @@
+#include <assert.h>
+#include <dlfcn.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <poll.h>
+#include <stdarg.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <sys/epoll.h>
```
sporksmith, line 9: We could probably drop a lot of these now, but I wouldn't spend too much time figuring out exactly which ones. Eventually we could rerun the iwyu tool again.

PR #1099, opened 2021-02-16
Remove stale gitignore entries, add new entries and comments

robgjansen:


robgjansen:
I've asked for review from everyone to see if there are other entries I should add as long as I am modifying gitignore now. Let me know.

RWails:
Thanks Rob! I have just ignored "Cargo.lock"s floating around because of laziness

Could you add `install/*` as well?

RWails (approved):

.gitignore
```diff
@@ -1,12 +1,13 @@
...
+/local
+# ide-specific settings
+.vscode
 cscope.*
 # ctags index
 /tags
 /src/tags
+# other temp files
+*.swp
+*Cargo.lock
```
stevenengler, line 13: Can you replace this line with `/src/support/logger/rust_bindings/Cargo.lock`? The other `Cargo.lock` files should be checked into the repository.

@sporksmith Any thoughts about how we should handle the lock files? My understanding is they should be checked in to the repo for executable projects, and we should run `cargo update` every once in a while to update it. Do you agree, or is there anything else we should consider?

stevenengler (approved):

.gitignore
robgjansen: FWIW, since the Cargo.toml already lists dependency versions (or version ranges), I'm fine with either approach: ignoring the Cargo.locks, or checking them in as Steven suggests.

sporksmith: Based on https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries, checking in the Cargo.lock files probably makes sense. While we have some internal "libraries", we don't expect them to be used by projects outside of this repo.

> Can you replace this line with /src/support/logger/rust_bindings/Cargo.lock?

What's the reasoning for not checking in that `Cargo.lock`?

stevenengler: My reasoning for not checking in the logging library's `Cargo.lock` is that the shadow_rs library's `Cargo.lock` already includes all of the pinned dependencies for the logging library. As far as I know the logging library's `Cargo.lock` isn't used for anything other than when building documentation, running `cargo check`, or when generating bindings.

robgjansen: Done, thanks!

PR #1100, opened 2021-02-17
Pin CPUs in phold test for efficiency

robgjansen:
This PR reduces *worst case* runtime of the phold tests.

Here is some unscientific results:

## Before this PR

Debug mode, CPU pinning off:

image

Release mode, CPU pinning off:

image

## After this PR

Debug mode, CPU pinning on:

image

Release mode, CPU pinning on:

image

## Commentary (off-channel):

- pinning off is beneficial when you are not running with multiple workers. without pinning, a plugin can “steal” an extra CPU that’s available on the machine
- when debugging is on, there is a lot more work being done on my machine and it’s probably overloaded running all 4 tests in parallel, so in that case pinning helps
- it seems like overall we will win when testing by enabling pinning, because the major decrease in debug mode runtime is worth the minor increase in release mode runtime


RWails (approved):
No objections, looks good!

PR #1102, opened 2021-02-19
Fix tcp test

robgjansen:
Fixes errno time of check bugs in the tcp test case. These will be an issue when we start preloading `gettimeofday`.


stevenengler (approved):
Looks good!

PR #1110, opened 2021-02-19
Add some determinism fixes and re-enable tests

robgjansen:
Also adds a preload for `gettimeofday` and `time`.

One of the determinism tests still fails, so it is still disabled for now.


src/shim/shim_logger.c
```diff
@@ -17,10 +17,11 @@ typedef struct _ShimLogger {
     FILE* file;
 } ShimLogger;
 
-static uint64_t _simulation_nanos;
+static uint64_t _simulation_nanos = 0;
 void shimlogger_set_simulation_nanos(uint64_t simulation_nanos) {
     _simulation_nanos = simulation_nanos;
 }
+uint64_t shimlogger_get_simulation_nanos() { return _simulation_nanos; }
```
stevenengler, line 24: Is there a reason why sometimes the `shimlogger_get_simulation_nanos()` is used, and sometimes the variable is accessed directly?

stevenengler (approved):

src/shim/shim_logger.c
robgjansen: Within the shim logger, `_simulation_nanoseconds` can be used directly because it is local to the shim logger. I made an access function so that I can retrieve it externally from `preload_syscall.c`.

PR #1111, opened 2021-02-19
Adding missing syscalls, poll handler and tests

robgjansen:


src/main/host/syscall/poll.c
```diff
@@ -0,0 +1,241 @@
...
+            epoll_control(sys->epoll, EPOLL_CTL_ADD, pfd->fd, cdesc, &epev);
+        }
+    }
+}
+
+static SysCallReturn _syscallhandler_pollHelper(SysCallHandler* sys, PluginPtr fds_ptr, nfds_t nfds,
+                                                const struct timespec* timeout) {
+    // Get the pollfd struct in our memory
+    struct pollfd* fds =
+        process_getWriteablePtr(sys->process, sys->thread, fds_ptr, nfds * sizeof(*fds));
```
stevenengler, line 125: Should this be `process_getMutablePtr`?

src/main/host/syscall/poll.c
```diff
@@ -0,0 +1,241 @@
...
+done:
+    // Clear epoll for the next poll
+    epoll_reset(sys->epoll);
+    return (SysCallReturn){.state = SYSCALL_DONE, .retval.as_i64 = num_ready};
+}
+
+static int _syscallhandler_checkPollArgs(PluginPtr fds_ptr, nfds_t nfds) {
+    if (nfds >= INT_MAX) {
+        debug("nfds was out of range [0, INT_MAX], returning EINVAL");
+        return -EINVAL;
```
stevenengler, line 173: Should this be just `nfds > INT_MAX`?

stevenengler (approved):
I left a few minor comments, but there's one main thing I'm confused about.

If the plugin calls poll() but none of the fds are ready and it blocks, when do the `revents` get set later? I see that they get set if poll() returns immediately without blocking, but if it does block and waits for the epoll object to become readable, are the `revents` members updated correctly?

robgjansen:
> If the plugin calls poll() but none of the fds are ready and it blocks, when do the revents get set later? I see that they get set if poll() returns immediately without blocking, but if it does block and waits for the epoll object to become readable, are the revents members updated correctly?

The way the syscall handlers are set up is that whenever we need to block, in this case because there are no events available on any of the fds, then we keep the plugin in the "paused" state and we go back and let Shadow process events for other plugins/hosts. Later, when either epoll notices that a new event is available or the timeout expires, then we _execute the syscall again_, and that will cause `syscallhandler_poll` to get run again, and it will set the revents info on that second time around.

src/main/host/syscall/poll.c
robgjansen: INT_MAX was a somewhat arbitrary choice of a "system limit" that shadow will impose. But I suppose `>` makes as much and maybe more sense than `>=` here.

src/main/host/syscall/poll.c
robgjansen: Oh yeah, whoops. Not sure how `process_getWriteablePtr` was working actually, because we do need to read AND write so `process_getMutablePtr` should be used. Good catch!

PR #1114, opened 2021-02-22
Define macros for consistent & concise test defs

robgjansen:
This will (i) make it easier to define new tests concisely, (ii) make all of the test cases actually fail when they should, and (iii) make the tests more consistent.


stevenengler (approved):
Looks great! This will definitely help keep the tests more consistent.

The only downside is that it might be difficult to specify a custom timeout or other property for the test without needing to duplicate the macro. Is there a way that the macro can return the test name so that extra `set_property()` calls can be made on it?

robgjansen:
Well, for the shadow tests, there will be 3 test names. The macro populates the `TEST_NAMES` variable, so you could loop over those and set properties on each one, or the ones that match the method you want. Also, in circumstances where something crazy is needed, we could just fall back to specifying the tests manually in those cases.

PR #1116, opened 2021-02-22
Convert rust tests to use test macros

robgjansen:


src/test/eventfd/CMakeLists.txt
```diff
@@ -1,18 +1,2 @@
...
-        | ${CMAKE_BINARY_DIR}/src/main/shadow --interpose-method=${Method} -l debug -d eventfd.shadow-${Method}.data -\
-        "
-    )
-    list(APPEND TestNames eventfd-shadow-${Method})
-endforeach()
-
-set_property(TEST ${TestNames} PROPERTY ENVIRONMENT "RUST_BACKTRACE=1")
-# need to check the test's return code, not just shadow's (see shadow/shadow#902)
-set_property(TEST ${TestNames} PROPERTY FAIL_REGULAR_EXPRESSION "main error code '.*' for process")
+add_linux_tests(BASENAME eventfd COMMAND sh -c "../target/debug/test_eventfd --libc-passing")
```
stevenengler, line 1: Is there a reason why some of the non-shadow tests run in a shell now and others don't (for example test/socket/getpeername)?

stevenengler (approved):

src/test/eventfd/CMakeLists.txt
robgjansen: I wanted to be consistent and run them all in a shell, but for some reason the `test/socket/*` tests failed when doing that. I'm not sure why it fails for `test/socket/*` but not for the other tests in `test/*`.

robgjansen: Nevermind, it's working now!

PR #1117, opened 2021-02-23
Convert c tests to use test macros

robgjansen:


src/test/epoll/CMakeLists.txt
```diff
@@ -1,43 +1,9 @@
...
-    COMMAND sh -c "\
-    ${yaml2xml} ${CMAKE_CURRENT_SOURCE_DIR}/epoll-writeable.test.shadow.config.yaml --output - \
-    | ${CMAKE_BINARY_DIR}/src/main/shadow --interpose-method=preload -l debug -d epoll-writeable-preload.shadow.data - \
-    "
-)
+add_linux_tests(BASENAME epoll COMMAND test-epoll)
+add_shadow_tests(BASENAME epoll METHODS ptrace-only ptrace preload)
 
+add_executable(test-epoll-writeable test_epoll_writeable.c)
+add_shadow_tests(BASENAME epoll-writeable METHODS ptrace-only ptrace preload)
```
stevenengler, line 9: Can you add a TODO note here about running the `test-epoll-writeable` test outside of Shadow? I don't see why it was not previously enabled, and I think it should be using `launch_with_unused_ports.py` and `test-launcher`.

stevenengler (approved):
Looks good, and much easier to read :)

Issue #867, opened 2020-07-01
shadow-preload tests hang in coverage configuration


sporksmith:
> **Describe the issue**

In coverage builds, the shadow-preload tests hang.

> **To Reproduce**

```
./setup build --clean --test --debug --coverage
./setup install
./setup test shutdown-shadow-preload
```

> **Operating System (please complete the following information):**

Ubuntu 18.04

> **Shadow (please complete the following information):**

https://github.com/shadow/shadow/pull/863

sporksmith:
After locally changing the `vfork` in `thread_preload.c` to `fork` to workaround a [gdb bug](https://sourceware.org/bugzilla/show_bug.cgi?id=24454):

```
BASE=`realpath .`
cd build/src/test/shutdown
gdb $BASE/build/src/main/shadow
set follow-fork-mode child
run "--interpose-method=preload" "-l" "debug" "-d" "shutdown.shadow-preload.data" "../../../../src/test/shutdown/shutdown.test.shadow.config.xml"
^C
bt
```

output:
```
#0  __strcmp_sse2_unaligned () at ../sysdeps/x86_64/multiarch/strcmp-sse2-unaligned.S:31
#1  0x00007ffff7bc0a09 in _shmemfilenode_findName (file_nodes=0x60b7c0, name=0x7fffffffe420 "/shadow_shmemfile_2410671.287020804-5960")
    at /home/jnewsome/projects/shadow/src/main/shmem/shmem_allocator.c:52
#2  0x00007ffff7bc1249 in shmemserializer_blockDeserialize (serializer=0x610e40, serial=0x7fffffffe408)
    at /home/jnewsome/projects/shadow/src/main/shmem/shmem_allocator.c:501
#3  0x00007ffff7bbde7f in shmemserializer_globalBlockDeserialize (serial=0x7fffffffe408)
    at /home/jnewsome/projects/shadow/src/main/shmem/shmem_allocator.h:249
#4  0x00007ffff7bbdfa6 in shim_shmemHandleCloneString (ev=0x7fffffffe400) at /home/jnewsome/projects/shadow/src/shim/shim_shmem.c:24
#5  0x00007ffff7bb9c0e in _shadow_syscall_event (syscall_event=0x7fffffffe5b0)
    at /home/jnewsome/projects/shadow/src/shim/preload_syscalls.c:130
[...]

PR #1119, opened 2021-02-23
Rename interpose modes to hybrid, ptrace, and preload

robgjansen:
`ptrace-only` is a bit clumsy and we want to change the modes to be wholly distinct.


src/main/core/slave.c
```diff
@@ -338,12 +338,12 @@ static gchar** _slave_generateEnvv(Slave* slave, InterposeMethod interposeMethod
...
     }
 
-    if (interposeMethod == INTERPOSE_PRELOAD_ONLY) {
-        envv = g_environ_setenv(envv, "SHADOW_INTERPOSE_METHOD", "PRELOAD_ONLY", 0);
-    } else if (interposeMethod == INTERPOSE_PTRACE_ONLY) {
-        envv = g_environ_setenv(envv, "SHADOW_INTERPOSE_METHOD", "PTRACE_ONLY", 0);
-    } else if (interposeMethod == INTERPOSE_PRELOAD_PTRACE) {
-        envv = g_environ_setenv(envv, "SHADOW_INTERPOSE_METHOD", "PRELOAD_PTRACE", 0);
+    if (interposeMethod == INTERPOSE_PRELOAD) {
+        envv = g_environ_setenv(envv, "SHADOW_INTERPOSE_METHOD", "PRELOAD", 0);
```
sporksmith, line 342: codecov thinking this line is uncovered seems concerning. Could be a bug in our coverage analysis, but could you double-check that the `preload` test cases are being generated and are actually using preload-mode (e.g. by checking the shadow log output)?

Could there be a bug in the `add_shadow_tests` macro that's not actually generating all the test cases?

sporksmith (approved):

src/main/core/slave.c
robgjansen: TL;DR You're right, the preload tests are not running in coverage mode because we have [configured them not to run](https://github.com/shadow/shadow/blob/cefa1d2cfcfefc88dd86d249dc5fcf879b63c29d/ci/container_scripts/test.sh#L7) due to #867. I found no indication of errors introduced in the PR or in the recently merged test macros.

----

There are 174 total tests that run on my local machine, including 35 preload mode tests.

```
$ cat build/Testing/Temporary/LastTest.log | grep "interpose-method" | grep arg | grep "hybrid" | wc -l
42
$ cat build/Testing/Temporary/LastTest.log | grep "interpose-method" | grep arg | grep "ptrace" | wc -l
44
$ cat build/Testing/Temporary/LastTest.log | grep "interpose-method" | grep arg | grep "preload" | wc -l
35
```

I checked several CTest files, e.g., `build/src/test/socket/listen/CTestTestfile.cmake` and it appears that all of the preload tests are generated correctly.

I also spot checked some of the Github action logs for the non-coverage tests that passed, and they all appear to be running preload tests too.

The coverage container test only ran 139 total tests, none of which are preload mode. This agrees with my above stats, where indeed the 35 preload tests are missing (174-35=139).

sporksmith: Oops, right :man_facepalming: . Sorry about that!

PR #1122, opened 2021-02-25
Increased phold timeout and set PROCESSORS to 2

stevenengler:
The `phold-shadow-hybrid` test was failing for me since the timeout was too low, and setting the `PROCESSORS` property to 2 seems to make the test time more reliable.

### DIFF
--- a/src/test/phold/CMakeLists.txt
+++ b/src/test/phold/CMakeLists.txt
@@ -73,4 +73,10 @@ set_tests_properties(
     phold-threaded-shadow-hybrid
     phold-threaded-shadow-ptrace
     phold-threaded-shadow-preload
-        PROPERTIES TIMEOUT 60)
+        PROPERTIES TIMEOUT 120)
+
+set_tests_properties(
+    phold-threaded-shadow-hybrid
+    phold-threaded-shadow-ptrace
+    phold-threaded-shadow-preload
+        PROPERTIES PROCESSORS 2)


robgjansen:
I have a stash to make some changes to phold, including tweaking the params so we send fewer messages and consume less CPU. With those changes, the phold tests complete reliably in <10 seconds in my slow VM. I think my changes are preferable to increasing the timeout, because we don't really need to be running a stress test in every CI instance and it will make all of the CI instances finish the CI scripts more quickly.

(If for some reason we still do want a version of phold to act as a stress test, I would argue that we run such a test in a single CI instance that doesn't also run the other tests.)

WIth the above in mind, would you still like to merge this in the meantime while waiting for me to turn my stash into a PR?

stevenengler:
I forgot about that. I'll close this, but I still think setting `PROCESSORS` to 2 for the threaded tests might be a good idea.

sporksmith:
> I forgot about that. I'll close this, but I still think setting `PROCESSORS` to 2 for the threaded tests might be a good idea.

Oh cool, I didn't know about the [PROCESSORS](https://cmake.org/cmake/help/latest/prop_test/PROCESSORS.html) directive. Agree that setting it to 2 (equal to the # of workers) makes sense.

robgjansen:
It does seem like a good idea to set the processors.

Note that we already run 4 tests in parallel with `ctest -j4`, even though the github action runners [have 2 CPUs](https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources) (not sure if that means we have 4 hyperthreads or not).

stevenengler:
Also since pinning is now enabled for the tests, maybe we want to disable running the phold tests in parallel (set `RUN_SERIAL`). Otherwise I think we will have multiple Shadow instances pinning to the same CPU cores.

sporksmith:
> Note that we already run 4 tests in parallel with ctest -j4

In the CI scripts we override this to `nproc`, which should return 2 in the GH CI: https://github.com/shadow/shadow/blob/dev/ci/container_scripts/test.sh#L27

> Also since pinning is now enabled for the tests, maybe we want to disable running the phold tests in parallel (set `RUN_SERIAL`)

IIUC using `PROCESSORS=2` with `-j2` would effectively force the test to be serialized in the GH CI case, without forcing it to be necessarily be serialized when running with a higher `-j` flag. If that's correct, setting `PROCESSORS` to whatever the `-w` flag is in a given test should ensure we're not over-subscribed, while using all available cores on bigger machines?

sporksmith:
Ideally we'd set `-j` to `nproc`/2 in the presence of hyper-threading. AFAICT there's not a *super* clean way to identify that case, but at least that shouldn't be an issue in the GH CI. https://unix.stackexchange.com/questions/33450/checking-if-hyperthreading-is-enabled-or-not

stevenengler:
> If that's correct, setting `PROCESSORS` to whatever the `-w` flag is in a given test should ensure we're not over-subscribed, while using all available cores on bigger machines?

Won't the CPU pinning policy be the same for each Shadow insacne, which would mean that all Shadow instances pin to the exact same cores? So for example if you had 8 cores, all tests with `-w 2` would run shadow pinned to the same cores?

sporksmith:
> Won't the CPU pinning policy be the same for each Shadow instance, which would mean that all Shadow instances pin to the exact same cores? So for example if you had 8 cores, all tests with -w 2 would run shadow pinned to the same 2 cores?

Oh, I think you're right. In the absence of creating some way of globally coordinating between independent concurrent shadow runs (ew), `RUN_SERIAL` is probably the right approach for any test using cpu pinning.

stevenengler:
CMake does have a [`PROCESSOR_AFFINITY`](https://cmake.org/cmake/help/latest/prop_test/PROCESSOR_AFFINITY.html) option, but I haven't looked into what method it uses or if we can get the mask easily from within Shadow.

sporksmith:
> Ideally we'd set `-j` to `nproc`/2 in the presence of hyper-threading. AFAICT there's not a _super_ clean way to identify that case, but at least that shouldn't be an issue in the GH CI. https://unix.stackexchange.com/questions/33450/checking-if-hyperthreading-is-enabled-or-not
I did find a relatively clean way of getting the number of physical cores. It probably still makes sense to use that instead of `nproc` to set `-j` in the CI script.

Using python seems a bit heavy-weight, but every other shell-oriented approach I could find involved grepping human-readable files, which is likely to be fragile.

https://stackoverflow.com/a/37619393

```
python -c "import psutil; print(psutil.cpu_count(logical=False))"
```

PR #1125, opened 2021-02-26
Added warning when debug logging enabled, but not built in debug mode

stevenengler:


robgjansen (approved):
Looks good!

(I think I would place the added block after the call to `shadow_logger_setEnableBuffering`, but not a big deal either way.)

PR #1128, opened 2021-02-27
Ptrace vfork

sporksmith:
Use vfork in thread_ptrace

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,6 +158,7 @@ endif(SHADOW_WERROR STREQUAL ON)
 
 if(SHADOW_COVERAGE STREQUAL ON)
     add_compile_options(--coverage)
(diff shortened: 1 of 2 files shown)


robgjansen (approved):

sporksmith:
Hmm, I was expecting the timeouts in the coverage build to go away once #1131 was merged, but they didn't.

Looking more closely, there are a lot of lines like:

```
profiling: /__w/shadow/shadow/build/src/main/CMakeFiles/shadow.dir/main.c.gcda: cannot merge previous run count: corrupt object tag (0x01000000)
```

I suspect that the profiling code ends up corrupting its in-memory data structures due to the code in between the `vfork` and the `exec`. I'll arrange to use `fork` in coverage builds.

PR #1131, opened 2021-03-01
Use test macros in phold and timerfd tests, fix bugs

robgjansen:
This PR finishes converting the remaining tests to the new macros:
- Using the new test macros, the phold test previously failed because it did not ever return 0. This fixes the test so that we can specify a phold runtime independent of the stop time for the simulation.
- The timerfd test was failing and we didn't notice it. Using the new test macros made the failing test apparent. Now we also test it using all interpose methods.


stevenengler:
Related: shadow/shadow#902 for the phold and timerfd tests.

src/main/host/descriptor/timer.c
```diff
@@ -149,7 +149,11 @@ static SimulationTime _timer_timespecToSimTime(const struct timespec* config, gb
         /* the time that was passed in represents an emulated time, so we need to adjust */
         EmulatedTime emNanoSecs = (EmulatedTime)(config->tv_sec * SIMTIME_ONE_SECOND);
         emNanoSecs += (EmulatedTime) config->tv_nsec;
-        simNanoSecs = EMULATED_TIME_TO_SIMULATED_TIME(emNanoSecs);
+        if(emNanoSecs >= EMULATED_TIME_OFFSET) {
+            simNanoSecs = EMULATED_TIME_TO_SIMULATED_TIME(emNanoSecs);
+        } else {
+            simNanoSecs = 0;
```
sporksmith, line 155: How do we get here, and why do we want to set to zero in that case?

src/test/CMakeLists.txt
```diff
@@ -131,9 +137,7 @@ add_test(
...
-## TODO swap the following lines when all tests have been converted to the macros
-#set_tests_properties(leakcheck-shadow-grep PROPERTIES DEPENDS "${ALL_SHADOW_TESTS}")
-set_tests_properties(leakcheck-shadow-grep PROPERTIES DEPENDS "bind-shadow-ptrace;bind-shadow-preload;clone-shadow-ptrace;config-convert-shadow;cpp-shadow-ptrace;cpp-shadow-preload;epoll-shadow-ptrace;epoll-writeable-shadow-ptrace;eventfd-shadow-ptrace;eventfd-shadow-preload;file-shadow-ptrace;mmap-shadow-ptrace;phold-shadow-ptrace;phold-threaded-shadow-ptrace;random-shadow-ptrace;random-shadow-preload;shutdown-shadow-ptrace;shutdown-shadow-preload;sleep-shadow-ptrace;sleep-shadow-preload;sockbuf-shadow-ptrace;socket-shadow-ptrace;socket-shadow-preload;bind-shadow-ptrace;bind-shadow-preload;listen-shadow-ptrace;listen-shadow-preload;getsockname-shadow-ptrace;getsockname-shadow-preload;accept-shadow-ptrace;accept-shadow-preload;timerfd-shadow-ptrace;udp-uniprocess-shadow-ptrace;udp-uniprocess-shadow-preload;udp-shadow-ptrace;unistd-shadow-ptrace;unistd-shadow-preload")
+set_tests_properties(leakcheck-shadow-grep PROPERTIES DEPENDS "${ALL_SHADOW_TESTS}")
```
sporksmith, line 140: Woohoo!

src/test/phold/phold-parallel.yaml
```diff
@@ -0,0 +1 @@
+phold.yaml
```
sporksmith, line 1: Huh; I guess this is an 'include'? I didn't know yaml supported that; [SO](https://stackoverflow.com/questions/528281/how-can-i-include-a-yaml-file-inside-another) seems to think it doesn't :). Since this isn't standard yaml could probably use a comment

src/test/phold/phold-serial.yaml
```diff
@@ -0,0 +1 @@
+phold.yaml
```
sporksmith, line 1: Comment how this works?

sporksmith (approved):

sporksmith:
Oh huh, just noticed this is marked as a draft. I got an email notification for the PR and misinterpreted it as a review request :man_facepalming:

src/test/phold/phold-parallel.yaml
robgjansen: It's a symbolic link checked into the repo.

robgjansen:
Thanks for the preemptive review :)

PR #1132, opened 2021-03-01
No need for shadow to notify plugin that shmem operations are done

sporksmith:
thread_preload was notifying the shim shmem operations were done before
sending back the results of each syscall. The shim doesn't do anything
in response to this other than send back an acknowledgment message.

Removing this drops execution time from 20s to 17s in the "workload"
syscall microbenchmark with 512 hosts.


RWails (approved):
Yep, in favor of removing those code. I think in some prior iterations we were using the SHMEM_COMPLETE directive to mark when transfers were completed, but that code has been removed. I don't think there's any need for it now.

PR #1133, opened 2021-03-01
Serialize tests using cpu-pinning

sporksmith:


sporksmith:
Redundant with #1131, but merging this will get tests passing so we can merge other PRs in the meantime

robgjansen:
Am just seeing this now. #1131 is ready pending the CI tests, which are now running. If they fail, I'll approve this. If they pass, I'll merge #1131 and then I guess we close this without merging?

robgjansen:
#1131 is merged now. Do we still want to make additional tweaks?

For phold, we still use a 20 second timeout when logging at info level. In this configuration, the coverage tests finished in less than 10 seconds for me.

PR #1142, opened 2021-03-03
Moved leak check tests to the 'add_shadow_tests' macro

stevenengler:
There are two issues with the leak check tests:

1. If one of the tests causes Shadow to leak memory, it's difficult to tell which test caused it.
2. It parses the `LastTest.log` file to look for all `ObjectCounter: counter diffs:` lines, but if the last test run didn't run all of the tests (for example using `./setup test some-specific-test`, then it's not obvious that running `./setup test leakcheck` will only check that one test and not all of the tests.

I moved the grep to inside the `add_shadow_tests` macro so that we can easily see if a specific test is causing Shadow to leak memory (the test will fail immediately). It is also much simpler than the current leak check tests, but I don't know if there are any advantages to the current approach.


robgjansen (approved):
`+3, -40 lines`
amazing!!!

Issue #1134, opened 2021-03-01
waitpid is O(# children), resulting in O((# sim processes)^2) in thread_ptrace


sporksmith:
Broken out from #1041 

Using a [syscall-heavy-microbenchmark](https://github.com/sporksmith/phantom-paper/blob/realtime-experiment/experiments/realtime/workload.c), I did some perf-based profiling of ptrace mode.

Without cpu-pinning, we spend 20% of our time in [wait_consider_task inside the kernel](https://github.com/torvalds/linux/blob/219d54332a09e8d8741c1e1982f5eae56099de85/kernel/exit.c#L1279). It turns out that `waitpid` ends up using the `wait4` syscall, which ends up in [do_wait](https://github.com/torvalds/linux/blob/219d54332a09e8d8741c1e1982f5eae56099de85/kernel/exit.c#L1441), which checks *every* child and *every* ptraced process; not just the one whose pid was supplied.

From `perf report` (there are 4 symmetric worker threads; the total counts are ~4x these):

```
-    7.21%     6.55%         24363  worker-3         [kernel.vmlinux]         [k] wait_cons▒
   - 6.55% __GI___clone (inlined)                                                          ▒
        start_thread                                                                       ▒
        worker_run                                                                         ▒
      - event_execute                                                                      ▒
         - 6.54% process_continue                                                          ▒
              thread_resume                                                                ▒
              _thread_cleanupSysCallCondition (inlined)                                    ▒
            - threadptrace_resume                                                          ▒
               - 6.41% _threadptrace_doAttach (inlined)                                    ▒
                    _threadptrace_nextChildState                                           ▒
                    __waitpid                                                              ▒
                    entry_SYSCALL_64                                                       ▒
                    do_syscall_64                                                          ▒
                    __x64_sys_wait4                                                        ▒
                    __do_sys_wait4                                                         ▒
                  - kernel_wait4                                                           ▒
                     - 6.06% do_wait                                                       ▒
                          6.03% wait_consider_task 
```

[...]

PR #1143, opened 2021-03-04
Fork from a proxy thread

sporksmith:
In ptrace_thread, use a proxy thread to be the parent of plugin processes, and always ptrace-detach inactive plugins. This avoid quadratic behavior due to waitpid being O(n).

Mitigates #1134

### DIFF
--- a/src/main/host/thread_ptrace.c
+++ b/src/main/host/thread_ptrace.c
@@ -20,6 +20,7 @@
 #include "main/host/shimipc.h"
 #include "main/host/thread_protected.h"
 #include "main/host/tsc.h"
+#include "main/utility/fork_proxy.h"
 #include "shim/ipc.h"
 #include "support/logger/logger.h"
 
@@ -34,6 +35,22 @@
 // Using PTRACE_O_TRACECLONE causes the `clone` syscall to fail on Ubuntu 18.04.
 // We instead add the CLONE_PTRACE flag to the clone syscall itself.
 #define THREADPTRACE_PTRACE_OPTIONS (PTRACE_O_EXITKILL | PTRACE_O_TRACESYSGOOD | PTRACE_O_TRACEEXEC)
+
+// `waitpid` is O(n) in the # of child threads and tracees
+// <https://github.com/shadow/shadow/issues/1134>. We work around it by
+// spawning processes on a ForkProxy thread, keeping them off the worker
+// thread's child list, and by detaching inactive plugins to keep them off the
+// worker thread's tracee list.
+//
+// Each worker thread gets its own proxy thread so that forking simulated
+// processes can be parallelized.
+static bool _useONWaitpidWorkarounds = true;
+OPTION_EXPERIMENTAL_ENTRY("disable-o-n-waitpid-workarounds", 0, G_OPTION_FLAG_REVERSE,
+                          G_OPTION_ARG_NONE, &_useONWaitpidWorkarounds,
+                          "Disable performance workarounds for waitpid being O(n). Beneficial to "
+                          "disable if waitpid is patched to be O(1) or in some cases where it'd "
+                          "otherwise result in excessive detaching and reattaching",
+                          NULL)
 
 static char SYSCALL_INSTRUCTION[] = {0x0f, 0x05};
 
@@ -238,6 +255,7 @@ static void _threadptrace_memcpyToPlugin(ThreadPtrace* thread,
 const void* threadptrace_getReadablePtr(Thread* base, PluginPtr plugin_src,
                                         size_t n);
 static void _threadptrace_ensureStopped(ThreadPtrace* thread);
+static void _threadptrace_doAttach(ThreadPtrace* thread);
 
 static ThreadPtrace* _threadToThreadPtrace(Thread* thread) {
     utility_assert(thread->type_id == THREADPTRACE_TYPE_ID);
@@ -297,10 +315,10 @@ static const char* _syscall_regs_to_str(const struct user_regs_struct* regs) {
     return buf;
 }
 
-static pid_t _threadptrace_fork_exec(const char* file, char* const argv[],
-                                     char* const envp[]) {
+static pid_t _threadptrace_fork_exec(const char* file, char* const argv[], char* const envp[]) {
     pid_t shadow_pid = getpid();
 
+    // fork requested process.
 #ifdef SHADOW_COVERAGE
     // The instrumentation in coverage mode causes corruption in between vfork
     // and exec. Use fork instead.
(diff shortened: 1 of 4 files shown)


sporksmith:
GH's CI is unhappy right now, but I've locally verified that at least one of the configurations passes in the `ci/run.sh` script

sporksmith:
While getting the kernel patch working I learned a couple things that I want to try before merging. Mainly: it turns out that while you can't call `ptrace` on tasks that are traced by another thread (even if that thread is in the same process as yourself), you *can* call waitpid on tasks whose parent or tracer is a different thread in the same process as yourself (if you don't set __WNOTHREAD).

This means:

* We should be able to avoid zombie tasks after getting the plugin exit ptrace stop. We might be already - I need to double-check whether the child might already be reaped after the ptrace-stop is processed in the case where the real parent task is in the same process as the current task. If it's not, we should be able to just call `waitpid` again to reap it.

* Since we can wait on traced tasks that are ptrace-attached to another process, the idea of having dedicated ptracer thread(s) should be simpler and more performant than I'd thought before. It'd only need to proxy `ptrace` calls (which don't block), not `waitpid` calls (which do). It'd still mean a thread-hop on `ptrace` calls, but that might not be too bad. This could be added to the functionality of the forkproxy thread. In that case we wouldn't need to do any ptrace reattachments.

* Small tweak - we can use __WNOTHREAD in the waitpid inside the fork proxy to avoid traversing child lists of *other* fork proxy threads. This *probably* won't matter in practice since the implementation of waitpid checks the current task's children before checking the children of other tasks in the thread group, but could end up helping if that implementation changes later.

sporksmith:
>     * We should be able to avoid zombie tasks after getting the plugin exit ptrace stop. We might be already - I need to double-check whether the child might already be reaped after the ptrace-stop is processed in the case where the real parent task is in the same process as the current task. If it's not, we should be able to just call `waitpid` again to reap it.

Turns out it's even better - there were no zombies in the first place. When the tracer is in the same process as the real parent, the kernel reaps the child after handling the exit ptrace-event. i.e. no extra `waitpid` needed.

>     * Since we can wait on traced tasks that are ptrace-attached to another process, the idea of having dedicated ptracer thread(s) should be simpler and more performant than I'd thought before. It'd only need to proxy `ptrace` calls (which don't block), not `waitpid` calls (which do). It'd still mean a thread-hop on `ptrace` calls, but that might not be too bad. This could be added to the functionality of the forkproxy thread. In that case we wouldn't need to do any ptrace reattachments.

Thinking through this some more - we won't want to do this until if/when the O(1) waitpid kernel patch lands. Otherwise we'll get O(n) waitpid calls again since the tracer threads would have O(n) tracees attached.

>     * Small tweak - we can use __WNOTHREAD in the waitpid inside the fork proxy to avoid traversing child lists of _other_ fork proxy threads. This _probably_ won't matter in practice since the implementation of waitpid checks the current task's children before checking the children of other tasks in the thread group, but could end up helping if that implementation changes later.

Since the current task's children are checked first anyway, it probably doesn't make sense to use this relatively esoteric feature.

src/main/host/thread_ptrace.c
```diff
@@ -297,69 +299,41 @@ static const char* _syscall_regs_to_str(const struct user_regs_struct* regs) {
...
-        default: {
-            // parent
-            info("started process %s with PID %d", file, pid);
-            return pid;
-        }
+// To be called in plugin child processes before `exec`
+void _childPreExecCb() {
+    // Disable RDTSC
+    if (prctl(PR_SET_TSC, PR_TSC_SIGSEGV, 0, 0, 0) < 0) {
+        error("prctl: %s", g_strerror(errno));
```
stevenengler, line 332: Does anything bad happen by accessing the logger from within the child of vfork? And if the child process aborts due to `error()`, does the parent process also abort?

stevenengler (approved):
What advantage does this approach have compared to forking and detaching in the worker thread?

src/main/utility/fork_proxy.c
```diff
@@ -0,0 +1,125 @@
...
+                         char* const envp[]) {
+    forkproxy->file = file;
+    forkproxy->argv = argv;
+    forkproxy->envp = envp;
+    if (sem_post(&forkproxy->sem_begin) != 0) {
+        error("sem_post: %s", g_strerror(errno));
+    }
+    if (sem_wait(&forkproxy->sem_done) != 0) {
+        error("sem_wait: %s", g_strerror(errno));
+    }
```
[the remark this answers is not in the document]
sporksmith: Done

src/main/utility/fork_proxy.h
```diff
@@ -0,0 +1,28 @@
...
+
+// Fork and exec the specified program. Uses `fork` when SHADOW_COVERAGE is
+// defined, and `vfork` otherwise.
+pid_t forkproxy_forkExec(ForkProxy* forkproxy, const char* file, char* const argv[],
+                         char* const envp[]);
+
+// Destroy the forkproxy, including stopping its thread and freeing its memory.
+// Doesn't kill processes that were started by the ForkProxy, but will result
+// in them being re-parented if they're still alive.
+void forkproxy_delete(ForkProxy* forkproxy);
```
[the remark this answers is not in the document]
sporksmith: Right. Deleted the declaration.

sporksmith (commented):
PTAL - I ended up adding a flag to disable these workarounds, which also resulted in me refactoring back into a monolithic fork-exec function. I think the end result is better anyway than having split it across the ForkProxy and parent and child callbacks.

Sorry, I also rebased onto the latest dev without thinking about it :/

sporksmith:
> What advantage does this approach have compared to forking and detaching in the worker thread?

Hopefully clearer now in comments, but we want to keep the plugin processes out of the worker thread's child lists, so that they don't make the worker thread's calls to waitpid more expensive.

src/main/host/thread_ptrace.c
```diff
@@ -312,55 +330,61 @@ static pid_t _threadptrace_fork_exec(const char* file, char* const argv[],
     switch (pid) {
         case -1: {
             error("fork: %s", g_strerror(errno));
-            return -1;
+            break;
         }
```
stevenengler, line 334: (Optional) I'm guessing this `break` is here to avoid compiler warnings, but if we need to add an extra line anyways, I think this is clearer as `abort()` or `exit()` just to show that we don't want it to continue past this point. (And the same after the `execvpe()`.)

stevenengler (approved):

src/main/host/thread_ptrace.c
sporksmith: Agreed; done

PR #1145, opened 2021-03-04
Improve rust syscalls/descriptors

stevenengler:
Rearranged the Rust syscall helper code, added Rust wrappers for the `pread64` and `pwrite64` syscalls, added the `Debug` trait to some Rust types, and set the `CLOSED` (and remove the `ACTIVE`) flag when the pipe is closed.

### DIFF
--- a/src/main/host/descriptor/mod.rs
+++ b/src/main/host/descriptor/mod.rs
@@ -114,7 +114,7 @@ impl From<FileStatus> for c::Status {
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub enum NewStatusListenerFilter {
     Never,
     OffToOn,
@@ -303,6 +303,20 @@ impl PosixFile {
     }
 }
 
+impl std::fmt::Debug for PosixFile {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Pipe(_) => write!(f, "Pipe")?,
+        }
+        write!(
+            f,
+            "(status: {:?}, flags: {:?})",
+            self.status(),
+            self.get_flags()
+        )
+    }
+}
+
 bitflags::bitflags! {
     // Linux only supports a single descriptor flag:
     // https://www.gnu.org/software/libc/manual/html_node/Descriptor-Flags.html
@@ -311,7 +325,7 @@ bitflags::bitflags! {
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct Descriptor {
     file: Arc<AtomicRefCell<PosixFile>>,
     flags: DescriptorFlags,
--- a/src/main/host/descriptor/mod.rs
+++ b/src/main/host/descriptor/mod.rs
@@ -260,6 +260,12 @@ impl IsSend for PosixFile {}
 impl IsSync for PosixFile {}
 
 impl PosixFile {
+    pub fn close(&mut self, event_queue: &mut EventQueue) -> SyscallReturn {
+        match self {
+            Self::Pipe(f) => f.close(event_queue),
+        }
+    }
+
     pub fn read(&mut self, bytes: &mut [u8], event_queue: &mut EventQueue) -> SyscallReturn {
         match self {
             Self::Pipe(f) => f.read(bytes, event_queue),
@@ -329,6 +335,7 @@ bitflags::bitflags! {
 pub struct Descriptor {
     file: Arc<AtomicRefCell<PosixFile>>,
     flags: DescriptorFlags,
+    count: Arc<()>,
 }
 
 impl Descriptor {
@@ -336,6 +343,7 @@ impl Descriptor {
         Self {
             file,
             flags: DescriptorFlags::empty(),
+            count: Arc::new(()),
         }
     }
 
@@ -349,6 +357,10 @@ impl Descriptor {
 
     pub fn set_flags(&mut self, flags: DescriptorFlags) {
         self.flags = flags;
+    }
+
+    pub fn get_fd_count(&self) -> usize {
+        Arc::<()>::strong_count(&self.count)
     }
 }
 
--- a/src/main/host/descriptor/mod.rs
+++ b/src/main/host/descriptor/mod.rs
@@ -343,8 +343,13 @@ bitflags::bitflags! {
 
 #[derive(Clone, Debug)]
 pub struct Descriptor {
+    /// The PosixFile that this descriptor points to.
     file: Arc<AtomicRefCell<PosixFile>>,
+    /// Descriptor flags.
     flags: DescriptorFlags,
+    /// A count of how many open descriptors there are with reference to this file. Since a
+    /// reference to the file can be held by other objects like an epoll file, it should
+    /// be true that `Arc::strong_count(&self.count)` <= `Arc::strong_count(&self.file)`.
     count: Arc<()>,
 }
 
--- a/src/main/host/descriptor/mod.rs
+++ b/src/main/host/descriptor/mod.rs
@@ -350,7 +350,7 @@ pub struct Descriptor {
     /// A count of how many open descriptors there are with reference to this file. Since a
     /// reference to the file can be held by other objects like an epoll file, it should
     /// be true that `Arc::strong_count(&self.count)` <= `Arc::strong_count(&self.file)`.
-    count: Arc<()>,
+    open_count: Arc<()>,
 }
 
 impl Descriptor {
@@ -358,7 +358,7 @@ impl Descriptor {
         Self {
             file,
             flags: DescriptorFlags::empty(),
-            count: Arc::new(()),
+            open_count: Arc::new(()),
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/main/host/descriptor/mod.rs
```diff
@@ -303,6 +309,20 @@ impl PosixFile {
     }
 }
 
+impl std::fmt::Debug for PosixFile {
```
sporksmith, line 312: Is this substantially different from what derive(Debug) would get us? I *think* it already stringifies enum values

src/main/host/descriptor/mod.rs
```diff
@@ -311,17 +331,19 @@ bitflags::bitflags! {
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct Descriptor {
     file: Arc<AtomicRefCell<PosixFile>>,
     flags: DescriptorFlags,
+    count: Arc<()>,
```
sporksmith, line 338: This seems unusual, and could use a comment.

I suppose we want to be able to get to the reference count given a reference to the `Descriptor` itself and not the `Arc` wrapping it?

src/main/host/descriptor/pipe.rs
```diff
@@ -43,6 +43,16 @@ impl PipeFile {
         self.flags = flags;
     }
 
+    pub fn close(&mut self, event_queue: &mut EventQueue) -> SyscallReturn {
+        // set the closed flag and remove the active flag
+        self.copy_status(
+            FileStatus::CLOSED | FileStatus::ACTIVE,
```
sporksmith, line 49: Why do we both remove and add the CLOSED flag?

src/main/host/syscall/unistd.rs
```diff
@@ -27,32 +29,42 @@ pub fn close(sys: &mut c::SysCallHandler, args: &c::SysCallArgs) -> c::SysCallRe
...
                     sys as *mut c::SysCallHandler,
                     args as *const c::SysCallArgs,
-                )
-            };
+                );
+            },
+        };
+
+        // if this is the last remaining descriptor
+        if desc.get_fd_count() == 1 {
```
sporksmith, line 43: Could this be implemented by implementing Drop, instead of keeping an internal explicit count?

src/main/host/syscall/unistd.rs
```diff
@@ -84,19 +106,34 @@ pub fn read(sys: &mut c::SysCallHandler, args: &c::SysCallArgs) -> c::SysCallRet
...
+        CompatDescriptor::Legacy(_) => unsafe {
+            return c::syscallhandler_read(
+                sys as *mut c::SysCallHandler,
+                args as *const c::SysCallArgs,
+            );
+        },
+    };
 
-    read_helper(sys, args, fd, buf_ptr, buf_size, offset)
+    read_helper(sys, fd, desc, buf_ptr, buf_size, offset)
```
sporksmith, line 128: control flow might be a bit easier to follow as:

```
return match desc {
  CompatDescriptor::New(d) => read_helper(...),
  CompatDescriptor::Legacy(_) => unsafe { c::syscallhandler_read(...) }
}
```

src/main/host/syscall/unistd.rs
```diff
@@ -107,18 +144,34 @@ fn read_helper(
...
+    buf_ptr: c::PluginPtr,
+    buf_size: libc::size_t,
+    offset: libc::off_t,
+) -> c::SysCallReturn {
     // need a non-null buffer
     if buf_ptr.val == 0 {
         return SyscallReturn::Error(nix::errno::Errno::EFAULT).into();
     }
 
+    // we can only seek on files, otherwise its a pipe error
```
sporksmith, line 170: Got a bit confused here until I realized that (at least currently) this function is only called for pipe descriptors. If that'll always be the case, then maybe `desc` should specifically be a pipe descriptor type? If it's not, then maybe some code validating that it's a pipe descriptor (for now)?

src/main/host/syscall/unistd.rs
```diff
@@ -183,18 +251,34 @@ fn write_helper(
...
```
sporksmith, line 277: Same comment about clarifying/checking that this really is a pipe descriptor

sporksmith (changes requested):

src/main/host/descriptor/mod.rs
stevenengler: The `derive(Debug)` impl for an enum will also print whatever objects it contains. This means that all file types (ex: `PipeFile`) then also need to impl `Debug`, and if you `println!("{:?}", desc)`, it will print out a huge amount of information. I'm assuming that if someone does a `println!("{:?}", desc)`, then they're probably only concerned about the high-level details like the descriptor/file type, status, and flags, and not all the details about its contents (for example the pipe's internal buffer, event source, event handle, etc).

src/main/host/descriptor/mod.rs
stevenengler: It's essentially a reference count for the descriptor object and not the file object. For example, if the plugin calls `fd1 = socket(AF_INET, ...); fd2 = dup(fd1);`, there will be two `Descriptor` objects pointing to the same file object. If the plugin then adds the socket `PosixFile` to an epoll instance, then there will still only be two `Descriptor` objects, but three references to the `PosixFile`. If the plugin then calls `close(fd1); close(fd2);`, both (all) `Descriptor` objects will be removed, but there will still be a reference to the file object in the epoll instance.

We want to run `close()` on the file object when all descriptors have been closed, even if there are still references to the `PosixFile` object.

If this make sense, I'll add a comment to explain what that `Arc` is being used for.

src/main/host/descriptor/pipe.rs
stevenengler: I'm not sure what you mean here. This sets the mask for the `copy_status` operation as `FileStatus::CLOSED | FileStatus::ACTIVE`, and sets the value under that mask to `FileStatus::CLOSED`. So this adds the `CLOSED` flag, and removes the `ACTIVE` flag.`

src/main/host/syscall/unistd.rs
stevenengler: I had started doing it this way, but the problem is that you can't pass arguments to `Drop::drop()`. My two concerns are:

1. Since calling `close()` on a descriptor/file object may need to emit events (for example if it sets its status flag to `CLOSED`), it needs access to an `EventLoop` object to add the event to. We could make a new `EventLoop` object within the descriptor's `Drop::drop()` implementation and process any events there immediately, but this drop could happen while we're already holding references to other objects, in which case we could end up trying to borrow those same objects within the `Drop::drop()` and end up with a runtime error.
2. Right now we only pass an `EventLoop` struct to `Descriptor::close()`, but if we need to pass the Host object or Process object in the future, then that will be a problem.

I'd be okay with using the `Drop::drop()` for now if you think it's better, as the above are only hypothetical issues at this point.

src/main/host/descriptor/mod.rs
sporksmith: Ah, makes sense

src/main/host/descriptor/pipe.rs
sporksmith: Ah ok. `copy_status` is a little difficult to follow, especially at the call sites, but I'm not sure how to improve it.

stevenengler: Yeah I agree, and I'm not sure how to improve it either. I could set the flags separately, but currently it would require two calls to `handle_status_change()`, which would then emit two events.

src/main/host/syscall/unistd.rs
stevenengler: Good idea, I made this change.

src/main/host/syscall/unistd.rs
stevenengler: I modified this a bit to pass the offset on to the file object (for example the `PipeFile::read()`). Then since pipes don't support seeking, it returns the ESPIPE error from there). Hopefully it makes more sense now.

src/main/host/descriptor/mod.rs
sporksmith: Could you add a comment clarifying exactly what this count is? Sounds like it's the number of *open* Descriptors with a reference to the file?

src/main/host/syscall/unistd.rs
sporksmith: >it needs access to an EventLoop object to add the event to. We could make a new EventLoop object within the descriptor's Drop::drop() implementation and process any events there immediately,

Seems fine, particularly since that's what we're doing here as well...?

> this drop could happen while we're already holding references to other objects, in which case we could end up trying to borrow those same objects within the Drop::drop() and end up with a runtime error.

I think this possibility isn't too bad. Unlike, say, Java, in Rust objects are dropped at a deterministic time (roughly when they go out of scope). So I *think* it'd end up being close to equivalent to explicitly doing the cleanup here, or at least if it failed it'd fail predictably and could reasonably be debugged.

> Right now we only pass an EventLoop struct to Descriptor::close(), but if we need to pass the Host object or Process object in the future, then that will be a problem.

`drop` also gets a reference to `self`. Should the `Host` and `Process` be reachable from there?

All that said, from our conversation on slack, if we're keeping an explicit count here instead of creating a separate object, it seems like a Drop won't work.

I'm ok with this ending up being a special case where it's easier to keep an explicit count and do explicit cleanup; I just don't want it to become the default :)

sporksmith (approved):

src/main/host/descriptor/mod.rs
stevenengler: I've added some comments, and also renamed this to `open_count`.

PR #1148, opened 2021-03-05
Ensure exit on LOGLEVEL_ERROR

sporksmith:


stevenengler (approved):

stevenengler:
In debug mode the Shadow logger will override this behaviour, but will still abort in `utility_assert()` -> `utility_handleError()`.

```c
if (level == LOGLEVEL_ERROR) {
    ...

    /* now abort, but get a backtrace */
    utility_assert(FALSE && "failure due to error-level log message");
}
```

sporksmith:
> (Just a note) In debug mode the Shadow logger will override this behaviour, but will still abort in `utility_assert()` -> `utility_handleError()`.
> 
> ```c
> if (level == LOGLEVEL_ERROR) {
>     ...
> 
>     /* now abort, but get a backtrace */
>     utility_assert(FALSE && "failure due to error-level log message");
> }
> ```

Yeah, I was initially planning on changing it, but decided to leave it in place since the `utility_assert` also prints a stack backtrace. Maybe that behavior would be worth adding to the base logger as well, but it'd take a little work since currently `utility_assert` is part of Shadow itself.

PR #1151, opened 2021-03-05
Disabled coverage tests in CI

stevenengler:
Coverage tests are currently [failing in the CI](https://github.com/shadow/shadow/issues/1152), so this PR disables them until we figure out how to fix them.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -29,10 +29,12 @@ jobs:
         cc: ['gcc', 'clang']
         buildtype: ['debug', 'release']
         include:
+          # TODO: re-enable coverage tests once we figure out why they're failing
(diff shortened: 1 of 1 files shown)


.github/workflows/run_tests.yml
```diff
@@ -29,10 +29,11 @@ jobs:
         cc: ['gcc', 'clang']
         buildtype: ['debug', 'release']
         include:
+          # TODO: re-enable coverage tests once we figure out why they're failing
```
sporksmith, line 32: Could you add a link here to #1152 for reference?

sporksmith (approved):

.github/workflows/run_tests.yml
stevenengler: Added

PR #1154, opened 2021-03-07
Do not use subroutine execution timers by default

### LINKED ISSUE #1141: Significant time spent in g_get_monotonic_time
`g_get_monotonic_time` is relatively slow, and we call it relatively often.

Using the "workload" benchmark [here](https://github.com/sporksmith/phantom-paper/tree/d8c485b58c78c2e6b06c98abef06bedffb714fbb/experiments/realtime), with `#define WORK_ITR 1` in workload.c, at shadow commit 0e13d3f4e3c974ed9a939032efc06f17c7066958,

```
sudo chrt -f 1 perf record -o ptrace.pin.256.perf --call-graph=dwarf /home/jnewsome/.shadow/bin/shadow --interpose-method=ptrace -w 4 -t host --pin-cpus config.xml > ptrace
sudo perf report --call-graph=graph,0.5,callee -n -i ptrace.pin.256.perf
```

We spend 16% of *each worker thread* in `do_syscall_64`, mostly by way of `__vdso_clock_gettime`, which is called from `g_get_monotonic_time`:

```
-   16.33%     2.18%          9218  worker-2         [kernel.vmlinux]                      ▒
   - 16.32% do_syscall_64                                                                  ▒
      - 16.30% entry_SYSCALL_64                                                            ▒
         - 13.18% __vdso_clock_gettime                                                     ▒
              __GI___clock_gettime (inlined)                                               ▒
            - g_get_monotonic_time                                                         ▒
               - 4.59% g_timer_stop                                                        ▒
                  + 1.60% _schedulerpolicyhostsingle_pop                                   ▒
                  + 1.54% _schedulerpolicyhostsingle_push                                  ▒
                  + 1.45% event_execute                                                    ▒
               - 4.22% g_timer_continue                                                    ▒
                  + 1.49% event_execute                                                    ▒
                  + 1.47% _schedulerpolicyhostsingle_pop                                   ▒
                  + 1.25% _schedulerpolicyhostsingle_push                                  ▒
               - 2.18% g_timer_start                                                       ▒
                  + 1.42% _syscallhandler_pre_syscall (inlined)                            ▒
                  + 0.76% process_continue                                                 ▒
               - 2.17% g_timer_elapsed                                                     ▒
                  + 1.44% _syscallhandler_post_syscall (inlined)                           ▒
                  + 0.73% process_continue                                                 ▒
         + 1.55% ptrace                                                                    ▒
         + 0.77% __GI___sched_yield (inlined)                                              ▒
         + 0.58% __waitpid
```

The timers in `_syscallhandler_pre_syscall` and `_syscallhandler_post_syscall` could probably be disabled in release builds.

If we need the timers in the scheduler, perhaps we could ensure `clock_gettime` gets called with `CLOCK_MONOTONIC_COARSE`; that *should* let `__vdso_clock_gettime` avoid making a syscall, making it much cheaper.

robgjansen:
Shadow includes internal performance timers to help us track execution
time of various subroutines. The timer-related functions have shown
up on performance profiles as causing significant overhead. These timers
are now disabled by default.

We add a new --use-perf-timers option to the setup script, and a new
USE_PERF_TIMERS option to cmake that allows us to toggle the performance
timers back on if desired.

Closes #1141

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,6 +117,7 @@ option(SHADOW_EXPORT "export service libraries and headers (default: OFF)" OFF)
 option(SHADOW_WERROR "turn compiler warnings into errors. (default: OFF)" OFF)
 option(SHADOW_COVERAGE "enable code-coverage instrumentation. (default: OFF)" OFF)
 option(SHADOW_USE_C_SYSCALLS "use only the C syscall handlers. (default: OFF)" OFF)
+option(SHADOW_USE_PERF_TIMERS "compile in timers for tracking the run time of various internal operations. (default: OFF)" OFF)
 
 ## display selected user options
 MESSAGE(STATUS)
@@ -128,6 +129,7 @@ MESSAGE(STATUS "SHADOW_EXPORT=${SHADOW_EXPORT}")
 MESSAGE(STATUS "SHADOW_WERROR=${SHADOW_WERROR}")
 MESSAGE(STATUS "SHADOW_COVERAGE=${SHADOW_COVERAGE}")
 MESSAGE(STATUS "SHADOW_USE_C_SYSCALLS=${SHADOW_USE_C_SYSCALLS}")
+MESSAGE(STATUS "SHADOW_USE_PERF_TIMERS=${SHADOW_USE_PERF_TIMERS}")
 MESSAGE(STATUS "-------------------------------------------------------------------------------")
 MESSAGE(STATUS)
 
@@ -170,7 +172,13 @@ else()
 endif(SHADOW_COVERAGE STREQUAL ON)
 
 if(SHADOW_USE_C_SYSCALLS STREQUAL ON)
+    message(STATUS "C syscalls enabled")
     add_definitions(-DUSE_C_SYSCALLS)
+endif()
+
+if(SHADOW_USE_PERF_TIMERS STREQUAL ON)
+    message(STATUS "Perf timers enabled")
+    add_definitions(-DUSE_PERF_TIMERS)
 endif()
 
 if($ENV{VERBOSE})
--- a/setup
+++ b/setup
@@ -72,6 +72,11 @@ def main():
         action="store_true", dest="do_use_c_syscalls",
         default=False)
 
+    parser_build.add_argument('--use-perf-timers',
+        help="compile in timers for tracking the run time of various internal operations",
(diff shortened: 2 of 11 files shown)


robgjansen:
One of the timer objects is causing the rust bindings to change. Do we need to do something special to deal with this so things work correctly when using the new `--use-perf-timers` setup script option?

sporksmith (commented):
> One of the timer objects is causing the rust bindings to change. Do we need to do something special to deal with this so things work correctly when using the new `--use-perf-timers` setup script option?

Yeah, I think we'll run into problems with this as-is. [This test assertion](https://github.com/shadow/shadow/pull/1154/files#diff-43915958aedf246ec3bbc6abb14f599b0c4f4cd9fda493be9fdfdc3ff8c30808L693) is going to fail if the flag has a different value when building the bindings as when building everything else. Worse, if the rust code later ends up accessing any of the members that come after the timers and there's a mismatch, it'll access the wrong offset into the struct.

The easiest short term solution is probably just to leave the timers as members of the struct so that the size of the struct doesn't depend on this flag.

The more robust solution is to build the bindings on the fly as part of the build instead of checking them in; then we can always build the bindings with the same compiler flags as everything else. That'd also save us some other minor nuisances (having to remember to rebuild and check in bindings when appropriate, ensuring that our local build of those bindings matches exactly what's generated from the CI environment), but would add the relatively heavy-weight bindgen (and/or cbindgen? I always get them confused) crates as dependencies to the build. I'm mildly in favor of going this route in the longer term, but I don't think we need to block this PR on it.

@stevenengler wdyt?

stevenengler:
> @stevenengler wdyt?

Yeah I'm not sure what's best here. The way we do Rust-C bindings now won't work with things that are `#ifdef`'d out. Leaving the member in the structure is the easiest way to get this working.

I had a reason for why generating the bindings at build-time might cause problems, but I can't remember it anymore. We used to not be able to bootstrap the bindings because of the circular dependencies between the C code and Rust code (we needed to have the older bindings in order to generate new bindings, and we couldn't generate the new bindings from scratch), but I think I fixed that in a previous PR.

I'm okay with trying it (generating the bindings at build time) and seeing if we run into issues.

robgjansen:
(Once approved, I plan to squash before merging.)

sporksmith (approved):

PR #1159, opened 2021-03-08
Use number of system CPU cores for number of test jobs

stevenengler:
Still use a minimum of 4 jobs, but use more if we have more CPU cores.

This could potentially make race conditions in the tests more common since some non-Shadow tests still bind to the same ports, but those are bugs in the tests that we can fix.

### DIFF
--- a/setup
+++ b/setup
@@ -126,7 +126,8 @@ def main():
         help="number of tests to run simultaneously during testing",
         metavar="N", type=int,
         action="store", dest="njobs",
-        default=4)#multiprocessing.cpu_count())
+        # use a minimum of 4 jobs by default
+        default=max(4, multiprocessing.cpu_count()))
 
     parser_test.add_argument('-r', '--rerun-failed',
         help="Run only the tests that failed previously",


setup
```diff
@@ -126,7 +126,8 @@ def main():
         help="number of tests to run simultaneously during testing",
         metavar="N", type=int,
         action="store", dest="njobs",
-        default=4)#multiprocessing.cpu_count())
+        # use a minimum of 4 jobs by default
+        default=max(4, multiprocessing.cpu_count()))
```
robgjansen, line 130: This means if we run on a machine with 1 CPU core, we still run 4 jobs. Is that really what we want? Or should we just always set the default to `multiprocessing.cpu_count()`?

There was a reason we did not use `multiprocessing.cpu_count()`, but I think it had something to do with the elf-loader build. As long as each test case is set to consume the correct number of cores, I think `multiprocessing.cpu_count()` should be OK.

stevenengler: From Justin @ https://github.com/shadow/shadow/commit/f040e2b2cad98617093f76eb02601006c7b1a349

> Even if the CPU doesn't have 4 threads, it's still faster because of the blocking from I/O, or the sleep call, etc. This could probably be set even higher, but it's already down to ~6 seconds anyway.

So I think it's good to still stick with 4 unless there's a reason to make it lower.

jtracey: It was because the slowest tests were a few constrained by IO, so empirically, running 4 threads (even on my dual core laptop) was faster, since it allowed the more numerous but faster CPU constrained tests to run at the same time as the few slow IO constrained tests. No idea whether that's still the case today. (As for why we didn't use the strategy of this PR, tests only took about a minute or two to run with 4 threads, so we just didn't bother scaling beyond that.)

robgjansen: OK, great, 4 makes sense then! Thanks for the reminder!

robgjansen (approved):

PR #1160, opened 2021-03-09
Added more thorough checks to sendto-recvfrom test

stevenengler:
Compares the hash of the sent/received bytes, uses a buffer of pseudo-random bytes, and increases the sleep time to work with large Shadow values of `SYSCALL_IO_BUFSIZE`.


robgjansen (approved):
Checking the contents is a great addition!

PR #1163, opened 2021-03-10
Use a unique file name in poll tests

stevenengler:
Hopefully this fixes the test errors we've had in the CI.

This PR also sets different seeds for the 'file' and 'memory' tests since they use `mkstemp`.


robgjansen (approved):
👍

PR #1165, opened 2021-03-10
Add option to disable the MemoryManager

sporksmith:
Add option to disable the MemoryManager, and fix some bugs in the fallback memory-access paths.

* Fix buffer size requested in getdents and getdents64
* Add missing memory-flush after writing the `clear_child_tid` pointer

### DIFF
--- a/src/main/host/syscall_handler.c
+++ b/src/main/host/syscall_handler.c
@@ -50,6 +50,13 @@
 #define SYS_copy_file_range 326
 #endif
 
+static bool _useMM = true;
+OPTION_EXPERIMENTAL_ENTRY("disable-memory-manager", 0, G_OPTION_FLAG_REVERSE, G_OPTION_ARG_NONE,
+                          &_useMM,
+                          "Disable the MemoryManager. This can be useful for debugging, but will "
+                          "hurt performance in most cases.",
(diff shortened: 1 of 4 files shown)


src/main/host/syscall_handler.c
```diff
@@ -50,6 +50,12 @@
 #define SYS_copy_file_range 326
 #endif
 
+static bool _useMM = true;
+OPTION_EXPERIMENTAL_ENTRY("no-memory-manager", 0, G_OPTION_FLAG_REVERSE, G_OPTION_ARG_NONE, &_useMM,
```
robgjansen, line 54: FYI, I think Tor's options use the `--enable-XXX` and `--disable-XXX` format to enable and disable boolean features. Probably not worth changing here, since I don't think we use this consistently for the other options anyway.

robgjansen (approved):
👍  Thanks for the quick debugging help!

src/main/host/syscall_handler.c
sporksmith: Done.

I think one reason I previously went with integer types here is so that in case the default changes we don't need to change the flag. I suppose we could always just define both the `enable-X` and `disable-X` flags though.

PR #1169, opened 2021-03-10
Miscellaneous bugfixes for running tor

### LINKED ISSUE #1136: Possible issues when running Tor
Running a minimal Tor network config in phantom  at b72c1cf24e1651fb675d429ba857d3a60d4d5edc produces the following warnings in the `shadow.log` file:

1. `[clone.c:69] [syscallhandler_clone] Unhandled clone flags 0x80000`
2. `[ioctl.c:90] [syscallhandler_ioctl] We do not support ioctl request 35147 on descriptor 12`
3. `[protected.c:73] [_syscallhandler_validateDescriptor] descriptor handle '100' is closed`  
(Is not really an issue - covered by #1101)
4. `[socket.c:1201] [syscallhandler_socket] unsupported socket domain "1", we only support AF_INET`  
(I _think_ Tor can function without using UNIX sockets - covered by #1035.)
5. The dirauth seems to go to sleep on a futex at 3 minutes into the sim, and never wakes up again:

```
00:00:17.114349 [thread-0] 00:03:03.250000000 [debug] [4uthority~100.0.0.1] [syscall_handler.c:135] [_syscallhandler_pre_syscall] SYSCALL_HANDLER_PRE(tor,pid=1000): handling syscall 202 futex
00:00:17.114350 [thread-0] 00:03:03.250000000 [debug] [4uthority~100.0.0.1] [futex.c:146] [syscallhandler_futex] futex called with addr=0x22bde68 op=128 (operation=0 and options=128) and val=3
00:00:17.114353 [thread-0] 00:03:03.250000000 [debug] [4uthority~100.0.0.1] [futex.c:159] [syscallhandler_futex] Handling FUTEX_WAIT operation 0
00:00:17.114356 [thread-0] 00:03:03.250000000 [debug] [4uthority~100.0.0.1] [futex.c:48] [_syscallhandler_futexWaitHelper] Futex value is 3, expected value is 3
00:00:17.114357 [thread-0] 00:03:03.250000000 [debug] [4uthority~100.0.0.1] [futex.c:89] [_syscallhandler_futexWaitHelper] Dynamically created a new futex object for futex addr 0x3e80000022bde68
00:00:17.114360 [thread-0] 00:03:03.250000000 [debug] [4uthority~100.0.0.1] [futex.c:96] [_syscallhandler_futexWaitHelper] Futex blocking for wakeup without timeout
00:00:17.114361 [thread-0] 00:03:03.250000000 [debug] [4uthority~100.0.0.1] [syscall_handler.c:149] [_syscallhandler_post_syscall] SYSCALL_HANDLER_POST(tor,pid=1000): syscall 202 futex result: state=BLOCK code=0(n/a) in 0.000012 seconds
00:00:17.114366 [thread-0] 00:03:03.250000000 [debug] [4uthority~100.0.0.1] [syscall_condition.c:248] [_syscallcondition_logListeningState] Process 4uthority.tor.1000 thread 0x29fc710 started listening for status on futex 0x3e80000022bde68
```

robgjansen:
resolves #1136

### DIFF
--- a/src/main/host/futex.c
+++ b/src/main/host/futex.c
@@ -7,6 +7,7 @@
 
 #include <errno.h>
 #include <glib.h>
+#include <stdbool.h>
 
 #include "main/core/support/definitions.h"
 #include "main/core/support/object_counter.h"
@@ -19,7 +20,8 @@ struct _Futex {
     // The unique physical address that is used to refer to this futex
     PluginPhysicalPtr word;
     // Listeners waiting for wakups on this futex
-    // Used as a set, where both keys and values are of type StatusListener*
+    // The key is a listener of type StatusListener*, the value is a boolean that indicates
+    // whether or not a wakeup has already been performed on the listener.
     GHashTable* listeners;
     // Manage references
     int referenceCount;
@@ -30,7 +32,7 @@ Futex* futex_new(PluginPhysicalPtr word) {
     Futex* futex = malloc(sizeof(*futex));
     *futex = (Futex){.word = word,
                      .listeners = g_hash_table_new_full(
-                         g_direct_hash, g_direct_equal, NULL, (GDestroyNotify)statuslistener_unref),
+                         g_direct_hash, g_direct_equal, (GDestroyNotify)statuslistener_unref, NULL),
                      .referenceCount = 1,
                      MAGIC_INITIALIZER};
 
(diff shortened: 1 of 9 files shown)


src/main/host/futex.c
```diff
@@ -73,15 +75,29 @@ unsigned int futex_wake(Futex* futex, unsigned int numWakeups) {
...
+    unsigned int numWoken = 0;
+
+    while (item && (numWoken < numWakeups)) {
         StatusListener* listener = item->data;
 
         // Only call if the listener is still valid
         if (g_hash_table_contains(futex->listeners, listener)) {
-            statuslistener_onStatusChanged(listener, STATUS_FUTEX_WAKEUP, STATUS_FUTEX_WAKEUP);
+            // If this listener was already woken up, skip it this time
+            bool did_wakeup =
```
sporksmith, line 88: Would it make sense to add some `pending_wake` flag in the listener itself, instead of keeping an auxiliary table here?

I think that might be a behavior change if a task gets woken by multiple conditions before it runs again, but I'm not sure which behavior is right :)

src/main/host/syscall/protected.h
```diff
@@ -69,7 +69,7 @@ struct _SysCallHandler {
  * send/recv or read/write syscall. It would be more efficient to dynamically
  * compute how much we can read/write rather than using this static size.
  * TODO: remove this when we switch to dynamic size calculations. */
-#define SYSCALL_IO_BUFSIZE (1024 * 16) // 16 KiB
+#define SYSCALL_IO_BUFSIZE (1024 * 1024 * 10) // 10 MiB
```
sporksmith, line 72: Did this end up mattering for correctness? Worth a comment if so

sporksmith (approved):

src/main/host/syscall/protected.h
robgjansen: I'm pretty sure it didn't matter, but I left it in since we wanted it anyway.

src/main/host/futex.c
robgjansen: Hmm. The pending flag is already handled in the syscall condition:

https://github.com/shadow/shadow/blob/2d6ed8dc5840958aeda2ab264669f6f00a447e0c/src/main/host/syscall_condition.c#L336-L339

The problem is that the futex does not have visibility into the state of the syscall condition and so cannot query the pending flag. Maybe we could try to propagate the state of the pending flag back down the call stack, but we already use a table in futex anyway so storing the boolean is little extra complexity.

robgjansen: There may be a cleaner way to do this, but I'm going to avoid that rabbit hole for now.

PR #1173, opened 2021-03-12
Remove unnecessary string in tcp test

stevenengler:
After this, we can merge it into dev.


robgjansen (approved):
Yes, thank you!! I'm glad we have a review process in place now so that stuff like this is less likely sneak in from contributors.

Issue #1157, opened 2021-03-08
Rename utility_assert() to debug_assert()


stevenengler:
We should rename `utility_assert()` to `debug_assert()` since it only runs when Shadow is built in debug mode. Making this issue so that we don't forget about it.

stevenengler:
We decided not to change this since a regular `assert()` also only runs in debug mode, so having `utility_assert()` also only run in debug mode isn't unexpected.

PR #1178, opened 2021-03-12
Rename utility_assert() to debug_assert()

stevenengler:
Closes shadow/shadow#1157.

### DIFF
--- a/src/main/core/controller.c
+++ b/src/main/core/controller.c
@@ -81,7 +81,7 @@ struct _Controller {
 //}
 
 Controller* controller_new(Options* options) {
-    utility_assert(options);
+    debug_assert(options);
 
     /* Don't do anything in this function that will cause a log message. The
      * global engine is still NULL since we are creating it now, and logging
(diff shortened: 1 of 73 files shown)


sporksmith (approved):
I should've thought of this earlier, but libc's `assert` can also be compiled out in non-debug builds (if `NDEBUG` is defined), making `debug_assert` not the most descriptive contrast with `assert`. IIUC the main benefit of our custom macro is that it prints the backtrace before dieing. Maybe `shadow_assert` or `shd_assert` would be more descriptive, since really its most defining characteristic is that it's an assert whose definition we control?

I don't want to make you redo work or hold this up on bikeshedding though. `debug_assert` is a better name than `utility_assert`, so I'd be happy for this to be merged :)

stevenengler:
Ah, I didn't know about the `NDEBUG` macro. My reasoning for changing this was that it wasn't clear that `utility_assert()` is always compiled out of debug builds, whereas I thought `assert()` was never compiled out.

So I think we can keep the original name then, or we can change it to `shadow_assert()` if you think it's better, but I have no opinion now.

Also, what do you think about changing the `utility_assert()` macro to be defined based on `NDEBUG` instead of `DEBUG`? The end result is the same since we set `NDEBUG` in `CMakeLists.txt`, but I think this way is more consistent:

```
#ifdef DEBUG
#define utility_assert(expr) \
do { \
...
} while (0)
#else
#define utility_assert(expr)
#endif
```

we use:


```
#ifndef NDEBUG
#define utility_assert(expr) \
do { \
...
} while (0)
#else
#define utility_assert(expr)
#endif
```

sporksmith:
> So I think we can keep the original name then, or we can change it to shadow_assert() if you think it's better, but I have no opinion now.

Adding `debug` to the name wouldn't be a bad thing since it *is* something that's commonly overlooked. Maybe `shadow_debug_assert`? But yeah, I don't feel that strongly about this either; I'm fine with leaving as-is.

> Also, what do you think about changing the utility_assert() macro to be defined based on NDEBUG instead of DEBUG? The end result is the same since we set NDEBUG in CMakeLists.txt, but I think this way is more consistent:

More consistent with `assert`? Yes, I suppose so. No strong feelings either way :)

Issue #344, opened 2017-06-19
Add FIONBIO support for setting non-blocking socket mode


bobbyvinon:
ioctl with FIONBIO currently not supported, but i think it's not very hard to implement it support.

// set non-blocking
int opt = 1;
ioctl(sock, FIONBIO, &opt);

robgjansen:
Agreed, this one shouldn't be too hard to add. Thanks!

PR #1179, opened 2021-03-13
Small ioctl and sendto bug fixes, and implement FIONBIO for Nginx

stevenengler:
Fixes two socket bugs and implements `FIONBIO` for `ioctl()` to set a socket as non-blocking, which is required for Nginx.

Closes shadow/shadow#344.

Nginx can be used with the following (still missing some syscalls and socket options, but it works):

```
$ shadow --disable-memory-manager shadow-nginx.yaml
```

nginx.conf:
```
error_log stderr;

# shadow wants to run nginx in the foreground
daemon off;

# shadow doesn't support fork()
master_process off;
worker_processes 0;

# don't use the system pid file
pid nginx.pid;

events {
  # we're not using any workers, so this is the maximum number
  # of simultaneous connections we can support
  worker_connections 1024;
}

http {
  include             /etc/nginx/mime.types;
  default_type        application/octet-stream;

  # shadow does not support sendfile() nor TCP_NODELAY
  sendfile off;
  tcp_nodelay off;

  access_log off;

  server {
    listen 80;

    location / {
      root /var/www/html;
      index index.nginx-debian.html;
    }
  }
}
```

shadow-nginx.yaml:
```
options:
  stoptime: 10
topology:
- graphml: |
    <graphml xmlns="http://graphml.graphdrawing.org/xmlns" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://graphml.graphdrawing.org/xmlns http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd">
        <node id="poi-1">
          <data key="d0">US</data>
          <data key="d1">10240</data>
          <data key="d2">10240</data>
        </node>
          <data key="d3">50.0</data>
          <data key="d4">0.0</data>
        </edge>
      </graph>
    </graphml>
plugins:
- id: nginx
  path: /usr/sbin/nginx
- id: curl
  path: /usr/bin/curl
hosts:
- id: server
  quantity: 1
  processes:
  - plugin: nginx
    starttime: 1
    arguments: -c /path/to/working-dir/nginx.conf -p /path/to/working-dir/shadow.data/hosts/server
- id: client
  quantity: 20
  processes:
  - plugin: curl
    starttime: 3
    arguments: server --verbose
```

### DIFF
--- a/src/main/host/descriptor/descriptor.c
+++ b/src/main/host/descriptor/descriptor.c
@@ -237,3 +237,8 @@ void descriptor_addFlags(LegacyDescriptor* descriptor, gint flags) {
     MAGIC_ASSERT(descriptor);
     descriptor->flags |= flags;
 }
+
+void descriptor_removeFlags(LegacyDescriptor* descriptor, gint flags) {
+    MAGIC_ASSERT(descriptor);
+    descriptor->flags &= ~flags;
+}
--- a/src/main/host/descriptor/descriptor.h
+++ b/src/main/host/descriptor/descriptor.h
@@ -36,6 +36,7 @@ gint* descriptor_getHandleReference(LegacyDescriptor* descriptor);
 gint descriptor_getFlags(LegacyDescriptor* descriptor);
 void descriptor_setFlags(LegacyDescriptor* descriptor, gint flags);
 void descriptor_addFlags(LegacyDescriptor* descriptor, gint flags);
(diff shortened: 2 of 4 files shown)


sporksmith:
Woah, cool!

robgjansen (approved):
Looks great! I'm excited about Nginx support 👍

PR #1183, opened 2021-03-15
Improvements to the CompatDescriptor

stevenengler:
Instead of dropping the CompatDescriptor internally when the descriptor is deregistered, the descriptor table now returns an owned reference to the CompatDescriptor so that we can drop it ourselves. This lets us add a nicer `close()` method to the CompatDescriptor.

### DIFF
--- a/src/main/bindings/rust/wrapper.rs
+++ b/src/main/bindings/rust/wrapper.rs
@@ -400,7 +400,10 @@ extern "C" {
     ) -> ::std::os::raw::c_int;
 }
 extern "C" {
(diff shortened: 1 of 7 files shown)


stevenengler:
@sporksmith In shadow/shadow#1145 you mentioned implementing `Drop` to close the file when all descriptor references are dropped. This PR still doesn't quite do that, but it adds a method `close(self, &mut EventQueue)` which will drop the descriptor object. I'm still not sure if we want to run this automatically with `impl Drop`, but if we decide to in the future, it should be very easy to implement now.

sporksmith (approved):

PR #1184, opened 2021-03-15
Run small Tor network as part of CI tests

stevenengler:
This PR adds a small tor experiment to the CMake tests that is not run by default. This test is run in the CI whenever the dev branch is updated, or if run manually using a `workflow_dispatch` event.


robgjansen:
The `v3bw` file was supposed to be loaded by the authority. The `tor.authority.torrc` file has the following option:

```
V3BandwidthsFile shadow.data/hosts/torflowauthority/v3bw
```

Maybe we can just put the v3bw file in the 4uthority dir instead, and update this path? Or maybe we don't actually need the v3bw file after all? (I'm surprised that Tor does not fail since the above file no longer exists at the specified path.)

-----

In `full.config.yaml` it does look like the `torbridgeclient` is using the `4uthority` relay as the bridge instead of a separate bridge relay:

```
...--UseBridges 1 --Bridge 100.0.0.1:9111
```

So removing the bridge dir seems ok.

stevenengler:
> The `v3bw` file was supposed to be loaded by the authority. The `tor.authority.torrc` file has the following option:
> 
> ```
> V3BandwidthsFile shadow.data/hosts/torflowauthority/v3bw
> ```
> 
> Maybe we can just put the v3bw file in the 4uthority dir instead, and update this path? Or maybe we don't actually need the v3bw file after all? (I'm surprised that Tor does not fail since the above file no longer exists at the specified path.)

I added this file back.

> In `full.config.yaml` it does look like the `torbridgeclient` is using the `4uthority` relay as the bridge instead of a separate bridge relay:
> 
> ```
> ...--UseBridges 1 --Bridge 100.0.0.1:9111
> ```
> 
> So removing the bridge dir seems ok.

Should I re-add the bridge and update the bridge line for torbridgeclient to use it?

robgjansen:
> Should I re-add the bridge and update the bridge line for torbridgeclient to use it?

I don't think we need it because the authority already fills the bridge role. I don't think we need a bridge that does not also act as an authority, but if you are excited to get that working and want to add it, go ahead.

stevenengler:
> > Should I re-add the bridge and update the bridge line for torbridgeclient to use it?
> 
> I don't think we need it because the authority already fills the bridge role. I don't think we need a bridge that does not also act as an authority, but if you are excited to get that working and want to add it, go ahead.

I'm fine with leaving it out, just wanted to know if you had a preference.

stevenengler:
@robgjansen Before merging, I'll disable the `on: pull_request` trigger for this action. There is also a commit specifically to show changes I made to the config files you sent me, so let me know if you don't agree with any of them.

src/test/tor/minimal/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+                 LOGLEVEL info
+                 ARGS
+                   --pin-cpus
+                   --workers 2
+                   --data-template "shadow.data.template"
+                 PROPERTIES
+                   TIMEOUT 180
+                   LABELS tor
+                   CONFIGURATIONS tor
+                   RUN_SERIAL TRUE)
```
robgjansen, line 37: I think we should check that `Bootstrapped 100` appears 9 times in the Tor log files (once per Tor process) and `stream-success` 40 times in the tgen log files (10 downloads per tgen client) after the simulation finishes. Are we doing this somewhere that I am overlooking?

robgjansen (changes requested):

robgjansen (approved):
Thank you! Looks good (assuming the tests pass).

PR #1187, opened 2021-03-16
Added Rust wrappers around process_get*Ptr() functions

stevenengler:
Ideally the memory manager would provide this interface, and these new enums would be passed all the way from there. But until we write the Process and Thread objects in Rust, these wrappers will be good enough.

The PosixFile `read()` and `write()` methods now accept an `Option` for the buffer. Different Linux file types handle NULL buffers differently, so pass this information along to the file object.

I'm open to changing the names for the enums, but I can't think of any better names.

### DIFF
--- a/src/main/host/descriptor/pipe.rs
+++ b/src/main/host/descriptor/pipe.rs
@@ -55,7 +55,7 @@ impl PipeFile {
 
     pub fn read(
         &mut self,
-        bytes: &mut [u8],
+        bytes: Option<&mut [u8]>,
         offset: libc::off_t,
         event_queue: &mut EventQueue,
     ) -> SyscallReturn {
@@ -69,6 +69,11 @@ impl PipeFile {
             return SyscallReturn::Error(nix::errno::Errno::EBADF);
         }
 
+        let bytes = match bytes {
+            Some(b) => b,
+            None => return SyscallReturn::Error(nix::errno::Errno::EFAULT),
+        };
+
         let num_read = self.buffer.borrow_mut().read(bytes, event_queue);
 
         SyscallReturn::Success(num_read as i32)
@@ -76,7 +81,7 @@ impl PipeFile {
 
     pub fn write(
         &mut self,
-        bytes: &[u8],
+        bytes: Option<&[u8]>,
         offset: libc::off_t,
         event_queue: &mut EventQueue,
     ) -> SyscallReturn {
@@ -89,6 +94,11 @@ impl PipeFile {
         if !self.mode.contains(FileMode::WRITE) {
             return SyscallReturn::Error(nix::errno::Errno::EBADF);
         }
+
+        let bytes = match bytes {
+            Some(b) => b,
+            None => return SyscallReturn::Error(nix::errno::Errno::EFAULT),
+        };
 
         let num_written = self.buffer.borrow_mut().write(bytes, event_queue);
 
--- a/src/main/host/syscall/mod.rs
+++ b/src/main/host/syscall/mod.rs
@@ -1,5 +1,7 @@
-use atomic_refcell::AtomicRefCell;
+use std::convert::TryInto;
 use std::sync::Arc;
+
+use atomic_refcell::AtomicRefCell;
 
 use crate::cshadow as c;
 use crate::host::descriptor::{CompatDescriptor, FileStatus, PosixFile};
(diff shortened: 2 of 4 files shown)


src/main/host/descriptor/pipe.rs
```diff
@@ -69,14 +69,19 @@ impl PipeFile {
             return SyscallReturn::Error(nix::errno::Errno::EBADF);
         }
 
+        let bytes = match bytes {
```
sporksmith, line 72: Optional throughout: more succinct as

```
let bytes = bytes.ok_or(SyscallReturn::Error(nix::errno::Errno::EFAULT))
```

Or maybe avoid constructing the error in the ok case (but probably a wash in this case)
```
let bytes = bytes.ok_or_else(|| SyscallReturn::Error(nix::errno::Errno::EFAULT))
```

src/main/host/syscall/mod.rs
```diff
@@ -72,3 +74,115 @@ pub fn get_descriptor(
...
+    Null,
+    /// Length was 0.
+    ZeroLen,
+    /// Length was too small to represent the generic type `T`.
+    LenTooSmall,
+    /// A local aligned non-null address for the plugin data.
+    Address(*mut T),
+}
+
+pub fn get_readable_ptr<T>(
```
sporksmith, line 100: It'd be nice if we could encapsulate the raw pointers into a smaller piece of the code. The MemoryManager has some APIs like this (`get_ref` etc) but those don't fall back to the Thread methods on a miss.

Would it make sense to introduce a Process object here on the Rust side that wraps a Process*, and provides more idiomatic Rust apis? (Or maybe a Process *trait*, with an implementation that wraps a C process, as we have with Thread)

The return types could use `Result` with the appropriate reference as the `Ok` type, and a single custom error type enum replacing the two you have above.

This will potentially force us to deal with some ownership and mutability issues, but we probably *want* that rather than giving up safety by working with raw pointers. (If we really wanted to dodge those, I think maybe these APIs could return references with 'static lifetime instead of the lifetime of the Process object, but that'd again be giving up some safety)

sporksmith (changes requested):

src/main/host/descriptor/pipe.rs
stevenengler: This line replaces an `Option<&[u8]>` with a `Result<&[u8], SyscallReturn>`, but we still need to unpack the type to either return the `SyscallReturn`, or destructure the enum to get the `&[u8]`.

sporksmith: Ah right. Maybe we ought to break the Error state out of SyscallReturn so that we can use Result types throughout, and here use the `?` operator (We'd need to define `From<nix::errno::Errno>` for our error type, assuming we didn't use that directly as our error type)

Seems orthogonal to this PR though, so nevermind for now :)

stevenengler: My reasoning for not doing it that way is that `SyscallReturn::Error` represents an error state for the plugin and not Shadow. So having a Shadow function return a `Result::Err` when it's not actually a Shadow error might be confusing. And then, how do we return an actual Shadow error if we're already using a `Result::Err` to return the plugin error state?

I'm not against doing it this way, especially if it cuts down on the amount of code. It's just not clear to me which way is better in the long term.

src/main/host/syscall/mod.rs
stevenengler: > It'd be nice if we could encapsulate the raw pointers into a smaller piece of the code. The MemoryManager has some APIs like this (get_ref etc) but those don't fall back to the Thread methods on a miss.

There are three main issues:
1. Like you mentioned, we need to access a path that has the fallback option, so we can't access the memory manager directly.
2. The MemoryManager's methods like get_ref() and get_slice() don't work well for types like `sockaddr` which later has to be re-interpreted as types of possibly larger sizes. We could get a u8 slice, and then convert the bytes to the various sockaddr types, but this doesn't really seem any safer than just using a pointer. Is there a nicer way of dealing with this?
3. You cannot borrow more than one reference at a time from the memory manager. You have a comment `TODO: Consider tracking borrowed memory ranges so that we can safely allow mutable borrowed references, with a run-time check [...]` so we would need to implement that.

I think if we want safety here, we need to restructure how we manage plugin memory. We need to remove the implicit `thread_flushPtrs()` so that we can give the references proper lifetimes. We need a manager for the MemoryManager and the fallback path to make sure that they don't return aliased pointers and that they're aligned. Or we'd need to move the fallback path out of `process_getWriteablePtr()` and into MemoryManager, and implement this runtime bookkeeping in MemoryManager instead. This should also be compatible with the C code.

> Would it make sense to introduce a Process object here on the Rust side that wraps a Process*, and provides more idiomatic Rust apis? (Or maybe a Process trait, with an implementation that wraps a C process, as we have with Thread)

We'd want to avoid ever creating two Process objects with the same pointer. I'm not sure how we can guarantee this. How about if the C Process object has a `bool rustWrapperExists` and the rust Process object sets this when created, and removes it when dropped?

> The return types could use Result with the appropriate reference as the Ok type, and a single custom error type enum replacing the two you have above.
>
> This will potentially force us to deal with some ownership and mutability issues, but we probably want that rather than giving up safety by working with raw pointers. (If we really wanted to dodge those, I think maybe these APIs could return references with 'static lifetime instead of the lifetime of the Process object, but that'd again be giving up some safety)

We could do this, but we'd need to figure out the memory management issue first. I don't think we should return a reference with a `'static` lifetime as that seems more dangerous to me than returning a pointer.

src/main/host/descriptor/pipe.rs
sporksmith: > My reasoning for not doing it that way is that SyscallReturn::Error represents an error state for the plugin and not Shadow.

I think that's ok. Right now there is no mechanism for returning a Shadow error, so we wouldn't actually lose anything through this transformation.

> So having a Shadow function return a Result::Err when it's not actually a Shadow error might be confusing.

Somewhat subjective, but IMO using `Result:Err` for this sort of use-case is reasonably idiomatic.

> And then, how do we return an actual Shadow error if we're already using a Result::Err to return the plugin error state?

We'd only need to do that in the case of some kind of recoverable Shadow error. So far though most Shadow errors aren't recoverable, and we just want to `panic` (well, `error`) right where the error happened rather than bubbling it up through the call stack.

If we really wanted to be able to represent Shadow errors though, we could use an error type like:

```
enum SyscallError {
  PluginError(nix::errno::Errno),
  ShadowError1(...),
  ShadowError2(...),
}
```

src/main/host/syscall/mod.rs
sporksmith: > Like you mentioned, we need to access a path that has the fallback option, so we can't access the memory manager directly.

Yup, I think we'll want to expose the Process APIs in rust and wrap them so they return references.

> The MemoryManager's methods like get_ref() and get_slice() don't work well for types like sockaddr which later has to be re-interpreted as types of possibly larger sizes. We could get a u8 slice, and then convert the bytes to the various sockaddr types, but this doesn't really seem any safer than just using a pointer. Is there a nicer way of dealing with this?

Good point. Still, I think it's better to return references that can just be used as references in the common case, and only in some exceptional cases have the handling code have to delve back into casting/transmuting. Alternatively we could still provide APIs that return pointers, so the syscall handlers that do need pointers could just use those, but I think that'd be weakening any exclusivity guarantees etc.

> You cannot borrow more than one reference at a time from the memory manager. You have a comment TODO: Consider tracking borrowed memory ranges so that we can safely allow mutable borrowed references, with a run-time check [...] so we would need to implement that.

Maybe. I have been wondering if we can get away without it. A syscall handler could just "check out" one reference at a time and return it before getting the next. For pointers to primitives or smallish structs it could just copy locally before releasing the reference if needed.

> We could do this, but we'd need to figure out the memory management issue first. I don't think we should return a reference with a 'static lifetime as that seems more dangerous to me than returning a pointer.

Fair enough.

I'm on the fence whether it makes sense to try to add these APIs now. On one hand it's nontrivial and will block your work in progress, on the other hand doing it now would save work in the long run vs. implementing more handlers that operate on raw pointers and then migrating them to reference-based APIs later.

stevenengler:
@sporksmith I've changed the return type here. Most of the added code in this PR will be replaced in a follow-up PR, but I still want to add this code as it forces us to explicitly consider all of the possible plugin pointer cases (null pointer, zero length, etc), which I think will be useful.

sporksmith (approved):

PR #1190, opened 2021-03-18
Fix issues when exiting threaded processes

### LINKED ISSUE #1166: Ptrace hangs during shutdown of Tor plugins
On d8ea39e010f52c8077dfcf3b63b0255c5f6a270a, at the end of a Tor simulation, the plugin processes are still running and Shadow has to send them signals to shut them down. During this step, ptrace hangs. Here is the backtrace:

```
0x00007f11e5d51ccf in wait4 () from /lib64/libc.so.6
(gdb) bt
#0  0x00007f11e5d51ccf in wait4 () from /lib64/libc.so.6
#1  0x00000000005e7205 in _waitpid_spin (pid=1552335, wstatus=0x7ffd4f35f034, options=0)
    at /home/rjansen/dev/shadow/src/main/host/thread_ptrace.c:528
#2  0x00000000005e8e89 in threadptrace_terminate (base=0x2b59400) at /home/rjansen/dev/shadow/src/main/host/thread_ptrace.c:1012
#3  0x00000000005dff49 in thread_terminate (thread=0x2b59400) at /home/rjansen/dev/shadow/src/main/host/thread.c:107
#4  0x00000000005dca77 in _process_reapThread (process=0x2a56430, thread=0x2b59400) at /home/rjansen/dev/shadow/src/main/host/process.c:138
#5  0x00000000005dcc40 in _process_terminate_threads (proc=0x2a56430) at /home/rjansen/dev/shadow/src/main/host/process.c:197
#6  0x00000000005ddbb8 in process_stop (proc=0x2a56430) at /home/rjansen/dev/shadow/src/main/host/process.c:461
#7  0x00000000005e1c93 in host_freeAllApplications (host=0x2a54d20) at /home/rjansen/dev/shadow/src/main/host/host.c:396
#8  0x00000000005db733 in _worker_freeHostProcesses (host=0x2a54d20, worker=0x2a65c50) at /home/rjansen/dev/shadow/src/main/core/worker.c:341
#9  0x00007f11e620df9f in g_queue_foreach (queue=<optimized out>, func=0x5db6ff <_worker_freeHostProcesses>, user_data=0x2a65c50)
    at ../glib/gqueue.c:281
#10 0x00000000005db7be in worker_freeHosts (hosts=0x2a410e0) at /home/rjansen/dev/shadow/src/main/core/worker.c:355
#11 0x00000000005fccdc in _scheduler_stopHosts (scheduler=0x2a4b180) at /home/rjansen/dev/shadow/src/main/core/scheduler/scheduler.c:110
#12 0x00000000005fe306 in scheduler_awaitFinish (scheduler=0x2a4b180) at /home/rjansen/dev/shadow/src/main/core/scheduler/scheduler.c:610
#13 0x00000000005db0e9 in worker_run (data=0x2a5df90) at /home/rjansen/dev/shadow/src/main/core/worker.c:196
#14 0x00000000005da885 in slave_run (slave=0x2a47ce0) at /home/rjansen/dev/shadow/src/main/core/slave.c:554
#15 0x00000000005d8c58 in master_run (master=0x2a3d5e0) at /home/rjansen/dev/shadow/src/main/core/master.c:465
#16 0x00000000005d7215 in _main_helper (options=0x2a371e0) at /home/rjansen/dev/shadow/src/main/core/main.c:165
#17 0x00000000005d7571 in main_runShadow (argc=5, argv=0x7ffd4f35f5d8) at /home/rjansen/dev/shadow/src/main/core/main.c:279
#18 0x00000000005d6966 in main (argc=5, argv=0x7ffd4f35f5d8) at /home/rjansen/dev/shadow/src/main/main.c:9
```

sporksmith:
Fixes #1166


src/main/host/thread_ptrace.c
```diff
@@ -809,20 +831,67 @@ static void threadptrace_flushPtrs(Thread* base) {
...
+            //
+            // If this task was the thread leader, _threadptrace_reapProcess
+            // should be called later, which will get the exit code.
+            // Otherwise, there's no way to get the exit code now.
+            thread->childState = THREAD_PTRACE_CHILD_STATE_EXITED;
+            return;
+        } else if (errno == EPERM) {
+            // Experimentally we can alternatively get EPERM here if the thread
+            // has disappeared. Haven't found any documented reason why.
+            debug("Got EPERM when attempting to reattach");
```
robgjansen, line 848: I wonder if this could happen if the thread exited and then another process was started by the system with the same tid. Though I don't think that is likely if tids are monotonically increasing (until they wrap).

sporksmith: Yeah, agreed on both counts - might explain it but seems unlikely.

I was thinking about this some more and think I'm going to try a different approach. In general when something happens that kills the entire process (namely `exit_group` being called or being killed by a signal), the threads die at unpredictable times. After iteratively adding more and more "exception handling" to handle the thread going away at different points, I think in principle it could happen in any ptrace or waitpid call. 

I'm thinking instead when we know the entire process is killed, we should not try resuming any other threads in the process again. e.g. I'll add an exit_group syscall handler that marks the process dead before making the syscall. When the Process code returns from the thread it can check if it's dead, and if so just do the bare minimum cleanup of other threads. We can also skip writing to the tid mutex in this case, avoiding potential trouble of writing to the `/proc/x/mem` file of a dead process.

robgjansen: Ahh, there is mention of handing `exit_group` in #826, but we never got there (yet). :)

sporksmith:
This is currently failing the determinism tests in the coverage build. It looks like these use the default interposition mechanism, which is currently hybrid mode. I expect #1195 will "fix" this, by implicitly switching these to use ptrace mode.

I'm still trying to find why it is failing in hybrid mode, but I'm not sure it's worth investing much effort in it right now. Basically we're getting to an unexpected state (trying to make a native syscall but we didn't save the plugin's register values, which we need). The state machine in hybrid mode is a bit tricky. I'm a little tempted to stop testing hybrid mode altogether since it's likely to be largely rewritten before we want to use it again anyway.

sporksmith:
Actually I'm starting another PR to update the determinism tests to use the macros (unless there's a reason I don't see not to?), test all modes (maybe excluding hybrid mode with a comment when this is merged), and add `RUN_SERIAL` to the `-w2` versions

src/main/host/thread_ptrace.c
```diff
@@ -1079,41 +1114,27 @@ bool threadptrace_isRunning(Thread* base) {
...
     }
 
     StopReason reason = _getStopReason(wstatus);
+    _threadptrace_updateChildState(thread, reason);
 
-    if (reason.type != STOPREASON_EXITED_NORMAL && reason.type != STOPREASON_EXITED_SIGNAL) {
-        error("Expected process %d to exit after SIGKILL, instead received status %d",
-              thread->base.nativePid, wstatus);
+    if (threadptrace_isRunning(&thread->base)) {
+        warning("Process %d to have exited, instead received status %d", thread->base.nativePid,
```
robgjansen, line 1129: I think you're missing a verb, e.g., "_Expected_ process %d to have exited"?

robgjansen (approved):

src/main/host/thread_preload.c
```diff
@@ -336,15 +341,10 @@ void threadpreload_terminate(Thread* base) {
 
     utility_assert(thread->base.nativePid > 0);
 
-    pid_t rc = waitpid(thread->base.nativePid, &status, WNOHANG);
+    // Process should already be dead; reap the pid
+    pid_t rc = waitpid(thread->base.nativePid, &status, 0);
```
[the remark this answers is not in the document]
sporksmith: Done

src/main/host/thread_ptrace.c
sporksmith: This is actually dead code; deleted

sporksmith:
Fixing your comments cascaded into consolidating where the process is waited on and the exit code  obtained into Process, but it's a big enough change that I'm going to make it a separate PR

PR #1195, opened 2021-03-19
Use ptrace mode by default

sporksmith:
Hybrid mode is a poor default since it's currently slower, and
development on it is on hold.

ptrace seems like the most reasonable default right now, since it's
faster than hybrid mode and supports threads.


robgjansen (approved):
👍

PR #1201, opened 2021-03-22
ptrace: get registers on-demand

sporksmith:
This fixes a known bug where we sometimes get to thread_nativeSyscall without already having saved registers. Moving register-fetching to when we need them should help protect against such bugs in the future, and could save us work of fetching them in cases where we *don't* actually need them later.


sporksmith:
@stevenengler I was still having trouble consistently reproducing the bug locally - probably would be worth verifying on your machine since you seemed to be hitting more consistently.

stevenengler (approved):
Fixes the tests for me, thanks! I ran them several times in both debug and release mode and no tests failed.

PR #1202, opened 2021-03-22
Removed the second test run with '--rerun-failed' in CI script

stevenengler:
Shadow should be stable enough now that we don't need to filter out flaky tests, and if there are any, we probably want to know about them.


stevenengler:
I've re-run the tests three times and they haven't failed.

sporksmith (approved):

Issue #1198, opened 2021-03-22
Support for chdir/fchdir syscalls is broken


stevenengler:
The `chdir` syscall is enabled with `NATIVE(chdir)`, but if a plugin process changes its working directory, any files it creates/opens with a relative path will be relative to Shadow's working directory, not the plugin's.

The `fchdir` syscall is enabled with `HANDLE(fchdir)`, which runs `file_fchdir(File*)` and eventually `fchdir()` directly. This will change the working directory of Shadow itself, not the working directory of the plugin process.

https://github.com/shadow/shadow/blob/ed5682c462165da7386dda114538b52096212044/src/main/host/descriptor/file.c#L564-L576

stevenengler:
These syscalls were removed from Shadow in #1204 so this is no longer an issue.

PR #1204, opened 2021-03-22
Set the plugin's working directory to the host directory, and disable support for chdir/fchdir

### LINKED ISSUE #1192: Set the working directory of each plugin process to its corresponding 'hosts/hostname' directory
When Shadow runs its plugin/simulated processes, each process uses the same working directory as Shadow. Instead we should set each process' working directory to its host directory (for example `shadow.data/hosts/hostname`).

One advantage of this approach is that the Shadow config file wouldn't need to set the `DataDirectory` for each tor process explicitly depending on its hostname. Instead the data directory could just be in the current working directory.

stevenengler:
Adds support for process-specific working directories, sets the default working directory of plugin processes to their corresponding host directory, and disables the chdir/fchdir syscalls as they don't currently work correctly. These changes will hopefully help make it easier to support chdir/fchdir in the future.

Closes #1192 and #1198.


src/main/host/process.c
```diff
@@ -101,6 +101,9 @@ struct _Process {
     SimulationTime startTime;
     SimulationTime stopTime;
 
+    /* absolute path to the process' working directory */
```
sporksmith, line 104: Nit - process's. A trailing ' is only used with plural nouns. e.g. processes'

src/main/host/process.c
```diff
@@ -351,8 +359,15 @@ static File* _process_openStdIOFileHelper(Process* proc, int fd, gchar* fileName
...
     File* stdfile = file_new();
+
+    char* cwd = getcwd(NULL, 0);
+    if (!cwd) {
+        error("getcwd unable to allocate string buffer, error %i: %s", errno, strerror(errno));
+    }
+
     int errcode = file_open(stdfile, fileName, O_WRONLY | O_CREAT | O_TRUNC,
-                            S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH);
+                            S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH, cwd);
```
sporksmith, line 369: Would it make sense to make the stdout and stdin filenames relative to the process's workingDir, and then pass proc->workingDir here?

src/main/host/thread.c
```diff
@@ -78,13 +78,13 @@ void thread_unref(Thread* thread) {
     }
 }
 
-void thread_run(Thread* thread, gchar** argv, gchar** envv) {
+void thread_run(Thread* thread, gchar** argv, gchar** envv, const char* workingDir) {
```
sporksmith, line 81: Optional - the thread could get the workingDir from its process instead of wiring it through here. OTOH the same could probably be said of `argv` and `envv`, so maybe it makes sense to pass `workingDir` here for consistency.

src/main/host/thread_ptrace.c
```diff
@@ -371,6 +372,10 @@ static pid_t _threadptrace_fork_exec(const char* file, char* const argv[], char*
             if (prctl(PR_SET_TSC, PR_TSC_SIGSEGV, 0, 0, 0) < 0) {
                 error("prctl: %s", g_strerror(errno));
             }
+            // Set the working directory
+            if (workingDir != NULL && chdir(workingDir) < 0) {
```
sporksmith, line 376: Probably doesn't make sense to silently ignore a NULL workingDir, since that'd be a programming error. Maybe `utility_assert(workingDir)`?

src/main/host/thread_preload.c
```diff
@@ -143,6 +142,11 @@ static pid_t _threadpreload_fork_exec(ThreadPreload* thread, const char* file, c
         case 0: {
             // child
 
+            // Set the working directory
+            if (workingDir != NULL && chdir(workingDir) < 0) {
```
sporksmith, line 146: Probably doesn't make sense to silently ignore a NULL workingDir, since that'd be a programming error. Maybe `utility_assert(workingDir)`?

sporksmith (approved):

src/main/host/process.c
stevenengler: Fixed

src/main/host/process.c
stevenengler: Since we want to always create these files in the same directory regardless of the working directory of the process, I think using Shadow's working directory makes it clearer that we don't want it to be relative to the process's working directory. These paths are actually absolute paths anyways so the working directory doesn't matter in practice.

src/main/host/thread.c
stevenengler: I'd prefer to keep them consistent. And these members aren't accessible through accessors from the thread, so I think it's better to keep these private to the process.

src/main/host/thread_ptrace.c
stevenengler: Done

src/main/host/thread_preload.c
stevenengler: Done

PR #1205, opened 2021-03-22
Count syscalls

robgjansen:
This allows us to count the frequency with which each syscall is made by each processing running in shadow. No counting is done unless the new `--count-syscalls` option is specified. If the option is given, we track the counts for each syscall and log the counts at the end of the simulation.

### DIFF
--- a/src/main/utility/counter.rs
+++ b/src/main/utility/counter.rs
@@ -2,28 +2,44 @@
  * The Shadow Simulator
  * See LICENSE for licensing information
  */
+
+/*!
+A counter that can be used to count frequencies of a set of objects. The counter starts
+with no keys. When an unknown key is incremented, the counter adds a new key to an
+internal map and sets the count for that key to 1. When a known key is incremented, the
+count for that key increases. The state of the counter can be extracted by converting it
+to a string, which lists the counts for all keys sorted with the heaviest hitters first.
+Currently, only String types are supported, but we may eventually support counting
+generic types.
+*/
+
 use std::collections::HashMap;
 use std::iter::FromIterator;
 
+/// The main counter object that maps individual keys to count values.
 pub struct Counter {
+    // TODO: convert this so we could count generic types instead of Strings
     items: HashMap<String, u64>,
 }
 
 impl Counter {
+    /// Initializes a new counter map that starts with no keys.
     pub fn new() -> Counter {
         Counter {
             items: HashMap::new(),
         }
     }
 
+    /// Increment the counter value for the key given by id.
+    /// Returns the value of the counter after it was incremented.
     pub fn add_one(&mut self, id: &str) -> u64 {
         let count = self.items.entry(id.to_string()).or_insert(0);
         *count = *count + 1;
         *count
     }
 
-    // Used for unit test
-    #[allow(dead_code)]
+    /// Returns the counter value for the key given by id, or 0 if the key has not
+    /// previously been incremented.
     pub fn get_value(&mut self, id: &str) -> u64 {
         match self.items.get(&id.to_string()) {
             Some(val) => *val,
@@ -31,6 +47,9 @@ impl Counter {
         }
     }
 
+    /// Returns a string representation of the counter. The string starts with "Counts:"
+    /// and then lists known keys and values as a space-separated list of "key=value"
+    /// pairs sorted by value with the largest value first.
     fn to_string(&mut self) -> String {
         // Sort the counts with the heaviest hitters first
         let mut item_vec = Vec::from_iter(&self.items);
@@ -77,6 +96,8 @@ mod export {
         counter.add_one(id.to_str().unwrap())
     }
 
+    /// Creates a new string representation of the counter, e.g., for logging.
+    /// The returned string must be free'd by passing it to counter_free_string.
     #[no_mangle]
     pub extern "C" fn counter_alloc_string(counter: *mut Counter) -> *mut c_char {
         assert!(!counter.is_null());
@@ -88,6 +109,7 @@ mod export {
         CString::new(string).unwrap().into_raw()
     }
 
+    /// Frees a string previously returned from counter_alloc_string.
     #[no_mangle]
     pub extern "C" fn counter_free_string(counter: *mut Counter, ptr: *mut c_char) {
         assert!(!counter.is_null());
--- a/src/main/utility/counter.rs
+++ b/src/main/utility/counter.rs
@@ -1,0 +1,154 @@
+/*
+ * The Shadow Simulator
+ * See LICENSE for licensing information
+ */
+use std::collections::HashMap;
+use std::iter::FromIterator;
+
+pub struct Counter {
+    items: HashMap<String, u64>,
+}
+
+impl Counter {
+    pub fn new() -> Counter {
+        Counter {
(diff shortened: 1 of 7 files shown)


src/main/utility/counter.rs
```diff
@@ -0,0 +1,178 @@
...
+        let count = self.items.entry(id.to_string()).or_insert(0);
+        *count = *count + 1;
+        *count
+    }
+
+    /// Returns the counter value for the key given by id, or 0 if the key has not
+    /// previously been incremented.
+    // Used for unit test
+    #[allow(dead_code)]
+    fn get_value(&mut self, id: &str) -> u64 {
```
sporksmith, line 45: Optional - seems like a reasonable method to make public, even if we don't currently use it. If so we should be able to drop the `#[allow(dead_code)]`

src/main/utility/counter.rs
```diff
@@ -0,0 +1,178 @@
...
+        match self.items.get(&id.to_string()) {
```
sporksmith, line 46: Unless I'm mistaken, you should be able to pass `id` here without converting to `String`.

src/main/utility/counter.rs
```diff
@@ -0,0 +1,178 @@
...
+            Some(val) => *val,
+            None => 0,
+        }
+    }
+
+    /// Returns a string representation of the counter. The string starts with "Counts:"
+    /// and then lists known keys and values as a space-separated list of "key=value"
+    /// pairs sorted by value with the largest value first.
+    fn to_string(&mut self) -> String {
```
sporksmith, line 55: This is fine, but fyi it's probably a little more idiomatic to implement the [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) trait. That would allow these objects to be used directly with formatters, and would end up also providing a `to_string` method via the [ToString](https://doc.rust-lang.org/std/string/trait.ToString.html) trait

src/main/utility/counter.rs
```diff
@@ -0,0 +1,178 @@
...
+    /// The returned string must be free'd by passing it to counter_free_string.
+    #[no_mangle]
+    pub extern "C" fn counter_alloc_string(counter: *mut Counter) -> *mut c_char {
+        assert!(!counter.is_null());
+
+        let counter = unsafe { &mut *counter };
+        let string = counter.to_string();
+
+        // Transfer ownership back to caller
+        CString::new(string.as_str()).unwrap().into_raw()
```
sporksmith, line 111: Optional - I think you can convert String to &str a little more succinctly with `&`. e.g. `&string`

sporksmith (changes requested):

src/main/utility/counter.rs
robgjansen: Got an error:

```
error[E0277]: the trait bound `std::vec::Vec<u8>: std::convert::From<&std::string::String>` is not satisfied
   --> utility/counter.rs:109:9
    |
109 |         CString::new(&string).unwrap().into_raw()
    |         ^^^^^^^^^^^^ the trait `std::convert::From<&std::string::String>` is not implemented for `std::vec::Vec<u8>`
    | 
   ::: /home/rjansen/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/ffi/c_str.rs:404:19
    |
404 |     pub fn new<T: Into<Vec<u8>>>(t: T) -> Result<CString, NulError> {
    |                   ------------- required by this bound in `std::ffi::CString::new`
    |
    = help: the following implementations were found:
              <std::vec::Vec<T> as std::convert::From<&[T]>>
              <std::vec::Vec<T> as std::convert::From<&mut [T]>>
              <std::vec::Vec<T> as std::convert::From<[T; N]>>
              <std::vec::Vec<T> as std::convert::From<std::borrow::Cow<'a, [T]>>>
            and 6 others
    = note: required because of the requirements on the impl of `std::convert::Into<std::vec::Vec<u8>>` for `&std::string::String`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
error: could not compile `shadow-rs`.
```

src/main/utility/counter.rs
robgjansen: Done.

src/main/utility/counter.rs
robgjansen: Done.

src/main/utility/counter.rs
sporksmith: Oops, yeah I'd mis-guessed exactly what `CString::new` takes.

Looking at the [docs](https://doc.rust-lang.org/std/ffi/struct.CString.html#creating-a-cstring) though, it looks like we can pass `string` by value, which should "steal/move" the `String`'s heap-allocated string data, instead of having to allocate and copy it from a `&str`. `CString::new(string).unwrap().into_raw()`

robgjansen: Yay, that worked!

src/main/utility/counter.rs
```diff
@@ -0,0 +1,178 @@
...
+    fn new() -> Counter {
+        Counter {
+            items: BTreeMap::new(),
+        }
+    }
+
+    /// Increment the counter value for the key given by id.
+    /// Returns the value of the counter after it was incremented.
+    fn add_one(&mut self, id: &str) -> u64 {
+        let count = self.items.entry(id.to_string()).or_insert(0);
```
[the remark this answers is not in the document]
robgjansen: We think we want to go the `BTreeMap.get_mut` route as that better aligns with supporting generic keys in the future.

src/main/utility/counter.rs
```diff
@@ -24,45 +25,76 @@ pub struct Counter {
...
 
-        // Create a string representation of the counts
-        let mut string = String::from("Counts:");
-        for item in item_vec.iter() {
-            string.push_str(format!(" {}={}", item.0, item.1).as_str());
+        // Create a string representation of the counts by iterating over the items.
+        let len = item_vec.len();
+
+        // Opening bracket.
+        match write!(f, "{{") {
```
sporksmith, line 75: I think you can use the `?` operator to bubble up formatter errors. e.g. `write!(f, "{{")?`

src/main/utility/counter.rs
```diff
@@ -24,45 +25,76 @@ pub struct Counter {
...
+            Ok(_) => (),
+            Err(e) => return Err(e),
         }
-        string
+
+        for i in 0..len {
+            if i == (len - 1) {
```
sporksmith, line 81: Would be a little nicer to dedupe by unconditionally writing the item, and then conditionally writing the `,`

sporksmith (approved):

src/main/utility/counter.rs
robgjansen: I was going to do that, but that takes 2 writes for all except the last item instead of one. Or maybe I don't have to worry about write overhead because the compiler will optimize it anyway?

src/main/utility/counter.rs
robgjansen: I knew there had to be a cleaner way - thanks!

src/main/utility/counter.rs
robgjansen: OK, done now.

PR #1210, opened 2021-03-23
Remove the default value for '--data-template' and stop auto-deleting the data directory

### LINKED ISSUE #1191: Remove the default value for '--data-template', stop auto-deleting `--data-directory`
Shadow's `--data-template` option copies a template directory to the data directory before running an experiment. By default, this is set to `shadow.data.template`. If someone's not familiar with this feature of Shadow, this default behaviour might not be very intuitive. Instead we should not set a default data template directory so that the user must set it explicitly.

For backwards-compatibility reasons, if the `--data-template` option was not set but a `shadow.data.template` directory exists, we should log a warning to let the user know that this default behaviour changed.

-------

We discussed removing this option entirely and requiring the user to copy the template directory manually before running the experiment, and also to not automatically delete the `shadow.data` directory before running the experiment, since this could be considered outside of Shadow's responsibility. But there are some other problems to consider. For example, someone could run an experiment, and then immediately run it again, but since the files from the previous experiment would still exist, the results from the second experiment might be different which could be confusing to the user. It's not immediately obvious which approach is best.

stevenengler:
Closes #1191.

### DIFF
--- a/src/main/core/controller.c
+++ b/src/main/core/controller.c
@@ -465,6 +465,10 @@ gint controller_run(Controller* controller) {
                     controller->bootstrapEndTime, managerSeed, element->preloadPath.string->str,
                     element->environment.isSet ? element->environment.string->str : NULL);
 
+    if (controller->manager == NULL) {
+        error("unable to create manager");
+    }
+
     message("registering plugins and hosts");
 
     /* register the components needed by each manager.
--- a/src/main/core/manager.c
+++ b/src/main/core/manager.c
@@ -218,8 +218,7 @@ Manager* manager_new(Controller* controller, Options* options, SimulationTime en
     manager->hostsPath = g_build_filename(manager->dataPath, "hosts", NULL);
 
     if (g_file_test(manager->dataPath, G_FILE_TEST_EXISTS)) {
-        gboolean success = utility_removeAll(manager->dataPath);
-        utility_assert(success);
+        error("data directory '%s' already exists", manager->dataPath);
     }
 
     gchar* templateDataPath = options_getDataTemplatePath(options);
--- a/src/main/core/manager.c
+++ b/src/main/core/manager.c
@@ -222,13 +222,15 @@ Manager* manager_new(Controller* controller, Options* options, SimulationTime en
         utility_assert(success);
     }
 
-    gchar* templateDataPath =
-        g_build_filename(manager->cwdPath, options_getDataTemplatePath(options), NULL);
-    if (g_file_test(templateDataPath, G_FILE_TEST_EXISTS)) {
-        gboolean success = utility_copyAll(templateDataPath, manager->dataPath);
-        utility_assert(success);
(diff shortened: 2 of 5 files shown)


src/main/core/manager.c
```diff
@@ -215,17 +215,39 @@ Manager* manager_new(Controller* controller, Options* options, SimulationTime en
...
+        critical("data directory already exists");
+        return NULL;
     }
 
-    gchar* templateDataPath =
-        g_build_filename(manager->cwdPath, options_getDataTemplatePath(options), NULL);
-    if (g_file_test(templateDataPath, G_FILE_TEST_EXISTS)) {
-        gboolean success = utility_copyAll(templateDataPath, manager->dataPath);
-        utility_assert(success);
+    const gchar* templateDataPath = options_getDataTemplatePath(options);
```
sporksmith, line 222: Might be worth breaking this out into a helper function. If so you can use early returns to unnest some of this logic. e.g.

```
if (!g_file_test(absTemplatePath, G_FILE_TEST_EXISTS) {
  critical("...")
  return
}
...
```
(Or if we upgrade these to `error` we can unnest either way)

src/main/core/controller.c
```diff
@@ -465,6 +465,10 @@ gint controller_run(Controller* controller) {
                     controller->bootstrapEndTime, managerSeed, element->preloadPath.string->str,
                     element->environment.isSet ? element->environment.string->str : NULL);
 
+    if (controller->manager == NULL) {
```
sporksmith, line 468: It's a bit surprising to have to check a `new` function for a NULL return. Since this is the only call site it's probably not that big of a deal in practice, but for an error that's ultimately going to result in exiting the simulation we might as well just `error` at that point rather than bubbling errors back up.

If we do want to make such a method fallible, we should probably call it something else; maybe `tryNew`

sporksmith (changes requested):

src/main/core/manager.c
```diff
@@ -215,17 +215,39 @@ Manager* manager_new(Controller* controller, Options* options, SimulationTime en
     manager->hostsPath = g_build_filename(manager->dataPath, "hosts", NULL);
 
     if (g_file_test(manager->dataPath, G_FILE_TEST_EXISTS)) {
-        gboolean success = utility_removeAll(manager->dataPath);
-        utility_assert(success);
+        critical("data directory already exists");
```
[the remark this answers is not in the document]
stevenengler: Added the path

src/main/core/manager.c
stevenengler: I made these return error, and now they're not longer nested.

src/main/core/controller.c
stevenengler: I made all of the new code call error() instead of returning NULL, but I left the NULL check here since `manager_new()` can already return NULL from the existing code (since only a single manager can exist at a time), so I think it's still good to have this check.

sporksmith (approved):

PR #1214, opened 2021-03-24
work stealing sched: Fix swapped parameter order setting cpu affinity

sporksmith:
This fixes a bug affecting performance when using the work stealing scheduler with cpu pinning.


robgjansen (approved):
🤯

robgjansen:
Oh, does this bug exists on the main branch too?

sporksmith:
> Oh, does this bug exist on the main branch too? If so, should we merge it to main, and then merge main to dev?

Nope - this is where the Thread base code is setting the plugin process's affinity to match the worker. Not applicably on main since the plugin executes on the worker thread :)

PR #1219, opened 2021-03-25
Make shadow logger registration thread-safe

### LINKED ISSUE #1218: fork-proxy threads self-register using non-thread-safe shadow_logger_register
I believe this is the cause of some of the non-deterministic failures of multi-worker tests. Based on inspecting a core dump after such a failure, the logger's hash table seems to be getting corrupted.

While we could protect the hash table with a mutex, this would cause contention between multiple threads trying to log that the shadow logger is specifically meant to avoid.

I think we want to change the API so that threads always self-register, and the logger stores its thread-specific metadata in thread-local storage instead of a hash table.

```
Core was generated by `/home/jnewsome/projects/shadow/build/src/main/shadow --data-directory=determini'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00000000006290d5 in shadow_logger_logVA (logger=0xeab600, level=LOGLEVEL_INFO, fileName=0xa6e9ef "/home/jnewsome/projects/shadow/src/main/host/thread_ptrace.c", 
    functionName=0xa6ed37 "_threadptrace_fork_exec", lineNumber=390, format=0xa6e75e "started process %s with PID %d", vargs=0x7f07323996f0)
    at /home/jnewsome/projects/shadow/src/main/core/logger/shadow_logger.c:176
176	    MAGIC_ASSERT(threadData);
[Current thread is 1 (Thread 0x7f073239a700 (LWP 1089397))]
warning: Missing auto-load script at offset 0 in section .debug_gdb_scripts
of file /home/jnewsome/.shadow/bin/shadow.
Use `info auto-load python-scripts [REGEXP]' to list them.
(gdb) bt
#0  0x00000000006290d5 in shadow_logger_logVA (logger=0xeab600, level=LOGLEVEL_INFO, fileName=0xa6e9ef "/home/jnewsome/projects/shadow/src/main/host/thread_ptrace.c", 
    functionName=0xa6ed37 "_threadptrace_fork_exec", lineNumber=390, format=0xa6e75e "started process %s with PID %d", vargs=0x7f07323996f0)
    at /home/jnewsome/projects/shadow/src/main/core/logger/shadow_logger.c:176
#1  0x00000000006299e6 in _shadow_logger_log_cb (logger=0xeab600, level=LOGLEVEL_INFO, fileName=0xa6e9ef "/home/jnewsome/projects/shadow/src/main/host/thread_ptrace.c", 
    functionName=0xa6ed37 "_threadptrace_fork_exec", lineNumber=390, format=0xa6e75e "started process %s with PID %d", vargs=0x7f07323996f0)
    at /home/jnewsome/projects/shadow/src/main/core/logger/shadow_logger.c:346
#2  0x0000000000a6a2a9 in logger_log (logger=0xeab600, level=LOGLEVEL_INFO, fileName=0xa6e9ef "/home/jnewsome/projects/shadow/src/main/host/thread_ptrace.c", 
    functionName=0xa6ed37 "_threadptrace_fork_exec", lineNumber=390, format=0xa6e75e "started process %s with PID %d") at /home/jnewsome/projects/shadow/src/support/logger/logger.c:139
#3  0x000000000060f4de in _threadptrace_fork_exec (file=0xebf6d0 "/home/jnewsome/projects/shadow/build/src/test/determinism/../phold/test-phold", argv=0xeb3520, envp=0x7f0728008600, 
    workingDir=0xec03b0 "/home/jnewsome/projects/shadow/build/src/test/determinism/determinism2a-shadow-ptrace.data/hosts/peer1") at /home/jnewsome/projects/shadow/src/main/host/thread_ptrace.c:390
#4  0x0000000000625d4f in forkproxy_fn (void_forkproxy=0x7f0728006e60) at /home/jnewsome/projects/shadow/src/main/utility/fork_proxy.c:55
#5  0x00007f073743e609 in start_thread (arg=<optimized out>) at pthread_create.c:477
#6  0x00007f0736c2f293 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
(gdb) print threadData
$1 = (LoggerThreadData *) 0x1c0008d0
(gdb) print *threadData
Cannot access memory at address 0x1c0008d0
```

sporksmith:
Since workers concurrently create ForkProxy threads, we need to be able
to register those threads with the logger in a thread-safe way. This
adds a GRWLock to protect the shadow logger's hash table of
thread-specific data.

I considered doing away with the hash table entirely and using
thread-local storage, but then we wouldn't have a way to flush all of
the thread buffers at shutdown.

Fixes #1218


robgjansen:
The RW lock is fine, but note that I ended up having threads only flush their own log buffers in 9f6b6e519177c06df3e2275d7607f2830df1b1c1 so I think thread-local storage might have worked out OK.

sporksmith:
Ah, I'd misread the [ThreadData destructor](https://github.com/shadow/shadow/blob/a2473facca9586208f173a18d5d94641ba0941ca/src/main/core/logger/shadow_logger.c#L73); I thought it was flushing any remaining buffers, but yeah it's just freeing them. That does mean any remaining buffers wouldn't be freed at shutdown, but it shouldn't matter in practice nor be considered a leak by tooling since they're still reachable from a statically allocated pointer.

I'm inclined to leave as-is, but yeah looks like either way should work.

stevenengler (approved):
If there's still a concern about the performance, maybe it would be better to just not log anything in the forkproxy thread, or return the error back from `forkproxy_forkExec()` since almost all logging in `forkproxy_fn()` and `_threadptrace_fork_exec` are `error()` calls anyways.

PR #1224, opened 2021-03-29
Handle hot-path time syscalls in the shim

### LINKED ISSUE #1141: Significant time spent in g_get_monotonic_time

robgjansen:
Shadow stores the latest sim time in memory shared with the shim. When time-related syscalls are executed, the shim side can read the latest simulation time directly from shared memory and avoid the more expensive inter-process syscall serviced by shadow.

Fixes #1141

### DIFF
--- a/src/shim/shim_syscall.c
+++ b/src/shim/shim_syscall.c
@@ -11,24 +11,22 @@
 #include <sys/time.h>
 #include <time.h>
 
+#include "main/core/support/definitions.h" // for SIMTIME definitions
 #include "shim/shim.h"
 #include "shim/shim_syscall.h"
 #include "support/logger/logger.h"
-
-#define SIMTIME_NANOS_PER_SEC 1000000000l
-#define SIMTIME_MICROS_PER_NANOSEC 1000l
 
 // We store the simulation time using timespec to reduce the number of
 // conversions that we need to do while servicing syscalls.
 static struct timespec _cached_simulation_time = {0};
 
 void shim_syscall_set_simtime_nanos(uint64_t simulation_nanos) {
-    _cached_simulation_time.tv_sec = simulation_nanos / SIMTIME_NANOS_PER_SEC;
-    _cached_simulation_time.tv_nsec = simulation_nanos % SIMTIME_NANOS_PER_SEC;
+    _cached_simulation_time.tv_sec = simulation_nanos / SIMTIME_ONE_SECOND;
+    _cached_simulation_time.tv_nsec = simulation_nanos % SIMTIME_ONE_SECOND;
 }
 
 uint64_t shim_syscall_get_simtime_nanos() {
-    return (uint64_t)(_cached_simulation_time.tv_sec * SIMTIME_NANOS_PER_SEC) +
+    return (uint64_t)(_cached_simulation_time.tv_sec * SIMTIME_ONE_SECOND) +
            _cached_simulation_time.tv_nsec;
 }
 
@@ -118,7 +116,7 @@ bool shim_syscall(long syscall_num, long* rv, va_list args) {
 
             if (tp) {
                 tp->tv_sec = simtime_ts->tv_sec;
-                tp->tv_usec = simtime_ts->tv_nsec / SIMTIME_MICROS_PER_NANOSEC;
+                tp->tv_usec = simtime_ts->tv_nsec / SIMTIME_ONE_MICROSECOND;
                 debug("gettimeofday() successfully copied time");
             }
             *rv = 0;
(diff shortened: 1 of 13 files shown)


src/shim/shim_syscall.c
```diff
@@ -0,0 +1,134 @@
...
+#include <stdbool.h>
+#include <sys/syscall.h>
+#include <sys/time.h>
+#include <time.h>
+
+#include "shim/shim.h"
+#include "shim/shim_syscall.h"
+#include "support/logger/logger.h"
+
+#define SIMTIME_NANOS_PER_SEC 1000000000l
```
sporksmith, line 18: Why the ending 1's? And should we move these constants to a header shared between shadow and the shim to ensure they stay sync'd?

src/shim/shim_syscall.c
```diff
@@ -0,0 +1,134 @@
...
+    _cached_simulation_time.tv_sec = simulation_nanos / SIMTIME_NANOS_PER_SEC;
+    _cached_simulation_time.tv_nsec = simulation_nanos % SIMTIME_NANOS_PER_SEC;
+}
+
+uint64_t shim_syscall_get_simtime_nanos() {
+    return (uint64_t)(_cached_simulation_time.tv_sec * SIMTIME_NANOS_PER_SEC) +
+           _cached_simulation_time.tv_nsec;
+}
+
+bool shim_syscall_is_supported(long syscall_num) {
```
sporksmith, line 35: Instead of trying to keep this in sync with which syscalls are supported in `shim_syscall`, could the caller just try calling `shim_syscall` and fallback if it returns false?

sporksmith (approved):
Looks great!

Might we want a command-line flag to disable this for benchmarking/debugging purposes?

src/shim/shim_syscall.c
robgjansen: It's an `l` (a lower case `L`). I think I refactored this from a previously defined constant to this `#define`, but yeah we should use an existing definition or create one (e.g. in `definitions.h`).

src/shim/shim_syscall.c
robgjansen: Yep, that was one wrinkle I noticed but left in place. But I agree and I'll change it.

PR #1225, opened 2021-03-29
GitHub workflow maintenance

stevenengler:
Removed an old commented workflow, renamed some workflows/jobs, and moved the coverage test to its own workflow.


robgjansen:
@sporksmith what do you think about leaving the name empty (as it is now) vs. setting it to “tor” and “shadow”?

It will look like either, e.g.:
- `Shadow Tests / (fedora:32, clang, release) (pull_request)`
or
- `Shadow Tests / shadow (fedora:32, clang, release) (pull_request)`

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,73 @@
...
+    branches: [dev]
+  pull_request:
+    types: [opened, synchronize]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  codecov:
+    runs-on: ubuntu-latest
```
sporksmith, line 21: Just a thought - I think ubuntu 20 is available "natively" now, and is actually the new default? So hypothetically we could specify 20 here and not use a container? Otoh using a container doesn't add much overhead afaik, and would keep it closer to the other workflows.

sporksmith:
> @sporksmith what do you think about leaving the name empty (as it is now) vs. setting it to “tor” and “shadow”?

Either way is fine with me :)

.github/workflows/coverage.yml
stevenengler: I think the coverage test requires 20.10 for clang 11, but GitHub only has LTS releases: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

robgjansen (approved):

PR #1229, opened 2021-03-30
Extensions for counters for tracking global syscall counts

robgjansen:
With these changes, it will be easier to profile hot-path syscalls - the global totals for each syscall will be logged at the end of the simulation.

### DIFF
--- a/src/main/core/manager.c
+++ b/src/main/core/manager.c
@@ -11,6 +11,7 @@
 #include <stddef.h>
 #include <sys/resource.h>
 
+#include "main/bindings/c/bindings.h"
 #include "main/core/controller.h"
 #include "main/core/logger/shadow_logger.h"
 #include "main/core/manager.h"
@@ -59,6 +60,9 @@ struct _Manager {
 
     /* global object counters, we collect counts from workers at end of sim */
     ObjectCounter* objectCounts;
+
+    // Global syscall counter, we collect counts from workers at end of sim
+    Counter* syscall_counter;
 
     /* the parallel event/host/thread scheduler */
     Scheduler* scheduler;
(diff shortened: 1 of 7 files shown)


src/main/utility/counter.rs
```diff
@@ -80,6 +151,24 @@ impl Display for Counter {
...
+}
+
+impl Sub for Counter {
+    type Output = Self;
+    /// Combines two counters by subtracting all values for all keys of `other` from `self`.
+    fn sub(mut self, other: Self) -> Self {
+        self.sub_counter(&other);
+        self
+    }
+}
```
stevenengler, line 170: Does it make sense to subtract counters? For example if counter `a` has keys that counter `b` doesn't, then I think `assert_eq!(b-a+a, b)` would fail, which might be unexpected.

stevenengler (approved):

src/main/core/manager.c
```diff
@@ -643,6 +654,24 @@ void manager_countObject(ObjectType otype, CounterType ctype) {
...
 }
 
+void manager_add_syscall_counts(Manager* manager, Counter* syscall_counts) {
+    MAGIC_ASSERT(manager);
+    _manager_lock(manager);
+    // This is created on the fly, so that if we did not enable counting mode
+    // then we don't need to create the counter object.
+    if (!manager->syscall_counter) {
+        manager->syscall_counter = counter_new();
+    }
```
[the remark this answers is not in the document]
robgjansen: There is some logic that depends on whether or not the C pointer is NULL - when we disable object counting, we want the C pointer to be NULL so we don't try to aggregate it at the end of the sim.

src/main/utility/counter.rs
robgjansen: I think subtracting makes sense, but subtracting 1-5 to equal 0 is weird. I'll change the value type to `i64` instead of `u64` to support negative values too.

PR #1231, opened 2021-03-30
Merge main

sporksmith:


robgjansen (approved):
So we're adding a bunch of commits to dev that originally were committed to dev, but now they'll appear in the history twice?

sporksmith:
> So we're adding a bunch of commits to dev that originally were committed to dev, but now they'll appear in the history twice?

Yeah, it's a little unfortunate, but I don't see a better way to proceed. We would've ended up with a cleaner history if we did the rename in `main` first and then merged into `dev` but ¯\_(ツ)_/¯

PR #1235, opened 2021-03-30
Use new Rust coverage instrumentation

stevenengler:
This is just a test


stevenengler:
Closing for now since this may have worked, but because of #1236, we aren't getting code coverage metrics on Rust code so I can't tell if this PR is working or not.

PR #1239, opened 2021-03-30
Rust object counters

robgjansen:
Use rust counter to count objects and remove c counter.


src/main/utility/counter.rs
```diff
@@ -235,6 +240,17 @@ mod export {
         counter.sub_counter(other)
     }
 
+    #[no_mangle]
+    pub extern "C" fn counter_equals_counter(counter: *mut Counter, other: *mut Counter) -> bool {
+        assert!(!counter.is_null());
+        assert!(!other.is_null());
```
stevenengler, line 246: I think these can just be const pointers.

src/main/utility/counter.rs
```diff
@@ -422,6 +438,25 @@ mod tests {
         assert_eq!(counter_a, counter_sum);
     }
 
+    #[test]
+    fn test_counter_equality() {
+        let mut counter_a = Counter::new();
+        counter_a.set_value("read", 1);
+        counter_a.set_value("write", 2);
```
stevenengler, line 445: Might also be good to add a test case with a value at 0. For example:

```rust
let mut counter_d = Counter::new();
counter_d.set_value("read", 1);
counter_d.set_value("write", 2);
counter_d.set_value("open", 1);
counter_d.sub_value("open", 1);

assert_eq!(counter_a, counter_d);
```

stevenengler (approved):
Looks good!

PR #1242, opened 2021-03-31
thread_ptrace: Make shutdown more robust

sporksmith:
In particular this fixes the deadlock at end of shutdown that happens in
tor simulations in the max-concurrency branch (not yet merged) when
using more workers than the max-concurrency. I don't think it's directly
related to that, but I haven't been able to reproduce it in any other
scenario.

### DIFF
--- a/src/main/host/thread_ptrace.c
+++ b/src/main/host/thread_ptrace.c
@@ -531,7 +531,9 @@ static void _threadptrace_enterStateSignalled(ThreadPtrace* thread,
 }
 
 static void _threadptrace_enterStateExited(ThreadPtrace* thread) {
-    debug("enterStateExited");
+    debug("enterStateExited for thread %d.%d native %d.%d", thread_getProcessId(&thread->base),
+          thread_getID(&thread->base), thread_getNativePid(&thread->base),
+          thread_getNativeTid(&thread->base));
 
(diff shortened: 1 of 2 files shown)


src/main/host/thread_ptrace.c
```diff
@@ -142,7 +142,7 @@ static StopReason _getStopReason(int wstatus) {
             .type = STOPREASON_EXITED_NORMAL,
             .exited_normal.exit_code = WEXITSTATUS(wstatus),
         };
-        debug("STOPREASON_EXITED_NORMAL: %d", rv.exited_normal.exit_code);
+        warning("STOPREASON_EXITED_NORMAL: %d", rv.exited_normal.exit_code);
```
robgjansen, line 145: Why is it a warning if the process exited normally?

robgjansen (approved):
👍

src/main/host/thread_ptrace.c
sporksmith: Catching this stop-reason reaps the child, which means reaping the child will fail in Process. Instead we want to detach on the ptrace exit event and let Process get the "normal" exit.

We already have a warning in `_threadptrace_updateChildState` though, so I reverted it. I'd bumped it up to a warning while debugging.

PR #1245, opened 2021-04-01
Add a shadow-specific name lookup syscall

### LINKED ISSUE #1094: Count number of getaddrinfo calls and log at exit
Our `getaddrinfo` handler in `syscall_libraries.c` currently scans `/etc/hosts` on every call, which turns out to be a lot of overhead in our phold tests when there are lots of hosts. We can work around this in phold by using static IP addresses instead of doing `getaddrinfo` lookups, but this overhead could be an issue with other applications like Tor so we want to make sure we notice if it becomes an issue again.

We want to count the number of `getaddrinfo` calls that are being handled in `syscall_libraries.c`, so we can compute the fraction relative to the total number of syscalls being handled in `syscall_handler.c`. The count should be logged either periodically (eg once per second) or at the end of a simulation.

robgjansen:
Added a shadow-specific syscall to perform hostname to address lookups. The new syscall takes precedence over scanning the `/etc/hosts` file. Scanning the hosts file is still done as a fallback in case the new syscall fails.

This was done primarily to support phold experiments that make lots of `getaddrinfo` calls where we need the work performed to be constant no matter the number of hosts we run in the experiment. 

Also refactored the shadow-specific syscall handling code into a single module, so that the syscalls go through the syscall handler. This means we can count the frequency with which shadow-specific syscalls are made (using the `--enable-syscall-counters` option).

Closes #1094


src/shim/preload_libraries.c
```diff
@@ -419,8 +450,14 @@ int getaddrinfo(const char* node, const char* service,
...
     if (add_ipv4) {
-        _getaddrinfo_add_matching_hosts_ipv4(
-            res, &tail, node, add_tcp, add_udp, add_raw, port);
+        // Try a Shadow syscall first to avoid scanning the /etc/hosts file.
+        uint32_t addr;
+        if (_syscall_hostname_to_addr_ipv4(node, &addr)) {
+            _getaddrinfo_appendv4(res, &tail, add_tcp, add_udp, add_raw, addr, port);
+        } else {
+            // Fall back to scanning /etc/hosts.
+            _getaddrinfo_add_matching_hosts_ipv4(res, &tail, node, add_tcp, add_udp, add_raw, port);
```
sporksmith, line 459: Do we expect to have to fall back to the hosts file in any "normal" case, or is this an attempt to recover from a bug in shadow? Should there be a warning log entry here?

sporksmith (approved):
Cool! And nice cleanup :)

src/shim/preload_libraries.c
robgjansen: I've never seen the new syscall fail, I just didn't want to remove the /etc/hosts scanning functionality yet. I'll add a warning, I think that's appropriate.

PR #1255, opened 2021-04-02
Memory manager usage

sporksmith:
I experimentally verified that when mmap'ing anonymous regions, the MemoryManager was effectively pre-allocating the space in /dev/shm. i.e. the allocation showed up in `df` and in the shared section of `free`. It turns out that posix_fallocate and fallocate pre-reserve the requested space. ftruncate OTOH appaars to provide the behavior we want of making the given region accessible, but only actually allocating space for pages that are written to.


robgjansen (approved):
Looks great!

PR #1257, opened 2021-04-03
Fix descriptor bugs and support dup() on regular files for Python

stevenengler:
Fixes a few descriptor bugs and implements `dup()` for regular files, which is required for Python. (Python uses `dup()` on stdin, stdout, and stderr to check if they are valid descriptors. See [`create_stdio()` in CPython](https://github.com/python/cpython/blob/442ad74fc2928b095760eb89aba93c28eab17f9b/Python/pylifecycle.c#L2108).)

Shadow is still missing syscalls to use things like the built-in http server (I see futex errors, missing socket options, and IPv6 issues). Python also make heavy use of `select()`, so things like [`time.sleep()`](https://github.com/python/cpython/blob/b2a91e0c9ee18b50cc86b21211c2258520a9f5d0/Modules/timemodule.c#L360) don't work in Shadow.

```
options:
  stoptime: 10
topology:
- graphml: |
    <graphml xmlns="http://graphml.graphdrawing.org/xmlns" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://graphml.graphdrawing.org/xmlns http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd">
        <node id="poi-1">
          <data key="d0">US</data>
          <data key="d1">10240</data>
          <data key="d2">10240</data>
        </node>
          <data key="d3">50.0</data>
          <data key="d4">0.0</data>
        </edge>
      </graph>
    </graphml>
plugins:
- id: python
  path: /usr/bin/python3
hosts:
- id: test
  quantity: 1
  processes:
  - plugin: python
    starttime: 1
    arguments: -c print(1234)
```

### DIFF
--- a/src/main/host/descriptor/file.c
+++ b/src/main/host/descriptor/file.c
@@ -149,6 +149,38 @@ File* file_new() {
     return file;
 }
 
+File* file_dup(File* file, int* dupError) {
+    MAGIC_ASSERT(file);
+
+    int newFd;
+
+    // only dup the os fd if it's valid
+    if (file->osfile.fd >= 0) {
+        newFd = dup(file->osfile.fd);
+
+        if (newFd < 0) {
+            *dupError = errno;
+            return NULL;
+        }
+    } else {
+        newFd = file->osfile.fd;
+    }
+
+    File* newFile = file_new();
+
+    newFile->type = file->type;
+
+    newFile->osfile.fd = newFd;
+    newFile->osfile.flags = file->osfile.flags;
+    newFile->osfile.mode = file->osfile.mode;
+    newFile->osfile.abspath = strdup(file->osfile.abspath);
+
+    // CLOEXEC is a descriptor flag and it is not copied during a dup()
+    newFile->osfile.flags &= ~O_CLOEXEC;
+
+    return newFile;
+}
+
 static char* _file_getConcatStr(const char* prefix, const char sep, const char* suffix) {
     char* path = NULL;
     if (asprintf(&path, "%s%c%s", prefix, sep, suffix) < 0) {
--- a/src/main/host/descriptor/file.c
+++ b/src/main/host/descriptor/file.c
@@ -235,6 +235,11 @@ int file_openat(File* file, File* dir, const char* pathname, int flags, mode_t m
     }
 #endif
 
+    int fd = _file_getFD(file);
+    if (fd < 0) {
+        error("Cannot openat() on an unregistered descriptor object with fd %d", fd);
+    }
+
     /* The default case is a regular file. We do this first so that we have
      * an absolute path to compare for special files. */
     char* abspath = _file_getPath(file, dir, pathname, workingDir);
@@ -292,7 +297,8 @@ int file_openat(File* file, File* dir, const char* pathname, int flags, mode_t m
     /* The os-backed file is now ready. */
     descriptor_adjustStatus(&file->super, STATUS_DESCRIPTOR_ACTIVE, TRUE);
 
-    return _file_getFD(file);
+    /* We checked above that fd is non-negative. */
+    return fd;
 }
 
 int file_open(File* file, const char* pathname, int flags, mode_t mode, const char* workingDir) {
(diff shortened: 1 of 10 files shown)


src/main/host/descriptor/file.c
```diff
@@ -292,7 +292,7 @@ int file_openat(File* file, File* dir, const char* pathname, int flags, mode_t m
     /* The os-backed file is now ready. */
     descriptor_adjustStatus(&file->super, STATUS_DESCRIPTOR_ACTIVE, TRUE);
 
-    return _file_getFD(file);
+    return 0;
```
robgjansen, line 327: Add a comment here, or document in the header file what `file_open` and `file_openat` are expected to return. (It's useful since the return logic is slightly different from `open` and `openat`.)

src/main/host/syscall/unistd.c
```diff
@@ -297,9 +297,27 @@ SysCallReturn syscallhandler_close(SysCallHandler* sys,
...
 
 SysCallReturn syscallhandler_dup(SysCallHandler* sys,
-                                   const SysCallArgs* args) {
-    warning("Cannot dup legacy descriptors");
-    return (SysCallReturn){.state = SYSCALL_DONE, .retval.as_i64 = -EOPNOTSUPP};
+                                 const SysCallArgs* args) {
+    gint fd = args->args[0].as_i64;
+    LegacyDescriptor* desc = process_getRegisteredLegacyDescriptor(sys->process, fd);
+
+    LegacyDescriptorType dType = descriptor_getType(desc);
```
robgjansen, line 304: What if the plugin passes in a bad FD? I think `desc` will be NULL then, and `descriptor_getType` will fail the magic assertion.

Instead, do we want to return `-EBADF` if `desc` is NULL?

For example, `_syscallhandler_readHelper` does this:

```c
    /* Get the descriptor. */
    LegacyDescriptor* desc = process_getRegisteredLegacyDescriptor(sys->process, fd);
    if (!desc) {
        return (SysCallReturn){.state = SYSCALL_DONE, .retval.as_i64 = -EBADF};
    }
```

src/main/host/syscall/unistd.c
```diff
@@ -297,9 +297,27 @@ SysCallReturn syscallhandler_close(SysCallHandler* sys,
 }
 
 SysCallReturn syscallhandler_dup(SysCallHandler* sys,
-                                   const SysCallArgs* args) {
-    warning("Cannot dup legacy descriptors");
-    return (SysCallReturn){.state = SYSCALL_DONE, .retval.as_i64 = -EOPNOTSUPP};
+                                 const SysCallArgs* args) {
+    gint fd = args->args[0].as_i64;
```
robgjansen, line 301: I generally like to add a `debug` message right after setting the fd in this line, to log the fd that we got on the syscall, before doing other stuff that may cause problems.

src/test/file/test_file.c
```diff
@@ -259,6 +259,17 @@ static void _test_tmpfile() {
     assert_nonneg_errno(fclose(file));
 }
 
+static void _test_dup() {
+    g_auto(AutoDeleteFile) adf = _create_auto_file();
+    int fd, fd2;
+
+    assert_nonneg_errno(fd = open(adf.name, O_RDONLY));
+    assert_nonneg_errno(fd2 = dup(fd));
```
robgjansen, line 267: Do we want to expand this test, to make sure that something we wrote to `fd` _before_ the dup is actually readable from `fd2` _after_ the dup?

robgjansen (changes requested):
Thanks for organizing the commits - that made it easy to review :)

Requesting changes, primarily to address the NULL pointer issue.

src/main/host/descriptor/file.c
stevenengler: I changed the return value back, but made it an error to call `file_openat()` if you haven't registered the descriptor yet.

src/main/host/syscall/unistd.c
stevenengler: Thanks, I forgot to check this.

src/main/host/syscall/unistd.c
stevenengler: Added.

src/test/file/test_file.c
stevenengler: I added more to this test.

robgjansen (approved):
Looks great, thanks!

Issue #1220, opened 2021-03-26
Organize and remove legacy options


robgjansen:
We've collected a set of options that may be rarely used. As we are migrating to a new option and config format for Shadow 2.0, we are going to remove legacy options that we don't think are that useful for most users.

Here are the legacy options I am pretty confident we can remove:

```
Simulation Examples
  --test                                  Run basic benchmark tests
  --tgen                                  PLACEHOLDER - Run basic data transfer simulation

System Options
  --cpu-precision=TIME                    round measured CPU delays to the nearest TIME, in microseconds (negative value to disable fuzzy CPU delays) [200]
  --cpu-threshold=TIME                    TIME delay threshold after which the CPU becomes blocked, in microseconds (negative value to disable CPU delays) (experimental!) [-1]
  --tcp-ssthresh=N                        Set TCP ssthresh value instead of discovering it via packet loss or hystart [0]
  --tcp-windows=N                         Initialize the TCP send, receive, and congestion windows to N packets [10]

Application Options:
  -p, --preload=VALUE                     LD_PRELOAD environment VALUE to use for function interposition (/path/to/lib:...) [None]
```

I'm less confident about these, and we might want to keep them around. How about we move them to the "Experimental" section?

```
System Options
  --interface-batch=TIME                  Batch TIME for network interface sends and receives, in microseconds [5000]
  --interface-buffer=N                    Size of the network interface receive buffer, in bytes [1024000]
  --interface-qdisc=QDISC                 The interface queuing discipline QDISC used to select the next sendable socket ('fifo' or 'rr') ['fifo']
  --socket-recv-buffer=N                  Initialize the socket receive buffer to N bytes [174760]
  --socket-send-buffer=N                  Initialize the socket send buffer to N bytes [131072]
```

These we do want, but I would move them to the "Experimental" section too:

```
Application Options:
  -r, --runahead=TIME                     If set, overrides the automatically calculated minimum TIME workers may run ahead when sending events between nodes, in milliseconds [0]
  -t, --scheduler-policy=SPOL             The event scheduler's policy for thread synchronization ('thread', 'host', 'steal', 'threadXthread', 'threadXhost') ['steal']
  -n, --interpose-method=METHOD           Which interposition method to use ('hybrid', 'preload', 'ptrace') ['ptrace']
  -z, --pin-cpus                          Use experimental CPU pinning
```

I think those options not listed here are still useful to keep.

stevenengler:
[...]

PR #1263, opened 2021-04-05
Remove command-line options and configuration fields

stevenengler:
Removes the following:

Command-line options:
- `--preload`
- `--valgrind`
- `--test`
- `--tgen`
- `--cpu-precision`
- `--cpu-threshold`
- `--tcp-ssthresh`
- `--tcp-windows`

Configuration fields:
- `cpufrequency`

Part of #1220.


src/main/core/controller.c
```diff
@@ -348,20 +348,16 @@ static void _controller_registerHostCallback(ConfigurationHostElement* he, Contr
...
         if (params->cpuFrequency == 0) {
             params->cpuFrequency = 2500000; // 2.5 GHz
             debug("both configured and raw manager cpu frequencies unavailable, using 2500000 KHz");
         }
 
-        gint defaultCPUThreshold = options_getCPUThreshold(controller->options);
-        params->cpuThreshold = defaultCPUThreshold > 0 ? defaultCPUThreshold : 0;
-        gint defaultCPUPrecision = options_getCPUPrecision(controller->options);
-        params->cpuPrecision = defaultCPUPrecision > 0 ? defaultCPUPrecision : 0;
+        params->cpuThreshold = 0;
```
robgjansen, line 353: Should this be 0 or -1? It looks like -1 was the default before to disable cpu delays?

src/main/core/support/options.c
```diff
@@ -172,8 +169,6 @@ Options* options_new(gint argc, gchar* argv[]) {
       { "socket-recv-buffer", 0, 0, G_OPTION_ARG_INT, &(options->initialSocketReceiveBufferSize), sockrecv->str, "N" },
       { "socket-send-buffer", 0, 0, G_OPTION_ARG_INT, &(options->initialSocketSendBufferSize), socksend->str, "N" },
       { "tcp-congestion-control", 0, 0, G_OPTION_ARG_STRING, &(options->tcpCongestionControl), "Congestion control algorithm to use for TCP ('aimd', 'reno', 'cubic') ['reno']", "TCPCC" },
-      { "tcp-ssthresh", 0, 0, G_OPTION_ARG_INT, &(options->tcpSlowStartThreshold), "Set TCP ssthresh value instead of discovering it via packet loss or hystart [0]", "N" },
-      { "tcp-windows", 0, 0, G_OPTION_ARG_INT, &(options->initialTCPWindow), "Initialize the TCP send, receive, and congestion windows to N packets [10]", "N" },
```
robgjansen, line 182: The [getting started tutorial doc](https://github.com/shadow/shadow/blob/dev/docs/2-Getting-Started-Tutorial.md#example-experiment) has an example that tweaks the `--tcp-windows` option. We should update that example experiment so it no longer references a non-existent option.

src/main/core/support/options.c
```diff
@@ -385,11 +378,6 @@ gboolean options_shouldExitAfterShmCleanup(Options* options) {
     return options->shouldExitAfterShmCleanup;
 }
 
-const gchar* options_getPreloadString(Options* options) {
```
robgjansen, line 438: Huh, this wasn't called anywhere?

robgjansen (approved):
Always happy to see a bunch of old code removed :)

src/main/core/controller.c
stevenengler: The `params->cpuThreshold = defaultCPUThreshold > 0 ? defaultCPUThreshold : 0` line sets a lower limit of 0, so the -1 would end up as 0 here.

src/main/core/support/options.c
stevenengler: Nope, it looks like all uses were removed in the initial e41c4a997.

src/main/core/support/options.c
stevenengler: I updated this example to use different queuing disciplines instead. Let me know if you think there's a better option to use.

PR #1266, opened 2021-04-06
better parameters handling - handles quotes as well

### LINKED ISSUE #1265: Process configuration arguments enclosed with double quotes (") are not properly handled
**Describe the issue**
Some plugin may require parameters that contain a space and that need to be enclosed within double quotes (").
At the moment, the process.c module doesn't handle such special case because it just breaks down the string by spaces, resulting in a wrong argv vector passed to the plugin.

**To Reproduce**
- Extract shadowtor-minimal-config from shadow-plugin-tor.
- Remove the following two lines from `conf/tor.hiddenserver.torrc`:
```
HiddenServiceDir shadow.data/hosts/hiddenserver/hs
HiddenServicePort 80 127.0.0.1:8080
```
- Update shadow.config.xml and add those two parameters `--HiddenServiceDir shadow.data/hosts/hiddenserver/hs --HiddenServicePort &quot;80 127.0.0.1:8080&quot;` to the tor plugin arguments as follows (notice how the quotes `"` have been encoded to `&quot;` so that they can be embedded within the xml attribute `arguments`):
```
    <host id="hiddenserver" bandwidthdown="102400" bandwidthup="102400" >
        <process plugin="tgen" starttime="1" arguments="conf/tgen.hiddenserver.graphml.xml" />
        <process plugin="tor" preload="tor-preload" starttime="900" arguments="--Address ${NODEID} --Nickname ${NODEID} --DataDirectory shadow.data/hosts/${NODEID} --GeoIPFile ~/.shadow/share/geoip --defaults-torrc conf/tor.common.torrc -f conf/tor.hiddenserver.torrc --BandwidthRate 1024000 --BandwidthBurst 1024000 --HiddenServiceDir shadow.data/hosts/hiddenserver/hs --HiddenServicePort &quot;80 127.0.0.1:8080&quot;" />
        <process plugin="torctl" starttime="901" arguments="localhost 9051 STREAM,CIRC,CIRC_MINOR,ORCONN,BW,STREAM_BW,CIRC_BW,CONN_BW"/>
    </host>
```
- Run the simulation and observe that it fails with `** Stopping Shadow, returning code -1 (error)`.
- Open the `shadow.data/hosts/hiddenserver/stdout-hiddenserver.tor.1001.log` log file, and observe the following error:
```Jan 01 00:15:00.000 [warn] Command-line option '127.0.0.1:8080"' with no value. Failing.```

This is because the string `--HiddenServicePort &quot;80 127.0.0.1:8080&quot;` is split into three tokens, i.e.
```
--HiddenServicePort
"80
127.0.0.1:8080"
```
while it should be split into two tokens (see https://www.gnu.org/software/bash/manual/bash.html#Double-Quotes) as well having the quotes stripped:
```
--HiddenServicePort
80 127.0.0.1:8080
```

 - OS and version: post the output of `lsb_release -d`  
`Description:    Ubuntu 18.04.5 LTS`
 - Kernel version: post the output of `uname -a`  
`Linux ****** 4.15.0-137-generic #141-Ubuntu SMP Fri Feb 19 13:46:27 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`

 - Version: post the output of `shadow --version`  
`Shadow v1.13.2-696-g1c7c8da0 2021-04-04--17:14:57 running GLib v2.56.4 and IGraph v0.7.1
Shadow was built from branch main on 2021-04-06--10:25:50 in Release mode with compile options: -ggdb;-fno-omit-frame-pointer;-O3 and link options: -Wl,--no-as-needed,-rpath=/home/msimioni/.shadow/lib,-dynamic-linker=/home/******/.shadow/lib/ldso -z lazy
For more information, visit https://shadow.github.io or https://github.com/shadow`
`Shadow-Tor`

**Additional context**

### PR DESCRIPTION
**Pull request by marcosimioni**
closes #1265

Couple of questions:

- rather than manually splitting & parsing, I've used g_shell_parse_argv() here (see https://developer.gnome.org/glib/stable/glib-Shell-related-Utilities.html#g-shell-parse-argv). this will also handle special cases like quoted arguments containing spaces (not not expansions and other special cases, see the referenced docs). do you have any concern?

- should I add a test case? I don't see any test cases for process.c, I'm not sure if you want them or not.

### DIFF
--- a/src/main/host/process.c
+++ b/src/main/host/process.c
@@ -774,10 +774,9 @@ static void _process_handleTimerResult(Process* proc, gdouble elapsedTimeSec) {
 #endif
 
 static gint _process_getArguments(Process* proc, gchar** argvOut[]) {
-    /* first argument is the name of the program */
     const gchar* pluginName = _process_getPluginName(proc);
 
-    /* parse the full argument string into separate strings */
+    /* build the full argument string (with plugin name as first argument) */
     gchar* arguments = NULL;
     if(proc->arguments && proc->arguments->len > 0 && g_ascii_strncasecmp(proc->arguments->str, "\0", (gsize) 1) != 0) {
         arguments = g_strconcat(pluginName, " ", proc->arguments->str, NULL);
@@ -790,6 +789,7 @@ static gint _process_getArguments(Process* proc, gchar** argvOut[]) {
     /* a pointer to an array that holds pointers */
     gchar** argv = NULL;
 
+    /* parse the full argument string into argc and argv */
     GError *error = NULL;
     if (!g_shell_parse_argv(arguments, &argc, &argv, &error)) {
         error("unable to parse arguments for plugin '%s': %s", pluginName, error->message);
--- a/src/main/host/process.c
+++ b/src/main/host/process.c
@@ -784,7 +784,7 @@ static gint _process_getArguments(Process* proc, gchar** argvOut[]) {
 
     /* parse the full argument string into separate strings */
     if(proc->arguments && proc->arguments->len > 0 && g_ascii_strncasecmp(proc->arguments->str, "\0", (gsize) 1) != 0) {
-        arguments = g_strconcat(arguments, " ", proc->arguments->str);
+        arguments = g_strconcat(arguments, " ", proc->arguments->str, NULL);
     }
 
     /* setup for creating new plug-in, i.e. format into argc and argv */
--- a/src/main/host/process.c
+++ b/src/main/host/process.c
@@ -776,15 +776,15 @@ static void _process_handleTimerResult(Process* proc, gdouble elapsedTimeSec) {
 static gint _process_getArguments(Process* proc, gchar** argvOut[]) {
     gchar* threadBuffer;
 
-    gchar* arguments = NULL;
-
     /* first argument is the name of the program */
     const gchar* pluginName = _process_getPluginName(proc);
-    arguments = g_strdup(pluginName);
 
     /* parse the full argument string into separate strings */
+    gchar* arguments = NULL;
     if(proc->arguments && proc->arguments->len > 0 && g_ascii_strncasecmp(proc->arguments->str, "\0", (gsize) 1) != 0) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/main/host/process.c [hunk 1]
```diff
@@ -776,35 +776,30 @@ static void _process_handleTimerResult(Process* proc, gdouble elapsedTimeSec) {
...
-        gchar* argumentString = g_strdup(proc->arguments->str);
-        gchar* token = strtok_r(argumentString, " ", &threadBuffer);
-        while(token != NULL) {
-            gchar* argument = g_strdup((const gchar*) token);
-            g_queue_push_tail(arguments, argument);
-            token = strtok_r(NULL, " ", &threadBuffer);
-        }
-        g_free(argumentString);
+        arguments = g_strconcat(arguments, " ", proc->arguments->str, NULL);
     }
```
robgjansen, line 788: I think this is leaking memory. The pointer returned by `g_strdup` is being overwritten after the `g_strconcat` call and is never `free()`d. How about this instead:

```c
    gchar* arguments = NULL;
    if(proc->arguments && proc->arguments->len > 0 && g_ascii_strncasecmp(proc->arguments->str, "\0", (gsize) 1) != 0) {
        arguments = g_strconcat(pluginName, " ", proc->arguments->str, NULL);
    } else {
        arguments = g_strdup(pluginName);
    }
    ....
    g_free(arguments);
```

robgjansen (changes requested):
This is a great improvement, thanks for the contribution! I'm requesting changes to fix the memory leak.

robgjansen:
> I've used g_shell_parse_argv() here

Seems ok to me.

> should I add a test case?

I would be nice to add something to one of the tests in src/test, but in this case we don't have a good way to test this so I think we can skip it on this one.

src/main/host/process.c
marcosimioni: thanks for pointing that out. done!

marcosimioni:
quick q - should I switch this PR to dev? or should it stay on main?

robgjansen (approved):
Looks great! We want to merge to `main`, which I'll do shortly :) Thank you!

PR #1267, opened 2021-04-06
Make getaddrinfo return DGRAM sockets if requested

robgjansen:


robgjansen:
We basically want to ignore this commit in dev. I think we can do that with

```
git checkout dev
git merge --strategy ours main
```

when we merge main into dev.

RWails (approved):

PR #1270, opened 2021-04-06
Removed the TCP congestion control option

stevenengler:
There is only a single congestion control method implemented in the dev branch (reno), so we should remove this option until we support others. Part of #1220.

### DIFF
--- a/src/main/host/descriptor/tcp.c
+++ b/src/main/host/descriptor/tcp.c
@@ -2555,21 +2555,11 @@ TCP* tcp_new(guint receiveBufferSize, guint sendBufferSize) {
 
     Options* options = worker_getOptions();
     guint32 initial_window = 10;
-    const gchar* tcpCC = options_getTCPCongestionControl(options);
     gint tcpSSThresh = 0;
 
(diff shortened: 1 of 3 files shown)


src/main/host/descriptor/tcp.c
```diff
@@ -2555,21 +2555,12 @@ TCP* tcp_new(guint receiveBufferSize, guint sendBufferSize) {
...
-        case TCP_CC_RENO:
-            tcp_cong_reno_init(tcp);
-            break;
-        case TCP_CC_UNKNOWN:
-            error("Failed to initialize TCP congestion control for %s", tcpCC);
-            break;
-    }
+    /* in the future we'd like to support more congestion control types
+     * and allow it to be set as a host option */
+    TCPCongestionType congestionType = TCP_CC_RENO;
```
robgjansen, line 2562: I think `congestionType` is an unused variable, unless I missed it?

robgjansen (approved):

src/main/host/descriptor/tcp.c
stevenengler: Oops, fixed.

Issue #1221, opened 2021-03-26
Add default single-node topology in case user does not provide one


robgjansen:
We decided that if the user does not provide a topology, we'll load in a default one internally stored as a string. We can store such a default topology using something like this:

https://github.com/shadow/shadow/blob/0a62c035b149af060390fdcc669ba45219878a36/src/main/core/support/examples.c#L10-L50

Note, the above should be updated and simplified because some of the attributes no longer exist or are unnecessary.

stevenengler:
We now include a basic topology we're currently calling "plain".

```yaml
topology: plain
```

https://github.com/shadow/shadow/blob/f6f25cd9c21ab9379a6fd3ea46beffd642f35bc8/src/main/core/support/configuration.rs#L671-L697

Issue #1161, opened 2021-03-09
Allow variable suffixes for bandwidth config values


robgjansen:
Currently, Shadow interprets the values given in the `bandwidthup` and `bandwidthdown` config attributes as `KiB/s`.

https://github.com/shadow/shadow/blob/482458a2ff0573d4b5eec0c588852af46aead24a/src/main/core/master.c#L376-L377

We should do something like [TGen does in it's `tgenoptionparser_parseBytes` function](https://github.com/shadow/tgen/blob/47d5eb385195eb22a87b05984116447ad32ef8d7/src/tgen-optionparser.c#L197), where
- a raw integer is interpreted as bytes, and
- a suffix can be given to instruct Shadow how to interpret the config value.

stevenengler:
This behaviour is now supported in the new Rust version. It differs slightly from the TGen parsing, but is very similar. The main differences are that the Shadow parsing is case-sensitive, and Shadow uses bits-per-second rather than bytes-per-second for bandwidth.

Issue #1147, opened 2021-03-05
Remove the plugins list from the root of the configuration file


stevenengler:
The current configuration file format requires you to list plugins and their paths in the configuration root, and then refer to those plugins when specifying processes. For example:

```yaml
plugins:
- id: myplugin
  path: tor/src/app/tor
hosts:
- id: mynode
  quantity: 1
  processes:
  - plugin: myplugin
    starttime: 1
```

It would be simpler to remove the plugins list at the root of the configuration file and require proceses to specify the path directly. For example:

```yaml
hosts:
- id: mynode
  quantity: 1
  processes:
  - plugin: tor/src/app/tor
    starttime: 1
```

robgjansen:
In v2, we `fork`/`exec` an executable file rather than `dlopen`ing a library, so they aren't really plugins anymore in the usual sense. Therefore, I think we should also rename the `plugin` keyword to something else, e.g., `exec_path`.

stevenengler:
One reason for leaving this list of executables is that we could allow the user to override their paths using cli options. For example if we want to run using different versions of tor, we could set the default path as `/usr/sbin/tor` but allow it to be overridden with maybe `shadow --plugin-path tor,~/tor-0.4.5/src/app/tor`.

stevenengler:
The new YAML format does not have a plugin list.

Issue #1146, opened 2021-03-04
Remove old Shadow configuration options


stevenengler:
Some of the configuration options have new names:
- `kill` -> `stoptime`
- `time` -> `starttime`

Before Shadow v2.0, we should remove the old options.

https://github.com/shadow/shadow/blob/551c65c17c835723ad4bda52d7cfe78036809e1f/src/main/core/support/configuration.c#L567-L571

https://github.com/shadow/shadow/blob/551c65c17c835723ad4bda52d7cfe78036809e1f/src/main/core/support/configuration.c#L641-L642

https://github.com/shadow/shadow/blob/551c65c17c835723ad4bda52d7cfe78036809e1f/src/main/core/support/configuration.c#L784

stevenengler:
These are no longer supported in the new YAML format.

PR #1278, opened 2021-04-08
(Part 3) Command-line parsing and YAML config files in Rust

### LINKED ISSUE #773: More human-readable Shadow config

stevenengler:
Replaces Shadow's existing command-line and configuration file parsing code with Rust versions using [Serde](https://serde.rs/) and [Clap](https://docs.rs/clap/2.33.3/clap/) (version 3.0.0-beta.2). Shadow's XML configuration file format has been replaced by YAML in the process.

This PR will be easier to review as individual commits.

Closes #773.

Changed CLI options:

```
Old options                                                 New options
-----------                                                 -----------

-?, --help                                                  -h, --help
--help-all
--help-sys
--help-experimental
--interface-buffer=N                    [1024000]           --interface-buffer                  ["1024000 B"]
--interface-qdisc=QDISC                 ['fifo']            --interface-qdisc                   ["fifo"]
--socket-recv-buffer=N                  [174760]            --socket-recv-buffer                ["174760 B"]
                                                            --socket-recv-autotune              [true]
--socket-send-buffer=N                  [131072]            --socket-send-buffer                ["131072 B"]
                                                            --socket-send-autotune              [true]
--max-concurrency=N                     [-1]                --max-concurrency
--set-sched-fifo                                            --use-sched-fifo                    [false]
--disable-object-counters                                   --use-object-counters               [true]
--disable-shim-syscall-handler                              --use-shim-syscall-handler          [true]
--disable-o-n-waitpid-workarounds                           --use-o-n-waitpid-workarounds       [true]
--send-explicit-block-message=[0|1]     [1]                 --use-explicit-block-message        [true]
--enable-syscall-counters                                   --use-syscall-counters              [false]
--disable-memory-manager                                    --use-memory-manager                [true]
--preload-spin-max=N                    [8096]              --preload-spin-max                  [8096]
                                                            --bootstrap-end-time                ["0 sec"]
                                                            --heartbeat-interval                ["1 sec"]
                                                            --stop-time
                                                            --bandwidth-down
                                                            --bandwidth-up
                                                            --city-code-hint
                                                            --country-code-hint
                                                            --ip-hint
                                                            --pcap-directory
                                                            --type-hint
-c, --exit-after-shm-cleanup                                --shm-cleanup
-d, --data-directory=PATH               ['shadow.data']     -d, --data-directory                ["shadow.data"]
-e, --data-template=PATH                [None]              -e, --template-directory
-g, --gdb                                                   -g, --gdb
-h, --heartbeat-frequency=N             [1]                 --heartbeat-interval                ["1 sec"]
-i, --heartbeat-log-info=LIST           ['node']            --host-heartbeat-log-info           [["node"]]
-j, --heartbeat-log-level=LEVEL         ['message']         --host-heartbeat-log-level          ["message"]
                                                            --host-log-level
-l, --log-level=LEVEL                   ['message']         -l, --log-level                     ["message"]
-r, --runahead=TIME                     [0]                 --runahead
-s, --seed=N                            [1]                 --seed                              [1]
-t, --scheduler-policy=SPOL             ['steal']           --scheduler-policy                  ["steal"]
-n, --interpose-method=METHOD           ['ptrace']          --interpose-method                  ["ptrace"]
-z, --pin-cpus                                              --use-cpu-pinning                   [false]
-w, --workers=N                         [0]                 -w, --workers                       [0]
-v, --version                                               -V, --version
                                                            --show-build-info
                                                            --show-config
```

Changed configuration options:

```
bootstraptime         bootstrap_end_time
starttime             start_time
stoptime              stop_time
arguments             args
iphint                ip_hint
citycodehint          city_code_hint
countrycodehint       country_code_hint
typehint              type_hint
loglevel              log_level
heartbeatloglevel     heartbeat_log_level
heartbeatloginfo      heartbeat_log_info
pcapdir               pcap_directory
bandwidthdown         bandwidth_down (unit changed from KiB/s to bits)
bandwidthup           bandwidth_up (unit changed from KiB/s to bits)
heartbeatfrequency    heartbeat_interval
socketrecvbuffer      socket_recv_buffer
socketsendbuffer      socket_send_buffer
interfacebuffer       interface_buffer
```

Changed topology options:

```
bandwidthup      bandwidth_up (type changed to "string", and unit changed to bits)
bandwidthdown    bandwidth_down (type changed to "string", and unit changed to bits)
packetloss       packet_loss
countrycode      country_code
citycode         city_code
```

Miscellaneous changes:

- added a `quantity` field for processes
- renamed the configuration file top-level `options` field to `general`, and added new experimental options under the top-level `experimental` field, and new host default options under the top-level `host_defaults` field.
- hosts in the configuration file are now specified using a map/dictionary, and not a list of hosts
- moved some existing options into the `experimental` group (#1220)
- topology in the configuration file is no longer a list
- most of the host options have moved into an `options` field
- added a `plain` topology option (in addition to the existing `graphml` and `path` options), which uses a simple built-in topology (#1221)
- process arguments in the configuration file can now be passed as a list of arguments
- the existing XML/YAML conversion script has been modified to convert to the new YAML format, and no longer converts from YAML to XML
- added a new script for converting the topology graphml file to the new format
- removed the configuration file plugin list, and now each process specifies the path to the executable to run (#1147)
- disabled the XML/YAML conversion tests
- the `logpcap` and `pcapdir` options have been combined into a single option `pcap_directory`
- removed some of the short option flags (`-h`, `-i`, `-j`, etc)
- removes legacy configuration options suck as `kill` and `time` (#1146)
- values can be provided with a unit value (for example "1 Mbit", "1 megabit", etc) and if a unit isn't given, will default to the base unit (#1161)
- environment field is now per-process rather than global
- arguments are no longer split by spaces

To do after this PR:
- update tornettools to support new formats
- add PR for new/modified conversion scripts
- update and re-enable the XML/YAML conversion tests
- update documentation ([3.1-Shadow-Config.md](https://github.com/shadow/shadow/blob/dev/docs/3.1-Shadow-Config.md), [3.2-Network-Config.md](https://github.com/shadow/shadow/blob/dev/docs/3.2-Network-Config.md))


src/main/routing/topology.c
```diff
@@ -241,6 +243,30 @@ static gboolean _topology_isValidEdgeAttributeKey(const gchar* attrName, EdgeAtt
...
+    if (igraph_cattribute_has_attr(&top->graph, IGRAPH_ATTRIBUTE_VERTEX, name)) {
+        const gchar* value = igraph_cattribute_VAS(&top->graph, name, vertexIndex);
+        if (value != NULL && value[0] != '\0') {
+            if (valueOut != NULL) {
+                int64_t bandwidth = parse_bandwidth(value);
+                if (bandwidth >= 0) {
+                    // parse_bandwidth() returns bits-per-second, but shadow works with KiB/s
+                    bandwidth /= 8 * 1024;
+                    *valueOut = bandwidth;
+                    return TRUE;
```
robgjansen, line 261: I think we are planning to remove KiB/s and just pass around bits or bytes. Do we need a TODO here for when we address that?

src/test/CMakeLists.txt
```diff
@@ -70,13 +70,12 @@ macro(add_shadow_tests)
       
       set(SHADOW_TEST_COMMAND sh -c "\
          rm -rf ${SHADOW_TEST_NAME}.data \
-         && ${yaml2xml} ${SHADOW_TEST_SHADOW_CONFIG} --output - \
-         | ${CMAKE_BINARY_DIR}/src/main/shadow \
+         && ${CMAKE_BINARY_DIR}/src/main/shadow \
```
robgjansen, line 73: Woo! :)

src/test/clone/clone.yaml
```diff
@@ -1,32 +1,8 @@
...
+  stop_time: 10
+topology: plain
 hosts:
-- id: host
-  quantity: 1
-  processes:
-  - plugin: test_clone
-    starttime: 1
-    arguments: ''
+  host:
```
robgjansen, line 5: Super-nit: can we change the name of the host to something that is more apparently a hostname instead of `host`?

src/test/determinism/determinism1.test.shadow.config.yaml
```diff
@@ -1,32 +1,28 @@
...
+      <key attr.name="bandwidth_up" attr.type="string" for="node" id="d2" />
+      <key attr.name="bandwidth_down" attr.type="string" for="node" id="d1" />
+      <key attr.name="country_code" attr.type="string" for="node" id="d0" />
       <graph edgedefault="undirected">
         <node id="poi-1">
           <data key="d0">US</data>
-          <data key="d1">10240</data>
-          <data key="d2">10240</data>
+          <data key="d1">81920 Kibit</data>
+          <data key="d2">81920 Kibit</data>
```
robgjansen, line 15: I think you want this to be `Kibibit` instead of `Kibit`?
https://en.wikipedia.org/wiki/Kibibit

This same problem applies to all of the yaml files in the PR that contain topology graphml code. I'm only noting it here, but you should change it everywhere.

src/test/tor/minimal/tor-minimal.yaml
```diff
@@ -30,114 +28,112 @@ topology:
...
-    arguments: --Address torhiddenclient --Nickname torhiddenclient --DataDirectory .
-      --GeoIPFile ~/.shadow/share/geoip --defaults-torrc ../../../conf/tor.common.torrc
-      -f ../../../conf/tor.client.torrc --BandwidthRate 1024000 --BandwidthBurst 1024000
-  - plugin: tgen
-    starttime: 1500
-    arguments: ../../../conf/tgen.torhiddenclient.graphml.xml
+  fileserver:
+    options:
+      bandwidth_down: 819200 Kibit
+      bandwidth_up: 819200 Kibit
```
robgjansen, line 35: Some more `Kibit` -> `Kibibit` conversions here.

src/test/tor/minimal/tor-minimal.yaml
```diff
@@ -30,114 +28,112 @@ topology:
...
+    processes:
+    - path: ~/.shadow/bin/tgen
+      args: ../../../conf/tgen.server.graphml.xml
+      start_time: 1
+  hiddenserver:
+    options:
+      bandwidth_down: 819200 Kibit
+      bandwidth_up: 819200 Kibit
```
robgjansen, line 43: Should we just set this bandwidth as the default for everyone in the `host_defaults` part of the config? Then we wouldn't need to duplicate it for each host.

Or even better, the topology only contains a single node, and that node already has the correct bandwidth, so I think all the hosts will already get assigned that bandwidth even if we don't specify it as a host option or host_default option.

robgjansen (changes requested):
This looks great! It simplifies a lot of things and will make it much simpler to configure experiments. I'm looking forward to using this new format in my experiments :)

I don't have any complex changes, but I'd like to look at any changes you make anyway :)

src/test/determinism/determinism1.test.shadow.config.yaml
stevenengler: Do you want to change this to `Kibibit` just as a preference, or is there an issue with `Kibit`? Right now the configuration parsing will accept units in both `kibibit` (the full unit name) and `Kibit` (the unit symbol).

src/test/tor/minimal/tor-minimal.yaml
stevenengler: Some hosts have bandwidths of 819200 kibibit (102400 KiB/s), some have 81920 kibibit (10240 KiB/s), clients have no defined bandwidths, and the topology node has a bandwidth of 81920 kibibit. Is there a reason why the original configuration had these hosts with different bandwidths? And why some hosts had bandwiths 10x greater than the topology node? Since this is just a test, it should be fine giving them all the same bandwidth, right?

src/test/determinism/determinism1.test.shadow.config.yaml
robgjansen: Oh shit, this is my bad, I had never seen `Kibit` before and I missed this part on the wiki page:

> The unit symbol of the kibibit is Kibit

So never mind, I think how you have it is fine.

src/test/tor/minimal/tor-minimal.yaml
robgjansen: Never mind, `Kibit` is fine.

src/test/tor/minimal/tor-minimal.yaml
robgjansen: Right, I don't think the bandwidths matter much in this small test. I think they can all be the same. Not sure why they were different before - probably because I decided that many many years ago and never changed it.

src/test/determinism/determinism1.test.shadow.config.yaml
stevenengler: > So never mind, I think how you have it is fine.

Okay, and if you would still prefer "kibibit" just let me know, it's an easy change.

src/main/core/main.c
```diff
@@ -150,56 +147,83 @@ gint main_runShadow(gint argc, gchar* argv[]) {
...
     }
 
+    char* configName = clioptions_getConfig(options);
+
+    /* read config from file or stdin */
+    ConfigFileOptions* configFile = NULL;
+    if (strcmp(configName, "-") != 0) {
+        configFile = configfile_parse(configName);
+    } else {
+        configFile = configfile_parse("/dev/stdin");
```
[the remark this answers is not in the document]
stevenengler: Added a comment.

src/main/core/main.c
```diff
@@ -150,56 +147,83 @@ gint main_runShadow(gint argc, gchar* argv[]) {
...
+        clioptions_free(options);
+        return EXIT_FAILURE;
+    }
+
+    /* generate the final shadow configuration from the config file and cli options */
+    ConfigOptions* config = config_new(configFile, options);
+    configfile_free(configFile);
+
+    if (clioptions_getShowConfig(options)) {
+        config_showConfig(config);
```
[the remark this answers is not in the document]
stevenengler: Added.

src/main/core/support/config_handlers.c
```diff
@@ -0,0 +1,24 @@
...
+
+void runConfigHandlers(const ConfigOptions* config) {
+    if (experimentalOptions == NULL) {
+        return;
+    }
+    for (int i = 0; i < experimentalOptions->len; i++) {
+        ConfigHandlerFn fn = g_array_index(experimentalOptions, ConfigHandlerFn, i);
+        fn(config);
+    }
+    g_array_free(experimentalOptions, true);
```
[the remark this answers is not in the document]
stevenengler: Good idea, added.

src/main/core/support/config_handlers.h
```diff
@@ -0,0 +1,19 @@
...
+
+typedef void (*ConfigHandlerFn)(const ConfigOptions*);
+
+void addConfigHandler(ConfigHandlerFn fun_ptr);
+void runConfigHandlers(const ConfigOptions* config);
+
+// Used to generate unique symbol names. See
+// https://stackoverflow.com/questions/1597007/creating-c-macro-with-and-line-token-concatenation-with-positioning-macr
+#define OPTIONS_TOKENPASTE(x, y) x##y
+#define OPTIONS_TOKENPASTE2(x, y) OPTIONS_TOKENPASTE(x, y)
```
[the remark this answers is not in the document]
stevenengler: I don't think we can `undef` them since they're used by the `ADD_CONFIG_HANDLER` macro. (The build fails when I undef them.)

src/main/routing/topology.c
stevenengler: Added a TODO (and in `_controller_registerHostCallback()` also).

src/test/clone/clone.yaml
stevenengler: Changed to `testnode` like the other tests.

src/test/tor/minimal/tor-minimal.yaml
stevenengler: Changed the topology bandwidth to `1 Gbit` (rounded up from the ~800 Mbit that most hosts were using) and unset all of the other bandwidths.

robgjansen (approved):
Looks great! Thanks for the changes :)

PR #1281, opened 2021-04-12
For pipes size zero reads & writes are no-ops

sporksmith:
Split out of #1277

### DIFF
--- a/src/main/host/descriptor/channel.c
+++ b/src/main/host/descriptor/channel.c
@@ -107,6 +107,13 @@ static gssize channel_sendUserData(Transport* transport, PluginVirtualPtr buffer
     /* the read end of a unidirectional pipe can not write! */
     utility_assert(channel->type != CT_READONLY);
 
+    /* Zero-size writes of pipes aren't very clearly specified. The pipe(2)
+     * documentation for O_DIRECT indicates that a size-zero write is a no-op
+     * with O_DIRECT; experimentally they are also a no-op without it. */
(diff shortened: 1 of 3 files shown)


src/main/host/descriptor/channel.c
```diff
@@ -133,6 +140,12 @@ static gssize channel_receiveUserData(Transport* transport, PluginVirtualPtr buf
     /* the write end of a unidirectional pipe can not read! */
     utility_assert(channel->type != CT_WRITEONLY);
 
+    /* Zero-size reads of pipes aren't very clearly specified, but
+     * experimentally they are a no-op. */
+    if (nBytes == 0) {
+        return 0;
+    }
```
stevenengler, line 147: I'm not 100% sure if this is correct for reading. If this is a non-blocking pipe, it might first check if there are bytes to read and return EAGAIN if not.

stevenengler (approved):
Looks good. I'm not 100% sure this works correctly for non-blocking sockets (for example reading non-blocking sockets with `len == 0` will sometimes return EAGAIN), but either way I think it's an improvement. Maybe running the new test case a second time using non-blocking pipe fds would be good too.

src/main/host/descriptor/channel.c
sporksmith: I added a second size-zero read to the test, which validates that it doesn't block. I think it's reasonable to infer it would also not return EAGAIN for a nonblocking descriptor.

PR #1287, opened 2021-04-13
Changes to experimental options

stevenengler:
This PR has two changes:

1. The `--send-explicit-block-message` option (which defaulted to 1) has been renamed to `--disable-explicit-block-message` (which defaults to false).
2. The variables representing two of the experimental options have been inverted to better match their corresponding option.

The default behaviour of Shadow has not changed.

These are changes that will be helpful when adding Rust option/config support.


robgjansen (approved):

stevenengler:
We decided to remove the `--disable` and `--enable` prefixes from options, so this PR isn't needed anymore.

PR #1289, opened 2021-04-13
max-concurrency: Fix backwards error message

### LINKED ISSUE #1284: max-concurrency, worker check in SP_PARALLEL_HOST_STEAL inconsistent with error message
https://github.com/shadow/shadow/blob/1c7c8da06c4f1c887daccaa81231499b037a5157/src/main/core/scheduler/scheduler.c#L162

If `nWorkers > _maxConcurrency` is true, the following error is printed:

`Host stealing scheduler is incompatible with --max-concurrency > --workers`

which is *opposite* of what's being checked.

sporksmith:
Also remove redundant condition.

Fixes #1284


RWails (approved):

RWails:
Looks good!

PR #1290, opened 2021-04-14
thread_ptrace: use unbuffered IO by default

sporksmith:
Buffering only helps when doing small sequential accesses. For syscalls that do large accesses (read, write), buffering just adds an extra copy.

Smaller accesses are generally small structs; not many handlers access more than one, and even then it only helps if they happen to be sequential in memory.

### DIFF
--- a/src/main/host/thread_ptrace.c
+++ b/src/main/host/thread_ptrace.c
@@ -56,6 +56,13 @@ OPTION_EXPERIMENTAL_ENTRY("disable-o-n-waitpid-workarounds", 0, G_OPTION_FLAG_RE
                           "otherwise result in excessive detaching and reattaching",
                           NULL)
 
+static bool _useBufferedIo = false;
+OPTION_EXPERIMENTAL_ENTRY(
+    "enable-ptrace-buffered-io", 0, 0, G_OPTION_ARG_NONE, &_useBufferedIo,
+    "Use buffered IO when reading plugin memory through /proc. This introduces some extra copying "
+    "but may help performance when making small sequential accesses.",
+    NULL)
+
 // Because of <https://github.com/shadow/shadow/issues/1134> we also always use __WNOTHREAD when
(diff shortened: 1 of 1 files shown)


src/main/host/thread_ptrace.c
```diff
@@ -56,6 +56,13 @@ OPTION_EXPERIMENTAL_ENTRY("disable-o-n-waitpid-workarounds", 0, G_OPTION_FLAG_RE
                           "otherwise result in excessive detaching and reattaching",
                           NULL)
 
+static bool _useBufferedIo = false;
```
robgjansen, line 59: Steve just renamed a bunch of these `_useXXX` switch variables, so maybe we should follow his pattern?

https://github.com/shadow/shadow/pull/1287/files#diff-458bdb269062900e2d0ac153268cbd70b89d5c811f1fd9445d2e49f21c4b5033R55

Although I think he was just aiming to make it so that the default for all of the options is `false`. Anyway, I guess `_enableBufferedIo` would be more consistent with his change?

robgjansen (approved):

src/main/host/thread_ptrace.c
sporksmith: Good point; done

sporksmith: > We decided to remove the --disable and --enable prefixes from options, so this PR isn't needed anymore.

I guess I should still use the `--enable` prefix here, and they'll all be changed in another PR?

stevenengler: I closed that PR since I thought we wanted to not use the `--enable` or `--disable` prefixes anymore.

robgjansen:
For posterity: I ran a Tor 5% net with memory manager disabled using Jim’s byte queue commit aed8b717e8f88b0f9c9cff1ff7c4ea2d20c26efb.

[run_time.pdf](https://github.com/shadow/shadow/files/6327107/run_time.pdf)

It looks like unbuffered IO indeed is the winner.

PR #1291, opened 2021-04-19
Avoid library calls after vfork

### LINKED ISSUE #1243: Error conditions after vfork() cause Shadow to freeze
When running Shadow, it froze (required `SIGKILL` to stop) with the last log line:

```
forking new thread with environment '[snip]', arguments 'test-epoll', and working directory '[snip]'
```

After changing the code to use `fork()` rather than `vfork()`, I could see that the forked process was getting killed by a `SIGSEGV` (`STOPREASON_SIGNAL: 11`). This `SIGSEGV` seems to have been caused by the combination of disabling the RDTSC (with the `PR_TSC_SIGSEGV` option), and then later calling `error()`. I was able to fix this `SIGSEGV` by changing `error()` to `printf()` and then calling `exit(1)`. After these changes, I was able to figure out that the path to my plugin was incorrect so `execvpe()` was failing.

I think we should avoid calling the Shadow logging methods after `prctl()` (when using `fork()` or `vfork()`). I think we should also avoid calling any logging methods (Shadow logging or printf) after `vfork()` since the program is in a delicate state, and it's difficult to debug when things go wrong. If we really want to log, maybe we should instead have a helper program so that we can `vfork()`, immediately `execvpe()`, and then set the remaining options in the helper program which can do another `execvpe()` later to start the plugin.

Edit: We also need to be careful about Shadow in multi-threaded mode. If another thread is in the middle of allocating memory or logging or anything else requiring locks when the vfork happens, the vforked process will hang forever.

sporksmith:
In between vfork and exec we share an address space with the parent
process. This can lead to surprising, difficult to debug behavior.

We should restrict ourselves to only making syscalls; e.g. functions in
`man` section 2. Higher level functions could try to allocate,
manipulate locks, etc, which is problematic.

Fixes #1243

### DIFF
--- a/src/main/host/thread_preload.c
+++ b/src/main/host/thread_preload.c
@@ -142,19 +142,16 @@ static pid_t _threadpreload_fork_exec(ThreadPreload* thread, const char* file, c
             // child
 
             // Set the working directory
-            utility_assert(workingDir != NULL);
             if (chdir(workingDir) < 0) {
-                error("chdir(%s): %s", workingDir, g_strerror(errno));
+                die_after_vfork();
             }
 
             int rc = execvpe(file, argv, envp);
             if (rc == -1) {
-                error("execvpe() call failed");
-                return -1;
+                die_after_vfork();
             }
-            while (1) {
-            } // here for compiler optimization
-            break;
+            // Unreachable
+            die_after_vfork();
         }
         default: // parent
             info("started process %s with PID %d", file, pid);
--- a/src/main/host/thread_ptrace.c
+++ b/src/main/host/thread_ptrace.c
@@ -366,30 +366,29 @@ static pid_t _threadptrace_fork_exec(const char* file, char* const argv[], char*
             // https://github.com/shadow/shadow/issues/903), so now we jump all
             // the way to SIGKILL.
             if (prctl(PR_SET_PDEATHSIG, SIGKILL) < 0) {
-                error("prctl: %s", g_strerror(errno));
+                die_after_vfork();
             }
             // Validate that Shadow is still alive (didn't die in between forking and calling
             // prctl).
             if (getppid() != shadow_pid) {
-                error("parent (shadow) exited");
+                die_after_vfork();
             }
             // Disable RDTSC
             if (prctl(PR_SET_TSC, PR_TSC_SIGSEGV, 0, 0, 0) < 0) {
-                error("prctl: %s", g_strerror(errno));
+                die_after_vfork();
             }
             // Set the working directory
-            utility_assert(workingDir != NULL);
             if (chdir(workingDir) < 0) {
-                error("chdir(%s): %s", workingDir, g_strerror(errno));
+                die_after_vfork();
             }
             // Become a tracee of the parent process.
             if (ptrace(PTRACE_TRACEME, 0, 0, 0) < 0) {
-                error("ptrace: %s", g_strerror(errno));
+                die_after_vfork();
             }
             if (execvpe(file, argv, envp) < 0) {
-                error("execvpe: %s", g_strerror(errno));
+                die_after_vfork();
             }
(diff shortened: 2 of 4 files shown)


src/main/host/thread_preload.c
```diff
@@ -142,19 +142,16 @@ static pid_t _threadpreload_fork_exec(ThreadPreload* thread, const char* file, c
...
 
             // Set the working directory
-            utility_assert(workingDir != NULL);
             if (chdir(workingDir) < 0) {
-                error("chdir(%s): %s", workingDir, g_strerror(errno));
+                die_after_vfork();
             }
 
             int rc = execvpe(file, argv, envp);
             if (rc == -1) {
```
stevenengler, line 150: `execvpe()` isn't part of `man 2`. Since Shadow forces the executable path to be an absolute path anyways, I think we should change this to be `execve()`.

src/main/host/thread_ptrace.c
```diff
@@ -366,30 +366,29 @@ static pid_t _threadptrace_fork_exec(const char* file, char* const argv[], char*
...
             }
             // Become a tracee of the parent process.
             if (ptrace(PTRACE_TRACEME, 0, 0, 0) < 0) {
-                error("ptrace: %s", g_strerror(errno));
+                die_after_vfork();
             }
             if (execvpe(file, argv, envp) < 0) {
-                error("execvpe: %s", g_strerror(errno));
+                die_after_vfork();
             }
```
stevenengler, line 390: Same as above.

src/main/utility/utility.c
```diff
@@ -362,3 +362,19 @@ int return_code_for_signal(int signal) {
...
+    // Ensure our saved errno doesn't get optimized away.
+    asm volatile(/*asm=*/"" : /*outputs=*/ : /*inputs=*/"irm"(saved_errno));
+
+    // abort and kill are higher-level functions that could attempt to access
+    // global memory, which could have surprising results. We resort to a bare
+    // `kill`.
+    kill(getpid(), SIGABRT);
+
+    // Convince the compiler that we really don't return.
+    abort();
```
stevenengler, line 379: What do you mean by "[...] and kill are higher-level functions that could attempt to access global memory [...] We resort to a bare `kill`"?

Do we ever expect the `kill()` to return and continue (I guess maybe if we have signal handlers)? Should the `abort()` be just `_exit()` instead?

stevenengler (approved):

src/main/host/thread_preload.c
sporksmith: The `exec` family of functions is explicitly whitelisted in vfork(2):

> (From  POSIX.1) The vfork() function has the same effect as fork(2), except that the behavior is undefined if the process created by vfork() either modifies any data other than a variable of type pid_t used to store the return value from vfork(), or returns from the function in which vfork() was called, or calls any other function before  successfully calling _exit(2) or one of the exec(3) family of functions.

Granted this also indicates that we're in undefined behavior land doing pretty much anything at all in between `vfork` and `exec*`, but I think they're being a bit conservative.

src/main/host/thread_ptrace.c
sporksmith: "

src/main/utility/utility.c
sporksmith: > What do you mean by "[...] and kill are higher-level functions that could attempt to access global memory [...] We resort to a bare kill"?

Oops - s/kill/raise/. Fixed.

> Do we ever expect the kill() to return and continue (I guess maybe if we have signal handlers)? Should the abort() be just _exit() instead to avoid any side-effects of abort() in that case?

No, we don't expect it. `kill` shouldn't return before the signal is delivered, so should never return. Went ahead and changed to _exit to be safe, though.

PR #1293, opened 2021-04-20
Adds a module to disable plugin ASLR

rwails:
Disabling ASLR should eliminate cases of non-determinism if a plugin branches on the memory address of any variable reference. I don't think there's any downsides.

Somewhat related:

We're accumulating a bunch of simulator settings in src/main/core/main.c (Lines 206 thru 226). Do these belong somewhere else?

### DIFF
--- a/src/main/core/main.c
+++ b/src/main/core/main.c
@@ -223,6 +223,9 @@ gint main_runShadow(gint argc, gchar* argv[]) {
         }
     }
 
+    // Disable address space layout randomization of processes forked from this
+    // one to ensure determinism in cases when an executable under simulation
+    // branch on memory addresses.
     disable_aslr();
 
     gint returnCode = _main_helper(options);
--- a/src/main/core/main.c
+++ b/src/main/core/main.c
@@ -23,6 +23,7 @@
 #include "main/core/support/options.h"
 #include "main/host/affinity.h"
 #include "main/shmem/shmem_cleanup.h"
+#include "main/utility/disable_aslr.h"
 #include "main/utility/utility.h"
 #include "shd-config.h"
 #include "support/logger/logger.h"
@@ -221,6 +222,8 @@ gint main_runShadow(gint argc, gchar* argv[]) {
             message("Successfully set real-time scheduler mode to SCHED_FIFO");
         }
     }
+
+    disable_aslr();
 
     gint returnCode = _main_helper(options);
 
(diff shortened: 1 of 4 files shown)


robgjansen (commented):
Seems OK to me, but I'd like Jim and Steve to take a look too, since it's a small change and they might have better feedback.

RWails:
Worth noting that dettrace makes a similar call to `personality()`, but also supplies the `PER_LINUX` flag. I don't know why. The thorough man documentation for `PER_LINUX` is "Linux." <https://man7.org/linux/man-pages/man2/personality.2.html>

See https://github.com/dettrace/dettrace/blob/657ff07bbfad5bac5ce680c412d66acb1008ae3b/src/dettrace.cpp#L401

robgjansen:
> The thorough man documentation for PER_LINUX is "Linux."

Oh, that makes sense then! Ha!

src/main/core/main.c
```diff
@@ -222,6 +223,8 @@ gint main_runShadow(gint argc, gchar* argv[]) {
         }
     }
 
+    disable_aslr();
```
stevenengler, line 226: A comment here might be useful to state that we disable ASLR for determinism reasons, and not because we need to (for the memory manager for example).

stevenengler (approved):
I don't know of any downsides to disabling ASLR (other than for security), so looks good to me. I can't tell what the `PER_LINUX` flag is for either, but it's [defined as 0](https://elixir.bootlin.com/linux/v3.3/source/include/linux/personality.h#L56) so shouldn't matter in practice.

src/main/core/main.c
sporksmith: +1

sporksmith (approved):

src/main/utility/disable_aslr.c
```diff
@@ -0,0 +1,20 @@
...
+
+#include <sys/personality.h>
+
+#include "support/logger/logger.h"
+
+void disable_aslr() {
+    int prev_persona = personality(ADDR_NO_RANDOMIZE);
+
+    if (prev_persona != -1) {
+        debug("ASLR disabled for processes forked from this parent process.");
```
robgjansen, line 14: Could we please change this to a `message` log? It's nice to get confirmation in the log file that features like this one are indeed enabled as expected.

src/main/core/main.c
RWails: fixed!

src/main/utility/disable_aslr.c
RWails: fixed!

PR #1299, opened 2021-04-21
Fix error detection in topology parsing

stevenengler:
The error state would get overwritten, even if there was previously an error.


robgjansen (approved):
Good catch!

PR #1300, opened 2021-04-22
Added value unit/prefix parsing and conversion

stevenengler:
This code is currently unused, but will be used in the future by the command line / configuration parsing. Part of #1161.

### DIFF
--- a/src/main/core/support/units.rs
+++ b/src/main/core/support/units.rs
@@ -1,0 +1,723 @@
+/*!
+Types for parsing/deserializing unit values.
+
+```
+let time = Time::from_str("10 min").unwrap();
+assert_eq!(time, Time::new(10, TimePrefix::Sec));
+
+assert_eq!(
+    time.convert(TimePrefix::Sec).unwrap(),
+    Time::from_str("600 sec").unwrap()
+);
+```
+*/
+
+use std::convert::TryInto;
+use std::fmt::{self, Debug, Display};
+use std::str::FromStr;
+
+use once_cell::sync::Lazy;
+use regex::Regex;
+use serde::de::{Deserialize, Deserializer, Visitor};
+use serde::ser::{Serialize, Serializer};
+
+/// The prefix of a unit value denoting magnitude. Common prefixes are
+/// SI prefixes (nano, micro, milli, etc).
+pub trait Prefix: Clone + Copy + Default + PartialEq + FromStr + Display + Debug {
+    /// The magnitude of this prefix relative to other prefixes of this type.
+    fn relative_magnitude(&self) -> u128;
+
+    /// An integer conversion factor.
+    fn conversion_factor(&self, to: Self) -> Result<u128, String> {
+        let from_mag = self.relative_magnitude();
+        let to_mag = to.relative_magnitude();
+        if from_mag % to_mag != 0 {
+            return Err("Conversion would lose precision".to_string());
+        }
+        Ok(from_mag / to_mag)
+    }
+
+    /// A floating point conversion factor.
+    fn conversion_factor_lossy(&self, to: Self) -> f64 {
+        let from_mag = self.relative_magnitude();
+        let to_mag = to.relative_magnitude();
+        from_mag as f64 / to_mag as f64
+    }
+}
+
+/// Common SI prefixes (including base-2 prefixes since they're similar).
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum SiPrefix {
+    Nano,
+    Micro,
+    Milli,
+    Base,
+    Kilo,
+    Kibi,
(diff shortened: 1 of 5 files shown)


stevenengler:
I update the code to take more options (for example "ms" can be specified as "milliseconds"), and bandwidth units have been changed (for example "Mbps" has been changed to "Mbit"). I think this now covers most of the same format as TGen, but unlike TGen is case-sensitive.

src/main/core/support/units.rs
```diff
@@ -0,0 +1,703 @@
+use std::convert::TryInto;
+use std::fmt::{self, Debug, Display};
+use std::str::FromStr;
+
+use once_cell::sync::Lazy;
+use regex::Regex;
+use serde::de::{Deserialize, Deserializer, Visitor};
+use serde::ser::{Serialize, Serializer};
+
+pub trait Prefixes: Clone + Copy + Default + PartialEq + FromStr + Display + Debug {
```
sporksmith, line 10: Nit throughout - it's more conventional to name a trait, type, or enum after a member of that type. e.g. `Prefix`, `SiPrefix`

sporksmith, line 10: Throughout - consider adding some doc comments to the `pub` elements.

src/main/core/support/units.rs
```diff
@@ -0,0 +1,703 @@
...
+    fn relative_magnitude(&self) -> u128;
+
+    /// An integer conversion factor.
+    fn conversion_factor(&self, to: Self) -> Result<u128, String> {
+        let from_mag = self.relative_magnitude();
+        let to_mag = to.relative_magnitude();
+        if from_mag % to_mag != 0 {
+            return Err("Conversion would lose precision".to_string());
+        }
+        Ok(from_mag / to_mag)
```
sporksmith, line 20: Check for overflow?

sporksmith (approved):
Looks great!

src/main/core/support/units.rs
sporksmith: Oops, nevermind; not possible here

src/main/core/support/units.rs
```diff
@@ -0,0 +1,703 @@
...
+    }
+}
+
+impl FromStr for SiPrefixes {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "n" | "nano" => Ok(Self::Nano),
+            "u" | "micro" => Ok(Self::Micro),
```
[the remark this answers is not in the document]
stevenengler: Added (for both parsing and display).

src/main/core/support/units.rs
```diff
@@ -0,0 +1,703 @@
...
+            Self::Gibi => write!(f, "Gi"),
+            Self::Tera => write!(f, "T"),
+            Self::Tebi => write!(f, "Ti"),
+            Self::Base => Ok(()),
+        }
+    }
+}
+
+impl Prefixes for SiPrefixes {
+    fn relative_magnitude(&self) -> u128 {
```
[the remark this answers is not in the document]
stevenengler: Added this base, but also am using `pow()` now to hopefully make it more readable.

src/main/core/support/units.rs
```diff
@@ -0,0 +1,703 @@
...
+    }
+}
+
+mod export {
+    use super::*;
+    use log::*;
+
+    /// Parses a string as bytes-per-second.
+    #[no_mangle]
+    pub extern "C" fn parse_bandwidth(s: *const libc::c_char, out: *mut u64) -> bool {
```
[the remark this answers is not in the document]
stevenengler: Done.

src/main/core/support/units.rs
stevenengler: Fixed

stevenengler: Added some more comments.

src/main/core/support/units.rs
```diff
@@ -0,0 +1,703 @@
...
+            Self::Giga => 1_000_000_000,
+            Self::Gibi => 1 << 30,
+            Self::Tera => 1_000_000_000_000,
+            Self::Tebi => 1 << 40,
+        }
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum TimePrefixes {
```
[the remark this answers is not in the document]
stevenengler: Yeah, originally I wasn't going to include minutes and hours since they're different units, but then figured I'd add them even though they don't really fit, which meant adding the suffix to the prefixes. I have an implementation that splits it up into "prefixes", "units", and "values" (values have units, and units may or may not have prefixes), but I feel that it might be too confusing. For now I'll just add comments.

PR #1304, opened 2021-04-26
Remove the 'preload' option

### LINKED ISSUE #1215: Remove 'preloadPath' configuration option
Since this functionality can be set with the `environment` option, we don't need this extra 'preloadPath' option anymore.

stevenengler:
Always get the preload path from the rpath. Closes #1215.

### DIFF
--- a/src/main/core/controller.c
+++ b/src/main/core/controller.c
@@ -11,6 +11,7 @@
 #if !defined __USE_LARGEFILE64
 #define __USE_LARGEFILE64
 #endif
+
 #include <errno.h>
 #include <glib.h>
 #include <glib/gstdio.h>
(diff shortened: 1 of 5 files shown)


src/main/core/controller.c
```diff
@@ -443,17 +507,26 @@ gint controller_run(Controller* controller) {
...
     ConfigurationShadowElement* element = configuration_getShadowElement(controller->config);
-    g_assert(element && element->preloadPath.isSet);
+
+    gchar* preloadPath = _controller_scanRPathForPreloadShim();
+    if (preloadPath != NULL) {
+        message("found %s at %s", PRELOAD_SHIM_LIB_STR, preloadPath);
+    } else {
+        critical("could not find %s in rpath", PRELOAD_SHIM_LIB_STR);
+        return 1;
+    }
```
robgjansen, line 517: Unless there is some advantage to doing this in the controller that I'm missing, I think the right place to do it is in the manager. Eventually, when we enhance shadow to utilize multiple machines, we'll have one manager per machine but only one single controller per simulation. Each machine (i.e., manager) will have to scan the rpath separately.

I think we can just move all of the code over to manager, and initiate the rpath scan in `manager_new`?

robgjansen (approved):
Looks good overall!

src/main/core/controller.c
stevenengler: Moved this over to the manager.

PR #1306, opened 2021-04-26
Rename some variables used for experimental options

stevenengler:
This does not change the behaviour of Shadow. It only changes some variable names (for example `_disable_shim_syscall_handler` to `_use_shim_syscall_handler`), and changes the GLib option's behaviour to match.

In the future our experimental option names (for example, currently `--disable-shim-syscall-handler`) will follow this format (`--use-shim-syscall-handler`), but this just changes the variable names for now.

### DIFF
--- a/src/main/core/main.c
+++ b/src/main/core/main.c
@@ -28,9 +28,9 @@
 #include "shd-config.h"
 #include "support/logger/logger.h"
 
-static bool _setSchedFifo = false;
+static bool _useSchedFifo = false;
 OPTION_EXPERIMENTAL_ENTRY(
-    "set-sched-fifo", 0, 0, G_OPTION_ARG_NONE, &_setSchedFifo,
+    "set-sched-fifo", 0, G_OPTION_FLAG_NONE, G_OPTION_ARG_NONE, &_useSchedFifo,
     "Use the SCHED_FIFO scheduler. Requires CAP_SYS_NICE. See sched(7), capabilities(7)", NULL)
 
 static Controller* shadowcontroller;
@@ -210,7 +210,7 @@ gint main_runShadow(gint argc, gchar* argv[]) {
         }
     }
 
-    if (_setSchedFifo) {
+    if (_useSchedFifo) {
         struct sched_param param = {0};
         param.sched_priority = 1;
         int rc = sched_setscheduler(0, SCHED_FIFO, &param);
--- a/src/main/host/shimipc.c
+++ b/src/main/host/shimipc.c
@@ -10,12 +10,12 @@
 
 // We use an int here because the option parsing library doesn't provide a way
 // to set a boolean flag to false explicitly.
-static gint _sendExplicitBlockMessage = true;
+static gint _useExplicitBlockMessage = true;
 OPTION_EXPERIMENTAL_ENTRY(
-    "send-explicit-block-message", 0, 0, G_OPTION_ARG_INT, &_sendExplicitBlockMessage,
+    "send-explicit-block-message", 0, G_OPTION_FLAG_NONE, G_OPTION_ARG_INT, &_useExplicitBlockMessage,
     "Send message to plugin telling it to stop spinning when a syscall blocks", "[0|1]")
 
-bool shimipc_sendExplicitBlockMessageEnabled() { return _sendExplicitBlockMessage; }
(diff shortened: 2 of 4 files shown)


src/main/core/main.c
```diff
@@ -28,9 +28,9 @@
 #include "shd-config.h"
 #include "support/logger/logger.h"
 
-static bool _setSchedFifo = false;
+static bool _useSchedFifo = false;
 OPTION_EXPERIMENTAL_ENTRY(
-    "set-sched-fifo", 0, 0, G_OPTION_ARG_NONE, &_setSchedFifo,
+    "set-sched-fifo", 0, 0, G_OPTION_ARG_NONE, &_useSchedFifo,
```
robgjansen, line 33: Replace the 0 in the arg preceding `G_OPTION_ARG_NONE` to `G_OPTION_FLAG_NONE`. (That makes it more obvious which arg to change if we want to change it to `G_OPTION_FLAG_REVERSE`  later.)

src/main/host/shimipc.c
```diff
@@ -10,12 +10,12 @@
 
 // We use an int here because the option parsing library doesn't provide a way
 // to set a boolean flag to false explicitly.
-static gint _sendExplicitBlockMessage = true;
+static gint _useExplicitBlockMessage = true;
```
robgjansen, line 13: While we're here, can we go ahead and change this to a `bool` option? That's how we're using it anyway so it's more appropriate. (I've been wishing it was a `bool` for a while.)

src/main/host/shimipc.c
```diff
@@ -10,12 +10,12 @@
...
 OPTION_EXPERIMENTAL_ENTRY(
-    "send-explicit-block-message", 0, 0, G_OPTION_ARG_INT, &_sendExplicitBlockMessage,
+    "send-explicit-block-message", 0, 0, G_OPTION_ARG_INT, &_useExplicitBlockMessage,
```
robgjansen, line 15: I think this would become `G_OPTION_ARG_NONE` if we change it to a `bool`?

robgjansen, line 15: Replace the 0 in the arg preceding `G_OPTION_ARG_INT` to `G_OPTION_FLAG_NONE`. (That makes it more obvious which arg to change if we want to change it to `G_OPTION_FLAG_REVERSE`  later.)

robgjansen (approved):
Just some minor nits, but looks good overall.

src/main/host/shimipc.c
stevenengler: The problem is that we can't change it to a bool without breaking the current behaviour. Right now `--send-explicit-block-message` defaults to true, so we'd either have to change the default to `false`, or rename the option to something like `--disable-explicit-block-message`.

The upcoming Rust changes will be removing all of this GLib stuff anyways, so I think it might be better to wait and make this breaking change with all of the other changes.

src/main/core/main.c
stevenengler: Added

src/main/host/shimipc.c
stevenengler: Added

src/main/host/shimipc.c
robgjansen: Why can't we keep the default as `true`, and then use `G_OPTION_FLAG_REVERSE` the same way we do with `_use_object_counters`?

robgjansen: > The upcoming Rust changes will be removing all of this GLib stuff anyways, so I think it might be better to wait and make this breaking change with all of the other changes.

OK, makes sense.

stevenengler: > Why can't we keep the default as true, and then use G_OPTION_FLAG_REVERSE the same way we do with _use_object_counters?

Then we'd have to rename the option, otherwise the default would be true, and `--send-explicit-block-message` would also set it to true.

As we talked about a couple of weeks ago, since we want the the CLI options to override the yaml config options, in the Rust version all experimental bool options will have to take a true/false option (for example `--send-explicit-block-message true`) since if the yaml file specifies  `send-explicit-block-message: true`, you will need to be able to disable it on the command line. So rather than having two options (`--send-explicit-block-message` and `--disable-explicit-block-message`), we'll have a single `--send-explicit-block-message` option that takes either true or false.

robgjansen: Got it, thanks!

PR #1309, opened 2021-04-27
(Part 1) Command-line parsing and YAML config files in Rust

stevenengler:
This PR contains the new Rust code for command-line and yaml configuration file parsing, but this code isn't yet being used.

See #1278 for a summary of all features, and what the intended result is.

I'm setting Jim as the reviewer on this PR, but feel free to look through it @robgjansen and leave comments if you think anything should change. We can always change things later instead in the follow-up PR.


src/main/core/support/configuration.rs
```diff
@@ -1,9 +1,517 @@
...
+    #[clap(long, value_name = "seconds")]
+    #[clap(about = GENERAL_HELP.get("stop_time").unwrap())]
+    stop_time: Option<units::Time<units::TimePrefixUpper>>,
+
+    /// Environment variables passed to all simulated processes (ex: "ENV_A=1;ENV_B=2")
+    #[clap(long, value_name = "env-string")]
+    #[clap(about = GENERAL_HELP.get("environment").unwrap())]
+    environment: Option<String>,
+
+    /// Initialize randomness for each thread using seed N
```
sporksmith, line 133: This makes it sound like every thread gets the same seed, which I think is not true. I think it's some kind of `f(seed, threadId)`?

src/main/core/support/configuration.rs
```diff
@@ -1,9 +1,517 @@
...
+    /// Run concurrently with N worker threads
+    #[clap(long, short = 'w', value_name = "N")]
+    #[clap(about = GENERAL_HELP.get("workers").unwrap())]
+    #[serde(default = "default_some_0")]
+    workers: Option<u32>,
+
+    #[clap(long, value_name = "seconds")]
+    #[clap(about = GENERAL_HELP.get("bootstrap_end_time").unwrap())]
+    #[serde(default = "default_some_time_0")]
+    bootstrap_end_time: Option<units::Time<units::TimePrefixUpper>>,
```
sporksmith, line 148: Maybe doesn't need to be in this PR, but do we eventually want getters that returns e.g. `std::time::Instant` for Rust-side code?

src/main/core/support/configuration.rs
```diff
@@ -1,9 +1,517 @@
...
+    /// Path to recursively copy during startup and use as the data-directory
+    #[clap(long, short = 'e', value_name = "path")]
+    #[clap(about = GENERAL_HELP.get("template_directory").unwrap())]
+    template_directory: Option<String>,
+}
+
+impl GeneralOptions {
+    /// Replace unset (`None`) values of `base` with values from `default`.
+    pub fn merge(base: &Self, default: &Self) -> Self {
+        let mut base = base.clone();
```
sporksmith, line 178: Nit - take `base` and `default` by value and let caller `clone` if they still need them? I don't expect a noticeable performance impact here, but it's a useful pattern in general for functions that would otherwise clone their parameters.

src/main/core/support/configuration.rs
```diff
@@ -1,9 +1,517 @@
...
+    /// most cases
+    #[clap(long, value_name = "bool")]
+    #[clap(about = EXP_HELP.get("use_memory_manager").unwrap())]
+    use_memory_manager: Option<bool>,
+
+    /// Use buffered IO when reading plugin memory through /proc. This introduces some extra copying
+    /// but may help performance when making small sequential accesses.
+    #[clap(long, value_name = "bool")]
+    #[clap(about = EXP_HELP.get("use_enable_ptrace_buffered_io").unwrap())]
+    use_enable_ptrace_buffered_io: Option<bool>,
```
sporksmith, line 242: `use_ptrace_buffered_io`?

src/main/core/support/configuration.rs
```diff
@@ -1,9 +1,517 @@
...
+    #[clap(long, value_name = "bool")]
+    #[clap(about = EXP_HELP.get("use_syscall_counters").unwrap())]
+    use_syscall_counters: Option<bool>,
+
+    /// Count object allocations and deallocations. If disabled, we will not be able to detect object memory leaks
+    #[clap(long, value_name = "bool")]
+    #[clap(about = EXP_HELP.get("use_object_counters").unwrap())]
+    use_object_counters: Option<bool>,
+
+    /// Max number of iterations to busy-wait on ICP sempahore before blocking
```
sporksmith, line 222: ICP -> IPC
sempahore -> semaphore

src/main/core/support/configuration.rs
```diff
@@ -1,9 +1,517 @@
...
+/// the Serde default.
+static EXP_HELP: Lazy<std::collections::HashMap<String, String>> =
+    Lazy::new(|| generate_help_strs(schema_for!(ExperimentalOptions)));
+
+#[derive(Debug, Clone, Clap, Serialize, Deserialize, Merge, JsonSchema)]
+#[clap(
+    help_heading = "EXPERIMENTAL (Unstable and may change or be removed at any time, regardless of Shadow version)"
+)]
+#[serde(default)]
+pub struct ExperimentalOptions {
```
sporksmith, line 194: Is there a reason the defaults aren't encoded inline as for GeneralOptions? It's nice to have them near the definitions, and I assume without it the defaults won't show up in `--help`?

src/main/core/support/configuration.rs
```diff
@@ -1,9 +1,517 @@
...
+            country_code_hint: None,
+            city_code_hint: None,
+            type_hint: None,
+            bandwidth_down: None,
+            bandwidth_up: None,
+        }
+    }
+
+    /// Replace unset (`None`) values of `base` with values from `default`.
+    pub fn merge(base: &Self, default: &Self) -> Self {
```
sporksmith, line 424: Take params by value?

src/main/core/support/configuration.rs
```diff
@@ -16,40 +524,230 @@ pub enum InterposeMethod {
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        serde_yaml::from_str(s)
+    }
+}
+
+impl std::fmt::Display for SchedulerPolicy {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match self {
+            Self::Thread => write!(f, "thread"),
```
sporksmith, line 555: Optional - `#[derive(Display)]`? https://docs.rs/enum-display-derive/0.1.1/enum_display_derive/

src/main/core/support/configuration.rs
```diff
@@ -16,40 +524,230 @@ pub enum InterposeMethod {
...
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct Quantity(u32);
+
+impl Default for Quantity {
+    fn default() -> Self {
+        Self(1)
+    }
+}
+
+impl std::ops::Deref for Quantity {
```
sporksmith, line 636: Nifty

src/main/core/support/configuration.rs
```diff
@@ -16,40 +524,230 @@ pub enum InterposeMethod {
...
+                name.clone(),
+                format!("{}{}[default: {}]", description, space, default),
+            ),
+            None => defaults.insert(name.clone(), description.to_string()),
+        };
+    }
+    defaults
+}
+
+fn tilde_expansion(path: &str) -> std::path::PathBuf {
```
sporksmith, line 729: Optional - use `shellexpand` crate? https://stackoverflow.com/questions/54267608/expand-tilde-in-rust-path-idiomatically

sporksmith, line 729: Probably worth some unit tests

src/main/core/support/simulation_time.rs
```diff
@@ -0,0 +1,35 @@
+/*!
```
sporksmith, line 1: Doesn't need to be this PR, but I think we'll eventually want some `From` implementations for converting between `SimulationTime` and `std::time::Instant` (and/or `std::time::Duration`)

sporksmith (changes requested):

src/main/core/support/configuration.rs
```diff
@@ -1,9 +1,517 @@
...
+
+    topology: Topology,
+
+    // we use a BTreeMap so that the hosts are sorted by their hostname (useful for determinism)
+    hosts: BTreeMap<String, HostOptions>,
+}
+
+impl ConfigOptions {
+    pub fn new(config_file: &ConfigFileOptions, options: &CliOptions) -> Self {
+        let mut config_file = config_file.clone();
```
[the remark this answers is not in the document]
stevenengler: I've removed the extra clone calls.

src/main/core/support/configuration.rs
stevenengler: I'm not sure what it does, for some of these I just copied the existing help text over. I've changed this now to just say "Initialize randomness using seed N".

src/main/core/support/configuration.rs
stevenengler: I added `impl From<Time<TimePrefix>> for Duration` and `impl From<Time<TimePrefixUpper>> for Duration`.

src/main/core/support/configuration.rs
stevenengler: Added.

src/main/core/support/configuration.rs
stevenengler: Fixed, thanks!

src/main/core/support/configuration.rs
stevenengler: Fixed.

src/main/core/support/configuration.rs
stevenengler: If the user doesn't provide an "experimental" section, Serde needs to be able to generate a default `ExperimentalOptions` struct, so it needs `impl Default`. The `#[serde(default)]` means that Serde will also be able to use any of those same defaults from the `Default` impl when the user leaves out individual fields (for example `preload_spin_max`).

There's an existing issue for this: https://github.com/serde-rs/serde/issues/1416

> I assume without it the defaults won't show up in --help?

They will still show up in `--help`.

src/main/core/support/configuration.rs
stevenengler: Done.

src/main/core/support/configuration.rs
stevenengler: I removed the `impl std::fmt::Display for SchedulerPolicy` here and instead added a function that converts directly to the type `c::SchedulerPolicyType`.

src/main/core/support/configuration.rs
stevenengler: I changed it to use this crate, but then remembered the main reason we didn't want to use it:

> Note: because I no longer have capacity to support it, I'm now looking for a new maintainer for this library. Until I'm able to find one, it is unlikely to receive new updates in any reasonably timely manner.

For now I just left in the existing `tilde_expansion()`, but in the future we can switch it out with a crate if we find one that's being maintained.

src/main/core/support/simulation_time.rs
stevenengler: We aren't able to implement `From` since `SimulationTime` (a `u64`) and `std::time::Duration` are both types implemented outside of the current crate. I added a function for this conversion though (`fn sim_time_to_duration(time: c::SimulationTime) -> std::time::Duration`).

sporksmith (approved):

src/main/core/support/configuration.rs
```diff
@@ -174,9 +173,8 @@ pub struct GeneralOptions {
 
 impl GeneralOptions {
     /// Replace unset (`None`) values of `base` with values from `default`.
-    pub fn merge(base: &Self, default: &Self) -> Self {
-        let mut base = base.clone();
-        base.merge(default.clone());
+    pub fn merge(mut base: Self, default: Self) -> Self {
```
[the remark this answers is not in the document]
stevenengler: > I missed this the first time, but I think it'd be more idiomatic to have these be `pub fn merge(self, default: Self) -> Self`?

The `merge()` function already exists from the `#[derive(Merge)]`, so we can't give it the name "merge". If you have other suggestions for what to name it, let me know.

> (Or maybe take `&mut self` and mutate?) Up to you

That is what the original design from the "merge" library does, but it makes the interface difficult if you want to actually mutate the second argument rather than the first. Then you need to do something like:

```rust
let temp = b.clone();
temp.merge(a);
let a = temp;
```

src/main/core/support/configuration.rs
```diff
@@ -85,26 +85,25 @@ pub struct ConfigOptions {
...
+            HostDefaultOptions::merge(options.host_defaults, config_file.host_defaults);
         config_file.experimental =
-            ExperimentalOptions::merge(&options.experimental, &config_file.experimental);
+            ExperimentalOptions::merge(options.experimental, config_file.experimental);
 
         // copy the host defaults to all of the hosts
         for (_, host) in &mut config_file.hosts {
-            host.options = HostDefaultOptions::merge(&host.options, &config_file.host_defaults);
+            host.options =
+                HostDefaultOptions::merge(host.options.clone(), config_file.host_defaults.clone());
```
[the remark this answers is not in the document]
stevenengler: The `host.options.clone()` is needed because `host` is behind a mutable reference. Rust doesn't currently allow you to iterate over values without consuming those values.

src/main/core/support/configuration.rs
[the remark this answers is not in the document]
sporksmith: If it returns a value, it'd be reasonable to name the function after what it returns. e.g. `merged_with`. I also just realized the semantics are the opposite of what I'd expect, particularly if the first parameter is some version of `self`. Maybe `overridden_with` to disambiguate, for something like `pub fn overridden_with(self, overrides: &Self) -> Self`? So at the caller it'd be something like `let settings = defaults.overriden_with(cli_args)`.

All up to you; I think the current version is ok and don't want to ask you to keep changing it :)

PR #1310, opened 2021-04-27
Legacy config/topology conversion scripts

stevenengler:
The existing XML/YAML conversion script has been modified to convert to the new YAML format, and no longer converts from YAML to XML. Also added a new script for converting the topology graphml file to the new format.

I added a commit for renaming `convert.py` to `convert_legacy_config.py`, so reviewing by commit may be easier.


src/tools/convert_legacy_topology.py
```diff
@@ -0,0 +1,75 @@
...
+        edges = x.findall('{}edge'.format(graphml_ns_prefix))
+        for y in nodes + edges:
+            for z in y.findall('{}data'.format(graphml_ns_prefix)):
+                if z.attrib['key'] in id_type_conversions:
+                    z.text = id_type_conversions[z.attrib['key']](z.text)
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description='Convert shadow topology files to the new format, and write the output to stdout')
```
sporksmith, line 63: this may go stale. How about something like "from Shadow 1.x format to Shadow 2.x format"?

sporksmith (approved):

src/tools/convert_legacy_topology.py
stevenengler: Good idea, added this.

PR #1313, opened 2021-04-28
Added support for ioctl tty requests such as TCGETS

stevenengler:
The two main descriptor types that shadow supports (files and sockets) aren't ttys, so these ioctl command should always return `ENOTTY`.


robgjansen (approved):
👍

PR #1325, opened 2021-05-07
Reimplement ShadowLogger in Rust and bound its memory usage

### LINKED ISSUE #1301: Logger queue grows unbounded, exceeds available RAM at debug level
When running a smallish network such as the Tor minimal example, and logging at the debug log level, log messages are being generated faster than they can be flushed to the file system over time. This means that if the experiment runs long enough, it uses all available system RAM.

sporksmith:
* Reimplement ShadowLogger in Rust
* Log system thread names instead of artificial thread names.

Fixes #1301


src/main/core/logger/log_wrapper.c
```diff
@@ -0,0 +1,29 @@
...
+        .log = _log,
+        .flush = _flush,
+        .destroy = rustlogger_destroy,
+        .isEnabled = _isEnabled,
+        .setLevel = _setLevel,
+    };
+    return logger;
+}
+
+void rustlogger_destroy(Logger* logger) { free(logger); }
```
robgjansen, line 29: FYI: I've recently come to appreciate files ending with a blank line rather than the last line not containing a newline. When adding new lines in the future, the diff will show that the previous last line in the file is modified even though the only thing changed was the addition of a newline character.

src/support/logger/logger.c
```diff
@@ -145,6 +143,22 @@ void logger_log(Logger* logger, LogLevel level, const gchar* fileName,
     }
 }
 
+void logger_setLevel(Logger* logger, LogLevel level) {
+    if (!logger) {
+
```
robgjansen, line 148: Was this empty line intentional? If so, and you want to keep the if statement, at least add a comment.

src/main/core/logger/shadow_logger.rs
```diff
@@ -0,0 +1,400 @@
...
+}
+
+impl TimeParts {
+    fn from_nanos(total_nanos: u128) -> Self {
+        let total_secs = u64::try_from(total_nanos / 1_000_000_000).unwrap();
+        let total_mins = u32::try_from(total_secs / 60).unwrap();
+        let hours = total_mins / 60;
+        let mins = total_mins - hours * 60;
+        let secs = total_secs - u64::from(total_mins) * 60;
+        let nanos = u64::try_from(total_nanos - u128::from(total_secs) * 1_000_000_000).unwrap();
```
robgjansen, line 43: The difference here between `total_*` and just `hours`, `mins`, etc. is unclear. I'm confused why we are converting from nanos to seconds back to nanos; don't we lose information due to rounding? Probably just needs a comment.

src/main/core/logger/shadow_logger.rs
```diff
@@ -0,0 +1,400 @@
...
+        .name("shadow-logger".to_string())
+        .spawn(move || SHADOW_LOGGER.logger_thread_fn())
+        .unwrap();
+
+    // Arrange to flush the logger on panic.
+    let default_panic_handler = std::panic::take_hook();
+    std::panic::set_hook(Box::new(move |panic_info| {
+        // Attempt to flush the logger. We want to avoid a recursive panic, so
+        // we flush the queue on the current thread instead of trying to send to
+        // send a command to the logger thread (because our thread-local sender
```
robgjansen, line 68: Nit: `send to send`

src/main/core/logger/shadow_logger.rs
```diff
@@ -0,0 +1,400 @@
...
+    }));
+
+    Ok(())
+}
+
+/// A logger specialized for Shadow. It attaches simulation context to log
+/// entries (e.g. sim time, running process, etc.). It's also designed for
+/// high performance to accomodate heavy logging from multiple threads.
+pub struct ShadowLogger {
+    // Channel used to send commands to second commands to the logger's thread.
```
robgjansen, line 82: `send commands to second commands`: it's unclear what this means.

src/main/core/logger/shadow_logger.rs
```diff
@@ -0,0 +1,400 @@
...
+            host_name,
+        });
+
+        if record.level() == Level::Error || self.records.len() > SYNC_FLUSH_QD_LINES_THRESHOLD {
+            // Unlike in Shadow's C code, we don't abort the program on Error
+            // logs. In Rust the same purpose is filled with `panic` and
+            // `unwrap`. C callers will still exit or abort via the support/logger wrapper.
+            //
+            // Flush *synchronously*, since we're likely about to crash one way or another.
+            self.flush_sync();
```
robgjansen, line 350: Wooo!

robgjansen (commented):
Looks reasonable to me! But let's have Steven also look at `src/main/core/logger/shadow_logger.rs`.

robgjansen:
@stevenengler I reviewed this but I think you should also review at least `src/main/core/logger/shadow_logger.rs`.

src/main/core/logger/log_wrapper.c
sporksmith: Done. Agreed; I'm used to vim doing it automatically. I'll see if there's a VSCode setting...

src/support/logger/logger.c
sporksmith: Added a comment

src/main/core/logger/shadow_logger.rs
sporksmith: Added comments and renamed variables to hopefully clarify. Also added a test.

src/main/core/logger/shadow_logger.rs
sporksmith: Done

src/main/core/logger/shadow_logger.rs
sporksmith: Done

src/main/Cargo.toml
```diff
@@ -26,3 +27,10 @@ regex = "1"
 schemars = "0.8"
 serde = { version = "1.0", features = ["derive"] }
 serde_yaml = "0.8"
+# TODO: switch to upstream crate if/when they merge and release
+# https://github.com/dylanmckay/vsprintf/pull/2
+#
+# Alternatively, just call libc's vsprintf directly when the unstable feature
+# `c_variadic` feature is stabilized, or if we decide to enable it anyway.
```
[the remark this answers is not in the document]
sporksmith: Done

src/main/core/logger/shadow_logger.rs
```diff
@@ -0,0 +1,426 @@
...
+                    } else {
+                        f
+                    })
+                    .unwrap_or("n/a"),
+                line = record
+                    .line
+                    .map(|l| format!("{}", l))
+                    .as_ref()
+                    .map(|s| s.as_str())
+                    .unwrap_or("n/a"),
```
stevenengler, line 264: Maybe just print "-1" instead of "n/a" here if not available to avoid the extra string allocation from `format!()`?

src/main/core/logger/shadow_logger.rs
```diff
@@ -0,0 +1,426 @@
...
+
+impl Log for ShadowLogger {
+    fn enabled(&self, metadata: &Metadata) -> bool {
+        metadata.level() <= log::max_level()
+    }
+
+    fn log(&self, record: &Record) {
+        if !self.enabled(record.metadata()) {
+            return;
+        }
```
stevenengler, line 343: Is the individual host log level taken into account anywhere?

src/main/core/logger/shadow_logger.rs
```diff
@@ -0,0 +1,426 @@
...
+    // Like the sender, needs a Mutex for ShadowLogger to be Sync.
+    // The Mutex is only locked once though by the logger thread, which keeps
+    // it locked for as long as it's running.
+    command_receiver: Mutex<Receiver<LoggerCommand>>,
+
+    // A lock-free queue for individual log records. We don't put the records
+    // themselves in the `command_sender`, because `Sender` doesn't support
+    // getting the queue length. Conversely we don't put commands in this queue
+    // because it doesn't support blocking operations.
+    records: SegQueue<ShadowLogRecord>,
```
stevenengler, line 124: Since the logger doesn't push more than `SYNC_FLUSH_QD_LINES_THRESHOLD` records, does it make sense to make this an `ArrayQueue` to avoid the extra allocation? Maybe this doesn't matter since there are already a few other string allocations per record.

stevenengler (approved):

src/main/core/logger/shadow_logger.rs
sporksmith: Yeah, I think I originally went with SegQueue because I started with a higher value for ASYNC_FLUSH_QD_LINES_THRESHOLD, and didn't want the fixed memory cost. With the current value of 100k it's not so bad, though.

src/main/core/logger/shadow_logger.rs
sporksmith: Broke up this `write!` to avoid the allocation while keepin "n/a" instead of "-1".

src/main/core/logger/shadow_logger.rs
sporksmith: Good catch! It is now :)

PR #1327, opened 2021-05-07
syscallhandler_shadow_hostname_to_addr_ipv4: null-terminate name

sporksmith:
The previous call to process_getReadablePtr wasn't including the
null-byte in the read. This happens to work in the MemoryMapper path,
since the NULL byte will still be there, but in the case where we do a
remote-read this could result in garbage at the end.

Switch to process_getReadableString, which ensures the returned  string
is null-terminated.


src/main/host/syscall/shadow.c
```diff
@@ -32,7 +32,12 @@ SysCallReturn syscallhandler_shadow_hostname_to_addr_ipv4(SysCallHandler* sys,
         return (SysCallReturn){.state = SYSCALL_DONE, .retval.as_i64 = -EINVAL};
     }
 
-    const char* name = process_getReadablePtr(sys->process, name_ptr, name_len);
+    const char* name;
+    int rv = process_getReadableString(
+        sys->process, name_ptr, name_len + 1 /* NULL byte */, &name, &name_len);
```
stevenengler, line 37: If the length of the string is already know in advance, `process_getReadablePtr(sys->process, name_ptr, name_len + 1)` seems more straightforward? But I'm good with it either way.

stevenengler (approved):
Nice catch!

src/main/host/syscall/shadow.c
sporksmith: Yeah, I could also go either way on this one, but chose `process_getReadableString` since it's a little more robust. e.g. if the caller side changes to include the  NULL byte in the length, `process_getReadableString` will still work properly, whereas process_getReadablePtr could fail if the extra byte isn't readable.

PR #1328, opened 2021-05-07
Enable the 'unsafe_op_in_unsafe_fn' lint

stevenengler:
The `unsafe_op_in_unsafe_fn` Rust lint [has been stabilized](https://github.com/rust-lang/rust/pull/79208) and I think we should use it. The lint is currently off by default, but they plan to make it a warning and possibly an error in the future.

Edit: Blocked on #1330.


stevenengler:
Oops, forgot that we're using an older nightly version. I'll see if we can update that.

sporksmith (approved):
Yay! Yeah it always bothered me that it wouldn't even *let* you use unsafe blocks in an unsafe function (or at least it would warn?), much less require it.

LGTM pending getting coverage working again

PR #1329, opened 2021-05-07
Align log levels with Rust's log crate

sporksmith:
Replaced usages of log levels:
* DEBUG -> INFO
* INFO -> DEBUG
* MESSAGE -> INFO
* CRITICAL -> ERROR

Added `panic` and `utility_panic`. The latter prints a backtrace but requires shadow's utility library, which in turn requires glib. Previous usage of log level ERROR now calls one of these, and logging to level ERROR by itself is no longer fatal.

Levels CRITICAL and MESSAGE are removed.


robgjansen (approved):
👍

Issue #1312, opened 2021-04-27
Tests 'pthreads' and 'determinism1' fail in Fedora 34


stevenengler:
The "pthreads" and "determinism1"  tests fail in Fedora 34.

```
The following tests FAILED:
	 25 - determinism1a-shadow-ptrace (Timeout)
	 26 - determinism1b-shadow-ptrace (Timeout)
	 27 - determinism1-compare-shadow-ptrace (Failed)
	 28 - determinism1a-shadow-hybrid (Timeout)
	 29 - determinism1b-shadow-hybrid (Timeout)
	 30 - determinism1-compare-shadow-hybrid (Failed)
	164 - pthreads-shadow-hybrid (Timeout)
	165 - pthreads-shadow-ptrace (Timeout)
```

Output from "pthreads-shadow-ptrace":
```
00:00:00.102967 [thread--1] 00:00:01.000000000 [message] [testnode~11.0.0.1] [process.c:437] [_process_start] starting process 'testnode.pthreads_plugin.1000'
00:00:00.102993 [thread--1] 00:00:01.000000000 [message] [testnode~11.0.0.1] [thread_ptrace.c:778] [threadptrace_run] forking new thread with environment 'SHADOW_SPAWNED=TRUE SHADOW_LOG_START_TIME=912058542 SHADOW_INTERPOSE_METHOD=PTRACE LD_PRELOAD=/__w/shadow/shadow/build/src/shim/libshadow-shim.so SHADOW_LOG_FILE=/__w/shadow/shadow/build/src/test/threads/pthreads-shadow-ptrace.data/hosts/testnode/testnode.pthreads_plugin.1000.shimlog SHADOW_SHM_BLK=8,134217728,24,/shadow_shmemfile_912.161422780-5141', arguments '/__w/shadow/shadow/build/src/test/threads/../target/debug/test_pthreads --shadow-passing', and working directory '/__w/shadow/shadow/build/src/test/threads/pthreads-shadow-ptrace.data/hosts/testnode'
00:00:00.103861 [thread-0] n/a [info] [n/a] [thread_ptrace.c:394] [_threadptrace_fork_exec] started process /__w/shadow/shadow/build/src/test/threads/../target/debug/test_pthreads with PID 5153
00:00:00.104002 [thread--1] 00:00:01.000000000 [message] [testnode~11.0.0.1] [process.c:458] [_process_start] process 'testnode.pthreads_plugin.1000' started
[...]

PR #1330, opened 2021-05-07
Update ubuntu/clang/rust versions required for coverage CI

stevenengler:
Hopefully we'll be good for another few months.

Blocked on #1332 and maybe #1312.


stevenengler:
We've solved this issue using a different approach in #1359

PR #1311, opened 2021-04-27
Support Fedora 34 instead of 33

stevenengler:
Related: #1254

Blocked by: #1312

PR #1331, opened 2021-05-07
Support newer glib and igraph versions

stevenengler:
Shadow doesn't currently build on Fedora 34 or Ubuntu 21.04. This PR allows them to build correctly. We don't yet test these platforms in the CI, but once we merge either #1330 or #1311 we'll need to fix this anyways, so I've made this its own PR. I've checked manually to make sure that both Fedora 34 or Ubuntu 21.04 build with this change.

This PR handles two deprecation warnings:

1. `igraph_i_set_attribute_table()` was renamed to `igraph_set_attribute_table()` in igraph 0.9.0
2. `g_memdup()` was replaced with `g_memdup2()` due to a [security issue](https://gitlab.gnome.org/GNOME/glib/-/issues/2319). This security issue does not affect Shadow's usage of this function since the problematic integer is set at compile time, but some static assertions were added to be sure.


src/main/routing/packet.c
```diff
@@ -16,6 +17,18 @@
...
 #include "support/logger/logger.h"
 
+/* g_memdup() is deprecated due to a security issue and has been replaced
+ * by g_memdup2(), but not all of our supported platforms support this yet.
+ * Since g_memdup() can be unsafe, using compat_g_memdup() can also be unsafe.
+ * !! make sure to limit the size to UINT_MAX !!
+ * https://gitlab.gnome.org/GNOME/glib/-/issues/2319
+ */
+#ifdef HAS_MEMDUP2
+#define compat_g_memdup g_memdup2
```
sporksmith, line 27: Could the `static_assert` go in the macro instead of the call sites?

sporksmith (approved):

PR #1338, opened 2021-05-10
futex_test: fix races for child taking lock

sporksmith:
When run natively, we can't guarantee that the futex won't have already
been released by the parent by the time we try to lock it.

### DIFF
--- a/src/test/futex/test_futex.c
+++ b/src/test/futex/test_futex.c
@@ -48,7 +48,8 @@ static void* _futex_wait_test_child(void* void_arg) {
     FutexWaitTestChildArg* arg = void_arg;
     _set_condition(&arg->child_started);
     trace("Child about to wait");
-    assert_true_errno(syscall(SYS_futex, &arg->futex, FUTEX_WAIT, UNAVAILABLE, NULL, NULL, 0) == 0);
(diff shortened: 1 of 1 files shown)


src/test/futex/test_futex.c
```diff
@@ -48,7 +48,8 @@ static void* _futex_wait_test_child(void* void_arg) {
     FutexWaitTestChildArg* arg = void_arg;
     _set_condition(&arg->child_started);
     trace("Child about to wait");
-    assert_true_errno(syscall(SYS_futex, &arg->futex, FUTEX_WAIT, UNAVAILABLE, NULL, NULL, 0) == 0);
+    assert_true_errno(!syscall(SYS_futex, &arg->futex, FUTEX_WAIT, UNAVAILABLE, NULL, NULL, 0) ||
+                      (errno == EAGAIN && !running_in_shadow()));
```
stevenengler, line 52: Wouldn't there still be a race condition where if `errno == EAGAIN`, the child might call `_set_condition(&arg->child_finished)` before the parent runs its `g_assert_true(!_get_condition(&arg.child_finished))`? Would it be better to just loop if `errno == EAGAIN`?

sporksmith: If we get EAGAIN here, it means the parent has already flipped the futex value to AVAILABLE on line 91.

stevenengler: Ah I see, thanks!

stevenengler (approved):

Issue #1297, opened 2021-04-20
Update some defaults for preload mode


robgjansen:
We want to set the default for both `--send-explicit-block-message` and `--preload-spin-max` to false.

robgjansen:
Closed in #1346

Issue #1271, opened 2021-04-06
Enable `--pin-cpus` by default and rename the option


stevenengler:
Since the `--pin-cpus` option is becoming an "experimental" option, we should instead enable CPU pinning by default and change the option to something like `--disable-cpu-pinning`.

robgjansen:
Remember this tidbit from @sporksmith 
> we'll need to explicitly disable it in our shadow_test macro to avoid them all pinning to the same core.

robgjansen:
Closed in #1346

PR #1346, opened 2021-05-11
Update default config options

robgjansen:
#1271, and #1297


src/main/core/support/configuration.rs
```diff
@@ -237,7 +237,7 @@ pub struct ExperimentalOptions {
     #[clap(about = EXP_HELP.get("use_shim_syscall_handler").unwrap())]
     use_shim_syscall_handler: Option<bool>,
 
-    /// Use CPU pinning
+    /// Pin each thread and any processes it executes to the same CPU to improve cache affinity
```
stevenengler, line 240: Maybe use "logical CPU core" to specify that you don't mean the same CPU package.

stevenengler (approved):

PR #1350, opened 2021-05-12
Enable 'unused-function' warning for C code

stevenengler:
This causes a warning if the C code has unused functions, which makes it more similar to how the Rust compiler works. This warning can also help find forgotten functions when refactoring.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -141,6 +141,13 @@ add_compile_options(-fno-omit-frame-pointer)
 add_compile_options(-Wreturn-type)
 add_compile_options(-Wswitch)
 
+if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+    # we only check for unused functions when builing in debug mode since some
+    # functions are only called when logging, which can be #ifdef'd out in
+    # release mode
+    add_compile_options(-Wunused-function)
+endif()
+
 if("${CMAKE_BUILD_TYPE}" STREQUAL "")
     set(CMAKE_BUILD_TYPE "Release")
 endif()
--- a/src/main/core/scheduler/scheduler.c
+++ b/src/main/core/scheduler/scheduler.c
@@ -385,7 +385,7 @@ static void _scheduler_assignHosts(Scheduler* scheduler) {
     g_mutex_unlock(&scheduler->globalLock);
 }
 
-static void _scheduler_rebalanceHosts(Scheduler* scheduler) {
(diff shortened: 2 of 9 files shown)


stevenengler:
Decided this isn't worthwhile since some functions are only called in debug mode and not release mode (for example `_tcp_stateToAscii()`), and we don't want to add `#pragma GCC diagnostic` lines for all of those.

sporksmith:
Maybe we could enable it only in debug builds?

stevenengler:
@sporksmith It seems that gcc on centos 7 does not support `#pragma GCC diagnostic ignored "-Wunused-function"` (it's ignored), so the only option would be to add `__attribute__((unused))` to each function. The downside is that gcc disables all warnings for functions which have the `__attribute__((unused))` attribute, which is not really what we want.

I'm fine with either merging this or closing it. Do you have a preference?

sporksmith:
Does `__attribute__((used))` work?

It'd be nice to turn this on, but probably not worth a lot of effort. Up to you :)

sporksmith:
Actually `used` is a little strange too, since it'll force the compiler to include it even if it's not actually used.

Is there a warning if something with the `unused` attribute is actually used? If so then at least we can count on the compiler forcing us to remove the attribute and un-suppress any other relevant warnings once something is actually used again.

stevenengler:
> Is there a warning if something with the `unused` attribute is actually used? If so then at least we can count on the compiler forcing us to remove the attribute and un-suppress any other relevant warnings once something is actually used again.

There's no warning if something with the `unused` attribute is actually used.

> The downside is that gcc disables all warnings for functions which have the `__attribute__((unused))` attribute, which is not really what we want.

I think I was reading documentation for an older gcc version, since I just tested a function with `__attribute__((unused))` and it does raise other warnings (for example `printf("%i %i", 4);` within the function will raises a `-Wformat` warning).

CMakeLists.txt
```diff
@@ -141,6 +141,11 @@ add_compile_options(-fno-omit-frame-pointer)
 add_compile_options(-Wreturn-type)
 add_compile_options(-Wswitch)
 
+if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+    # we only check for unused functions when builing in debug mode
```
sporksmith, line 145: Would be helpful to note why this is

sporksmith (approved):

CMakeLists.txt
stevenengler: Updated

Issue #1251, opened 2021-04-02
Rework scheduler options and defaults


sporksmith:
I think things are heading towards always using the "host" scheduler (`-t host`) with `--max-concurrency` set to what would previously have been the number of workers; e.g. the number of physical CPU cores available, and the number of workers (`-w`) set to something in between "number of hosts in the simulations" and "number of hosts in the simulation / ~100". This is all a bit confusing and fiddly for users.

Current proposal:

* Remove schedulers other than the host scheduler, reclaiming the `-t` flag
* Remove the "0 worker" special case that runs hosts on the main shadow thread. Running a single worker isn't much more overhead, and this would get rid of some extra complexity in the code.
* Rename `--max-concurrency` to `--processors` and `-p`, described as "Number of processors to utilize to execute simulated hosts". We decided against calling this something more specific like `--logical-processors`, since that's an implementation detail we'd like to be able to change without a major release.
* Add a flag `--hosts-per-thread`, and by default set the number of worker threads based on this and the number of hosts in the simulation. On one hand this leaks a bit about the "logical processor" implementation, but OTOH it probably does need to be a tunable parameter. Maybe this should be in the experimental/unstable section, and/or should *not* have a single-character-alias, so that we can later change it to silently do nothing.
* Keep `--workers` but drop `-w`, and again make it a flag that could in the future may be silently ignored. If set, this would override `--processors`

sporksmith:
Basically done in #1364, except "Remove the "0 worker" special case that runs hosts on the main shadow thread. Running a single worker isn't much more overhead, and this would get rid of some extra complexity in the code.", which we're deferring for now.

PR #1352, opened 2021-05-12
Disable forkproxy workaround by default

### LINKED ISSUE #1261: Remove or default-disable forkproxy waitpid workarounds
To address #1134, we implemented a workaround that uses proxy threads to fork the plugin threads (keeping the plugins out of the worker child list), and ptrace-detaches plugins on every blocking syscall (keeping the tracee list at size 1). While this workaround is a performance improvement overall for large simulations, the ptrace detaching-and reattaching adds substantial linear overhead.

Since then we've implemented a more performant solution of scaling the number of *worker* threads with the number of hosts, so that each worker thread only has a small number of children/tracees. While this has some extra memory and context-switching overhead, it generally appears to be a more performant solution since it doesn't require ptrace detaching-and-reattaching. (It also enables work stealing without detaching and reattaching).

We should probably at least disable the old workarounds by default before release. It *probably* makes sense to go a step further and remove them entirely, but we may need more evaluation to know for sure. e.g. it might turn out that in some cases it's better to take the performance hit of detaching and reattaching to avoid the extra memory cost of using more worker threads.

robgjansen:
We want the default mode to be that Shadow calculates the number
of processors to use based on the number of hosts specified in the
configuration file. (See #1251.) In this case, we do not want to
run the forkproxy workaround, because the fact that waitpid is O(n)
won't matter when each processor only runs a few pids.

Closes #1261

### DIFF
--- a/docs/3.1-Shadow-Config.md
+++ b/docs/3.1-Shadow-Config.md
@@ -260,10 +260,10 @@ Use the MemoryManager. It can be useful to disable for debugging, but will hurt 
 
 #### `experimental.use_o_n_waitpid_workarounds`
 
-Default: true  
+Default: false  
 Type: Bool
 
-Enable performance workarounds for waitpid being O(n). Beneficial to disable if waitpid is patched to be O(1) or in some cases where it'd otherwise result in excessive detaching and reattaching.
+Use performance workarounds for waitpid being O(n). Beneficial to disable if waitpid is patched to be O(1), if using one logical processor per host, or in some cases where it'd otherwise result in excessive detaching and reattaching.
 
 #### `experimental.use_object_counters`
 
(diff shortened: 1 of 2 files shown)


robgjansen:
@sporksmith can you take this over? I think you have a better understanding of the fork proxy code and the ptrace process exit code (which seems to be broken when disabling the fork proxy).

sporksmith:
@robgjansen sure, I'll take a look

sporksmith:
All tests passing! [You're welcome](https://anopensuitcase.com/wp-content/uploads/2016/11/maui-600x250.jpg)

sporksmith:
Forcing a re-run of the tor tests to double-check. I don't see the previous tor test failure, but I think GH throws it away after a rerun. The previous shadow tests failure was just a server timing out while installing deps.

stevenengler:
It doesn't look like the default was actually changed in this PR, the Tor tests should fail if `use_o_n_waitpid_workarounds` is set to false in `configuration.rs`.

robgjansen:
Shit, sorry, it looks like I incorrectly resolved a conflict when I cherry-picked this commit from my other branch :/

PR #1360, opened 2021-05-14
Deflake clone-linux

sporksmith:
Fixes the `clone-linux` failures I've been able to reproduce locally, and includes some other cleanup to avoid other potential failures.


sporksmith:
Tried switching from gcc atomic instrinsics to the C11 standard atomics, but centos 7 says no.

stevenengler (approved):

PR #1362, opened 2021-05-15
TypedPluginPtr::slice : allow empty slice at end of range

sporksmith:


src/main/host/syscall_types.rs
```diff
@@ -229,7 +229,9 @@ impl<T> TypedPluginPtr<T> {
             Bound::Unbounded => 0,
         };
         assert!(excluded_end <= self.count);
-        assert!(included_start < self.count);
+        // `<=` rather than `<`, to allow empty slice at end of ptr.
+        // e.g. `assert_eq!(&[1,2,3][3..3], &[])` passes.
+        assert!(included_start <= self.count);
```
stevenengler, line 234: Maybe worth adding an ` assert!(included_start <= excluded_end);` since `end_bound()` can be less than `start_bound()` (for example `10..5`).

stevenengler (approved):

src/main/host/syscall_types.rs
sporksmith: Yeah, I'd *thought* that trying to construct such a range would already panic, and therefore couldn't be received for the input. Now I can't find anything stating that though, and it runs and passes miri checks fine. So added :). https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=322d11e50af7b3ecbb7caaec89b0d680

PR #1364, opened 2021-05-17
Update scheduler options

sporksmith:
* The Experimental option `max-concurrency` is now the General option `parallelism`.
* The default scheduler policy is now Host.
* The General option `workers` is now the Experimental option `worker-threads`, and no longer has the `-w` shortcut. Users should use `parallelism` instead. If not set explicitly, the number of worker threads will be set to the number of hosts in the simulation.

### DIFF
--- a/docs/3.1-Shadow-Config.md
+++ b/docs/3.1-Shadow-Config.md
@@ -27,6 +27,7 @@ hosts:
 - [`general.data_directory`](#generaldata_directory)
 - [`general.heartbeat_interval`](#generalheartbeat_interval)
 - [`general.log_level`](#generallog_level)
+- [`general.parallelism`](#generalparallelism)
 - [`general.seed`](#generalseed)
 - [`general.stop_time`](#generalstop_time)
 - [`general.template_directory`](#generaltemplate_directory)
@@ -36,7 +37,6 @@ hosts:
 - [`experimental.interface_buffer`](#experimentalinterface_buffer)
 - [`experimental.interface_qdisc`](#experimentalinterface_qdisc)
 - [`experimental.interpose_method`](#experimentalinterpose_method)
-- [`experimental.max_concurrency`](#experimentalmax_concurrency)
 - [`experimental.preload_spin_max`](#experimentalpreload_spin_max)
 - [`experimental.runahead`](#experimentalrunahead)
 - [`experimental.scheduler_policy`](#experimentalscheduler_policy)
@@ -105,6 +105,14 @@ Default: "info"
 Type: "error" OR "warning" OR "info" OR "debug" OR "trace"
 
 Log level of output written on stdout. If Shadow was built in release mode, then messages at level 'trace' will always be dropped.
+
+#### `general.parallelism`
+
+Default: null  
+Type: Integer OR null
+
+How many parallel threads to use to run the simulation. Optimal performance is
+usually obtained with `nproc`, or sometimes `nproc/2` with hyperthreading.
 
 #### `general.seed`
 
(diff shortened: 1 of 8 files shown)


docs/3.1-Shadow-Config.md
```diff
@@ -27,16 +27,15 @@ hosts:
 - [`general.data_directory`](#generaldata_directory)
 - [`general.heartbeat_interval`](#generalheartbeat_interval)
 - [`general.log_level`](#generallog_level)
+- [`general.parallelism`](#general_parallelism)
```
robgjansen, line 30: The other options don't have an underscore between the `general` keyword and the name of the option. I have no idea why not, it's harder to read without the underscore. I assume there was a reason? Do we want to be consistent?

docs/3.1-Shadow-Config.md
```diff
@@ -106,6 +106,14 @@ Type: "error" OR "warning" OR "info" OR "debug" OR "trace"
 
 Log level of output written on stdout. If Shadow was built in release mode, then messages at level 'trace' will always be dropped.
 
+#### `general.parallelism`
+
+Default: null  
+Type: Integer OR null
+
+How many parallel threads to use to run the simulation. Optimal performance is
+usually obtained with `nproc`, or sometimes `nproc/2` with hyperthreading.
```
robgjansen, line 115: Do we need to explain what happens if set to null? Or maybe this isn't the place for that?

src/main/core/support/configuration.rs
```diff
@@ -313,7 +313,7 @@ impl Default for ExperimentalOptions {
             use_cpu_pinning: Some(true),
             interpose_method: Some(InterposeMethod::Ptrace),
             runahead: None,
-            scheduler_policy: Some(SchedulerPolicy::Steal),
+            scheduler_policy: Some(SchedulerPolicy::Host),
```
robgjansen, line 319: Is the `[default: "host"]` cli help string suffix auto-updated with these changes?

src/main/core/support/configuration.rs
```diff
@@ -1148,7 +1152,13 @@ mod export {
     pub extern "C" fn config_getWorkers(config: *const ConfigOptions) -> libc::c_uint {
         assert!(!config.is_null());
         let config = unsafe { &*config };
-        config.general.workers.unwrap()
+        match &config.experimental.worker_threads {
+            Some(w) => *w,
+            None => {
+                // By default use 1 worker per host.
+                config.hosts.len().try_into().unwrap()
```
robgjansen, line 1159: Much simpler than I had imagined :)

robgjansen (approved):
Looks great!

docs/3.1-Shadow-Config.md
sporksmith: Good catch! Yeah, we don't control how the anchors are generated for the headings, and unfortunately it omits the '.', so this was a broken link.

docs/3.1-Shadow-Config.md
sporksmith: I don't think there was a way for the user to explicitly set this to NULL. This reminded me though that there was some hidden logic that 0 would be silently treated as 1, and that `worker-threads=0` is special.

I cleaned this up by changing `parallel` to be of type `NonZeroU32`, and added the documentation about `worker-threads=0`.

src/main/core/support/configuration.rs
sporksmith: Yup!

src/main/core/support/configuration.rs
sporksmith: :)

PR #1367, opened 2021-05-19
Port DescriptorTable to Rust

sporksmith:
Wanted to start nailing down some more of how the object graph will work out in Rust; started to port Process to Rust, but looks like DescriptorTable needed to be ported first. (Though in hindsight I suppose the Rust Process could have just stored a `Box<DescriptorTable>`)


stevenengler:
> Sorry for spamming reviewer (un)assignment - I'm being wishy washy about whether to change to a Vec, which I suspect would be faster and use less memory in the common case. Going to sketch it out a bit more.

No problem. I think a vector is a good idea, it's probably faster in all cases, but I'm not sure at what size the memory usage becomes greater than a hash table. In a large tor experiment, what would the upper bound be on the number of file descriptors per process? Maybe ~2,000 sockets for a guard/exit?

> By the way - do you happen to know whether it's important to assign the lowest free descriptor first? Or is arbitrary order ok? e.g. if we create fd's 0, 1, 2, 3, 4, 5, then destroy 2, then destroy 3, is it ok if the next one created is 3? The C version sorts the free list to ensure the lowest free # is always returned, but the tests pass with arbitrary order.

The order of the descriptors is useful for i/o redirection. It's not as important now that there are `dup2` and `dup3` syscalls (which Shadow doesn't support), but some software (especially older software) relies on the ordering. For example, if you want to write stdout to a file, you would use:

```c
close(0);
open("filename");
```

And you could rely on the kernel giving the new file a fd of 0.

I think we should keep this behaviour as it is the expected behaviour and is part of posix, unless there's a strong reason not to. It also makes debugging error cases easier in tests when the program uses the same fd numbers both inside and outside shadow.

src/main/host/descriptor/descriptor_table.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    /// Insert a descriptor at `index`. If a descriptor is already present at
+    /// that index, it is unregistered from that index and returned.
+    pub fn set(&mut self, index: u32, descriptor: CompatDescriptor) -> Option<CompatDescriptor> {
+        let mut descriptor = descriptor;
+        super::export::compatdescriptor_setHandle(
+            &mut descriptor as *mut CompatDescriptor,
+            index.try_into().unwrap(),
+        );
```
stevenengler, line 78: `compatdescriptor_setHandle()` is only called from `descriptor_table.c`, so this `compatdescriptor_setHandle()` function could be removed entirely from shadow and replaced with just an inline:
```rust
if let CompatDescriptor::Legacy(d) = &mut descriptor {
    unsafe { c::descriptor_setHandle(d.ptr(), index) }
}
```

src/main/host/descriptor/descriptor_table.rs
```diff
@@ -0,0 +1,204 @@
...
+    ///
+    /// NOTE: that this consumes a reference to the descriptor, so if you are storing
+    /// it outside of the descriptor table you will need to ref it after calling
+    /// this function.
+    #[no_mangle]
+    pub unsafe extern "C" fn descriptortable_add(
+        table: *mut DescriptorTable,
+        descriptor: *mut CompatDescriptor,
+    ) -> c_int {
+        let table = unsafe { table.as_mut().unwrap() };
```
stevenengler, line 136: `table.as_mut()` -> `notnull_mut_debug(table).as_mut()`? (And same with the other `descriptortable_` functions below).

src/main/host/descriptor/descriptor_table.rs
```diff
@@ -0,0 +1,204 @@
...
+        unsafe { Box::from_raw(notnull_mut_debug(table)) };
+    }
+
+    /// Store a descriptor object for later reference at the next available index
+    /// in the table. The chosen table index is stored in the descriptor object and
+    /// returned. The descriptor is guaranteed to be stored successfully.
+    ///
+    /// NOTE: that this consumes a reference to the descriptor, so if you are storing
+    /// it outside of the descriptor table you will need to ref it after calling
+    /// this function.
```
stevenengler, line 130: These comments from `descriptor_table.h` are out of date. The `descriptor object` doesn't generally store the table index anymore (only the legacy descriptors do).

Also, the descriptor table consumes the descriptor, so it should not be stored or used after this function call.

src/main/host/descriptor/descriptor_table.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    /// Stop storing the descriptor so that it can no longer be referenced. The table
+    /// index that was used to store the descriptor is cleared from the descriptor
+    /// and may be assigned to new descriptors that are later added to the table.
+    /// Returns an owned pointer to the CompatDescriptor if the descriptor was found
+    /// in the table and removed, and NULL otherwise.
+    ///
+    /// NOTE: this will not unref the descriptor and you should do so manually.
+    #[no_mangle]
+    pub unsafe extern "C" fn descriptortable_remove(
```
stevenengler, line 149: The returned type `CompatDescriptor` doesn't have a reference count. Maybe the comment should say it should be freed with `compatdescriptor_free()`?

Also this function doesn't ever return NULL, but probably should following the earlier comment about `process_deregisterCompatDescriptor()`.

src/main/host/descriptor/descriptor_table.rs
```diff
@@ -0,0 +1,204 @@
...
+use crate::utility::notnull::*;
+use log::*;
+use std::convert::TryInto;
+
+/// Table of (file) descriptors. Typically owned by a Process.
+pub struct DescriptorTable {
+    descriptors: std::collections::HashMap<u32, CompatDescriptor>,
+    available_indices: Vec<u32>,
+    index_counter: u32,
+}
```
stevenengler, line 12: If we want to keep the available indices sorted so that we can always take the smallest, a BTreeSet might work well here.

stevenengler (changes requested):

src/main/host/descriptor/descriptor_table.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    /// Remove the descriptor at the given index and return it.  Panics if there
+    /// is no descriptor at `idx`.
+    pub fn remove(&mut self, idx: u32) -> CompatDescriptor {
+        let mut descriptor = self.descriptors.remove(&idx).unwrap();
+        self.available_indices.push(idx);
+        descriptor.set_handle(0);
+        descriptor
+    }
```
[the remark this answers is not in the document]
sporksmith: Done

src/main/host/descriptor/descriptor_table.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    /// Get the descriptor at `idx`, if any.
+    pub fn get(&self, idx: u32) -> Option<&CompatDescriptor> {
+        self.descriptors.get(&idx)
+    }
+
+    /// Insert a descriptor at `index`. If a descriptor is already present at
+    /// that index, it is unregistered from that index and returned.
+    pub fn set(&mut self, index: u32, descriptor: CompatDescriptor) -> Option<CompatDescriptor> {
+        let mut descriptor = descriptor;
```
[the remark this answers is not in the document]
sporksmith: I thought it'd be a little weird to have `mut` in the interface, since the caller shouldn't care. I checked `cargo doc` though, and it looks like the generated docs do ignore it, so it's mostly moot :). Done.

src/main/host/descriptor/descriptor_table.rs
sporksmith: I'd already created CompatDescriptor::set_handle (i.e. a Rust native version), but had missed a couple call sites. Fixed.

src/main/host/descriptor/descriptor_table.rs
sporksmith: Done

src/main/host/descriptor/descriptor_table.rs
sporksmith: `as_mut` and `as_ref` are safe to call on null pointers: https://doc.rust-lang.org/std/primitive.pointer.html#method.as_mut

src/main/host/descriptor/descriptor_table.rs
sporksmith: Done

sporksmith:
> is part of posix

Ok, yeah that clinches it for me :). I see `open(2)` specifically says it'll return the lowest available as well. Added back in this behavior and added a test for it.

> but I'm not sure at what point the memory usage of a spares vector becomes greater than a hash table

Yeah, that's the main drawback. While I still think a Vec is *probably* going to better in most cases, it won't be by that much, and could have unexpected poor behavior if it *is* sparse. Leaving it as a HashMap for this PR.

src/main/host/descriptor/descriptor_table.rs
```diff
@@ -2,49 +2,53 @@ use super::CompatDescriptor;
...
-                // `available_indices`.  Assuming this is rare, cheapest
-                // solution in memory and CPU is to just loop here.
-                continue;
+        let idx = if let Some(idx) = self.available_indices.iter().next() {
+            // Un-borrow from `available_indices`.
+            let idx = *idx;
+            // Take from `available_indices`
+            trace!("Reusing available index {}", idx);
+            self.available_indices.remove(&idx);
+            idx
```
stevenengler, line 37: Ah, it's too bad `BTreeSet::pop_first()` isn't stable yet.

stevenengler (approved):
Looks good!

src/main/host/descriptor/descriptor_table.rs
sporksmith: Yeah :/. I suppose we could backport it like I did for [stream_len](https://github.com/shadow/shadow/blob/dev/src/main/utility/stream_len.rs), but doesn't seem worth it for just this.

Issue #778, opened 2020-04-24
More human readable/editable topologies


sporksmith:
Broken out of #773. This is lowish priority, but we can consolidate ideas/discussion on the topic here.

Today the topologies in our examples look something like:

```xml
<topology><![CDATA[<?xml version="1.0" encoding="utf-8"?><graphml xmlns="http://graphml.graphdrawing.org/xmlns" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://graphml.graphdrawing.org/xmlns http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd">
  <key attr.name="packetloss" attr.type="double" for="edge" id="d6" />
  <key attr.name="latency" attr.type="double" for="edge" id="d5" />
  <key attr.name="packetloss" attr.type="double" for="node" id="d4" />
  <key attr.name="countrycode" attr.type="string" for="node" id="d3" />
  <key attr.name="bandwidthdown" attr.type="int" for="node" id="d2" />
  <key attr.name="bandwidthup" attr.type="int" for="node" id="d1" />
  <key attr.name="ip" attr.type="string" for="node" id="d0" />
  <graph edgedefault="undirected">
    <node id="isp">
      <data key="d0">0.0.0.0</data>
      <data key="d1">2251</data>
      <data key="d2">17038</data>
      <data key="d3">US</data>
      <data key="d4">0.0</data>
    </node>
    <edge source="isp" target="isp">
      <data key="d5">50.0</data>
      <data key="d6">0.01</data>
    </edge>
  </graph>
</graphml>
]]></topology>
```

The graphml format is generally visually noisy, and the networkx graphml exporter chooses opaque attribute-id's like "d6", forcing a reader to have to look back and forth between the `data` tags and `key` tags to understand what's happening.

graphml was chosen in part because it's a format that many tools understand. It's particularly useful that the `networkx` python package understands it (for generating graphs), and that the `igraph` library can import it (for use internally in Shadow).

sporksmith:
I was poking around networkx a bit to see if there's a way to produce nicer graphml output (with better `key` strings, in particular). There doesn't appear to be. However it *does* support other output formats, including some others [supported by igraph](https://igraph.org/c/doc/igraph-Foreign.html). Experimenting with `generate_example_config.py`, [gml](https://en.wikipedia.org/wiki/Graph_Modelling_Language) seems like a promising candidate:

```
graph [
  node [
    id 0
[...]

PR #1369, opened 2021-05-19
Switch from graphml to gml

stevenengler:
This changes igraph to read gml rather than graphml, the conversion scripts to write gml, and the test cases to use gml topologies. The conversion script now requires networkx version >= 2.5, so the config-convert test has been moved out of the required tests.

Part of #778.

Since igraph does not support graph attributes, the `prefers_direct_path` attribute is currently ignored. We plan to remove this attribute and [replace it](https://github.com/shadow/shadow/issues/778#issuecomment-843419313) with a shadow configuration option `use_shortest_path`.

### DIFF
--- a/ci/container_scripts/test.sh
+++ b/ci/container_scripts/test.sh
@@ -14,14 +14,16 @@ else
     EXCLUDE=""
 fi
 
-# On centos:7 we enable extra tests that currently require a patched libc.
-# https://github.com/shadow/shadow/issues/892
 EXTRA_FLAGS=""
 if [ "$CONTAINER" = "centos:7" ]
 then
+    # On centos:7 we enable extra tests that currently require a patched libc.
+    # https://github.com/shadow/shadow/issues/892
     CONFIG="ilibc"
 else
-    CONFIG=""
+    # On all other platforms, we run extra tests that we don't generally require (for
+    # example tests that require additional dependencies).
+    CONFIG="extra"
 fi
 
 # Array of flags to be passed on to setup script
@@ -38,6 +40,9 @@ FLAGS+=("-E" "$EXCLUDE")
 
 # Pass through an optional config-name, which can enable more tests
 FLAGS+=("-C" "$CONFIG")
+
+# Exclude tor tests as we test them in a different workflow
+FLAGS+=("-LE" "tor")
 
 FLAGS+=("--output-on-failure")
 
--- a/ci/container_scripts/test.sh
+++ b/ci/container_scripts/test.sh
@@ -36,13 +36,13 @@ FLAGS+=("-j$(nproc)")
 FLAGS+=("--")
 
 # We exclude some tests in some configurations.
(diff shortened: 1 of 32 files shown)


ci/container_scripts/test.sh
```diff
@@ -39,6 +41,9 @@ FLAGS+=("-E" "$EXCLUDE")
 # Pass through an optional config-name, which can enable more tests
 FLAGS+=("-C" "$CONFIG")
 
+# Exclude tor tests as we test them in a different workflow
+FLAGS+=("-LE" "tor")
```
robgjansen, line 45: Optional nit: prefer writing the long version of the flags if they exist (I don't remember what L and E mean).

src/main/routing/topology.c
```diff
@@ -353,7 +359,7 @@ static gboolean _topology_loadGraph(Topology* top, const gchar* graphPath) {
 
     _topology_lockGraph(top);
     info("reading graphml topology graph at '%s'...", graphPath);
```
robgjansen, line 361: Remove the "graphml" from the info log. (And maybe search to check if other log messages or comments refer to `graphml`).

src/test/config/convert/shadow.expected.yaml
```diff
@@ -1,28 +1,23 @@
...
-        </node>
-        <edge source="poi-1" target="poi-1">
-          <data key="d3">50.0</data>
-          <data key="d4">0.0</data>
-        </edge>
-      </graph>
-    </graphml>
+  gml: |
+    graph [
+      prefer_direct_paths 1
```
robgjansen, line 6: Does this need `directed 0` like shown in the other GML graphs?

robgjansen (approved):
Looks good! Minor issues only :)

src/test/config/convert/shadow.expected.yaml
stevenengler: The default for gml is undirected, so if the graph is undirected, networkx doesn't include the `directed` key. I'll update the xml to use the `edgedefault="directed"` attribute in this conversion test so that we can make sure networkx sets this correctly.

robgjansen: If you look at the other GML graphs in this commit, why do those list `directed 0`?

stevenengler: I hand-wrote those graphs and included it just to be explicit, but I can remove those if it's more clear.

robgjansen: Ahh, OK, this all makes sense now! I think you can just ignore this comment.

src/main/routing/topology.c
stevenengler: Updated these and also the temporary filename for the graph in controller.c.

ci/container_scripts/test.sh
stevenengler: Done.

PR #1374, opened 2021-05-24
Update network topology file documentation

### LINKED ISSUE #776: Clarify relationship between hosts and topology-nodes in documentation
Following up from discussion on #773 : I was looking over the docs to better understand the relationship between hosts and nodes. I have a few questions:

From https://github.com/shadow/shadow/blob/master/docs/3.1-Shadow-Config.md:

> bandwidthdown and bandwidthup optionally specify the downstream and upstream bandwidth capacities for this host, and override any default bandwidth values set in the cluster element corresponding to the cluster attribute. If not given, the default bandwidth values from the assigned cluster element are used.

As far as I can tell there is no `cluster` element. Is this left over from an old config format? Do these override the corresponding attributes on whatever node the host gets attached to?

> The iphint, geocodehint, and typehint attributes provide hints to Shadow's name and routing system on where to attach this host to the topology. These hints will be matched based on the values of the ip, geocode, and type of the vertices as specified in the topology file.

How is the matching done? Is the iphint used for an exact match against the ip address in the topology, or is there some kind of netmask logic? How do geocodehint and typehint work?

Presumably hosts get assigned IP addresses? Are those distinct from the node ip addresses in the topology?

robgjansen:
Closes #776

(I'm not sure if all of the links are working correctly, so we should double check that before merging.)

### DIFF
--- a/docs/3.2-Network-Config.md
+++ b/docs/3.2-Network-Config.md
@@ -1,48 +1,41 @@
-Shadow uses a graphml format to represent a network topology. The python module [networkx](http://networkx.github.io/) can be used to manipulate such a format. Shadow is distributed with some graphml topology files that were generated using data from a variety of sources. Because the topology is represented in a standard graphml format, it is easy to swap out specific measurements of latency, jitter, packetloss, etc, if desired.
+### Network Graph Overview
 
-### Vertices
+Processes running in Shadow do not have access to the internet; instead, processes running on Shadow virtual hosts utilize an internal routing module to communicate with other processes running on other virtual hosts in the simulation. The routing module is used to position virtual hosts within a network topology, to compute communication paths between virtual hosts, and to enforce network path characteristics like latency and packet loss.
 
-In addition to the default _id_ attribute, all vertices must have the _bandwidthup_ and _bandwidthdown_ attributes (given in KiB/s). Optional vertex attributes include _ip_, _citycode_, _countrycode_, and _packetloss_.
+Importantly, the routing module is currently used to _model_ the performance characteristics of internet paths; we do not _simulate_ the behavior of network routers (we do not run routing protocols like [BGP](https://en.wikipedia.org/wiki/Border_Gateway_Protocol)).
 
-You can control where hosts in your Shadow experiment get attached to the topology by setting the _iphint_, _citycodehint_, and _countrycodehint_ in your host attribute list in the Shadow config file. Hosts are always attached to the closest match to the best vertex following the hinted restrictions. See [3.1-Shadow-Config.md](3.1-Shadow-Config.md) for more details.
+This page describes the routing module and how it can be configured.
 
-### Edges
+#### Graph
 
-In addition to the default _source_ and _target_ attributes, all edges must have the _latency_, and _packetloss_ attributes. These are used when computing paths and routing packets between virtual hosts.
+Shadow represents a network topology over which processes can communicate using a [weighted graph](https://en.wikipedia.org/wiki/Graph_(discrete_mathematics)). The graph contains _vertices_ that abstractly represent network locations, and _edges_ representing network paths between those locations. Shadow requires that the graph is [connected](https://en.wikipedia.org/wiki/Connectivity_(graph_theory)) such that there exists at least one _path_ (a series of one or more edges) between every pair of vertices.
 
-### Routing
+#### Behavior
 
-If the topology is a complete graph, Shadow uses the single link between each vertex as the path. Otherwise, a routing path is approximated using Dijkstra's shortest path algorithm.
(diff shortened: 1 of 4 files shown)


stevenengler:
> (I'm not sure if all of the links are working correctly, so we should double check that before merging.)

If you click the `... > View File` button at the top-right of the file in the "Files changed" tab, you can view the rendered markdown. There are a couple of broken links, but I'll point them out in the review.

The tests are failing just because the new argument needs to be specified for `convert_topology` here:

https://github.com/shadow/shadow/blob/a0c1c8425a497e2e3e3e2c9e921ee8dd9e58b47f/src/tools/convert_legacy_config.py#L301-L307

docs/3.2-Network-Config.md
```diff
@@ -1,48 +1,39 @@
...
 
-In addition to the default _id_ attribute, all vertices must have the _bandwidthup_ and _bandwidthdown_ attributes (given in KiB/s). Optional vertex attributes include _ip_, _citycode_, _countrycode_, and _packetloss_.
+This page describes the routing module and how it can be configured.
 
-You can control where hosts in your Shadow experiment get attached to the topology by setting the _iphint_, _citycodehint_, and _countrycodehint_ in your host attribute list in the Shadow config file. Hosts are always attached to the closest match to the best vertex following the hinted restrictions. See [3.1-Shadow-Config.md](3.1-Shadow-Config.md) for more details.
+#### Graph
 
-### Edges
+Shadow represents a network topology over which processes can communicate using a [weighted graph](https://en.wikipedia.org/wiki/Graph_(discrete_mathematics)). The graph contains _vertices_ representing network routers, and _edges_ representing network paths between the routers. Shadow requires that the graph is [connected](https://en.wikipedia.org/wiki/Connectivity_(graph_theory)) such that there exists at least one _path_ (a series of one or more edges) between every pair of vertices.
```
stevenengler, line 9: One thing that originally confused me about the topology is that the vertices are only sort-of network routers. These vertices are used for determining the path of packets through the network, but attributes you set on these vertices don't act like attributes on a network router. For example if you set a bandwidth for a vertex, it's not actually a bandwidth for a router (since Shadow doesn't simulate routers). It's a default bandwidth for any hosts that are attached to the vertex. Would it be better to refer to them as just "network nodes" or "routing nodes"?

docs/network_graph_attributes.md
```diff
@@ -0,0 +1,152 @@
...
+- [`edge.jitter`](#edgejitter)
+- [`edge.packet_loss`](#edgepacket_loss)
+
+#### `graph.directed`
+
+Required: False  
+Default: `0`  
+Type: Integer
+
+Specifies the symmetry of the edges in the graph. It set to `0` (the default), the graph is an [undirected graph](https://en.wikipedia.org/wiki/Graph_(discrete_mathematics)): an edge between vertex `u` and vertex `v` is symmetric and can be used to construct a path both from `u` to `v` and from `v` to `u`. If set to `1`, the graph is a [directed graph](https://en.wikipedia.org/wiki/Directed_graph): an edge from vertex `u` to vertex `v` is assymmetric and can only be used to construct a path from `u` to `v` (a separate edge from `v` to `u` must be specified to compose a path in the reverse direction).
```
stevenengler, line 55: "It set to" -> "If set to"

docs/network_graph_attributes.md
```diff
@@ -0,0 +1,152 @@
...
+
+#### `vertex.ip_address`
+
+Required: False  
+Default: n/a
+Type: String
+
+An IP address at which the router represented by this vertex is located. This address can be used to control the placement of hosts in the network: after filtering vertices based on the city and country codes (as described above), we performance a [longest prefix match](https://en.wikipedia.org/wiki/Longest_prefix_match) on the remaining vertices by comparing the vertex `ip_address` with the host's [`ip_address_hint` host configuration value](3.1-Shadow-Config.md#host_defaultsip_address_hint) (if one is configured) and attach the host to the vertex with the closest match. We assign the host the address specified in `ip_address_hint` as long as that address has not yet been assigned to another host, otherwise we choose a unique address nearby to the requested address.
```
stevenengler, line 94: "we performance" -> "we perform"

docs/network_graph_attributes.md
```diff
@@ -0,0 +1,152 @@
...
+Type: String
+
+An optional, human-meaningful string description of the edge. The string may be used in log messages printed by Shadow.
+
+#### `edge.latency`
+
+Required: True  
+Type: Float
+
+The latency that will be added to packets traversing this edge. This value is used as a weight whle running Dijkstra's shortest path algorithm.
```
stevenengler, line 137: "whle" -> "while"

stevenengler (approved):

robgjansen:
Thanks for the review @stevenengler! I've fixed up all the mistakes you found and also clarified that we are not simulating routing but rather modeling network characteristics more abstractly.

PR #1376, opened 2021-05-25
Worker: make Rust Worker the canonical Worker object

sporksmith:
This is a step towards moving the object graph into Rust, so that we can
sort out ownership / sharing issues.

The Rust Worker object is currently just a thin wrapper around the
legacy C Worker object (now CWorker). It should be straightforward to
start moving more members out of the CWorker directly into the Worker,
but I'd like to prioritize getting all of the Shadow objects into at
least this state before converting any one object completely.

* The pieces of the Worker that are accessed from outside the Worker are
moved into parallel arrays in the WorkerPool. This allows the Worker
itself to be stored in a thread-local without any thread-safe wrappers,
for cheaper access.

* I removed support for 0-Workers, since it was already planned, and
it'd have further complicated the previous change.


src/main/host/tracker.h
```diff
@@ -25,6 +20,11 @@ enum _LogInfoFlags {
     LOG_INFO_FLAGS_RAM = 1<<2,
 };
 
+#include "main/core/support/definitions.h"
+#include "main/host/protocol.h"
+#include "main/routing/packet.h"
+#include "support/logger/log_level.h"
```
stevenengler, line 26: Should the `LogInfoFlags` enum be moved to a separate file so that it's not before the include statements?

stevenengler (approved):

src/main/host/tracker.h
sporksmith: Done. Oops, yeah meant to circle back and fix that.

PR #1380, opened 2021-05-26
Cleanup deprecated support for 0 worker threads

sporksmith:
Deprecation of support for 0-worker-threads was proposed in #1251 and
initially implemented in #1376. This cleans up some
no-longer-reachable-code.

In a similar spirit (that it's not worth supporting more code paths to
optimize a niche use-case), it also removes the special case of
not-shuffling the hosts when there is only 1 worker.


src/main/core/scheduler/scheduler_policy_type.h
```diff
@@ -7,8 +7,6 @@
 #define SHD_SCHEDULER_POLICY_TYPE_H_
 
 typedef enum {
-    /* one global unlocked priority queue */
-    SP_SERIAL_GLOBAL,
```
stevenengler, line 11: The reason why there was originally a C `SchedulerPolicyType` enum and a separate Rust `SchedulerPolicy` enum was that the configuration file didn't support specifying the serial global scheduler, but Shadow still needed this type internally. Now that the serial scheduler is removed and the two enums are the same, we could remove the `SchedulerPolicyType` enum and only use the Rust enum (with a `#[repr(C)]`). I don't think it's important for this PR, but just wanted to mention it.

stevenengler (approved):

PR #1385, opened 2021-05-28
More small documentation changes, and install cargo/rustc from rustup

stevenengler:
Mostly small changes, and the APT and DNF/YUM install instructions were re-ordered since the APT instructions are much simpler and shorter.

This also changes the installation docs to install cargo from rustup rather than from the distribution packages. On Debian stable the version of rustc is almost a year and a half old, and Ubuntu [doesn't recommend](https://wiki.ubuntu.com/FoundationsTeam/RustUpdates) using rustc and cargo from the distribution. Shadow currently uses features that aren't available in these versions, and I don't think there's a good reason to hold ourselves back to an older rust version.

### DIFF
--- a/docs/developer_guide.md
+++ b/docs/developer_guide.md
@@ -18,6 +18,8 @@ cargo install --force --version 0.18.0 cbindgen
 cargo install --force --version 0.57.0 bindgen
 ```
 
+The versions of bindgen and cbindgen you install should match the [versions installed in the CI](https://github.com/shadow/shadow/blob/dev/.github/workflows/lint.yml).
+
 ## Extra tests
 
 Shadow includes tests that require additional dependencies, such as Tor, TGen, and networkx. These aren't run by default, but are run as part of the CI tests. To run them locally, first make sure that both tor and tgen are located at `~/.shadow/bin/{tor,tgen}`. These can be symlinks to tor and tgen binaries elsewhere in the filesystem. You should also install all of Shadow's optional dependencies.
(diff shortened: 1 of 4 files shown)


docs/developer_guide.md
```diff
@@ -18,15 +18,21 @@ cargo install --force --version 0.18.0 cbindgen
 cargo install --force --version 0.57.0 bindgen
 ```
 
-## Tor tests
+The versions of bindgen and cbindgen you install should match the [versions installed in the CI](https://github.com/shadow/shadow/blob/dev/.github/workflows/lint.yml).
 
-Shadow includes tests that use Tor and TGen. These are not run by default. To run them, first make sure that both tor and tgen are located at `~/.shadow/bin/{tor,tgen}`. These can be symlinks to tor and tgen binaries elsewhere in the filesystem. It is recommended to build Shadow in release mode, otherwise the tests may take much longer to complete.
+## Extra tests
+
+Shadow includes tests that require additional dependencies (for example Tor, TGen, and networkx) and are not run by default, but are run as part of the CI tests. To run them locally, first make sure that both tor and tgen are located at `~/.shadow/bin/{tor,tgen}`. These can be symlinks to tor and tgen binaries elsewhere in the filesystem. You should also install all of Shadow's optional dependencies.
```
sporksmith, line 25: First sentence is a little awkward. Maybe break into two: "Shadow includes tests that require additional dependencies, such as Tor, TGen, and networkx. These aren't run by default, but are run as part of the CI tests."

sporksmith (approved):

docs/developer_guide.md
stevenengler: Fixed.

PR #1386, opened 2021-05-28
Pass references explicitly instead of getting from global Worker

sporksmith:
Adds a `context` module, which provides several *Context* structs,
intended to bundle together current relevant objects in the hierarchy.
These are meant to replace `worker_getActiveThread`, etc. Passing around
the current context explicitly instead of putting them in globals both
allows us to avoid interior mutability (and its associated runtime cost
and fallibility), and lets us keep a hierarchical object structure (e.g.
allow holding a mutable Process and a mutable Thread belonging to that
process simultaneously).

Most code (e.g. syscall handlers) can take a `ThreadContext` argument
and use that to manipulate anything on the Host. The *current* `Thread`
and `Process` should typically be accessed directly. e.g. since a
mutable reference to the current `Thread` exists at
`ThreadContext::thread`, it *cannot* also be accessible via
`ThreadContext::process` or `ThreadContext::host`.

The manner in which they're unavailable isn't implemented yet, but the
current plan is that they'll be temporarily removed from their
collections. e.g. something conceptually like:

```
impl Process {
    pub fn continue_thread(&mut self, host_ctx: &mut HostContext, tid: ThreadId) {
        let thread = self.threads.get_mut(tid).take();
        thread.continue(&mut host_ctx.add_process(self));
        self.threads.get_mut(tid).replace(thread);
    }
}
```

The Context objects are designed to allow simultaneously borrowing from
multiple of their objects.  This is currently implemented by exposing
their fields directly - Rust then allows each field to be borrowed
independently. This could alternatively be implemented by providing
methods that borrow some or all of their internal references
simultaneously.

Also:

* Move implementations of setActive* and getActive* to Rust Worker.  The
Rust Worker *doesn't* expose references to the active objects themselves
via the Rust API; only to some immutable info about those objects.

* The ShadowLogger now clones an Arc with the relevant Host info,
instead of copying the string name itself and rendering the IP address
from the calling thread. I wasn't sure whether this would actually be a
win or not, but experimentally `make test -j12` is ~2% faster cloning
the Arc than copying the data in both debug and release builds.

### DIFF
--- a/src/main/core/logger/shadow_logger.rs
+++ b/src/main/core/logger/shadow_logger.rs
@@ -1,5 +1,6 @@
 use crate::core::support::simulation_time::SimulationTime;
 use crate::core::worker::Worker;
+use crate::host::host::HostInfo;
 use crossbeam::queue::ArrayQueue;
 use log::{Level, Log, Metadata, Record, SetLoggerError};
 use log_bindings as c_log;
@@ -7,6 +8,7 @@ use once_cell::sync::Lazy;
 use std::cell::RefCell;
 use std::convert::TryFrom;
 use std::sync::mpsc::{Receiver, Sender};
+use std::sync::Arc;
 use std::sync::{Mutex, RwLock};
 use std::time::Duration;
 
@@ -244,15 +246,20 @@ impl ShadowLogger {
             } else {
                 write!(stdout, " n/a")?;
             }
+            write!(stdout, " [{level}]", level = record.level)?;
+            if let Some(host) = record.host_info {
+                write!(
+                    stdout,
+                    " [{hostname}:{ip}]",
+                    hostname = host.name,
+                    ip = host.default_ip,
+                )?;
+            } else {
+                write!(stdout, " [n/a]",)?;
+            }
             write!(
                 stdout,
-                " [{level}] [{host}] [{file}:",
-                level = record.level,
-                host = record
-                    .host_name
-                    .as_ref()
-                    .map(|s| s.as_str())
-                    .unwrap_or("n/a"),
+                " [{file}:",
                 file = record
                     .file
                     .map(|f| if let Some(sep_pos) = f.rfind('/') {
@@ -342,7 +349,7 @@ impl ShadowLogger {
 
 impl Log for ShadowLogger {
     fn enabled(&self, metadata: &Metadata) -> bool {
-        let filter = match Worker::with_active_host(|host| host.log_level()) {
+        let filter = match Worker::with_active_host_info(|host| host.log_level) {
             Some(Some(level)) => level,
             _ => log::max_level(),
         };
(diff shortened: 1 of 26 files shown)


src/main/core/worker.rs
```diff
@@ -1,188 +1,177 @@
...
-    pub fn with_active_process_mut<F, R>(f: F) -> R
-    where
-        F: FnOnce(&mut Process) -> R,
-    {
-        let mut process = unsafe {
-            let cprocess = cshadow::worker_getActiveProcess();
-            assert!(!cprocess.is_null());
-            Process::borrow_from_c(cprocess)
+    /// Set the currently-active Process.
+    pub fn set_active_process(process: &Process) {
```
stevenengler, line 116: If the active host is already set, should there be checks to make sure that this process belongs to that host (and same for the thread)?

src/main/core/worker.rs
```diff
@@ -1,188 +1,177 @@
...
+}
+
+struct ActiveThreadInfo {
+    #[allow(dead_code)]
+    id: ThreadId,
+    native_tid: Pid,
+}
+
 /// Worker context, capturing e.g. the current Process and Thread.
 pub struct Worker {
```
stevenengler, line 32: Is this comment still true? It seems like now the Worker is just a global cache for information about the current host, process, and thread, but not the host, process, or thread themselves (except for the legacy `c_` pointers)?

src/main/host/host.rs
```diff
@@ -1,7 +1,33 @@
...
+    // Store immutable info in an Arc, that we can safely clone into the Worker
+    // and ShadowLogger.
+    //
+    // Created on-demand for now, to avoid building unnecessarily for ephemeral
+    // Host objects.
+    //
+    // TODO: Consider caching a couple copies that we can "lend" out by value,
+    // without having to do any atomic operation?
+    info: OnceCell<Arc<HostInfo>>,
 }
```
stevenengler, line 31: A comment here might be useful to explain that the Rust `Host` is meant to be ephemeral, and there isn't a 1:1 relationship with a corresponding C `Host` (for the lifetime of a C `Host`, there will be many Rust `Host`s that point to it).

stevenengler (approved):
I'm having trouble following the design, I think there are too many layers of indirection to fit in my brain at once, but I know this is caused by integrating C with Rust and should become clearer as you continue moving things to Rust. So I'm good with just going with your judgement, and once it gets further along I'll look over all of the code to get a big-picture view again.

src/main/core/logger/shadow_logger.rs
```diff
@@ -244,15 +246,20 @@ impl ShadowLogger {
             } else {
                 write!(stdout, " n/a")?;
             }
+            write!(stdout, " [{level}]", level = record.level,)?;
```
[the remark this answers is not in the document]
sporksmith: Done

src/main/core/worker.rs
sporksmith: Done

PR #1390, opened 2021-06-01
Deny write access to all GitHub workflows

sporksmith:
GitHub recently added a mechanism for restricting the permissions of
what Workflow actions can do. Since none of our workflows need any
kind of write access to GitHub, it makes sense to only give them read
permission.

We could potentially tighten this further to only give read access to
some parts of Shadow, but since we currently don't keep any secrets in
GitHub, it doesn't seem worth trying to maintain more fine-grained
permissions.

### DIFF
--- a/.github/workflows/ci_maintenance.yml
+++ b/.github/workflows/ci_maintenance.yml
@@ -1,4 +1,5 @@
 name: CI Maintenance
+permissions: read-all
 
 on:
   schedule:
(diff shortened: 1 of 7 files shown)


stevenengler (approved):
Just leaving the link for future reference: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions

PR #1391, opened 2021-06-01
Raise fd limit to hard limit

stevenengler:
On my laptop, the file soft file limit is 1024 and the hard limit is 1,048,576, so I think it makes sense to increase the soft limit automatically as 1,048,576 should be enough for even medium-sized Tor simulations (a 1% network uses about 60,000 open files).


robgjansen (approved):
We think error makes sense so that we'll notice if something goes wrong when we don't expect it to.

PR #1393, opened 2021-06-01
Remove C usage of getActiveHost, getActiveProcess, getActiveThread

sporksmith:
This is conceptually a follow-up to #1386, but doesn't actually depend
on it. Once both this and #1386 are in, we can remove getActiveHost,
getActiveProcess, and getActiveThread entirely. This is helpful for
porting to Rust since we don't have or want those to be globals there.

In C code it suffices to pass around the "deepest" current context
object, since they all have "up" pointers. e.g. given a Thread we can
get the corresponding Process and Host. We might eventually want to
replace these parameters with their Context counterparts in #1386, but
this seems like a useful intermediate step.

### DIFF
--- a/src/main/routing/packet.c
+++ b/src/main/routing/packet.c
@@ -86,27 +86,32 @@ const gchar* protocol_toString(ProtocolType type) {
     }
 }
 
-Packet* packet_new(PluginVirtualPtr payload, gsize payloadLength, guint hostID, guint64 packetID) {
+Packet* packet_new(Host* host) {
     Packet* packet = g_new0(Packet, 1);
     MAGIC_INIT(packet);
 
     packet->referenceCount = 1;
 
-    packet->hostID = hostID;
-    packet->packetID = packetID;
-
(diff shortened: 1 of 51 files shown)


src/main/routing/packet.c
```diff
@@ -86,29 +86,32 @@ const gchar* protocol_toString(ProtocolType type) {
...
 
-    packet->hostID = hostID;
-    packet->packetID = packetID;
-
-    if (payload.val != 0 && payloadLength > 0) {
-        /* the payload starts with 1 ref, which we hold */
-        packet->payload = payload_new(payload, payloadLength);
-
-        /* application data needs a priority ordering for FIFO onto the wire */
-        packet->priority = host_getNextPacketPriority(worker_getActiveHost());
```
stevenengler, line 103: It doesn't look like the priority is set anywhere in the new code.

stevenengler (approved):

src/main/routing/packet.c
sporksmith: 😱 good catch!

PR #1397, opened 2021-06-03
Updated and reorganized 'getting_started.md'

stevenengler:
I split the 'getting_started.md' doc up into smaller documents and reorganized them a bit, and also added a shorter and hopefully more-relatable example before the TGen example.

Also removed the default topology `resource/topology.graphml.xml.xz`, and removed the `resource/examples` directory since I moved the contents into the documentation pages themselves.

You can view the new documents by clicking the "View file" button in the "Files changed" tab.


docs/getting_started.md
```diff
@@ -1,199 +1,102 @@
...
-```bash
-cd resource/examples/
-rm -rf shadow.data shadow.log qdisc-fifo.data qdisc-fifo.log qdisc-rr.data qdisc-rr.log
-shadow --interface-qdisc fifo --data-directory qdisc-fifo.data shadow.config.xml > qdisc-fifo.log
-shadow --interface-qdisc rr   --data-directory qdisc-rr.data   shadow.config.xml > qdisc-rr.log
-```
+### Simulation Output
 
-To parse these log files, we use the following scripts:
+Shadow will write simulation output to the `shadow.data` directory. Each host has its own directory under `shadow.data/hosts`. For example:
```
sporksmith, line 71: Something like:
to the data directory (in this example we'll assume the default value of `shadow.data`)
?

docs/getting_started.md
```diff
@@ -1,199 +1,102 @@
...
+<meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+<title>Directory listing for /</title>
+</head>
+<body>
+<h1>Directory listing for /</h1>
+...
 ```
 
-Then open the PDF file that was created to compare results from the experiments.
+Each host directory is also the [working directory](https://en.wikipedia.org/wiki/Working_directory) for the host's processes, which is why we specified `../../../server.py` as the path to the Python script in our Shadow configuration file (`./shadow.data/hosts/server/../../../server.py` → `./server.py`).
```
sporksmith, line 102: No action needed for this PR, but I wonder if we should consider some shortcut for "shadow's working directory" and/or "the directory of the shadow config file". e.g. `$CONFIGDIR/server.py`

docs/log_format.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Log Line Prefix
+
+Shadow produces simulator log messages in the following format:
+
+```text
+real-time [thread-id] virtual-time [logdomain-loglevel] [hostname~ip] [function-name] MESSAGE
+```
+
+**FIXME**: The `real-time` and `virtual-time` use different time units. Is this correct, or should we change this in Shadow?
```
sporksmith, line 11: You mean that real-time shows microsecond granularity, while virtual-time shows nano-second granularity? Yeah, I noticed and preserved this when porting the logger. I don't think finer granularity than microseconds for real-time is likely to be useful; and it probably makes sense to log sim time at nanosecond granularity since that's the internal granularity.

I'm inclined to leave as-is, though could be convinced to extend real-time to nanoseconds if we think it'd be helpful

Alternatively maybe it'd be clearer if we changed the separator after seconds to be `.`, so that the format of both could be expressed as `hours:minutes:seconds`? (Both fields should already be left-padded; e.g. 1 microsecond is 00:00:00:000001 not 00:00:00:1)

docs/log_format.md
```diff
@@ -0,0 +1,58 @@
...
+
+- `real-time`:  
+the wall clock time since the start of the experiment, represented as `hours:minutes:seconds:microseconds`
+- `thread-id`:  
+the ID of the worker thread that generated the message
+- `virtual-time`:  
+the simulated time since the start of the experiment, represented as `hours:minutes:seconds:nanoseconds`
+- `logdomain`:
```
sporksmith, line 19: We don't currently log such a field.

I *had* been thinking it might be useful to log contents of plugin writes to stdout, etc., in which case we could use such a field. OTOH it could also just be part of the log message itself.

OTOOH I had also been thinking it might make sense to log the current process and thread when applicable, though that might make more sense in the next field. e.g. `[hostname~ip.plugin-name.thread-id]`.

Assuming we don't want to hold up this change on new features, probably makes sense to just drop this field from the docs for now.

docs/log_format.md
```diff
@@ -0,0 +1,58 @@
...
+- `real-time`:  
+the wall clock time since the start of the experiment, represented as `hours:minutes:seconds:microseconds`
+- `thread-id`:  
+the ID of the worker thread that generated the message
+- `virtual-time`:  
+the simulated time since the start of the experiment, represented as `hours:minutes:seconds:nanoseconds`
+- `logdomain`:  
+either `shadow` or the name of one of the applications as specified in the _id_ tag of the _plugin_ element in the XML file (e.g., `tgen`, `tor`, `bitcoin`)
+- `loglevel`:  
+one of `error` < `critical` < `warning` < `message` < `info` < `debug`, in that order
```
sporksmith, line 22: I think this is now `ERROR` < `WARN` < `INFO` < `DEBUG` < `TRACE`

docs/parsing_statistics.md
```diff
@@ -0,0 +1,188 @@
+# Parsing Statistics
+
+Shadow logs simulator heartbeat messages that contain useful system information for each virtual node in the experiment. For example, Shadow logs the number of bytes sent/received, number of bytes allocated/deallocated, CPU usage, etc. You can parse these heartbeat log messages to get insight into the simulation. Details of these heartbeat messages can be found [here](log_format.md#heartbeat-messages).
+ 
+## Generating Traffic
+
+This example uses the [TGen traffic generator](https://github.com/shadow/tgen) to generate network traffic between hosts, and then shows results from Shadow's heartbeat messages.
+
+TGen is capable of modeling generic behaviors with an action-dependency graph in the standard GraphML format. If you don't have it installed, you can follow the [instructions here](https://github.com/shadow/tgen/#setup). The following example runs TGen with 10 clients that each download 10 files from a set of 5 servers over a simple network topology.
```
sporksmith, line 9: I only see 1 client and 1 server below? Should there be a `count` field?

sporksmith (approved):

docs/log_format.md
sporksmith: Actually the separator after seconds *is* `.`, not `:`

stevenengler: Oops, I meant to look into this before adding the PR. If it uses a `.` separator and is left-padded, then the current behaviour sounds good to me. I'll update the docs to just say `hours:minutes:seconds` for both.

docs/getting_started.md
stevenengler: Is this a tag we would look for in Shadow when parsing the config file?

docs/log_format.md
stevenengler: I removed this field, and also noticed the documentation was missing the `[file:line]` field so I added that to the documentation.

> I had been thinking it might be useful to log contents of plugin writes to stdout, etc., in which case we could use such a field. OTOH it could also just be part of the log message itself.

Do you mean logging shim messages to Shadow's log rather than the separate shim log file?

> OTOOH I had also been thinking it might make sense to log the current process and thread when applicable, though that might make more sense in the next field. e.g. [hostname~ip.plugin-name.thread-id].

Logging the plugin name and thread sounds like a good idea to me. A downside might be that the log lines are already very long and would get longer, but I think having more information is better than less.

docs/getting_started.md
```diff
@@ -1,199 +1,102 @@
...
-
-Since Shadow 2.0, applications can typically be run under Shadow without modification. Currently the only requirements are:
+The main Shadow binary executable, `shadow`, contains most of the simulator's code, including events and the event engine, the network stack, and the routing logic. Shadow's event engine supports multi-threading using the `-p` or `--parallelism` flags (or their corresponding [configuration file option](shadow_config.md#generalparallelism)) to simulate multiple hosts in parallel.
 
- + doesn't fork/exec processes
- + blocking calls (e.g. `sleep()`) are supported, but if nonblocking is used then it should poll I/O events using one of the `epoll`, `poll`, or `select` interfaces (see, e.g., `$ man epoll`)
+Shadow can typically run applications without modification, but there are a few limitations to be aware of:
 
-#### Traffic generation
+ - Applications must not fork or exec
```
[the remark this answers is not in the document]
stevenengler: That's a good way to word it. Added.

docs/getting_started.md
stevenengler: Added.

docs/log_format.md
stevenengler: Fixed.

docs/parsing_statistics.md
stevenengler: Yeah, I don't think the documentation follows the example. I changed the `count` filed in the TGen GraphML from 100 to 10, and added `count: 10` to the client host in the Shadow config yaml. I also changed the documentation to say 1 server.

PR #1398, opened 2021-06-03
Migrate WorkerC to Rust

sporksmith:
This PR migrates the rest of Worker's "top level" state into the Rust Worker object, allowing us to get rid of the legacy WorkerC object.

The Worker now directly owns a pointer to the WorkerPool, which isn't ported to Rust yet. Similarly, Worker methods that operate on the WorkerPool are still implemented in C (and work by fetching the WorkerPool* from the Rust Worker).


src/main/core/worker.rs
```diff
@@ -135,77 +148,80 @@ impl Worker {
...
+        Worker::with(|w| w.clock.now).flatten()
+    }
+
+    fn set_last_event_time(t: SimulationTime) {
+        Worker::with_mut(|w| w.clock.last.replace(t)).unwrap();
+    }
+
+    fn with<F, O>(f: F) -> Option<O>
+    where
+        F: FnOnce(&Worker) -> O,
```
stevenengler, line 209: Might be useful to have a comment explaining when this function returns `Some` and when it returns `None`.

stevenengler (approved):
> This PR migrates the rest of Worker's "top level" state into the Rust Worker object, allowing us to get rid of the legacy WorkerC object.

Cool! I think having this in Rust and using the context objects will be much nicer to work with.

src/main/core/worker.rs
sporksmith: Done

PR #1404, opened 2021-06-07
Change install prefix to '~/.local'

stevenengler:


robgjansen (approved):

robgjansen:
Do we want to change`src/test/tor/minimal/tor-minimal.yaml` so that each relay uses:
- `--defaults-torrc` to load common configs
- `-f` to load a local, node-specific config file (rather than a share conf in ../../config/blah.torrc)?

stevenengler:
> Do we want to change`src/test/tor/minimal/tor-minimal.yaml` so that each relay uses:
> 
>     * `--defaults-torrc` to load common configs
> 
>     * `-f` to load a local, node-specific config file (rather than a share conf in ../../config/blah.torrc)?

We talked about this in a meeting and decided to do this, first in this Shadow test, and later in tornettools. I'll leave this for a separate PR.

PR #1406, opened 2021-06-07
Updated example in 'parsing_statistics' doc

stevenengler:
The existing example comparing didn't show any difference between traffic queuing disciplines, so the new example uses different socket recv buffer sizes which show a noticeable difference. A small change was made to the plotting scripts to hide boundary effects when calculating a moving average.

### DIFF
--- a/docs/parsing_statistics.md
+++ b/docs/parsing_statistics.md
@@ -12,14 +12,26 @@ TGen is capable of modeling generic behaviors with an action-dependency graph in
 
 ```yaml
 general:
-  # stop after 1 simulated hour
-  stop_time: 3600
+  stop_time: 10 mins
 
 network:
   graph:
-    # use a built-in network graph containing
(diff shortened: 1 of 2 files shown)


docs/parsing_statistics.md
```diff
@@ -12,14 +12,27 @@ TGen is capable of modeling generic behaviors with an action-dependency graph in
 
 ```yaml
 general:
-  # stop after 1 simulated hour
-  stop_time: 3600
+  # stop after 600 simulated seconds
```
sporksmith, line 15: "stop after 10 minutes" ?

Actually might as well not repeat the number at all, especially now that the unit is in the yaml itself. Maybe just "How long to run the simulation", or drop completely

sporksmith (approved):

docs/parsing_statistics.md
stevenengler: Removed the comment.

PR #1407, opened 2021-06-09
Add units to topology latency

stevenengler:
The order of the weights generated by `_topology_extractEdgeWeights()` shouldn't change since the original function used `EANV()`, which is defined as:

https://github.com/igraph/igraph/blob/298c0ac9777869090de2b3bca94a4d17cd5564fa/include/igraph_attributes.h#L564-L565

```c
#define EANV(graph,n,vec) (igraph_cattribute_EANV((graph),(n), \
                           igraph_ess_all(IGRAPH_EDGEORDER_ID), (vec)))
```

And the new code uses:

```c
igraph_eit_create(&top->graph, igraph_ess_all(IGRAPH_EDGEORDER_ID), &edgeIterator)
```

So the edge selector is the same.

### DIFF
--- a/docs/network_graph_attributes.md
+++ b/docs/network_graph_attributes.md
@@ -22,7 +22,7 @@ graph [
     source 0
     target 0
     label "path from 1.2.3.4 to 1.2.3.4"
-    latency 10.0
+    latency "10 ms"
     jitter 0.0
     packet_loss 0.0
   ]
@@ -132,9 +132,9 @@ An optional, human-meaningful string description of the edge. The string may be 
 #### `edge.latency`
 
 Required: True  
-Type: Float
+Type: String
 
-The latency that will be added to packets traversing this edge. This value is used as a weight while running Dijkstra's shortest path algorithm.
+The latency that will be added to packets traversing this edge. This value is used as a weight while running Dijkstra's shortest path algorithm. The format of the string specifies the latency and its unit, e.g., `10 ms`. If a unit is not specified, it will be assumed that it is in the base unit of "seconds".
 
 #### `edge.jitter`
 
--- a/src/main/core/support/configuration.rs
+++ b/src/main/core/support/configuration.rs
@@ -724,15 +724,13 @@ const ONE_GBIT_SWITCH_GRAPH: &str = r#"graph [
   node [
     id 0
     ip_address "0.0.0.0"
-    country_code "XX"
     bandwidth_up "1 Gbit"
     bandwidth_down "1 Gbit"
   ]
   edge [
     source 0
     target 0
-    latency 1.0
-    jitter 0.0
+    latency "1 ms"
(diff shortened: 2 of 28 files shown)


src/main/core/support/configuration.rs
```diff
@@ -724,15 +724,13 @@ const ONE_GBIT_SWITCH_GRAPH: &str = r#"graph [
...
     ip_address "0.0.0.0"
-    country_code "XX"
     bandwidth_up "1 Gbit"
     bandwidth_down "1 Gbit"
   ]
   edge [
     source 0
     target 0
-    latency 1.0
-    jitter 0.0
```
robgjansen, line 735: I think we currently documented jitter as unused, but do we validate the latency value that gets configured? If we validate a specific format, then it would be better to make sure the format is in the correct (ie, follows the same format as latency). But if we don't validate it, then I think we can just ignore the format for now (since we documented jitter as unused).

docs/network_graph_attributes.md
```diff
@@ -132,9 +132,9 @@ An optional, human-meaningful string description of the edge. The string may be
 #### `edge.latency`
 
 Required: True  
-Type: Float
+Type: String
 
-The latency that will be added to packets traversing this edge. This value is used as a weight while running Dijkstra's shortest path algorithm.
+The latency that will be added to packets traversing this edge. This value is used as a weight while running Dijkstra's shortest path algorithm. The format of the string specifies the latency and its unit, e.g., `10 ms`. If a unit is not specified, it will be assumed that it is in the base unit of "seconds".
```
robgjansen, line 137: Do we document anywhere what the acceptable time units are? For example, ms, milliseconds, s, seconds, etc?

I mean time units in general, whenever a time is given (either here or in other parts of the config such as heartbeatinterval).

robgjansen (approved):

src/main/core/support/configuration.rs
stevenengler: Right now if 'jitter' is provided, shadow checks that it's numeric, so maybe we should change that to a string? And we do verify the format of 'latency', but it requires a unit. If we want to validate the 'jitter' field, what unit do we want it to use?

robgjansen: I'm wondering if it's better to allow it to be added to the graph but not validate it since I don't think it's used anywhere internally in Shadow right now. That way they can put whatever they want there and we ignore it for now, and maybe someday we can start using it.

stevenengler: Sounds good to me. Added in #1415.

docs/network_graph_attributes.md
stevenengler: We don't currently. I added this as a todo item in the 2.0 pre.1 project and will do it in a new PR.

PR #1411, opened 2021-06-09
Reorganize tor config files in tor minimal test

stevenengler:
This adds a `torrc-defaults` file for each host in the template directory. These files include `conf/tor.common.torrc` and one or more more `conf/tor.something.torrc` files. Since the bandwidth values are the same for each host, they were moved into `conf/tor.common.torrc`. This simplifies the Shadow config file.

The Shadow config file still has some command line options such as `--Nickname exit2` and `--Bridge 100.0.0.1:9111` that we could move into individual `torrc` files for each host in the template directory, but I decided to leave them in the Shadow config since these options use values that are specified elsewhere in the Shadow config. For example the `Nickname` should be the same as the Shadow host name, and the `Bridge` IP address should be the same as another host in the config file that specifies `ip_address_hint`.


robgjansen (approved):
This is great, thanks! Your reasons for keeping some of the Tor options in the shadow config make sense.

(Note that when setting up the structure for tornettools, we should include a `middle` and a `guard` or `entry` torrc files, and maybe remove the `non-exit` torrc; I could imagine wanting to configure something only on the guards but not the middles, and the `non-exit` torrc does not allow for that.)

stevenengler:
> This is great, thanks! Your reasons for keeping some of the Tor options in the shadow config make sense.
> 
> (Note that when setting up the structure for tornettools, we should include a `middle` and a `guard` or `entry` torrc files, and maybe remove the `non-exit` torrc; I could imagine wanting to configure something only on the guards but not the middles, and the `non-exit` torrc does not allow for that.)

One issue with doing it that ways is that there isn't really a way of specifying a relay as only a middle, since all relays (excluding exits, but that's a consensus setting and could technically change) will also be used as guards.

PR #1412, opened 2021-06-10
Make a pass through docs for prerelease

sporksmith:


docs/install_shadow.md
```diff
@@ -5,12 +5,17 @@ After building and testing Shadow, the install step is optional. If you do not w
 ```bash
 git clone https://github.com/shadow/shadow.git
 cd shadow
-./setup build --clean --test
-./setup test
+./setup build --clean --test -j`nproc`
+# Optional, but recommended to validate your build and environment.
+./setup test -j`nproc`
+# Optionally install (to ~/.local/bin by default). Can otherwise run the binary
+# directly at build/src/main/shadow.
```
stevenengler, line 12: I don't think we need the `-j$(nproc)` since the setup script should do this automatically:

https://github.com/shadow/shadow/blob/601c10933ec67f1d8bd1f7c18f91533be1a35709/setup#L130-L135

------

Also since Shadow is sensitive to things like container environments and possibly kernel versions (since we don't test that in our CI), I think it's good to not list the testing step as optional. If people don't want to run it then they won't regardless of what we say, but I think maybe people will run it more often if we don't say anything here.

docs/network_graph_attributes.md
```diff
@@ -4,7 +4,7 @@ The [network graph overview](network_config.md) provides a general summary of Sh
 
 ### Example Graph
 
-Below is an example of a simple network graph in the Shadow-supported GML format (note that GML calls _vertices_ as _nodes_).
+Below is an example of a simple network graph in the Shadow-supported GML format.
```
stevenengler, line 7: I think Rob had preferred `vertex` over `node` so that people don't get "node" and "host" mixed up, but the GML docs below should probably use "node" since that is the keyword used in GML. Maybe a comment can be added here to make sure the user doesn't confuse a node with a host.

docs/using_recompiled_libc.md
```diff
@@ -124,13 +128,16 @@ cd ~/rpmbuild/SPECS && rpmbuild -ba glibc.spec
...
+The patched libc can be injected into a plugin by setting `LD_LIBRARY_PATH` in each target plugin's environment. e.g.:
 
 ```
-<shadow environment="LD_LIBRARY_PATH=/path/to/glibc-2.27/build-tree/amd64-libc">
+...
+hosts:
+  my_host:
+    processes:
+      - path: /path/to/process
+      - environment: LD_LIBRARY_PATH=/path/to/glibc-2.27/build-tree/amd64-libc
```
stevenengler, line 139: There shouldn't be a `-` before environment (they're part of the same map/dictionary).

stevenengler (approved):

docs/network_graph_attributes.md
sporksmith: Hmm, yeah on second thought I think I'll revert this change and go back to vertex. The mismatch with "node" vs "vertex" in GML is a little unfortunate, but I agree "node" is maybe too ambiguous a term in general.

docs/using_recompiled_libc.md
sporksmith: Fixed

PR #1415, opened 2021-06-10
Ignore the jitter graph attribute

stevenengler:
See https://github.com/shadow/shadow/pull/1407#discussion_r648738932.

### DIFF
--- a/src/main/routing/topology.c
+++ b/src/main/routing/topology.c
@@ -590,7 +590,7 @@ static gboolean _topology_checkGraphAttributes(Topology* top) {
         if(_topology_isValidEdgeAttributeKey(name, EDGE_ATTR_LATENCY)) {
             isSuccess = isSuccess && _topology_checkAttributeType(name, type, IGRAPH_ATTRIBUTE_NUMERIC);
         } else if(_topology_isValidEdgeAttributeKey(name, EDGE_ATTR_JITTER)) {
-            isSuccess = isSuccess && _topology_checkAttributeType(name, type, IGRAPH_ATTRIBUTE_NUMERIC);
+            // we don't yet use the 'jitter' attribute, so ignore it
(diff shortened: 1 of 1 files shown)


sporksmith (approved):
This is probably an improvement over the current state (IIUC it would have rejected a jitter that included a unit?), though maybe it'd be better still to validate that it's a number + unit, so that we don't accept bad values (e.g. jitter "ack thpppt"), causing the corresponding topologies to stop working if/when we do start using jitter.

stevenengler:
> (IIUC it would have rejected a jitter that included a unit?)

Yep, it would have rejected any non-numeric value.

> though maybe it'd be better still to validate that it's a number + unit, so that we don't accept bad values (e.g. jitter "ack thpppt"), causing the corresponding topologies to stop working if/when we do start using jitter.

I agree, but the difficulty is that the number + unit validation requires you to specify the unit, and I don't know if we've decided on what unit we'll use for jitter. We could add custom validation code just for jitter that uses any unit, but maybe it would be better to just reject 'jitter' outright, and then when we enable it in the future, all existing topologies will still work?

stevenengler:
> > but the difficulty is that the number + unit validation requires you to specify the unit, and I don't know if we've decided on what unit we'll use for jitter.
> 
> Can't we accept any supported unit and later convert internally to e.g. nanos?

I'm used to jitter values being unit-less, but I think network jitter is always expressed in units of time? If so then we can parse it as a time value.

sporksmith:
OH yeah that's what I was thinking. It hadn't occurred to me that it wouldn't be a time unit, but I guess it could be a %.

But yeah, I'm in favor of just deciding now that it's time

stevenengler:
Closing, and once #1416 is merged, I'll reuse the time parsing function to parse the jitter which will validate the value.

PR #1416, opened 2021-06-10
Accept nanoseconds in the topology latency

stevenengler:
The atlas topology uses floating point latency values, so we need to support a higher precision than just milliseconds. Rounding to microseconds should be enough precision, so we should support microsecond values in Shadow.

### DIFF
--- a/src/main/bindings/c/bindings.h
+++ b/src/main/bindings/c/bindings.h
@@ -249,8 +249,8 @@ SimulationTime processoptions_getStopTime(const struct ProcessOptions *proc);
 // Parses a string as bits-per-second. Returns '-1' on error.
 int64_t parse_bandwidth(const char *s);
 
-// Parses a string as a time in milliseconds. Returns '-1' on error.
-int64_t parse_time_ms(const char *s);
+// Parses a string as a time in nanoseconds. Returns '-1' on error.
+int64_t parse_time_nanosec(const char *s);
 
 // Initialize a Worker for this thread.
 void worker_newForThisThread(WorkerPool *worker_pool,
--- a/src/main/core/support/units.rs
+++ b/src/main/core/support/units.rs
@@ -796,9 +796,9 @@ mod export {
         rv
     }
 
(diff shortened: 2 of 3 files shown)


sporksmith (commented):
It'd be nice if we could avoid dealing with floating point numbers. We're already stuck with them to some degree since that's what's in the atlas files, but wdyt of changing the internal unit to be microsec or even nanosec so that we at least don't need to use floating point internally? Maybe not worth it if it'd be a lot of work, though.

Btw from a quick glance at the latencies, it looks like some are actually sub-microsecond granularity. I suppose rounding to microseconds is probably fine though; I doubt the measurement error is < a microsecond.

```
$ xzcat ~/projects/tmodel-ccs2018.github.io/data/shadow/network/latency/2018-all-pairs.csv.xz | head -n2
id,src,src_city,src_country,dst,dst_city,dst_country,latency
0,62.251.17.184,2759794,NL,185.104.140.104,2938913,DE,4.4538675
```

stevenengler:
> It'd be nice if we could avoid dealing with floating point numbers. We're already stuck with them to some degree since that's what's in the atlas files, but wdyt of changing the internal unit to be microsec or even nanosec so that we at least don't need to use floating point internally? Maybe not worth it if it'd be a lot of work, though.

I agree, but since the plan is to move the topology code over to Rust, I figured I'd put this off until then.

> Btw from a quick glance at the latencies, it looks like some are actually sub-microsecond granularity. I suppose rounding to microseconds is probably fine though; I doubt the measurement error is < a microsecond.
> 
> ```
> $ xzcat ~/projects/tmodel-ccs2018.github.io/data/shadow/network/latency/2018-all-pairs.csv.xz | head -n2
> id,src,src_city,src_country,dst,dst_city,dst_country,latency
> 0,62.251.17.184,2759794,NL,185.104.140.104,2938913,DE,4.4538675
> ```

Since we're using i64 and double, I might as well make Shadow support nanoseconds, regardless of what we use in the atlas topology. I'll make this change.

sporksmith (approved):

PR #1417, opened 2021-06-10
Plugin -> Managed Process

sporksmith:
"Plugin" no longer makes sense for Shadow 2.0. Initially we talked about using "managed process" anywhere we would otherwise say "plugin". On further discussion, "virtual" process is a little better when talking about the more abstract concept of a process in the simulation vs the native process used to help implement it.

### DIFF
--- a/docs/developer_guide.md
+++ b/docs/developer_guide.md
@@ -59,15 +59,15 @@ gdb --pid=PID
 > continue
 ```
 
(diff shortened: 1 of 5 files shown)


stevenengler (approved):

robgjansen (approved):
Looks reasonable!

PR #1420, opened 2021-06-11
Add CI lint to check Shadow version in Cargo lockfile

stevenengler:
When we update Shadow's version in the Cargo.toml files, we also want to make sure that the Cargo.lock files are also updated. This lint checks only our package versions and not versions of our dependencies.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -40,6 +40,19 @@ jobs:
           (cd src/test && cargo fmt -- --check)
           (cd src/support/logger/rust_bindings && cargo fmt -- --check)
 
+  lint-cargo-lock:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Cargo update check
+        run: |
+          # This will return an error if any versions of local crates in the Cargo.lock
+          # are out of date compared to the crate versions in Cargo.toml. This can fail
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -40,6 +40,17 @@ jobs:
           (cd src/test && cargo fmt -- --check)
           (cd src/support/logger/rust_bindings && cargo fmt -- --check)
 
+  lint-cargo-lock:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Cargo update check
+        run: |
+          (cd src/main && cargo update --package shadow-rs)
```
sporksmith, line 49: Could probably use a comment explaining what this is trying to catch (and maybe what to do if this test fails).

I think it's ensuring that after we bump Shadow's version we run these commands to sync the version in the Cargo.lock files?

sporksmith (approved):

.github/workflows/lint.yml
stevenengler: I added a comment, and also changed the method that it uses to check since I found a better way. I also updated the "maintainer_playbook.md" doc to explain how to update the lock file versions when updating the Shadow version.

I'm remarking you as reviewer.

sporksmith (approved):

PR #1434, opened 2021-06-15
Port LogicalProcessors to Rust

sporksmith:
This was mostly an experiment of using c2rust to translate a single C file.

Also fixes:

* Attempted use of non-existent log level "message" when compiling with `--use-perf-timers`.
* Failing to take 'quantity' into account when counting the number of hosts. This led to creating too few worker threads in simulations that had hosts with quantity > 1.


src/main/utility/perf_timer.rs
```diff
@@ -0,0 +1,43 @@
+use basic_timer::{DurationType, Timer};
+use std::convert::TryInto;
+use std::time::Duration;
```
stevenengler, line 3: I think it's worth considering having our own timer rather than using basic_timer. The code for basic_timer is only 29 lines, and it doesn't have a license.

src/main/core/support/configuration.rs
```diff
@@ -1371,6 +1370,19 @@ mod export {
...
+    pub extern "C" fn config_getNHosts(config: *const ConfigOptions) -> u32 {
+        assert!(!config.is_null());
+        let config = unsafe { &*config };
+
+        let mut count = 0u32;
+        for (_name, host) in &config.hosts {
+            count += hostoptions_getQuantity(host as *const HostOptions)
+        }
+
+        count
```
stevenengler, line 1383: (Bikeshedding) Maybe nicer as just:

```
config.hosts.iter().map(|(_, host)| hostoptions_getQuantity(host)).sum()
```

src/main/bindings/c/bindings.h
```diff
@@ -201,6 +226,8 @@ void config_iterHosts(const struct ConfigOptions *config,
                       void (*f)(const char*, const struct ConfigOptions*, const struct HostOptions*, void*),
                       void *data);
 
+uint32_t config_getNHosts(const struct ConfigOptions *config);
+
```
stevenengler, line 230: This doesn't seem to be used from C anywhere. Do you plan to use this later, or can we avoid exporting it?

src/main/core/logical_processor.h
```diff
@@ -14,46 +14,44 @@
 /*
  * Logical processor on which a WorkerPool runs Worker threads.
  */
```
stevenengler, line 16: The `lps_` declarations in this file `src/main/core/logical_processor.h` are duplicated with `src/main/bindings/c/bindings.h`. Do we still need `src/main/core/logical_processor.h`?

src/main/core/logical_processor.rs
```diff
@@ -0,0 +1,183 @@
...
+impl LogicalProcessors {
+    pub fn new(n: usize) -> Self {
+        let mut lps = Vec::new();
+        for _ in 0..n {
+            lps.push(LogicalProcessor {
+                cpu_id: unsafe { cshadow::affinity_getGoodWorkerAffinity() },
+                ready_workers: SegQueue::new(),
+                done_workers: SegQueue::new(),
+                idle_timer: Mutex::new(PerfTimer::new()),
+            });
```
stevenengler, line 29: Performance timers were a compile-time option in the C code since they added a noticeable amount of overhead. Are these performance timers on a hot path? If so, should we make these enabled only when a Cargo "perf-timers" feature is configured?

src/main/core/logical_processor.rs
```diff
@@ -0,0 +1,183 @@
...
+    use super::*;
+
+    #[no_mangle]
+    pub extern "C" fn lps_new(n: libc::c_int) -> *mut LogicalProcessors {
+        Box::into_raw(Box::new(LogicalProcessors::new(n.try_into().unwrap())))
+    }
+    #[no_mangle]
+    pub unsafe extern "C" fn lps_free(lps: *mut LogicalProcessors) {
+        unsafe { Box::from_raw(lps) };
+    }
```
stevenengler, line 106: Check for null pointer here?

stevenengler (changes requested):
Marked as "request changes" mostly for the performance effects of not making the idle timer a compile-time option, and for the basic_timer code not having a license (I'm not sure how important this is in reality).

src/main/core/logical_processor.rs
sporksmith: Done

src/main/core/logical_processor.rs
sporksmith: Yeah, it doesn't make much difference in the current version since it's still up to the (C) caller to stop and start the timers. We'll need to tackle it eventually though; done.

src/main/core/logical_processor.h
sporksmith: Yeah, I was thinking of leaving it there since it's a little nicer to still have the "hand crafted" C header file for reference; we should get compilation errors if any of the declarations don't match those in the auto-generated header file. OTOH maybe having them in both places is confusing, and I suppose could lead to trouble if somehow no C files included both versions of the header. Removed.

src/main/bindings/c/bindings.h
sporksmith: No specific plans, but it doesn't hurt; IMO it makes sense to just be consistent with the other definitions here.

src/main/core/support/configuration.rs
sporksmith: I like it; done

src/main/utility/perf_timer.rs
sporksmith: Good point; done

stevenengler (approved):

Issue #1168, opened 2021-03-10
hybrid mode via seccomp


sporksmith:
Hybrid mode on the dev branch (currently e48ca0d3634c046a3c39583c4e08b496d2ad94f2) is very slow.

Hybrid mode attempts to be faster than ptrace mode by avoiding a syscall (and therefore a ptrace-stop) in communication between Shadow and the shim. This only happens when the plugin's request to shadow is responded to before the plugin reaches it's spin limit and falls back to a blocking `sem_wait`. If the plugin ends up calling `sem_wait`, then it'll end up making a `futex` syscall, which will result in a ptrace-stop. When that happens, the syscall ends up having strictly *more* overhead than in ptrace mode, since now it has to deal with the ptrace-stop hybrid mode was meant to avoid in the first place *and* handle the IPC request from the shim.

In hybrid mode's current form, this problem will be exacerbated in the configuration we now believe to be most performant: using as many workers as (non-hyperthreaded) CPUs, with plugin processes pinned to the same CPU as its worker. In this case the Shadow worker  won't get a chance to run to serve an IPC request until either the plugin process is pre-empted by the kernel (which will never happen if using sched_fifo), or it falls back to the blocking `sem_wait`. i.e. in this configuration hybrid mode will *never* be faster than ptrace mode.

sporksmith:
In [DetTrace](https://dl.acm.org/doi/10.1145/3373376.3378519), they use seccomp to allow a tracee to make some syscalls without a ptrace stop.

IIUC such policies can examine syscall arguments, so we could for example allow writes to a pipe of a specific file descriptor.

From ptrace(2):

```
   PTRACE_EVENT_SECCOMP stops (since Linux 4.8)
       Starting with Linux 4.8, the PTRACE_EVENT_SECCOMP stop was reordered to occur between syscall-entry-stop  and  syscall-exit-stop.
       Note  that  seccomp  no  longer  runs  (and  no  PTRACE_EVENT_SECCOMP  will  be  reported)  if  the system call is skipped due to
       PTRACE_SYSEMU.

       Functionally, a PTRACE_EVENT_SECCOMP stop functions comparably to a syscall-entry-stop (i.e., continuations using  PTRACE_SYSCALL
       will  cause  syscall-exit-stops, the system call number may be changed and any other modified registers are visible to the to-be-
       executed system call as well).  Note that there may be, but need not have been a preceding syscall-entry-stop.

[...]

Issue #892, opened 2020-07-16
Get all tests working in preload mode without recompiled libc


robgjansen:
In our multi-process prototype, we are using a recompiled libc in preload mode to ensure that we are handling all syscalls correctly. Ultimately, I don't think we want to require recompiling libc of users. Instead, I think we want to get all tests working without preloading the recompiled libc.

https://github.com/shadow/shadow/blob/dev/docs/using-recompiled-libc.md

PR #1452, opened 2021-06-17
thread_preload: use seccomp to interpose all syscalls

sporksmith:
Installs a seccomp filter from the shim in preload-mode, which catches any syscalls that would otherwise be missed. The filter allows syscalls that originate from the shim itself. Other syscalls cause a SIGSYS; in the SIGSYS handler we can process the syscall as appropriate (e.g. emulate it via Shadow). Can disable with `--use-seccomp=false`

Semi-fixes https://github.com/shadow/shadow/issues/1168. This is a somewhat different approach than described there, but if this approach works out, then we probably don't need the one described in #1168.

Recompiled libc no longer needed to interpose remaining syscalls, fixing https://github.com/shadow/shadow/issues/892

### DIFF
--- a/src/shim/shim.c
+++ b/src/shim/shim.c
@@ -3,6 +3,8 @@
 #include <assert.h>
 #include <errno.h>
 #include <inttypes.h>
+#include <linux/filter.h>
+#include <linux/seccomp.h>
 #include <pthread.h>
 #include <search.h>
 #include <stddef.h>
@@ -10,11 +12,13 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/prctl.h>
+#include <sys/syscall.h>
 #include <sys/types.h>
 #include <time.h>
 #include <unistd.h>
 
 #include "shim/ipc.h"
+#include "shim/preload_syscall.h"
 #include "shim/shim_event.h"
 #include "shim/shim_logger.h"
 #include "shim/shim_syscall.h"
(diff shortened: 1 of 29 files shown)


src/shim/shim.c
```diff
@@ -298,6 +302,125 @@ static void _shim_parent_init_ptrace() {
...
+                  },
+                  NULL) < 0) {
+        panic("sigaction: %s", strerror(errno));
+    }
+    sigset_t mask;
+    sigemptyset(&mask);
+    sigaddset(&mask, SIGSYS);
+    if (sigprocmask(SIG_UNBLOCK, &mask, NULL)) {
+        panic("sigprocmask: %s", strerror(errno));
+    }
```
stevenengler, line 340: Signal masks are per-thread I believe. Should this be run once for each thread?

src/shim/shim.c
```diff
@@ -298,6 +302,125 @@ static void _shim_parent_init_ptrace() {
...
+                                 ctx->uc_mcontext.gregs[REG_R10], ctx->uc_mcontext.gregs[REG_R8],
+                                 ctx->uc_mcontext.gregs[REG_R9]);
+    ctx->uc_mcontext.gregs[REG_RAX] = rv;
+}
+
+static void _shim_parent_init_seccomp() {
+    // Install signal sigsys signal handler, which will receive syscalls that
+    // get stopped by the seccomp filter.
+    if (sigaction(SIGSYS,
+                  &(struct sigaction){
```
stevenengler, line 325: Does it make sense to check for an existing `SIGSYS` handler first? It might make sense to emit a warning in that case.

src/shim/shim.c
```diff
@@ -298,6 +302,125 @@ static void _shim_parent_init_ptrace() {
...
+                      .sa_sigaction = _handle_sigsys,
+                      // SA_NODEFER: Allow recursive signal handling, to handle a syscall
+                      // being made during the handling of another.
```
stevenengler, line 328: For `SA_NODEFER`, is there an expected case where we'd receive a signal while handling another? It seems a lot safer to not do this unless we need to.

src/shim/shim.c
```diff
@@ -298,6 +302,125 @@ static void _shim_parent_init_ptrace() {
...
+        panic("prctl: %s", strerror(errno));
+    }
+
+    /* A bpf program to be loaded as a `seccomp` filter. Unfortunately the
+     * documentation for how to write this is pretty sparse. There's a useful
+     * example in samples/seccomp/bpf-direct.c of the Linux kernel source tree.
+     * The best reference I've been able to find is a BSD man page:
+     * https://www.freebsd.org/cgi/man.cgi?query=bpf&sektion=4&manpath=FreeBSD+4.7-RELEASE
+     */
+    struct sock_filter filter[] = {
```
stevenengler, line 354: Looks like a fun time figuring this out :)

src/shim/shim_syscall.c
```diff
@@ -78,8 +78,7 @@ bool shim_syscall(long syscall_num, long* rv, va_list args) {
                 *rv = 0;
             } else {
                 trace("found NULL timespec pointer in clock_gettime");
-                *rv = -1;
-                errno = EFAULT;
+                *rv = -EFAULT;
```
stevenengler, line 81: Does this change cause an issue when preload mode is used without seccomp (does the errno still need to be set somewhere)?

stevenengler (changes requested):

src/shim/shim.c
sporksmith: > Does it make sense to check for an existing SIGSYS handler first? It might make sense to emit a warning in that case.

Done

> Is there anything nice we can do if an application tries to do its own sandboxing and overwrites the sigsys signal handler (for example if someone runs tor with sandboxing enabled)?

I *think* the virtual process using seccomp itself won't *necessarily* interfere with our usage. Additional filters get added on; old ones never get removed. (I suppose if it blocks the syscalls we use for IPC things would break)

We would have a problem if our SIGSYS handler got blocked or overridden, though. For now I implemented handlers in the related syscalls to prevent that from happening. We *might* be able to handle an attempt to set a sigaction more transparently by having our handler also call the specified handler, but probably not worth it right now.

src/shim/shim.c
sporksmith: We need this for shim-side logging to work properly. Added a comment.

src/shim/shim.c
sporksmith: This code runs in the first thread, and the mask is inherited by new threads. Added a comment.

src/shim/shim.c
```diff
@@ -298,6 +302,125 @@ static void _shim_parent_init_ptrace() {
...
+        abort();
+    }
+    // Make the syscall via the *the shim's* syscall function (which overrides
+    // libc's).  It in turn will either emulate it or (if interposition is
+    // disabled), or make the call natively. In the latter case, the syscall
+    // will be permitted to execute by the seccomp filter.
+    long rv = shadow_raw_syscall(ctx->uc_mcontext.gregs[REG_RAX], ctx->uc_mcontext.gregs[REG_RDI],
+                                 ctx->uc_mcontext.gregs[REG_RSI], ctx->uc_mcontext.gregs[REG_RDX],
+                                 ctx->uc_mcontext.gregs[REG_R10], ctx->uc_mcontext.gregs[REG_R8],
+                                 ctx->uc_mcontext.gregs[REG_R9]);
```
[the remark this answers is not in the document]
stevenengler: Does the logging subsystem have comments anywhere saying that it shouldn't allocate memory? What about things like the `Vec` allocation in `rustlogger_log()` -> `vsprintf_raw()`?

stevenengler (approved):

PR #1453, opened 2021-06-17
Bump Tor versions in the CI

robgjansen:


robgjansen:
In the Tor 0.4.6.5 test that failed, there are 10 packet objects (and 10 payload objects) that did not get `free()`d. So the test fails due to a memory leak detected.

stevenengler:
The tor commit that breaks shadow is torproject/tor@2444629c.

stevenengler:
The packet leak issue was solved, but was found to be different from this issue.

This issue is at #1503.

stevenengler:
@robgjansen Could you rebase this, I think it's ready to merge.

stevenengler (approved):

Issue #1455, opened 2021-06-17
Handle or avoid nested syscalls in preload mode


sporksmith:
The reason our `signal` test currently doesn't pass for preload-mode is that:

* The virtual process calls kill(getpid(), SIGUSR1)
* Shadow sends a SHD_SHIM_EVENT_SYSCALL request, asking the managed process to call `kill` itself, with the native pid substituted for the virtual pid.
* The shim handler does so, which causes the test's corresponding signal handler to run
* The signal handler makes another syscall (exit) which ends up causing a message to be sent to shadow.
* Shadow is expecting SHD_SHIM_EVENT_SYSCALL_COMPLETE, and panics.

The simplest solution I can think of is to just call `kill` ourselves from Shadow instead of asking the plugin to do it on our behalf.

If there's a reason to have the managed process do it instead, the next simplest solution I can think of is to add a new message type like SHD_SHIM_EVENT_CONTINUE_WITH_SYSCALL, for cases like this one where Shadow needs to change the syscall parameters but doesn't need to know the result.

sporksmith:
#1457 seemed to fix the signal test, but caused the unistd test to sometimes start failing. I think maybe when we call kill from shadow, the signal might not be guaranteed to be delivered before our next interaction with the process.

More generally, the signal handler could still end up interfering with the shadow/shim communication protocol.

I'm now leaning towards the `SHD_SHIM_EVENT_CONTINUE_WITH_SYSCALL` approach. i.e. let the shim finish communicating with shadow for the syscall and *then* have it make the syscall. Experimentally it seems that when a process sends a signal to itself it *is* delivered synchronously.

sporksmith:
Fixed in #1881 . preload mode now supports signals, using an approach similar to `SHD_SHIM_EVENT_CONTINUE_WITH_SYSCALL` as described above. i.e. there's no recursion/nesting on the shadow side - the shim pushes new stack frames as needed to run signal handlers.

PR #1457, opened 2021-06-18
kill syscall: kill from Shadow instead of from the managed process

sporksmith:
Fixes https://github.com/shadow/shadow/issues/1455.


sporksmith:
Oops, I don't think this approach is going to work reliably either (and seems to occasionally break the unistd test).

I'm not sure exactly what's happening in the unistd test, but I think code in the signal handler could still end up interfering with the shadow/shim communication protocol.

PR #1458, opened 2021-06-18
Conveniently run the tor test manually

robgjansen:


docs/getting_started_tor.md
```diff
@@ -16,8 +16,7 @@ very simple Tor network:
 
 ```bash
 cd shadow/src/test/tor/minimal
-shadow --template-directory shadow.data.template tor-minimal.yaml > shadow.log
-./verify.sh
+./run.sh && ./verify.sh
```
stevenengler, line 19: Maybe include a link to the source files so that people reading the documentation can see what's happening here?

Is there an advantage to having this `run.sh` script compared to running the test directly with `./setup test -- --build-config extra --label tor`?

src/test/tor/minimal/run.sh
```diff
@@ -0,0 +1,4 @@
+# Ensure we have the correct permissions on the hs dir, or Tor won't run
+chmod 700 shadow.data.template/hosts/hiddenserver/hs
```
stevenengler, line 2: The "chmod" step is done as part of `./setup build` when it copies the files from `src/` to `build/src`. The build step also fixes the `AuthDirTestReachability` option depending on the tor version. When running shadow below, would it be better to use the template directory in `build/src` instead? Or maybe this script should run `(~/.local/bin/tor --list-torrc-options | grep -q 'AuthDirTestReachability') || find conf -type f -print0 | xargs -0 sed -i '/AuthDirTestReachability/s/^/#/g` itself?

stevenengler (approved):

docs/getting_started_tor.md
robgjansen: > Is there an advantage to having this run.sh script compared to running the test directly with ./setup test -- --build-config extra --label tor?

Maybe not. The main thing I want is a way for users to inspect the log files after running the experiment, which I think they can do using your setup script magic line. I didn't know you could run the Tor test that way, because I don't think it's documented anywhere or I missed it. The getting started tor markdown doc could be a good place to document it, rather than the custom run script.

src/test/tor/minimal/run.sh
robgjansen: > The "chmod" step is done as part of ./setup build when it copies the files from src/ to build/src. The build step also fixes the AuthDirTestReachability option depending on the tor version.

Oh...
- Somehow I got myself into a situation where the test was failing for me when I ran the run script without the chmod. Probably because I was moving around branches or something and didn't rebuild before running the test.
- I think it will be pretty easy for users to get into that situation, so doing whatever `chmod`s or tor version checks at runtime I think is the best way to maximize chance of success.
- It is not obvious to the user that they should need to make sure they have pinned a Tor version when they compile shadow. That is non-standard and fragile I think. What happens if they want to update Tor later but then run the shadow test again without rebuilding?

I guess some of these issues are arising because I am trying to hijack our CI test and turn it into an example that users can run to quickly get some results from a small Tor network.

stevenengler: Okay, I think having the run script makes sense. Maybe it shouldn't run the grep/find/sed command above then, so that it doesn't modify the torrc files and break the test's CMake file.

Also my PR to make tor not require special permissions on these files when running with `TestingTorNetwork` was never looked at, so I don't think this behaviour will be changing.

docs/getting_started_tor.md
stevenengler: The closest documentation is https://github.com/shadow/shadow/blob/main/docs/developer_guide.md#extra-tests, which isn't tor-specific, but shows how to run the extra tests (which include the tor test).

src/test/tor/minimal/run.sh
robgjansen: > Maybe it shouldn't run the grep/find/sed command above then, so that it doesn't modify the torrc files and break the test's CMake file.

Not sure I follow this part. I think you might mean we may want to modify the `CMakeLists.txt` file that does the find/replace of the `AuthDirTestReachability` option. If so, I think it would be reasonable to remove it and then just document that our test case requires that you run Tor v0.x.x.x or later. I think we already have a similar requirement since our torrc files use `ORPort 9111 IPv4Only`, so I would be fine with this. Probably should do it in a separate PR though.

> Also my PR to make tor not require special permissions on these files when running with TestingTorNetwork was never looked at, so I don't think this behaviour will be changing.

That's too bad :( I added a comment in support:
https://gitlab.torproject.org/tpo/core/tor/-/issues/40338

docs/getting_started_tor.md
robgjansen: Done in e40070c

src/test/tor/minimal/run.sh
stevenengler: > Not sure I follow this part. I think you might mean we may want to modify the CMakeLists.txt file that does the find/replace of the AuthDirTestReachability option. If so, I think it would be reasonable to remove it and then just document that our test case requires that you run Tor v0.x.x.x or later. I think we already have a similar requirement since our torrc files use ORPort 9111 IPv4Only, so I would be fine with this. Probably should do it in a separate PR though.

Our CI tests run all supported versions of tor (except the newest in #1453), which require us to handle the different `AuthDirTestReachability` behaviours. So I think we have to do this find/replace in the CMakeLists.txt file.

Maybe we should just copy these files to an examples directory and have the example and test separate?

robgjansen: I think we should leave it as it is in this PR. I don't think the Tor test is run by default when running `./setup test`, so most people won't run it anyway and there won't be any conflicts when they run the Tor example. If it does end up causing problems for people, we could change it up at that point.

robgjansen: Sound OK?

stevenengler: Sounds good

PR #1460, opened 2021-06-19
Some determinism fixes

### LINKED ISSUE #1294: Re-enable phold determinism test and get it to pass
The new shadow work in the dev branch has caused a regression on determinism. The first step is that we should re-enable our determinism test case that runs phold and ensures that those results are deterministic. See also #666 which documented issues we discovered and fixed with the classic (v1) architecture.

robgjansen:
This primarily includes a fix for a confirmed non-deterministic behavior I found in the epoll getEvents function (in c6039d3cace428349aae167c635cb4cfed9e120d).

I also updated a couple of other places in the code where we iterate a hash table (which is non-deterministic) - these were unconfirmed issues but I suspected they _might_ cause problems. (Probably best if we stop using glib hash table iterators in general, and instead prefer an iterator that guarantees a total ordering; rust probably makes this easier.)

Closes #1294

### DIFF
--- a/src/main/host/descriptor/epoll.c
+++ b/src/main/host/descriptor/epoll.c
@@ -251,13 +251,21 @@ static void _epoll_free(LegacyDescriptor* descriptor) {
 void epoll_clearWatchListeners(Epoll* epoll) {
     MAGIC_ASSERT(epoll);
 
+    /* Iterate the hash table in a determinsitic order. */
+    GList* watch_list = g_hash_table_get_values(epoll->watching);
+    GList* next_item = NULL;
+
+    /* Prepare the list for deterministic iteration. */
+    if(watch_list != NULL) {
+        watch_list = g_list_sort(watch_list, _epollwatch_compare);
+        next_item = g_list_first(watch_list);
+    }
+
     /* make sure none of our watch descriptors notify us anymore */
-    GHashTableIter iter;
-    gpointer key, value;
-    g_hash_table_iter_init(&iter, epoll->watching);
-    while(g_hash_table_iter_next(&iter, &key, &value)) {
-        EpollWatch* watch = value;
+    while(next_item != NULL) {
+        EpollWatch* watch = next_item->data;
         MAGIC_ASSERT(watch);
+
         statuslistener_setMonitorStatus(watch->listener, STATUS_NONE, SLF_NEVER);
 
         if (watch->watchType == EWT_LEGACY_DESCRIPTOR) {
@@ -265,6 +273,13 @@ void epoll_clearWatchListeners(Epoll* epoll) {
         } else if (watch->watchType == EWT_POSIX_FILE) {
             posixfile_removeListener(watch->watchObject.as_file, watch->listener);
         }
+
+        next_item = g_list_next(next_item);
+    }
+
+    /* Cleanup just the list but not the list values, which are owned by the hash table. */
+    if(watch_list) {
+        g_list_free(watch_list);
     }
 }
 
--- a/src/main/host/descriptor/epoll.c
+++ b/src/main/host/descriptor/epoll.c
@@ -252,7 +252,11 @@ static void _epoll_free(LegacyDescriptor* descriptor) {
 void epoll_clearWatchListeners(Epoll* epoll) {
     MAGIC_ASSERT(epoll);
 
-    /* Iterate the hash table in a deterministic order. */
+    /* Iterate the hash table in a deterministic order.
+     * It might be better to maintain the watching values in a sorted structure so we
+     * don't have to re-sort every time this function is called. One option is:
+     * https://developer.gnome.org/glib/2.68/glib-Balanced-Binary-Trees.html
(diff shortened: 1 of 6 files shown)


src/main/host/descriptor/epoll.c
```diff
@@ -223,20 +251,35 @@ static void _epoll_free(LegacyDescriptor* descriptor) {
 void epoll_clearWatchListeners(Epoll* epoll) {
     MAGIC_ASSERT(epoll);
 
+    /* Iterate the hash table in a determinsitic order. */
+    GList* watch_list = g_hash_table_get_values(epoll->watching);
```
sporksmith, line 255: Would it be better to use a data structure that maintains an ordering rather than sorting any time we need to iterate? e.g. https://developer.gnome.org/glib/2.68/glib-Balanced-Binary-Trees.html

src/main/host/descriptor/epoll.c
```diff
@@ -523,11 +566,30 @@ gint epoll_getEvents(Epoll* epoll, struct epoll_event* eventArray, gint eventArr
      * overflow. the number of actual events is returned in nEvents. */
     gint eventIndex = 0;
 
-    GHashTableIter iter;
-    gpointer key, value;
-    g_hash_table_iter_init(&iter, epoll->ready);
-    while(g_hash_table_iter_next(&iter, &key, &value) && (eventIndex < eventArrayLength)) {
-        EpollWatch* watch = value;
+    /*
+     * We need to guarantee that the events are returned in a determinstic order when the
```
sporksmith, line 570: *deterministic

src/main/host/descriptor/epoll.c
```diff
@@ -523,11 +566,30 @@ gint epoll_getEvents(Epoll* epoll, struct epoll_event* eventArray, gint eventArr
...
+     * simulation is run multiple times, so we cannot use hash table iterator.
+     * Using a list here has some potential performance implications:
+     * - O(n) to loop the hash table and create the list of values
+     * - O(n) to sort the list
+     * - O(n) for our iteration of the list
+     * We think that the ready list is typically small and so 3*O(n) will be small in practice.
+     * If this turns out not to be the case, we could consider maintaining a sorted list of the
+     * ready watch values alongside the epoll->ready hash table instead, which would allow us
+     * to reduce this function to 1*O(n) by avoiding creating and sorting the list here.
```
sporksmith, line 579: Again might be worth mentioning or using https://developer.gnome.org/glib/2.68/glib-Balanced-Binary-Trees.html

sporksmith (approved):
Approving now since the current state is definitely an improvement, but maybe worth a re-review if you do decide to switch to trees.

src/main/host/descriptor/epoll.c
```diff
@@ -129,6 +138,23 @@ static gboolean _epollkey_equal(gconstpointer ptr_1, gconstpointer ptr_2) {
...
+        /* watch_1 was reported longest ago and should come first. */
+        return -1;
+    } else if (watch_1->last_reported_event_time > watch_2->last_reported_event_time) {
+        /* watch_2 was reported longest ago and should come first. */
+        return 1;
+    } else {
+        /* Both were previously reported at the same time (or never reported yet),
+         * so now we fall back to the deterministic unique id ordering. */
+        return (watch_1->id < watch_2->id) ? -1 : (watch_1->id > watch_2->id) ? 1 : 0;
+    }
```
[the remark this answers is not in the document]
robgjansen: Done

src/main/host/descriptor/epoll.c
```diff
@@ -223,20 +251,35 @@ static void _epoll_free(LegacyDescriptor* descriptor) {
 void epoll_clearWatchListeners(Epoll* epoll) {
     MAGIC_ASSERT(epoll);
 
+    /* Iterate the hash table in a determinsitic order. */
```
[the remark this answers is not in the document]
robgjansen: Done

src/main/host/descriptor/epoll.c
robgjansen: Done

robgjansen:
I would like to check the performance implications of the changes in the PR before merging it.

robgjansen:
`Before` is at 6872439, `after` is this branch rebased on top of 6872439.

run_time

This PR appears to have a negligible performance impact.

src/main/host/descriptor/epoll.c
robgjansen: Added a note to consider trees. Since this PR appears to have a negligible performance impact, let's leave enhancements to epoll etc. for a future PR (maybe when we rewrite in rust).

src/main/host/descriptor/epoll.c
robgjansen: Added a note to consider trees. Since this PR appears to have a negligible performance impact, let's leave enhancements to epoll etc. for a future PR (maybe when we rewrite in rust).

PR #1462, opened 2021-06-21
Add an option to use a socket for ipc

sporksmith:
This was motivated by seccomp use-cases of being able to send a
file-descriptor over the socket. As it turns out, I'm not sure we need
that capability after all. It is a bit easier to whitelist the
corresponding IPC in the seccomp filter though, and a bit simpler and
more flexible in general.

It *should* have similar performance to the semaphore approach when
we're not spinning, but I've left it in as an option so we can verify.

### DIFF
--- a/src/main/bindings/c/bindings-opaque.h
+++ b/src/main/bindings/c/bindings-opaque.h
@@ -20,6 +20,13 @@ typedef enum InterposeMethod {
   INTERPOSE_METHOD_HYBRID,
 } InterposeMethod;
 
+typedef enum IpcMethod {
+  // Unix-domain socket
+  IPC_METHOD_SOCKET,
+  // Semaphore + shared memory
+  IPC_METHOD_SEMAPHORE,
+} IpcMethod;
+
 typedef enum QDiscMode {
   Q_DISC_MODE_FIFO,
   Q_DISC_MODE_ROUND_ROBIN,
--- a/src/main/bindings/c/bindings.h
+++ b/src/main/bindings/c/bindings.h
@@ -174,6 +174,8 @@ bool config_getUseCpuPinning(const struct ConfigOptions *config);
 
 enum InterposeMethod config_getInterposeMethod(const struct ConfigOptions *config);
 
+enum IpcMethod config_getIpcMethod(const struct ConfigOptions *config);
+
 bool config_getUseSchedFifo(const struct ConfigOptions *config);
 
 bool config_getUseOnWaitpidWorkarounds(const struct ConfigOptions *config);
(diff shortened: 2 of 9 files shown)


robgjansen:
If this ever moves forward, please mark @RWails as a reviewer (he has some thoughts on the cpp code).

sporksmith:
From @RWails on slack:

```
On our bare-metal machines running centos, context switching driven by a semaphore (futex syscall) is roughly 2x faster than context switching driven by UNIX domain sockets (sendto / recvfrom syscalls)---

    Approximately 3 microseconds for a parent -> child -> parent roundtrip when triggered by a semaphore post/wait
    Approximately 6 microseconds for a parent -> child -> parent roundtrip when triggered by a socket read/write

(The exact code that’s being measured here is <https://github.com/robgjansen/phantom-paper/blob/2b1a511d0039774aa337d9d9166180a49[…]riments/micro-benchmarks/context_switching/context_switching.cc>, which dispatches the switch and wait procedures of SocketSwitch and SemSwitch.)
```

I gave this a quick try with the mezzo-phold benchmark and got ~40% overhead (with zero CPU-load, so still very IPC heavy). Not quite as  bad as the 2x of the underlying switches, but still pretty significant.

```
$ cat config.yaml 
general:
  stop_time: 4 s
  bootstrap_end_time: 4 s
network:
  graph:
    type: 1_gbit_switch
hosts:
  host:
    quantity: 24
    processes:
    - path: phold/target/release/phold
      args: 24 host uniform 0 20 0 1337
      start_time: 1 s

$ hyperfine -p 'rm -rf shadow.data' -L type semaphore,socket '/home/jnewsome/.local/bin/shadow -p12 --interpose-method=preload --use-seccomp=true --ipc-method={type}  config.yaml'
Benchmark #1: /home/jnewsome/.local/bin/shadow -p12 --interpose-method=preload --use-seccomp=true --ipc-method=semaphore  config.yaml
  Time (mean ± σ):      8.266 s ±  0.241 s    [User: 49.973 s, System: 29.576 s]
  Range (min … max):    7.917 s …  8.628 s    10 runs
 
Benchmark #2: /home/jnewsome/.local/bin/shadow -p12 --interpose-method=preload --use-seccomp=true --ipc-method=socket  config.yaml
  Time (mean ± σ):     11.644 s ±  0.046 s    [User: 55.822 s, System: 55.438 s]
  Range (min … max):   11.588 s … 11.713 s    10 runs
 
Summary
  '/home/jnewsome/.local/bin/shadow -p12 --interpose-method=preload --use-seccomp=true --ipc-method=semaphore  config.yaml' ran
    1.41 ± 0.04 times faster than '/home/jnewsome/.local/bin/shadow -p12 --interpose-method=preload --use-seccomp=true --ipc-method=socket  config.yaml'
```

Gonna go ahead and just close this PR for now :)

PR #1463, opened 2021-06-21
Auto assign author and labels to pull requests

robgjansen:
Adds an action workflow to auto-assign the author and apply component labels when pull requests are opened.

### DIFF
--- a/.github/workflows/pr_metadata.yml
+++ b/.github/workflows/pr_metadata.yml
@@ -16,7 +16,8 @@ jobs:
     - name: Assign Author
       # Add the author as an assignee of the PR
       # https://github.com/marketplace/actions/auto-author-assign
-      uses: toshimaru/auto-author-assign@v1.3.5
+      # before updating the version of this untrusted action, check the code manually
+      uses: toshimaru/auto-author-assign@084de2241839f0a3433adfd18d2906e8e786380e
       with:
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
     - name: Apply Labels
--- a/.github/workflows/pr_metadata.yml
+++ b/.github/workflows/pr_metadata.yml
@@ -1,8 +1,10 @@
 # Auto-update metadata on pull requests
 # https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions
 name: "Pull Request Metadata"
+
 on:
-- pull_request_target
+  pull_request_target:
+    types: [opened]
 
 permissions:
   pull-requests: write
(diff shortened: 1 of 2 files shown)


.github/workflows/pr_metadata.yml
```diff
@@ -0,0 +1,23 @@
+# Auto-update metadata on pull requests
```
sporksmith, line 1: It's helpful to have a link here to the "schema":
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions

sporksmith (commented):
Nice!

.github/workflows/pr_metadata.yml
robgjansen: Done in 1b51ab0

.github/workflows/pr_metadata.yml
```diff
@@ -0,0 +1,23 @@
...
+name: "Pull Request Metadata"
```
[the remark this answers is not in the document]
robgjansen: Done in 1b51ab0

.github/workflows/pr_metadata.yml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  Triage:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Assign Author
+      # Add the author as an assignee of the PR
+      # https://github.com/marketplace/actions/auto-author-assign
+      uses: toshimaru/auto-author-assign@v1.3.5
+      with:
+        repo-token: "${{ secrets.GITHUB_TOKEN }}"
```
stevenengler, line 19: Maybe better to use a specific commit hash instead of a version so that the `v1.3.5` tag isn't unexpectedly swapped out with a malicious commit?

sporksmith (approved):

.github/workflows/pr_metadata.yml
```diff
@@ -0,0 +1,27 @@
+# Auto-update metadata on pull requests
+# https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions
+name: "Pull Request Metadata"
+on:
+- pull_request_target
```
stevenengler, line 5: This will re-add the labels and re-assign the author each time a commit is pushed, but I'm not sure if this is intentional. For example maybe the PR author doesn't think a tag is appropriate and removes it, but it's automatically added on the next commit or rebase. Or if the author doesn't want to be assigned to a PR.

Maybe restrict it to:

```yaml
on:
  pull_request_target:
    types: [opened]
```

stevenengler (approved):

.github/workflows/pr_metadata.yml
robgjansen: Done in 6364837

.github/workflows/pr_metadata.yml
robgjansen: Done in 80caa45

.github/workflows/pr_metadata.yml
robgjansen: This means if you push more commits to the PR later that change files in a new component, the new component tag will not be automatically added.

PR #1466, opened 2021-06-22
Fixed github labeler config

stevenengler:
Values that start with a `*` or `!` are not strings in yaml.


robgjansen (approved):
Thank you and sorry about that! (I tested a different syntax, but then changed it and should have tested again.)

stevenengler:
> Thank you and sorry about that! (I tested a different syntax, but then changed it and should have tested again.)

No problem, we all missed it in the review :)

Issue #1262, opened 2021-04-05
Implement futex operations that currently generate warnings in tor sims


sporksmith:
In Tor simulations we currently get some warnings about `futex` operations that we don't support yet. We return ENOSYS for these, and as far as we can tell, this is handled gracefully without adversely affecting the simulation.

We probably ought to implement these operations though to increase our assurance about correctness in Tor simulations, get rid of the warnings, and support other simulations where they might really be needed.

Previously I've run across that glibc semaphores use `FUTEX_WAIT_BITSET` and `FUTEX_WAKE_BITSET`, so we probably ought to implement those. There might be others that are generating warnings in the tor simulations, though.

sporksmith:
@stevenengler reports that the bitset operations are also used in Python, which currently ends up dead-locking.

robgjansen:
In a recent 5% Tor network experiment, I counted 876,113 warnings about missing futex operation 5, which on my machine maps to `FUTEX_WAKE_OP`. I did not see warnings for any other missing futex operations.

sporksmith:
@robgjansen are you still seeing warnings for these?

robgjansen:
Sorry for the delay! I'm not seeing this futex warning

https://github.com/shadow/shadow/blob/b0e1fd5e7b6c922cf0fad0ea0889c59bd55fcd0b/src/main/host/syscall/futex.c#L206

for either ptrace or preload mode when running Shadow in docker at v2.0.0-pre.4 (8f2ceee4dcb97ff12e6ff3d94bc54f109c2e9fd3).

sporksmith:
Thanks for the update! Going ahead and closing, then

PR #1470, opened 2021-06-23
Handle FUTEX_BITSET_MATCH_ANY case of futex bitset ops

sporksmith:
Progress on
https://github.com/shadow/shadow/issues/1262

### DIFF
--- a/src/main/host/syscall/futex.c
+++ b/src/main/host/syscall/futex.c
@@ -24,7 +24,8 @@
 ///////////////////////////////////////////////////////////
 
 static SysCallReturn _syscallhandler_futexWaitHelper(SysCallHandler* sys, PluginPtr futexVPtr,
-                                                     int expectedVal, PluginPtr timeoutVPtr) {
+                                                     int expectedVal, PluginPtr timeoutVPtr,
+                                                     TimeoutType type) {
     // This is a new wait operation on the futex for this thread.
     // Check if a timeout was given in the syscall args.
     const struct timespec* timeout;
@@ -96,7 +97,7 @@ static SysCallReturn _syscallhandler_futexWaitHelper(SysCallHandler* sys, Plugin
     Trigger trigger =
         (Trigger){.type = TRIGGER_FUTEX, .object = futex, .status = STATUS_FUTEX_WAKEUP};
     if (timeout) {
-        _syscallhandler_setListenTimeout(sys, timeout);
+        _syscallhandler_setListenTimeout(sys, timeout, type);
     }
     return (SysCallReturn){
         .state = SYSCALL_BLOCK, .cond = syscallcondition_new(trigger, timeout ? sys->timer : NULL)};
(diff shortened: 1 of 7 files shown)


src/main/host/syscall/futex.c
```diff
@@ -159,32 +160,45 @@ SysCallReturn syscallhandler_futex(SysCallHandler* sys, const SysCallArgs* args)
...
-            warning("We do not yet handle futex operation %i", operation);
-            result = -ENOSYS; // Invalid operation specified in futex_op
-            break; // return SYSCALL_DONE
-        }
+        case FUTEX_WAIT_REQUEUE_PI: break;
     }
 
-    return (SysCallReturn){.state = SYSCALL_DONE, .retval.as_i64 = result};
+    warning("Unhandled futex operation %i", operation);
+    return (SysCallReturn){.state = SYSCALL_DONE, .retval.as_i64 = -ENOSYS};
```
robgjansen, line 201: I think `int result = 0;` on line 158 is unused now and should be removed.

src/main/host/syscall_handler.c
```diff
@@ -506,7 +506,7 @@ SysCallReturn syscallhandler_make_syscall(SysCallHandler* sys,
     } else if (_syscallhandler_wasBlocked(sys)) {
         /* We were but are no longer blocked on a syscall. Make
          * sure any previously used listener timeouts are ignored.*/
-        _syscallhandler_setListenTimeout(sys, NULL);
+        _syscallhandler_setListenTimeout(sys, NULL, TIMEOUT_ABSOLUTE);
```
robgjansen, line 509: Isn't `TIMEOUT_RELATIVE` consistent with the code before, rather than `TIMEOUT_ABSOLUTE`?

---

Using `TIMEOUT_ABSOLUTE ` will cause us to pass `flags=TFD_TIMER_ABSTIME` to `timer_setTime`, whereas before we were passing `flags=0` to `timer_setTime`.

This will cause us to take the `if` statement in `_timer_setCurrentTime`, whereas before we took the `else` statement.

I *think* the end result is the same, but we should probably be consistent and use `TIMEOUT_RELATIVE` instead.

robgjansen (approved):
Did we check if Tor no longer prints the FUTEX warnings?

sporksmith:
> Did we check if Tor no longer prints the FUTEX warnings?

According to #1262, the tor warnings are from FUTEX_WAKE_OP, which still isn't implemented.

@stevenengler do you remember how to reproduce the deadlocks you saw in Python?

stevenengler:
> @stevenengler do you remember how to reproduce the deadlocks you saw in Python?

I think this shadow config should trigger it:

```yaml
general:
  stop_time: 10
network:
  graph:
    type: 1_gbit_switch
hosts:
  server:
    quantity: 1
    processes:
    - path: /usr/bin/python3
      args: -m http.server 80
      start_time: 3
  client:
    quantity: 3
    processes:
    - path: /usr/bin/curl
      args: -s server
      start_time: 5
```

sporksmith:
> I think this shadow config should trigger it:

Yup! Confirmed that config deadlocks on `main` and succeeds with the PR.

src/main/host/syscall/futex.c
sporksmith: Done

src/main/host/syscall_handler.c
sporksmith: Done

stevenengler:
> > I think this shadow config should trigger it:
> 
> Yup! Confirmed that config deadlocks on `main` and succeeds with the PR.

Cool! We can also make our getting started example simpler then by removing the Python script and using this CLI-only server instead.

PR #1471, opened 2021-06-23
seccomp: Disable Speculative Store Bypass mitigation

sporksmith:
By default, installing a seccomp filter sets the PR_SPEC_FORCE_DISABLE
bit (see prctl(2)). This results in a significant performance penalty.
Meanwhile Shadow is semi-cooperative with its virtual processes; it
doesn't try to protect itself or the system from malicious code. Hence,
it isn't worth paying this overhead.

Here we use the SECCOMP_FILTER_FLAG_SPEC_ALLOW to override this
behavior.

### DIFF
--- a/src/lib/shim/shim.c
+++ b/src/lib/shim/shim.c
@@ -431,8 +431,14 @@ static void _shim_parent_init_seccomp() {
         .filter = filter,
     };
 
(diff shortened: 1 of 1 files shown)


robgjansen (approved):
🚀

robgjansen:
Hmm, I wonder why the testing label got added. That seems like a bug.

PR #1473, opened 2021-06-24
Simplified the getting started doc

stevenengler:
We can now run the python HTTP server directly because of #1470.


sporksmith (approved):
Cool!

PR #1474, opened 2021-06-24
Updated PR labeler config

stevenengler:


.github/pr_labeler_conf.yml
```diff
@@ -20,7 +20,7 @@
 #  negation, you can write complex matching rules.
 "Component: Testing":
 - any: ["src/test/**/*"]
-  any: [".github/**/*.yml", "!.github/pr_labeler_conf.yml",
+- any: [".github/**/*.yml", "!.github/pr_labeler_conf.yml",
```
robgjansen, line 23: The examples on [the GitHub page](https://github.com/actions/labeler#match-object) show that multiple match objects do not have the extra dash, which is why I didn't add it?

stevenengler: The workflow fails since you cannot have two keys with the same name: https://github.com/shadow/shadow/runs/2906055933?check_suite_focus=true

robgjansen (approved):

PR #1475, opened 2021-06-24
interpose-method=preload thread support

### LINKED ISSUE #1454: Implement threads for preload+seccomp
Threads for preload mode were part of a bigger issue that we closed after deciding it was infeasible. It looks like it should be feasible for the (in-progress) preload+seccomp mode, though.

sporksmith:
thread_preload now implements `ThreadMethods`::clone. This is implemented by sending a message to the shim to set aside an ipc block for the thread, and then having the shim implement the clone syscall.

In the shim, when executing the clone syscall, the child thread *jumps* back to where the managed program made the original syscall. This is implemented by saving the instruction pointer (RIP) in the SIGSYS syscall handler.

This also adds a wrapper for thread local storage in the shim. In CentOS 7 this was needed to work around some of the implementation details of the native thread local storage. This indirection is no longer necessary at the moment now that we've dropped CentOS 7 support and all supported platforms have compatible thread local storage implementations. However in the future if we need to decouple ourselves form the native thread local storage we can now just change *our* thread local storage implementation to not rely on the native thread local storage, without having to (again) change all the individual variables using thread local storage.

Fixes #1454

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,8 @@
-# this file is also used with rsync --exclude-from
-# in 'ci/run.sh', so avoid globs
-.git
-build
-src/main/target
+# .dockerignore doesn't treat a bare basename as unrooted. e.g. just `target` is
+# equivalent to `./target`; it won't match `./src/test/target`.
+#
+# This file is also used with rsync --exclude-from in 'ci/run.sh', but rsync
+# appears to accept a super-set of patterns recognized by dockerignore.
+**/.git
+**/build
+**/target
--- a/src/lib/shim/shim_tls.c
+++ b/src/lib/shim/shim_tls.c
@@ -1,12 +1,16 @@
 #include "lib/shim/shim_tls.h"
 
+#include "lib/logger/logger.h"
 #include "lib/shim/shim.h"
 
 #include <assert.h>
 #include <stdalign.h>
 
-// XXX: Needs to be big enough for stack in _vshadow_raw_syscall until we get
-// rid of it.
+// Size of the thread-local stack in _vshadow_raw_syscall (4096*10) + an extra
+// kilobyte for the handful of other thread locals we use.  If the former
+// becomes a permanent fixture, we should make it a shared constant, but it
+// should eventually go away. In the meantime we catch at runtime if we try to
+// use more TLS than we've pre-reserved.
 #define BYTES_PER_THREAD (4096*10 + 1024)
 #define MAX_THREADS 100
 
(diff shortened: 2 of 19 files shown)


src/lib/shim/shim_tls.c
```diff
@@ -0,0 +1,63 @@
...
+static size_t _nextByteOffset = 0;
+
+// Index into _tlss.
+static int _tlsIdx = 0;
+
+// Take an unused TLS index, which can be used for a new thread.
+int shimtls_takeNextIdx() {
+    static int next = 0;
+    assert(next < MAX_THREADS);
+    return next++;
```
stevenengler, line 34: I think this should always be an error, rather than only in debug mode.

src/lib/shim/shim_tls.c
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// Use when switching threads.
+int shimtls_getCurrentIdx() {
+    static __thread bool idxInitd = false;
+    static __thread int idx;
```
stevenengler, line 40: Why are C thread-locals allowed here, but not elsewhere in the shim?

src/main/host/thread_ptrace.c
```diff
@@ -28,6 +28,12 @@
 
 #define THREADPTRACE_TYPE_ID 3024
 
+// glibc in centos 7 does not include the following, but it does use a supported
+// kernel (3.10 > 3.8)
+#if !defined(PTRACE_O_EXITKILL)
+#define PTRACE_O_EXITKILL (1 << 20)
+#endif
+
```
stevenengler, line 36: Is this still needed without CentOS 7?

stevenengler (approved):
Cool! Took me a little while to work though it, but looks good to me.

.dockerignore
```diff
@@ -2,4 +2,8 @@
 # in 'ci/run.sh', so avoid globs
 .git
 build
+src/build
 src/main/target
+src/test/target
+build/src/test/target
+build/src/main/target
```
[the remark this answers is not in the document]
sporksmith: > Do build/src/test/target and build/src/main/target need to be listed here when build is already listed?

Unlike gitignore, dockerignore treats a bare basename as rooted in the current directory. I checked again though and both it and rsync *do* support '**' wildcards, so I switched to that.

> And does anything buid into src/build

Hmm, I had some cmake cache files there, but after a clean build they're gone. Maybe some leftover from manual poking around.

src/lib/shim/shim_tls.c
```diff
@@ -0,0 +1,63 @@
+#include "lib/shim/shim_tls.h"
+
+#include "lib/shim/shim.h"
+
+#include <assert.h>
+#include <stdalign.h>
+
+// Needs to be big enough for stack in _vshadow_raw_syscall.
+#define BYTES_PER_THREAD (4096*10 + 1024)
```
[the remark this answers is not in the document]
sporksmith: Expanded comment with explanation.

src/lib/shim/shim_tls.c
sporksmith: Done

src/lib/shim/shim_tls.c
```diff
@@ -0,0 +1,63 @@
...
+
+// Index into _tlss.
+static int _tlsIdx = 0;
+
+// Take an unused TLS index, which can be used for a new thread.
+int shimtls_takeNextIdx() {
+    static int next = 0;
+    assert(next < MAX_THREADS);
+    return next++;
+}
```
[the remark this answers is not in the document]
sporksmith: Right, it could be. For now I figured I'd leave out that complexity until if and when we need it.

src/lib/shim/shim_tls.c
sporksmith: I have some ideas for how to do this without using the native thread local storage, but it adds some cost and complexity. Doesn't make sense to implement it now, but if we ever do again need to avoid using the native thread-local-storage this is the one place we'll have to change.

Added some comments.

src/main/host/thread_ptrace.c
sporksmith: Oops good catch. I ended up with a lot of merge conflicts after breaking out this PR to other PRs and merging them separately (each after some modification). I ended up just taking the final state of this PR and rebuilding on top of main from scratch, which meant I had to manually un-revert some changes. In hindsight this was probably not the best approach :)

PR #1477, opened 2021-06-25
_vshadow_raw_syscall: ensure new stack is aligned

sporksmith:


src/lib/shim/preload_syscall.c
```diff
@@ -152,7 +152,9 @@ static long _vshadow_raw_syscall(long n, va_list args) {
     // non-fatal signal.
     // TODO: Allocate dynamically and only do this until the MemoryManager is
     // initialized.
-    static __thread char new_stack[4096 * 10];
+    //
+    // C ABI requires 16-byte alignment for stack frames.
+    static __thread char new_stack[4096 * 10] __attribute__((aligned(16)));
```
stevenengler, line 157: Out of curiosity, did using an unaligned stack cause an error, or is this just something you noticed while reading?

stevenengler (approved):

src/lib/shim/preload_syscall.c
sporksmith: Yup - in optimized builds it'd sometimes end up trying to use vectorized instructions on misaligned stack-allocated variables, which results in a segfault

stevenengler: Okay, thanks!

PR #1478, opened 2021-06-25
signal syscalls: handle variable sized signal sets

sporksmith:
The signal handler syscalls accept variable-sized signal sets to support
future expansion. They also use a different struct layout than the
userspace `struct sigaction`, presumably for this reason (it puts the
signal set last).

### DIFF
--- a/src/main/host/syscall/signal.c
+++ b/src/main/host/syscall/signal.c
@@ -16,6 +16,38 @@
 #include "main/host/thread.h"
 #include "main/utility/syscall.h"
 
+// This should be compatible with the kernel's sigaction struct, which is what
+// the syscalls take in cases where the userspace wrappers take `struct
+// sigaction`. While a similar struct definition appears in the glibc source,
+// it's not exposed to users, so we duplicate it here. While in principle this
+// definition could change out from under us, it's more likely that the kernel
+// would introduce a new syscall number rather than breaking userspace programs
+// that still had the old definition.
+//
+// When reading/writing a userspace pointer, use kernel_sigaction_size to
+// calculate the true size. This is needed because the last field, `sa_mask`,
+// is actually variable size, and is specified with another parameter in the syscall.
+struct kernel_sigaction {
+    void* handler;
+    unsigned long sa_flags;
+    void (*sa_restorer)(void);
+    sigset_t sa_mask;
+};
+
+// Calculates the size of a corresponding `struct kernel_sigaction`. Returns 0
+// if the masksize is invalid.
+static size_t kernel_sigaction_size(size_t masksize) {
+    if (masksize < 4) {
+        warning("Got bad masksize: %zu < 4", masksize);
+    }
+    size_t sz = sizeof(struct kernel_sigaction) - sizeof(sigset_t) + masksize;
+    if (sz > sizeof(struct kernel_sigaction)) {
+        warning("Got bad sigaction size: %zu > %zu", sz, sizeof(struct kernel_sigaction));
+        return 0;
+    }
+    return sz;
+}
+
 ///////////////////////////////////////////////////////////
 // Helpers
 ///////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


src/main/host/syscall/signal.c
```diff
@@ -16,6 +16,34 @@
...
+//
+// When reading/writing a userspace pointer, use kernel_sigaction_size to
+// calculate the true size. This is needed because the last field, `sa_mask`,
+// is actually variable size, and is specified with another parameter in the syscall.
+struct kernel_sigaction {
+    void* handler;
+    unsigned long sa_flags;
+    void (*sa_restorer)(void);
+    sigset_t sa_mask;
+};
```
stevenengler, line 31: Is there a reason to use this struct rather than including the one from ` linux/signal.h`?

src/main/host/syscall/signal.c
```diff
@@ -126,76 +154,85 @@ SysCallReturn syscallhandler_tkill(SysCallHandler* sys, const SysCallArgs* args)
...
+    if (!sz) {
+        warning("Bad masksize %zu", masksize);
+        return (SysCallReturn){.state = SYSCALL_DONE, .retval = -EINVAL};
+    }
 
-SysCallReturn syscallhandler_rt_sigaction(SysCallHandler* sys, const SysCallArgs* args) {
-    utility_assert(sys && args);
-    if (!shimipc_getUseSeccomp()) {
+    struct kernel_sigaction action;
+    int rv = process_readPtr(sys->process, &action, actPtr, sz);
```
stevenengler, line 184: Since `sz` might be less than `sizeof(action)`, should this zero-initialize `action` first?

stevenengler (approved):

src/main/host/syscall/signal.c
sporksmith: Short answer: it's not there :). I expanded on the comment a bit.

```
$ grep -R kernel_sigaction /usr/include
<no output>
```

src/main/host/syscall/signal.c
```diff
@@ -16,6 +16,34 @@
...
+// if the masksize is invalid.
+static size_t kernel_sigaction_size(size_t masksize) {
+    if (masksize < 4) {
+        warning("Got bad masksize: %zu < 4", masksize);
+    }
+    size_t sz = sizeof(struct kernel_sigaction) - sizeof(sigset_t) + masksize;
+    if (sz >= sizeof(struct kernel_sigaction)) {
+        warning("Got bad sigaction size: %zu > %zu", sz, sizeof(struct kernel_sigaction));
+        return 0;
+    }
```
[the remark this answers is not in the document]
sporksmith: Good catch! Fixed.

src/main/host/syscall/signal.c
sporksmith: If our code is correct, we'll never end up accessing the uninitialized bits. Doesn't hurt to defensively zero it out, though; done.

src/main/host/syscall/signal.c
stevenengler: It seems to be there for me on ubuntu 20.04:

```c
#include <stddef.h>
#include <linux/signal.h>

int main() {
    struct sigaction test;
    return 0;
}
```

And I think it's just named `sigaction` in the kernel?

sporksmith: The glibc public interface uses `struct sigaction`; the kernel syscall API uses `struct kernel_sigaction`

stevenengler: I thought they both used the name `struct sigaction`, but glibc [uses the name `struct kernel_sigaction` internally](https://code.woboq.org/userspace/glibc/sysdeps/unix/sysv/linux/kernel_sigaction.h.html) to refer to their copy. But either way I think it's fine to redefine it, it shouldn't change anytime soon.

sporksmith: Sadly they are different structs :). `kernel_sigaction` has the mask last (presumably to support the run-time mask-size without changing the offsets of the other members).

https://github.com/sporksmith/glibc/blob/master/sysdeps/unix/sysv/linux/bits/sigaction.h#L27
```
struct sigaction
  {
    /* Signal handler.  */
#if defined __USE_POSIX199309 || defined __USE_XOPEN_EXTENDED
    union
      {
	/* Used if SA_SIGINFO is not set.  */
	__sighandler_t sa_handler;
	/* Used if SA_SIGINFO is set.  */
	void (*sa_sigaction) (int, siginfo_t *, void *);
      }
    __sigaction_handler;
# define sa_handler	__sigaction_handler.sa_handler
# define sa_sigaction	__sigaction_handler.sa_sigaction
#else
    __sighandler_t sa_handler;
#endif

    /* Additional set of signals to be blocked.  */
    __sigset_t sa_mask;

    /* Special flags.  */
    int sa_flags;

    /* Restore handler.  */
    void (*sa_restorer) (void);
  };
```

https://github.com/sporksmith/glibc/blob/master/sysdeps/unix/sysv/linux/kernel_sigaction.h#L9
```
struct kernel_sigaction
{
  __sighandler_t k_sa_handler;
  unsigned long sa_flags;
#ifdef HAS_SA_RESTORER
  void (*sa_restorer) (void);
#endif
  /* glibc sigset is larger than kernel expected one, however sigaction
     passes the kernel expected size on rt_sigaction syscall.  */
  sigset_t sa_mask;
};
```

sporksmith: Actually, looking at it again, I think the `sigaction` syscall does take `struct sigaction`, and it's only `sigaction_rt` that takes `struct sigaction`. i.e. they used to be the same and then rearranged it for future expandability (while leaving the glibc interface the same)

sporksmith: Will verify and prep a follow-up. I think modern libc always uses the `rt_sigaction` syscall, but older ones might still use `sigaction`...

sporksmith: Yup.

https://github.com/sporksmith/linux/blob/144c79ef33536b4ecb4951e07dbc1f2b7fa99d32/kernel/signal.c#L4360
```
SYSCALL_DEFINE3(sigaction, int, sig,
		const struct old_sigaction __user *, act,
	        struct old_sigaction __user *, oact)
{
```

https://github.com/sporksmith/linux/blob/master/include/linux/compat.h#L369
```
struct compat_old_sigaction {
	compat_uptr_t			sa_handler;
	compat_old_sigset_t		sa_mask;
	compat_ulong_t			sa_flags;
	compat_uptr_t			sa_restorer;
};
```

PR #1481, opened 2021-06-26
Notls

sporksmith:


sporksmith:
Now that we've dropped support for centos 7 we don't need this PR for now. Keeping it around for the moment since we will need this or something like it if we end up needing to decouple ourselves from the glibc/system implementation of thread local storage for whatever reason. e.g. a version of this approach also enables the bare `clone` unit tests to pass, which don't set up native thread local storage at all. (For that we'd need to always use our own lookup tables instead of using the native thread-local-storage outside of critical sections)

sporksmith:
I left some of this in #1475

PR #1482, opened 2021-06-28
Remove (broken) support for old signal APIs

sporksmith:
The older signal APIs took different structs than the new ones (e.g.
`sigaction` vs `rt_sigaction`). These syscall are all obsolete since
Linux 2.2 so it's probably not worth trying to maintain them.

### DIFF
--- a/src/main/host/syscall_handler.c
+++ b/src/main/host/syscall_handler.c
@@ -227,6 +227,11 @@ static void _syscallhandler_post_syscall(SysCallHandler* sys, long number,
         trace("native syscall %ld " #s, args->number);                         \
         scr = (SysCallReturn){.state = SYSCALL_NATIVE};                        \
         break
+#define UNSUPPORTED(s)                                                                             \
+    case SYS_##s:                                                                                  \
+        error("Returning error ENOSYS for explicitly unsupported syscall %ld " #s, args->number);  \
+        scr = (SysCallReturn){.state = -ENOSYS};                                                   \
+        break
 
 #ifdef USE_C_SYSCALLS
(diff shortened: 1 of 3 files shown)


src/main/host/syscall_handler.c
```diff
@@ -227,6 +227,11 @@ static void _syscallhandler_post_syscall(SysCallHandler* sys, long number,
         trace("native syscall %ld " #s, args->number);                         \
         scr = (SysCallReturn){.state = SYSCALL_NATIVE};                        \
         break
+#define UNSUPPORTED(s)                                                                             \
+    case SYS_##s:                                                                                  \
+        warning("unsupported syscall %ld " #s, args->number);                                      \
+        scr = (SysCallReturn){.state = -ENOSYS};                                                   \
+        break
```
stevenengler, line 234: What's the difference between this and the "unsupported" message in the default case? Maybe this message should be more explicit that this is a deliberately unsupported syscall, with no intention to support it in the future?

stevenengler (approved):

src/main/host/syscall_handler.c
sporksmith: Tweaked to make a bit more explicit. Mainly I wanted a way to explicitly have "known unsupported" syscall #'s in the `switch` statement. i.e. if someone later tries to add a NATIVE or HANDLE case for these syscall #'s, I want them to get a compiler error and be forced to see the comment about why it's unsupported :). The ~duplication of the error messaging is a little unfortunate, but the ways I could think of to dedupe didn't really seem better.

PR #1486, opened 2021-06-29
Drop support for centos 7

sporksmith:
Supporting CentOS 7 would significantly complicate the implementation of thread support for preload-based interposition: its version of glibc initializes thread-local-storage lazily in the new thread in a way that executes syscalls. This leads to an infinite recursive loop in the shim.

While there are potential ways to work around this issue (e.g. implement our own thread-local-storage), centos 7 no longer gets full updates as of [q4 2020](https://wiki.centos.org/About/Product), and has been preventing us from using modern C11 features such as C11 atomics.


sporksmith:
We'll need to disable the centos-8 test requirement in GH for this to be merged.

docs/supported_platforms.md
```diff
@@ -5,7 +5,7 @@ Shadow officially supports the following platforms:
   + Ubuntu 18.04 and 20.04
   + Debian 10
   + Fedora 33
-  + CentOS 7 and 8
+  + CentOS 8
```
robgjansen, line 8: I think we should add some specific details here to explain that not only is CentOS 7 unsupported, it is explicitly _broken_ based on our use of C11 atomics and seccomp thread support and you will need to use docker to run there.

robgjansen (approved):

docs/supported_platforms.md
sporksmith: Done

PR #1487, opened 2021-06-30
Ensure thread is really dead before notifying other threads that it is

sporksmith:
Progress towards #1454


stevenengler (approved):

robgjansen:
Wow, this unsolicited advertisement is very annoying. (We no longer use codecov or have it installed.) I may block the user from our org (let me know if you think I should not).

sporksmith:
> Wow, this unsolicited advertisement is very annoying. (We no longer use codecov or have it installed.) I may block the user from our org (let me know if you think I should not).

Yeah, bizarre. Maybe this report was expected when the PR was originally created, got hung up somehow, and just now got processed from some backlog? With an error now that we no longer have the app installed or now that an app is required where it wasn't before or ... ?

Blocking might cause confusion if we ever try using that particular tool again, but I guess that's pretty unlikely.

PR #1505, opened 2021-07-09
Change support from CentOS 8 to CentOS Stream 8

stevenengler:
CentOS 8 is only supported for another 5 months.


stevenengler:
Updated the reviewer to Rob since the required CI checks also needs to be updated.

robgjansen (approved):

PR #1510, opened 2021-07-13
spec-store-bypass: add a warning and documentation

### LINKED ISSUE #1489: Warn user if speculative stores are disabled
There's a speculative-store sidechannel that only affects processes that need to prevent code that has access to its memory locations from reading its data, such as browser sandboxes. The common mitigation is to disable speculative stores, which is expensive. In many situations this mitigation is conservatively turned on (such as seccomp, and by extension Docker), but isn't actually needed.

There's a `prctl`, `PR_GET_SPECULATION_CTRL` and `PR_SET_SPECULATION_CTRL` to check the status of some opt-in mitigations - currently only `PR_SPEC_STORE_BYPASS`. Notably, installing a seccomp policy sets this to PR_SPEC_FORCE_DISABLE by default, which can lead to quite large performance penalties in Shadow. Similarly, Docker uses seccomp by default, and uses it in a way that sets this to `PR_SPEC_FORCE_DISABLE`.

We should use `PR_GET_SPECULATION_CTRL` as startup to check whether this mitigation is turned on, and if so notify the user that performance will suffer. Ideally we should point them to some hints about opting out of it when using Docker or using seccomp directly.

Somewhat orthogonal, but we should also check whether this is getting turned on inside the tor process when it installs *its* seccomp policy.

sporksmith:
Fixes #1489.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -8,6 +8,7 @@ can fix it.
 
  * The Shadow simulator
    * [Design Overview](design_2x.md)
+     * [Non-goal: Security](security.md)
    * Installation Guide
      * [Supported Platforms](supported_platforms.md)
      * [Dependencies](install_dependencies.md)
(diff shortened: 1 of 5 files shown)


sporksmith:
The Tor test failure looks like unrelated flakiness; digging into it

stevenengler (approved):

PR #1511, opened 2021-07-13
childpid watcher: retry on EINTR

sporksmith:
epoll_wait doesn't automatically restart when a signal is received (see
signal(7)). We need to check for EINTR and retry.


sporksmith:
AFAIK this doesn't have anything to do with non-determinism/failure we were looking at earlier. I noticed it when running Shadow under gdb, since a ptrace-stop would then cause the watcher thread to panic.

stevenengler (approved):

PR #1513, opened 2021-07-14
Added documentation for the 'bootstrap_end_time' option

stevenengler:


sporksmith (approved):
I'd been meaning to find out what that did! Documentation LGTM, though obviously I can't vouch for its correctness :).

Btw I wonder if this could cause tor nodes to learn bad parameters during the bootstrap period? (Hopefully they'd adapt and recover though when the network gets slower + lossier)

stevenengler:
> I'd been meaning to find out what that did! Documentation LGTM, though obviously I can't vouch for its correctness :).
> 
> Btw I wonder if this could cause tor nodes to learn bad parameters during the bootstrap period? (Hopefully they'd adapt and recover though when the network gets slower + lossier)

It feels like it could cause the network to have bad info for things like bandwidth measurements and circuit timeouts, but I don't know whether this ends up being an issue in practice.

Issue #666, opened 2018-12-20
Shadow determinism is broken


robgjansen:
Running the same Shadow experiment twice results in non-deterministic behavior; so far I have observed the following:

- [x] hosts may get assigned different IP addresses
- [x] tgen uses `getpid` and logs Shadow's PID instead of the virtual process PID
- [x] packet events are executed out of order
- [x] gettid using `syscall(SYS_gettid, ...)` should be implemented
- [x] futexes are woken in a non-deterministic order
- [x] descriptors are notified in a non-deterministic order
- [x] ~~non-deterministic behavior when retransmitting packets~~ - no evidence in #2109

robgjansen:
I still need to verify that Tor experiments are deterministic with these changes.

RWails:
I'm able to reproduce non-deterministic results with the following experiment:

[exp.zip](https://github.com/shadow/shadow/files/2728395/exp.zip)

Experiment has two clients who write, one server who reads, all using non-blocking, epoll-managed I/O. Each host periodically logs how much data they've transmitted and corresponding timestamps.

Build the plugin using make.bash and run two instances using the run script.  I am able to reliably produce plugin outputs that differ.

robgjansen:
I updated @RWails test case and here I attach a new version for future reference.
[determinism_test.zip](https://github.com/shadow/shadow/files/2765780/determinism_test.zip)

robgjansen:
After some fixes in #1460, I'm still seeing non-deterministic behavior in the Tor minimal example.

Here is the latest way I've been checking:

```bash
# It's best to build shadow in release mode, but still enable trace logging at runtime.
# That will ensure that the experiment runs more quickly but we still get the desired
# packet-level log lines.
./setup build -ctj6

cd src/test/tor/minimal
chmod 700 shadow.data.template/hosts/hiddenserver/hs

shadow -d a -e shadow.data.template -l trace tor-minimal.yaml | grep "packet_add" > shadow.log
mv shadow.log a/.

shadow -d b -e shadow.data.template -l trace tor-minimal.yaml | grep "packet_add" > shadow.log
mv shadow.log b/.

for d in a/hosts b/hosts
do
    cd ${d}
    for name in *
    do
        cat ../shadow.log | grep "${name}:" | grep "packet_add" | cut -d' ' -f3- > ${name}.log
    done
    cd -
done

for p in a/hosts/*log
do
    host_packet_log=`basename ${p}`
[...]

PR #1515, opened 2021-07-14
Rdtsc in preload mode

sporksmith:
* Improve rdtsc emulation to get clock  speed via cpuid instead of non-deterministic measurement.
* Move rdtsc emulation into common library, where it can also be used by shim
* Emulate rdtsc in the shim when running in preload mode

Progress on #666


sporksmith:
Probably easiest to review one commit at a time - or at least before and after moving the tsc module in 55e1dc5

stevenengler (approved):
Looks good! I tried to look for any simple errors, but didn't go in-depth (didn't double check things with the Intel manual or anything).

It seems like using `cpuid(0x15, ...)` to get the clock frequency probably will work on all x86 CPUs, but I'm guessing the brand string method will fail on non-Intel x86 CPUs? Have you checked with any AMD documentation to see what they say about getting the clock frequency? Do we have any systems with AMD CPUs we can test this on?

sporksmith:
Good question. AFAICT from https://www.amd.com/system/files/TechDocs/25481.pdf, AMD doesn't support cpuid 0x15.  It *does* implement the same API for getting the brand string, but unlike Intel doesn't specify that the speed will be the last token.

I don't have any AMD systems handy to check on; I guess we'll cross that bridge when we get there.

PR #1524, opened 2021-07-19
TSC: Use CPU model map to find crystal frequency

sporksmith:
In https://github.com/shadow/shadow/issues/1519 we found that the
frequency from the CPU brand string can be substantially off from the
TSC rate (3% in this case), so it's worth implementing more of the cpuid
0x15h method of finding the TSC rate.

### DIFF
--- a/src/lib/tsc/tsc.c
+++ b/src/lib/tsc/tsc.c
@@ -50,7 +50,7 @@ static uint64_t _frequency_via_cpuid0x15() {
     }
     // From "cpuid": "An unsigned integer which is the nominal frequency of the
     // core crystal clock in Hz."
-    const unsigned int core = c;
+    unsigned int core = c;
     if (!core) {
         // From "cpuid": "If ECX is 0, the nominal core crystal clock frequency
         // is not enumerated". Gee, thanks.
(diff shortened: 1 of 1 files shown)


stevenengler (approved):
You may have seen this already, but Linux seems to use cpuid 0x16 as well for certain processors. Maybe it would help here too: https://lore.kernel.org/patchwork/patch/1071759/

sporksmith:
> You may have seen this already, but Linux seems to use cpuid 0x16 as well for certain processors. Maybe it would help here too: https://lore.kernel.org/patchwork/patch/1071759/

I hadn't!

I was poking around some more though and found that the current methods don't look like they'll work for AMD, and haven't been able to find a method that *will* reliably work - https://community.amd.com/t5/server-gurus-discussions/how-can-one-find-quot-p0-frequency-quot/td-p/292863

Also came across this article, which basically recommends getting it from the kernel (via a custom module because it isn't exposed by default): https://blog.trailofbits.com/2019/10/03/tsc-frequency-for-all-better-profiling-and-benchmarking/.

Given the complexity of finding the TSC rate, and that the only place we've observed to use the TSC to measure wallclock time is the vdso function (probably because of how complex it is), maybe we're better off just picking a simulated TSC rate and working on more thoroughly overriding vdso.

PR #1527, opened 2021-07-21
logger bugfix: Enforce log levels

sporksmith:
Fixes bug introduced in d167ccd6ee39bc7bdbe9e0d7020f79ab08aeaf6f


robgjansen (approved):
Note: This fix stops the `packet_addDeliveryStatus` string from being logged on every packet event regardless of the log level.

PR #1530, opened 2021-07-21
Improvements for preloading libraries

### LINKED ISSUE #1523: Provide a more ergonomic way of enabling extra preload libraries
Right now if a user wants to use extra LD_PRELOAD libraries (such as the soon-to-be-merged [shadow_openssl_rng](https://github.com/shadow/shadow/pull/1517), they must add it to `LD_PRELOAD` in each simulated process's `environment` in the simulation config. Because `LD_PRELOAD` doesn't support shell expansion, the normal installation path of `/home/<username>/.local/lib/...`  must be spelled out completely in the experiment config, including the literal username.

Maybe we should provide a more specialized config field for such preloads, and then handle it similarly to the shim: search for it in some expected places and then supply the full path where we found it to `LD_PRELOAD`.

robgjansen:
Changes:
- Adds a new `use_openssl_rng_preload` config option that defaults to true and allows control over our preloading of the openssl rng preload lib.
- The path to the openssl rng preload lib is found automatically by searching the rpath, as we do for the preload shim lib.
- For every entry given in the `LD_PRELOAD` key in the `environment` option for processes, we now expand the `~` if one appears as the first character in an entry. This allows users to avoid adding an absolute path to the config file.

I tested this locally in an ad-hoc manner and all features work as intended. It's a bit tricky to add tests to our framework because of the home dir absolute paths. (Unit tests would be ideal here, but the changes are mostly in C code.)

Closes #1523

### DIFF
--- a/src/main/core/manager.c
+++ b/src/main/core/manager.c
@@ -379,7 +379,7 @@ void manager_addNewVirtualHost(Manager* manager, HostParameters* params) {
 }
 
 static gchar** _manager_generateEnvv(Manager* manager, InterposeMethod interposeMethod,
-                                     const gchar* preloadShimPath, const gchar* environment) {
+                                     const gchar* environment) {
     MAGIC_ASSERT(manager);
 
     /* start with an empty environment */
@@ -405,11 +405,19 @@ static gchar** _manager_generateEnvv(Manager* manager, InterposeMethod interpose
     /* insert also the plugin preload entry if one exists.
      * precendence here is:
      *   - preload path of the shim
-     *   - preload values from LD_PRELOAD entries in the environment attribute of the shadow
-     * element*/
+     *   - preload path of the openssl rng lib
+     *   - preload values from LD_PRELOAD entries in the environment process option */
     GPtrArray* ldPreloadArray = g_ptr_array_new();
-    debug("adding shim path %s", preloadShimPath ? preloadShimPath : "null");
-    g_ptr_array_add(ldPreloadArray, g_strdup(preloadShimPath));
+
+    if (manager->preloadShimPath != NULL) {
+        debug("adding required preload shim path %s", manager->preloadShimPath);
+        g_ptr_array_add(ldPreloadArray, g_strdup(manager->preloadShimPath));
+    }
+
+    if (manager->preloadOpensslRngPath != NULL) {
+        debug("adding optional preload lib path %s", manager->preloadOpensslRngPath);
+        g_ptr_array_add(ldPreloadArray, g_strdup(manager->preloadOpensslRngPath));
+    }
 
     /* now we also have to scan the other env variables that were given in the shadow conf file */
 
@@ -468,8 +476,7 @@ void manager_addNewVirtualProcess(Manager* manager, const gchar* hostName, gchar
     InterposeMethod interposeMethod = config_getInterposeMethod(manager->config);
 
     /* ownership is passed to the host/process below, so we don't free these */
-    gchar** envv = _manager_generateEnvv(
-        manager, interposeMethod, manager->preloadShimPath, environment);
+    gchar** envv = _manager_generateEnvv(manager, interposeMethod, environment);
 
     Host* host = scheduler_getHost(manager->scheduler, hostID);
     host_continueExecutionTimer(host);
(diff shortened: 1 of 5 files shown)


src/main/core/manager.c
```diff
@@ -409,9 +442,40 @@ static gchar** _manager_generateEnvv(Manager* manager, InterposeMethod interpose
...
+                    gchar** paths = g_strsplit(preloadBuf->str, ":", 0);
+
+                    // Expand any paths that begin with '~'.
+                    for (int i = 0; paths != NULL && paths[i] != NULL; i++) {
+                        GString* pathBuf = g_string_new(NULL);
+
+                        if (!g_ascii_strncasecmp(paths[i], "~/", 2)) {
+                            g_string_printf(pathBuf, "%s%s", g_get_home_dir(), &paths[i][1]);
+                        } else if (!g_ascii_strncasecmp(paths[i], "~", 1)) {
+                            g_string_printf(pathBuf, "%s/%s", g_get_home_dir(), &paths[i][1]);
```
stevenengler, line 466: What's the purpose of this condition? This seems like it doesn't work for paths like `~steve/.local/bin`, which will turn into `/home/steve/steve/.local/bin`.

stevenengler (approved):

src/main/core/manager.c
robgjansen: I was trying to avoid expanding to `/home/steve//.local` (with two slashes), but I didn't realize that `~name` is usually expanded to `/home/name`.

I could just remove that condition and then we only support expanding `~/` but not `~`? Or keep the condition as is but change line 466 to
```
g_string_printf(pathBuf, "/home/%s", &paths[i][1]);
```

Which do you prefer?

src/main/core/manager.c
```diff
@@ -409,9 +442,40 @@ static gchar** _manager_generateEnvv(Manager* manager, InterposeMethod interpose
...
+                    }
+
+                    // Now split by ':' to handle each preload path.
+                    gchar** paths = g_strsplit(preloadBuf->str, ":", 0);
+
+                    // Expand any paths that begin with '~'.
+                    for (int i = 0; paths != NULL && paths[i] != NULL; i++) {
+                        GString* pathBuf = g_string_new(NULL);
+
+                        if (!g_ascii_strncasecmp(paths[i], "~/", 2)) {
```
sporksmith, line 463: Nit - I don't know of a benefit to hard-coding the string length when one of the strings being compared is hard-coded; the comparison will stop when the end of one of the strings is reached. i.e. I think this can be `if (!g_ascii_strcasecmp(paths[i], "~/")) {`

sporksmith (approved):
Maybe something to consider for a future PR - Now that you've generalized `_manager_scanRPathForLib`, it looks like it'd be easy to add a more general global and/or per-process list of preload-libraries to dynamically search for and add to LD_PRELOAD.

I guess there's not that much benefit when we only have 1 such optional preload lib, but it'd also be useful for "third party" preload libs meant for use with shadow, analogous to plugins in 1.x.

src/main/core/manager.c
robgjansen: I went withy the second approach :)

stevenengler: I'm not really sure what you mean here, or why there would be two slashes, but using `"/home/%s"` looks good to me.

src/main/core/manager.c
robgjansen: Done in ff4d95aedfe270fd40c18317504f919c7216d70a

stevenengler: I think the number is needed so that it only compares the first characters. For example I think `g_ascii_strncasecmp("~/test", "~/", 2) == 0`, but `g_ascii_strcasecmp("~/test", "~/") > 0`.

sporksmith: Oh, of course you're right. Sorry, my mistake!

I think g_str_has_prefix is the nicer thing to use here? https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-str-has-prefix

stevenengler: (I unresolved this since I think we want to change it back before merging.)

robgjansen: Good catch!

I tried again in 2e8d80a447c638cf38008a50371c14b489e8668a

stevenengler: Looks good to me.

robgjansen: Hmm, it looks like unresolving did not stop the auto-merge.

PR #1535, opened 2021-07-23
Set errno with appropriate value in _getaddrinfo_service 

cohosh:
Noticed this while debugging a test failure.


stevenengler (approved):
Thanks for the fix!

Issue #1532, opened 2021-07-22
pidfd_open is not supported in kernel 4.15 on Ubuntu 18.04


marcosimioni:
**Describe the issue**
I'm on the latest commit `103814bfff08b7ddfb1960bc9e263dbaeb437ecc`, shadow is building fine but some tests are failing (list below).

The latest working commit on my machine seems to be https://github.com/shadow/shadow/commit/4730119f9d644d6b2fcecb0fc190e245e4f6a3f7 and tests start failing on https://github.com/shadow/shadow/commit/022d3bf156ec434486660c3de081676f8c3e352e (and intermediate commits with slightly different behaviors, but those were WIP commits I guess).

Looking at https://github.com/shadow/shadow/runs/3129268581?check_suite_focus=true, I see that the CI is on `kernel 5.8.0-1036-azure`, while my machine is on kernel `4.15.0-137-generic`. Do you think that might be the root issue with those changes?

Or, by looking at the tests below, do you have any pointer?

```
64% tests passed, 58 tests failed out of 159                                             
                                                                                         
Total Test time (real) =  24.70 sec                                                      
                                                                                         
The following tests FAILED:                                                              
          1 - tsc_test (Failed)                                                          
          3 - rust-unit-tests (Timeout)                                                  
          6 - bindc-shadow-preload (Failed)                                              
         11 - config-stdin-shadow-preload (Failed)                                       
         14 - cpp-shadow-preload (Failed)                                                
         18 - determinism1a-shadow-preload (Failed)                                      
         19 - determinism1b-shadow-preload (Failed)                                      
         20 - determinism1-compare-shadow-preload (Failed)                               
         24 - determinism2a-shadow-preload (Failed)                                      
         25 - determinism2b-shadow-preload (Failed)                                      
         26 - determinism2-shadow-preload-compare (Failed)                               
         28 - environment-shadow-preload (Failed)                                        
         31 - epoll-shadow-preload (Failed)                                              
         33 - epoll-writeable-shadow-preload (Failed)  
         36 - eventfd-shadow-preload (Failed)                  
         39 - exit-shadow-preload (Failed)                      
         42 - file-shadow-preload (Failed)                     
         45 - futex-shadow-preload (Failed)                     
         48 - ifaddrs-shadow-preload (Failed)               
         51 - mmap-shadow-preload (Failed)                   
         54 - unaligned-shadow-preload (Failed)                                          
         56 - phold-serial-shadow-preload (Failed)                                       
[...]

PR #1539, opened 2021-07-23
ChildPidWatcher: remove dependence on pidfd_open

sporksmith:
pidfd_open was only added in Linux 5.3. There are still some platforms
supported by Shadow that may not have a new enough kernel to use it.

Fixes https://github.com/shadow/shadow/issues/1532

### DIFF
--- a/src/main/utility/childpid_watcher.rs
+++ b/src/main/utility/childpid_watcher.rs
@@ -394,10 +394,10 @@ mod tests {
         }
 
         // Child should be ready to be reaped.
-        assert_eq!(
-            waitpid(child, Some(WaitPidFlag::WNOHANG)).unwrap(),
-            WaitStatus::Exited(child, 42)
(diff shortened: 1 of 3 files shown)


src/main/utility/childpid_watcher.rs
```diff
@@ -166,51 +192,133 @@ impl ChildPidWatcher {
         }
     }
 
-    /// Call `callback` exactly once from another thread after the child `pid`
+    unsafe fn fork_watchable_internal(
+        &self,
+        fork_syscall: i64,
+        child_fn: impl FnOnce(),
+    ) -> Result<Pid, nix::Error> {
```
stevenengler, line 199: It's not clear to me why `fork_watchable_internal()` or `fork_watchable()` is `unsafe`.

stevenengler (approved):

PR #1547, opened 2021-07-29
Increase RLIMIT_NPROC to hard limit

### LINKED ISSUE #1534: Use `setrlimit()` to increase process limits
In Shadow we use `getrlimit()` to get the system file descriptor (`RLIMIT_NOFILE`) hard limit, and then use `setrlimit()` to increase shadow's limit to this maximum. There are other limits we could also increase, such as the process limit (`RLIMIT_NPROC`).

List of all possible limits:

- `RLIMIT_AS` - "This is the maximum size of the process's virtual memory (address space)"
- `RLIMIT_CORE` - "This is the maximum size of a core file in bytes that the process may dump"
- `RLIMIT_CPU` - "This is a limit, in seconds, on the amount of CPU time that the process can consume"
- `RLIMIT_DATA` - "This is the maximum size of the process's data segment (initialized data, uninitialized data, and heap)"
- `RLIMIT_FSIZE` - "This is the maximum size in bytes of files that the process may create"
- `RLIMIT_LOCKS` - (early Linux 2.4 only) - "This is a limit on the combined number of flock locks and fcntl leases that this process may establish"
- `RLIMIT_MEMLOCK` - "This is the maximum number of bytes of memory that may be locked into RAM"
- `RLIMIT_MSGQUEUE` - (since Linux 2.6.8) - "This is a limit on the number of bytes that can be allocated for POSIX message queues for the real user ID of the calling process"
- `RLIMIT_NICE` - "This specifies a ceiling to which the process's nice value can be raised using setpriority or nice"
- `RLIMIT_NOFILE` - "This specifies a value one greater than the maximum file descriptor number that can be opened by this process"
- `RLIMIT_NPROC` - "This is a limit on the number of extant process (or, more precisely on Linux, threads) for the real user ID of the calling process"
- `RLIMIT_RSS` - "This is a limit (in bytes) on the process's resident set (the number of virtual pages resident in RAM)"
- `RLIMIT_RTPRIO` - (since Linux 2.6.12, but see BUGS) - "This specifies a ceiling on the real-time priority that may be set for this process using sched_setscheduler and sched_setparam"
- `RLIMIT_RTTIME` - (since Linux 2.6.25) - "This is a limit (in microseconds) on the amount of CPU time that a process scheduled under a real-time scheduling policy may consume without making a blocking system call"
- `RLIMIT_SIGPENDING` - (since Linux 2.6.8) - "This is a limit on the number of signals that may be queued for the real user ID of the calling process"
- `RLIMIT_STACK` - "This is the maximum size of the process stack, in bytes"

stevenengler:
Closes #1534. I don't see any other limits we would want to increase.

### DIFF
--- a/src/main/core/main.c
+++ b/src/main/core/main.c
@@ -59,6 +59,32 @@ static void _main_logEnvironment(gchar** argv, gchar** envv) {
     }
 }
 
+static int _raise_rlimit(int resource) {
+    char* debug_name = "?";
+
+    /* can add more as needed */
+    switch (resource) {
+        case RLIMIT_NOFILE: debug_name = "RLIMIT_NOFILE"; break;
+        case RLIMIT_NPROC: debug_name = "RLIMIT_NPROC"; break;
+    }
+
+    struct rlimit lim = {0};
+    if (getrlimit(resource, &lim) != 0) {
+        error(
+            "Could not get rlimit for resource %s (%d): %s", debug_name, resource, strerror(errno));
+        return -1;
+    }
+
+    lim.rlim_cur = lim.rlim_max;
+    if (setrlimit(resource, &lim) != 0) {
+        error("Could not update rlimit for resource %s (%d): %s", debug_name, resource,
+              strerror(errno));
+        return -1;
+    }
+
+    return 0;
+}
+
 static void _check_mitigations() {
     int state = prctl(PR_GET_SPECULATION_CTRL, PR_SPEC_STORE_BYPASS, 0, 0, 0);
     if (state == -1) {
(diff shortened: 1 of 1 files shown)


src/main/core/main.c
```diff
@@ -59,6 +59,22 @@ static void _main_logEnvironment(gchar** argv, gchar** envv) {
     }
 }
 
+static int _raise_rlimit(int resource) {
+    struct rlimit lim = {0};
+    if (getrlimit(resource, &lim) != 0) {
+        error("Could not get rlimit: %s", strerror(errno));
```
robgjansen, line 65: Could we add the type of rlimit (files or procs, etc) that we are failing to raise? Ideally a string but even the int would be better than nothing.

src/main/core/main.c
```diff
@@ -59,6 +59,22 @@ static void _main_logEnvironment(gchar** argv, gchar** envv) {
...
+        return -1;
+    }
+
+    lim.rlim_cur = lim.rlim_max;
+    if (setrlimit(resource, &lim) != 0) {
+        error("Could not update rlimit: %s", strerror(errno));
```
robgjansen, line 71: Add limit type to the error message here too.

robgjansen (approved):

src/main/core/main.c
stevenengler: Added.

src/main/core/main.c
stevenengler: Added.

Issue #1540, opened 2021-07-26
Support syscalls necessary for Go runtime


cohosh:
**Describe the issue**
I'm working on getting Snowflake simulations running in Shadow, and for all of the Go binaries I'm running up against the same unsupported system calls. Here's an example from the broker:

```
00:00:00.079100 [worker-2] 00:00:01.000000000 [ERROR] [broker:11.0.0.1] [syscall_handler.c:506] [syscallhandler_make_syscall] Returning error 38 (ENOSYS) for unsupported syscall 204, which may result in unusual behavior
00:00:00.130658 [worker-2] 00:00:01.000000000 [ERROR] [broker:11.0.0.1] [syscall_handler.c:506] [syscallhandler_make_syscall] Returning error 38 (ENOSYS) for unsupported syscall 281, which may result in unusual behavior
00:00:00.131060 [worker-2] 00:00:01.000000000 [ERROR] [broker:11.0.0.1] [syscall_handler.c:506] [syscallhandler_make_syscall] Returning error 38 (ENOSYS) for unsupported syscall 24, which may result in unusual behavior
```
The unsupported system calls are:
- `sched_getaffinity (204)`
-  `epoll_pwait (281)`

It's also throwing an `ENOSYS` error for `sched_yield (24)`  which is strange because AFAICT from looking at the source code, it's implemented.

Since these are being called from the Go runtime library and I've tried setting `GOMAXPROCS=1` with no success, I suspect they are necessary for all Go programs (or at least all that have any goroutines). 

**To Reproduce**
Try running any Go program with goroutines

**Shadow (please complete the following information):**
 - Version: `Shadow 2.0.0-pre.4`

cohosh:
Also thank you so much for all this awesome Shadow work, I'm really excited about this project!

cohosh:
Also, fwiw it might be that `epoll_pwait` is the only one that's strictly necessary as it was almost certainly the cause of this go panic:
```
runtime: epollwait on fd 5 failed with 38
fatal error: runtime: netpoll failed

runtime stack:
runtime.throw(0x98829b, 0x17)
        /usr/local/go/src/runtime/panic.go:1116 +0x72
runtime.netpoll(0x4e94914f0000, 0x1)
        /usr/local/go/src/runtime/netpoll_epoll.go:130 +0x368
runtime.findrunnable(0xc000030000, 0x0)
        /usr/local/go/src/runtime/proc.go:2448 +0x785
runtime.schedule()
[...]

PR #1548, opened 2021-07-30
Fixed bug where epoll ready list contained non-ready items

stevenengler:
Fixes the issue where shadow gets stuck in #1540. The bug occurs with edge-triggered watch objects.

### DIFF
--- a/src/main/host/descriptor/epoll.c
+++ b/src/main/host/descriptor/epoll.c
@@ -5,6 +5,7 @@
  */
 #include "main/host/descriptor/epoll.h"
 
+#include <assert.h>
 #include <errno.h>
 #include <glib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


src/main/host/descriptor/epoll.c
```diff
@@ -646,6 +649,18 @@ gint epoll_getEvents(Epoll* epoll, struct epoll_event* eventArray, gint eventArr
...
+    gpointer val;
+
+    /* We modified some watched objects above, so remove any that are no longer ready. */
+    while (g_hash_table_iter_next(&iter, NULL, (gpointer)&val)) {
+        EpollWatch* watch = val;
+        if (!_epollwatch_isReady(watch)) {
+            g_hash_table_iter_remove(&iter);
+        }
+    }
+
```
robgjansen, line 663: I don't think we want to go through ALL of the watches when we only possibly changed a small number of them in the while loop above.

Instead of iterating the hash table here, how about we add the `isReady` check above immediately following line 634 (but still in the `if(_epollwatch_isReady(watch)) {` scope? That way we are only checking those events that we edit in the while loop, for which we may have caused the ready status to change (with the edits on line 629 and 633). If we have caused the event to become not ready, we could add it to a list or queue, and replace this hash table iterator with the list or queue iterator that removes those watches from the ready table.

Does this make sense?

robgjansen (changes requested):
I think we could be more efficient (see comment); let me know if you want to discuss.

src/main/host/descriptor/epoll.c
stevenengler: I changed it to store them in a list, and then later remove those from the hash table. Two downsides are (1) it requires memory allocations for the list, and (2) instead of looping over the hash table items, we need to iterate over the keys and perform hash table lookups for the items. But I don't think these downsides are very important.

robgjansen (approved):

PR #1583, opened 2021-08-19
Simplify syscall condition and timeouts

sporksmith:
This is primarily to make timeout handling easier to expose in Rust.

* Remove SyscallHandler::timer, since in practice it was redundant with SyscallCondition::timeout.
* Deduplicate code to read and validate a timespec

### DIFF
--- a/src/main/host/syscall/protected.c
+++ b/src/main/host/syscall/protected.c
@@ -16,6 +16,8 @@
 #include "main/host/descriptor/descriptor.h"
 #include "main/host/descriptor/tcp.h"
 #include "main/host/descriptor/timer.h"
+#include "main/host/syscall_condition.h"
+#include "main/host/thread.h"
 
 void _syscallhandler_setListenTimeout(SysCallHandler* sys, const struct timespec* timeout,
                                       TimeoutType type) {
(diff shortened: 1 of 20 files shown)


src/main/host/syscall_condition.c
```diff
@@ -33,7 +33,7 @@ struct _SysCallCondition {
     // The thread waiting for the signal
     Thread* thread;
     // If a task to deliver a signal has been scheduled
-    bool signalPending;
+    bool signalScheduled;
```
robgjansen, line 37: I think there are other parts of the code (in epoll maybe?) where "scheduled" is a state that turns back to false when the task later comes out of the event queue and is being run. Might be useful to either rename the bool, or add a comment to explain that once it becomes true we never set it to false again, or both.

src/main/host/syscall_condition.c
```diff
@@ -458,3 +459,5 @@ void syscallcondition_cancel(SysCallCondition* cond) {
     _syscallcondition_cleanupListeners(cond);
     _syscallcondition_cleanupProc(cond);
 }
+
+Timer* syscallcondition_timeout(SysCallCondition* cond) { return cond->timeout; }
```
robgjansen, line 463: Again, I think `_getTimeout` is a better name here.

src/main/host/syscall/protected.c
```diff
@@ -45,26 +47,47 @@ void _syscallhandler_setListenTimeoutMillis(SysCallHandler* sys,
     _syscallhandler_setListenTimeout(sys, &timeout, TIMEOUT_RELATIVE);
 }
 
-int _syscallhandler_isListenTimeoutPending(SysCallHandler* sys) {
+static const Timer* _syscallhandler_timeout(const SysCallHandler* sys) {
```
robgjansen, line 22: I think `_getTimeout` is a better name here - since `timeout` can be used as a verb, it could be mistaken for a setter (a mistake I made).

robgjansen (approved):
Looks great!

PR #1584, opened 2021-08-20
Add and use EmulatedTime type in Rust

sporksmith:
Unambiguously represents an instant in time in the simulation, analagous to std::time::Instant or std::time::SystemTime.


stevenengler (approved):
Looks good to me.

I'm not clear on why the emulated time is preferred to simulated time. My understanding was that simulated time is like an absolute time, and the emulated time is like the host's view of the time. So things like packet times or the simulation end time would be represented as a simulation time, and interactions with the host through syscalls would be represented as an emulated time. But maybe this distinction isn't needed since all hosts have a consistent view of the time. Is the plan for `SimulationTime` to disappear as more code moves to rust?

sporksmith:
> I'm not clear on why the emulated time is preferred to simulated time. My understanding was that simulated time is like an absolute time, and the emulated time is like the host's view of the time.

In the C code, both can be used to represent an instant of time in the simulation (either as simulation-time-units since sim-start, or since unix-epoch), but the SimulationTime type is also used to represent time intervals, making values of that type a bit ambiguous.

As with Rust's native time types, we make the inner "value" opaque, which makes the "epoch" that value is offset from also opaque. This effectively removes the difference between EmulatedTime and SimulationTime (when SimulationTime is used to represent an absolute time), so we don't need both. If you have a Rust EmulatedTime x you can get something like the C EmulatedTime by calling `x.duration_since(emulated_time::UNIX_EPOCH)`; or something like the C SimulationTime by calling `x.duration_since(emulated_time::SIMULATION_START)`. 

> So things like packet times or the simulation end time would be represented as a simulation time, and interactions with the host through syscalls would be represented as an emulated time.

That's what the C code does, but in Rust code we can use EmulatedTime for both. 

> But maybe this distinction isn't needed since all hosts have a consistent view of the time. Is the plan for SimulationTime to disappear as more code moves to rust?

Yup!

If you haven't worked with this style of time API before, the docs for Google's time API are pretty nice https://abseil.io/docs/cpp/guides/time. (We don't need to worry about an equivalent of their CivilTime thankfully; just Time and Duration)

PR #1616, opened 2021-08-30
Fixed documentation links

lnwor:
I just fixed a couple links that didn't get updated after the last commit to the shadow config specification


stevenengler (approved):
Thanks for catching these!

PR #1627, opened 2021-09-03
Re-schedule syscall condition after spurious wakeup

### LINKED ISSUE #1623: tgen getting stuck in epoll_wait after ~8m-15m simulated time in tor simulations
First noticed the problem running tor congestion control simulations in https://gitlab.torproject.org/jnewsome/sponsor-61-sims. Traced failed transfers to the tgen servers appearing to get deadlocked after somewhere around 8-15m of simulated time. The tgen output stops at that point, including heartbeats.

The problem occurs at shadow commit d64d07c65859090dc76a27fb3d9ef44ea9d756f3, but not on the previous commit b507cc1ed0fec74a0d02448401578f8b07add825.

After hacking up the minimal tor unit test a bit, I'm able to reproduce locally: https://github.com/sporksmith/shadow/tree/repro-bug. I extended the length of the simulation time, removed the non-tor clients since they're unrelated afaict. I added more tor clients, though I'm not sure that's essential. In this setup the bug takes a while longer to trigger - e.g. over an hour (see example below)

By attaching gdb to the fileserver tgen after it stops producing output, I was able to verify that it's hanging in a call to `epoll_wait`, with ~a timeout of 1 ms~ [edit: no timeout. One of the file descriptors watched by the epoll is a timerfd. See below.] I can confirm in the Shadow log that the process isn't scheduled again after that.

sporksmith:
Fixes #1623

### DIFF
--- a/src/main/host/syscall_condition.c
+++ b/src/main/host/syscall_condition.c
@@ -33,10 +33,9 @@ struct _SysCallCondition {
     Process* proc;
     // The thread waiting for the signal
     Thread* thread;
-    // Whether the condition has *ever* been triggered. Used to prevent multiple
-    // signals from a single condition, even if its triggered multiple times or
-    // ways.
-    bool signalTriggered;
+    // Whether there is currently a signal triggered for the condition.
+    // Used to ensure there is only ever one signal scheduled at once.
+    bool signalPending;
     // Memory tracking
     gint referenceCount;
     MAGIC_DECLARE;
@@ -310,6 +309,11 @@ static void _syscallcondition_signal(Host* host, void* obj, void* arg) {
     bool wasTimeout = (bool)arg;
     MAGIC_ASSERT(cond);
 
+    // The signal callback is executing here and now. Setting to false allows
+    // the callback to be scheduled again if the condition isn't canceled
+    // (which it will be, if we decide to actually run the process below).
+    cond->signalPending = false;
+
 #ifdef DEBUG
     _syscallcondition_logListeningState(cond, "signaling while");
 #endif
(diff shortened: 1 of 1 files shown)


sporksmith:
This seems to fix the bug in my local repro. Trying it out on the larger tor congestion control sim here https://gitlab.torproject.org/jnewsome/sponsor-61-sims/-/jobs/35639

sporksmith:
Just realized how https://github.com/shadow/shadow/pull/1583/commits/a18da0585c9f526877117adfc6ff8ceada556cca broke this in the first place. I'd removed `cond->signalPending = false;`, thinking it was effectively a no-op since we're no longer listening at that point. That's incorrect - we *are* still listening at that point, so flipping the boolean there does allow the signal to be scheduled again in the case that we decide not to actually run the process below.

Going to make a small simplifcation

sporksmith:
It's probably worth adding a focused unit test that catches this, but I probably won't get to it until I'm back on Tuesday. Would be good to merge this sooner to unbreak shadow @ head. Opened #1628 to add the regression test.

src/main/host/syscall_condition.c
```diff
@@ -310,6 +309,8 @@ static void _syscallcondition_signal(Host* host, void* obj, void* arg) {
     bool wasTimeout = (bool)arg;
     MAGIC_ASSERT(cond);
 
+    cond->signalPending = false;
```
robgjansen, line 312: Might we worth a comment:
The signal task is now being executed by the worker and is no longer pending in the host's event queue. This is necessary, since it's possible that we want to schedule another signal task before this condition object is destroyed.

robgjansen (approved):

src/main/host/syscall_condition.c
sporksmith: Done

PR #1631, opened 2021-09-07
Regression test for #1623

sporksmith:
I ended up having to make some ~unrelated changes to get the regression test working; probably easiest to review one commit at a time.

* Fixes build files to allow shadow-rs to depend on shadow-c (in particular, prevent the linker from dropping object files in shadow-c that are only depended on via shadow-rs).
* Move fcntl to Rust. This needs the dependency change so that the linker doesn't drop syscallhandler_fcntl, which rustsyscallhandler_fcntl falls back to for legacy descriptors.
* Add the regression test. The simplest way I could think to implement it is with a pipe. This needs the fcntl change so that fcntl works with pipes.


src/main/host/descriptor/channel.c
```diff
@@ -131,6 +131,12 @@ static gssize channel_sendUserData(Transport* transport, Thread* thread, PluginV
         descriptor_adjustStatus((LegacyDescriptor*)channel, STATUS_DESCRIPTOR_WRITABLE, FALSE);
     }
 
+    /* The read-end of the channel is now readable if data is present */
+    if (channel->bufferLength > 0) {
+        descriptor_adjustStatus(
+            (LegacyDescriptor*)channel->linkedChannel, STATUS_DESCRIPTOR_READABLE, TRUE);
+    }
```
stevenengler, line 138: I think this should already be done by the `channel_linkedWrite()` call above:

https://github.com/shadow/shadow/blob/1678d00cba5ef30c780efb67616a52a2b7e31828/src/main/host/descriptor/channel.c#L98-L99

sporksmith: Oops, just saw this. Yeah this is one of the things I fixed. Rebasing now...

stevenengler (approved):

PR #1632, opened 2021-09-08
Remove use-c-syscalls option

sporksmith:
This was added as a fallback when first experimenting with Rust syscall
handlers. Now that the latter is stable, it no longer makes sense to
maintain this extra code path.


stevenengler (approved):

sporksmith:
@robgjansen could you drop the "shadow (ubuntu:20.04, clang, use-c-syscalls)" CI requirement?

robgjansen:
> @robgjansen could you drop the "shadow (ubuntu:20.04, clang, use-c-syscalls)" CI requirement?

Done!

PR #1633, opened 2021-09-09
Rewrote shadow's main function in rust

stevenengler:
Most of the changed lines are just bindings. The main change is the rewriting of `main_runShadow()` in `core::main::run_shadow()`. The general flow should be the same, but a few things have been re-organized.

A few parts that can't easily be converted to rust (for example checking the glib version) remain in the C code, but the rest has moved to rust.

~This is currently using an unreleased version of nix for get/setrlimit support, so it either needs to wait for a new release, or needs a workaround.~ Using libc for now.


src/main/core/main.rs
```diff
@@ -0,0 +1,258 @@
...
+use clap::Clap;
+use nix::sys::{personality, signal};
+
+use crate::core::logger::log_wrapper::c_to_rust_log_level;
+use crate::core::logger::shadow_logger;
+use crate::core::support::configuration::{CliOptions, ConfigFileOptions, ConfigOptions};
+use crate::cshadow as c;
+
+/// Main entry point for the simulator.
+pub fn run_shadow<'a, I>(args: I) -> anyhow::Result<()>
```
sporksmith, line 14: Nitpick, but for code that's called from one place and never going to be on the hot path, perhaps better to simplify to just taking a concrete type here instead of making generic over iterators

src/main/core/main.rs
```diff
@@ -0,0 +1,258 @@
...
+    let log_level = c_to_rust_log_level(log_level).unwrap();
+    if log_level > log::STATIC_MAX_LEVEL {
+        log::warn!(
+            "Log level set to {}, but messages higher than {} have been compiled out",
+            log_level,
+            log::STATIC_MAX_LEVEL,
+        );
+    }
+
+    // before we run the simluation, clean up any orphaned shared memory
```
sporksmith, line 91: simulation

src/main/core/main.rs
```diff
@@ -0,0 +1,258 @@
...
+    // one to improve determinism in cases when an executable under simulation
+    // branch on memory addresses.
+    disable_aslr().context("Could not disable plugin address space layout randomization")?;
+    log::debug!("ASLR disabled for processes forked from this parent process");
+
+    // check sidechannel mitigations
+    if unsafe { c::main_sidechannelMitigationsEnabled() } {
+        log::warn!(
+            "Speculative Store Bypass sidechannel mitigation is enabled (perhaps by seccomp?). \
+             This typically adds ~30%% performance overhead."
```
sporksmith, line 122: I think this can just be a single `%` now, since it's not a C format string

src/main/core/main.rs
```diff
@@ -0,0 +1,258 @@
...
+        return Err(anyhow::anyhow!("Controller exited with code {}", rv));
+    }
+
+    Ok(())
+}
+
+fn pause_for_gdb_attach() -> anyhow::Result<()> {
+    let pid = nix::unistd::getpid();
+    log::info!(
+        "Pausing with SIGTSTP to enable debugger attachment (pid {})",
```
sporksmith, line 156: SIGSTOP

src/main/core/main.rs
```diff
@@ -0,0 +1,258 @@
...
+        pid
+    );
+    eprintln!(
+        "** Pausing with SIGTSTP to enable debugger attachment (pid {})",
```
sporksmith, line 160: SIGSTOP

sporksmith (approved):

src/main/core/main.rs
stevenengler: I made this generic so that we can easily drop in `std::env::args_os()` (which implements `Iter`) later when the real main function is eventually implemented in rust. But we can make this take a vector instead for now.

src/main/core/main.rs
stevenengler: Fixed.

src/main/core/main.rs
stevenengler: Fixed.

src/main/core/main.rs
stevenengler: Oops, good catch! I changed the `raise(SIGSTOP)` to `raise(SIGTSTP)` instead to match the original C code.

src/main/core/main.rs
stevenengler: See above comment.

Issue #1637, opened 2021-09-13
Support the Salticidae minimal C++ asynchronous network library (`FIONBIO`, `TCP_NODELAY`, and `CLOCK_PROCESS_CPUTIME_ID`)


bergerch:
**Describe the issue**
Dear developers,

I encountered a problem running an application that uses the [Salticidae minimal C++ asynchronous network library](https://github.com/Determinant/salticidae), to be concrete tried simulating [HotStuff](https://github.com/hot-stuff/libhotstuff), a blockchain consensus protocol. Unfortunately, it did not run out of the box, it seems some required functionality of the network library used by HotStuff does not go well yet, or is unsupported (?) in Shadow. Maybe you can help me to shed some light on the issue (I'm very interested if it can work now or in future):

The Shadow log basically warns because of an unsupported ioctl operation (I guess the the network library tries to open a socket?)

[syscallhandler_ioctl] We do not support ioctl request 21537 on descriptor 9 of type 6

**Shadow.log**
```
00:00:00.000001 [shadow] n/a [INFO] [n/a] [disable_aslr.c:14] [disable_aslr] ASLR disabled for processes forked from this parent process.
00:00:00.000023 [shadow] n/a [INFO] [n/a] [main.c:104] [_main_helper] Starting Shadow v2.0.0-pre.4-159-g2dbbeddc 2021-09-04--09:38:58 with GLib v2.64.6
00:00:00.000068 [shadow] n/a [INFO] [n/a] [main.c:111] [_main_helper] Shadow was built from branch main on 2021-09-07--16:16:32 in Release mode with compile options: -ggdb;-fno-omit-frame-pointer;-Wreturn-type;-Wswitch;-O3 and link options: -Wl,--no-as-needed
00:00:00.000069 [shadow] n/a [INFO] [n/a] [main.c:112] [_main_helper] For more information, visit https://shadow.github.io or https://github.com/shadow
00:00:00.000070 [shadow] n/a [INFO] [n/a] [main.c:113] [_main_helper] logging current startup arguments and environment
00:00:00.000081 [shadow] n/a [INFO] [n/a] [main.c:41] [_main_logEnvironment] arg: shadow
00:00:00.000081 [shadow] n/a [INFO] [n/a] [main.c:41] [_main_logEnvironment] arg: shadow.yaml
00:00:00.000084 [shadow] n/a [INFO] [n/a] [main.c:119] [_main_helper] startup checks passed, we are ready to start simulation
[...]

PR #1638, opened 2021-09-13
Support FIONBIO for eventfd descriptors

stevenengler:
For #1637.

### DIFF
--- a/src/main/host/syscall/ioctl.c
+++ b/src/main/host/syscall/ioctl.c
@@ -11,6 +11,7 @@
 
 #include "lib/logger/logger.h"
 #include "main/host/descriptor/descriptor.h"
+#include "main/host/descriptor/eventd.h"
 #include "main/host/descriptor/file.h"
 #include "main/host/descriptor/socket.h"
 #include "main/host/descriptor/tcp.h"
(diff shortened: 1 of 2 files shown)


src/main/host/syscall/ioctl.c
```diff
@@ -173,6 +174,32 @@ static int _syscallhandler_ioctlUDPHelper(SysCallHandler* sys, UDP* udp, int fd,
     return result;
 }
 
+static int _syscallhandler_ioctlEventDHelper(SysCallHandler* sys, EventD* eventd, int fd,
+                                             unsigned long request, PluginPtr argPtr) {
+    int result = -EINVAL;
+
+    switch (request) {
+        case FIONBIO: {
+            const int* val = process_getReadablePtr(sys->process, argPtr, sizeof(int));
```
sporksmith, line 183: Nit - for a small value like an int, `process_readPtr`  is preferable. It'll be faster in the case that the MM goes through the `readv` path, since it doesn't need to allocate and deallocate a buffer. It should be no slower in the case that the MM goes through the mapped path, since the cost to copy the int is negligible.

sporksmith: Remember to check for a NULL return value (or an error indicator from `process_readPtr`)

sporksmith (approved):

src/main/host/syscall/ioctl.c
stevenengler: I updated all of the ioctl syscall handlers to use process_readPtr and process_writePrt.

Issue #1358, opened 2021-05-13
Monitor resource usage and warn when remaining resources are low


stevenengler:
Shadow can use a lot of system resources like file descriptors, memory maps, ram, etc. We should monitor these resources while Shadow is running and log a warning message if we appear to be near the system's limits.

This should not be expensive since we only need to monitor the shadow process and not the plugins, so should only take a few syscalls to get the information we need.

It's undecided where the best place is to check the resource usage, but one option is in `_manager_heartbeat()` which logs every X simulated seconds.

PR #1644, opened 2021-09-15
Warn if memory or fd usage is high

stevenengler:
For #1358.

### DIFF
--- a/src/main/core/manager.c
+++ b/src/main/core/manager.c
@@ -83,6 +83,10 @@ struct _Manager {
 
     StatusBar_ShadowStatusBarState* statusBar;
 
+    time_t timeOfLastUsageCheck;
+    bool checkFdUsage;
+    bool checkMemUsage;
+
     MAGIC_DECLARE;
 };
 
@@ -283,6 +287,9 @@ Manager* manager_new(Controller* controller, const ConfigOptions* config, Simula
             warning("Status/progress bar will not be enabled since stderr is not a tty");
         }
     }
+
+    manager->checkFdUsage = true;
+    manager->checkMemUsage = true;
 
     return manager;
 }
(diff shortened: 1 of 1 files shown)


src/main/core/manager.c
```diff
@@ -646,6 +653,55 @@ static void _manager_heartbeat(Manager* manager, SimulationTime simClockNow) {
...
+        }
+
+        struct rlimit fd_lim;
+        if (getrlimit(RLIMIT_NOFILE, &fd_lim) != 0) {
+            warning("Unable to get fd limit: %s", g_strerror(errno));
+            manager->checkFdUsage = false;
+            return;
+        }
+
+        if (fd_count > fd_lim.rlim_cur * 90 / 100) {
```
sporksmith, line 678: I don't remember whether it's actually needed here, but parens would be helpful to the reader either way :). `fd_count > (fd_lim.rlim_cur * 90 / 100)`

src/main/core/manager.c
```diff
@@ -646,6 +653,55 @@ static void _manager_heartbeat(Manager* manager, SimulationTime simClockNow) {
...
+        long page_size = sysconf(_SC_PAGESIZE);
+        long avl_pages = sysconf(_SC_AVPHYS_PAGES);
+
+        if (page_size < 0 || avl_pages < 0) {
+            warning("Could not get memory usage information");
+            manager->checkMemUsage = false;
+            return;
+        }
+
+        long long avl_mem = page_size * avl_pages;
```
sporksmith, line 695: I believe this will multiply two `long`s, truncating to a `long`, before assigning to the `long long`. I think what you mean is: `long long avl_mem = (long long)page_size * avl_pages;`

Also, probably clearer to use an explicit number of bits, such as int64_t.

src/main/core/manager.c
```diff
@@ -646,6 +653,55 @@ static void _manager_heartbeat(Manager* manager, SimulationTime simClockNow) {
...
+
+        // if less than 500 MiB available
+        if (avl_mem < 500 * 1024 * 1024) {
```
sporksmith, line 698: Parens would be helpful here

src/main/core/manager.c
```diff
@@ -663,6 +719,13 @@ void manager_run(Manager* manager) {
         /* do some idle processing here if needed */
         _manager_heartbeat(manager, windowStart);
 
+        /* check resource usage every 30 seconds */
+        time_t current_time = time(NULL);
+        if (current_time != (time_t)-1 && current_time > manager->timeOfLastUsageCheck + 30) {
```
sporksmith, line 724: Parens would be helpful here

sporksmith (approved):

src/main/core/manager.c
stevenengler: Added.

src/main/core/manager.c
stevenengler: Added.

src/main/core/manager.c
stevenengler: Added.

src/main/core/manager.c
stevenengler: Good catch! And updated the type.

PR #1650, opened 2021-09-18
Improve the descriptor close process

### LINKED ISSUE #1596: Repeated FIN packets without acknowledgement
**Describe the issue**

TCP endpoints don't seem to ACK the FIN packet. In the following screenshot showing a small HTTP request from a client (11.0.0.1) to a server (11.0.0.4), the client sends the GET request, the server sends the response, the client sends a FIN, and then the server sends a FIN. But neither the server nor client acknowledge the FIN, so they continue sending FIN messages to each other once every second.

1629780829_grim

**To Reproduce**

```yaml
general:
  stop_time: 20
network:
  graph:
    type: gml
    inline: |
      graph [
        node [
          id 0
          host_bandwidth_down "140 Mbit"
          host_bandwidth_up "18 Mbit"
        ]
        edge [
          source 0
          target 0
          latency "100 ms"
          packet_loss 0.01
        ]
      ]
host_defaults:
  pcap_directory: 'pcaps'
hosts:
  server:
    network_node_id: 0
    quantity: 1
    processes:
    - path: /usr/bin/python3
      args: -m http.server 80
      start_time: 3
  client:
    network_node_id: 0
    quantity: 3
    processes:
    - path: /usr/bin/curl
      args: -s server
      start_time: 5
```

Using the shadow version from #1595.

stevenengler:
This PR makes some improvements to decouple the descriptor table and legacy descriptor objects. It should improve simulation accuracy and make it easier to write rust syscalls that need to work with legacy descriptors. It also closes all descriptors when the process ends, but doesn't disassociate them, so this appears to fix #1596.

### DIFF
--- a/src/main/host/descriptor/descriptor.c
+++ b/src/main/host/descriptor/descriptor.c
@@ -69,8 +69,12 @@ void descriptor_ref(gpointer data) {
     LegacyDescriptor* descriptor = data;
     MAGIC_ASSERT(descriptor);
 
+    // should not increment the strong count when there are only weak references left
+    utility_assert(descriptor->refCountStrong > 0);
+
     (descriptor->refCountStrong)++;
-    trace("Descriptor %i ref++ to %i", descriptor->handle, descriptor->refCountStrong);
+    trace("Descriptor %i strong_ref++ to %i (weak_ref=%i)", descriptor->handle,
+          descriptor->refCountStrong, descriptor->refCountWeak);
 }
 
 void descriptor_unref(gpointer data) {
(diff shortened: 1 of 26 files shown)


src/main/host/descriptor/descriptor.c
```diff
@@ -58,22 +69,48 @@ void descriptor_ref(gpointer data) {
...
+
+    (descriptor->refCountWeak)--;
+    trace("Descriptor %i weak_ref-- to %i", descriptor->handle, descriptor->refCountWeak);
 
-    utility_assert(descriptor->referenceCount >= 0);
-    if(descriptor->referenceCount == 0) {
-        gint handle = descriptor->handle;
+    utility_assert(descriptor->refCountWeak >= 0);
+    if (descriptor->refCountStrong == 0 && descriptor->refCountWeak == 0) {
         _descriptor_free(descriptor);
```
sporksmith, line 114: It looks like `descriptor_cleanup` is never called if the last reference is a weak ref?

sporksmith (approved):
Looks like a nice improvement!

src/main/host/descriptor/descriptor.c
stevenengler: `descriptor_cleanup()` should have been called when the last strong reference reached 0.

src/main/host/descriptor/descriptor.c
```diff
@@ -58,22 +69,48 @@ void descriptor_ref(gpointer data) {
...
 
 void descriptor_unref(gpointer data) {
     LegacyDescriptor* descriptor = data;
     MAGIC_ASSERT(descriptor);
 
-    (descriptor->referenceCount)--;
-    trace("Descriptor %i ref-- to %i", descriptor->handle,
-          descriptor->referenceCount);
+    (descriptor->refCountStrong)--;
+    trace("Descriptor %i ref-- to %i", descriptor->handle, descriptor->refCountStrong);
```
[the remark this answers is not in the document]
stevenengler: Added this logging.

src/main/host/descriptor/descriptor.c
```diff
@@ -70,49 +70,72 @@ void descriptor_ref(gpointer data) {
     MAGIC_ASSERT(descriptor);
 
     (descriptor->refCountStrong)++;
```
sporksmith, line 72: Is it an error to increment from 0, even if a non-zero weak count has kept the object alive? If so, maybe add a utility_assert?

src/main/host/descriptor/descriptor.c
sporksmith: Ah yeah, I'd misread that code. The new structure makes it clearer.

sporksmith (approved):

src/main/host/descriptor/descriptor.c
stevenengler: I think it would be best as an error. I added a `utility_assert()` here.

PR #1653, opened 2021-09-19
Fixed '--version' and '--help' parsing

### LINKED ISSUE #1464: Update version string that is printed on --version
- [x] When running `shadow --version` and `shadow --show-build-info` Shadow prints different version information, the latter option producing a more specific version that includes the git shorthash and the number of commits since the previous tag. It would be nice if `--version` could include the more specific version too.

- [x] Also, `--version` should end in one newline instead of 2 as it currently does, I think.

### LINKED ISSUE #1652: Shadow does not print usage or help info
Clean build at bbf7d014a5240d131a51a2b5b03426e76f5ee784

```
$ ./build/src/main/shadow --help
** Shadow did not complete successfully: Could not parse command-line options
** See the log for details

$ ./build/src/main/shadow --version
** Shadow did not complete successfully: Could not parse command-line options
** See the log for details
```

All of the tests pass successfully and passing a `yml` files does cause it to run correctly.

stevenengler:
Closes #1652 and closes #1464.

### DIFF
--- a/src/main/bindings/c/bindings-opaque.h
+++ b/src/main/bindings/c/bindings-opaque.h
@@ -39,9 +39,6 @@ typedef struct CliOptions CliOptions;
 
 typedef struct CompatDescriptor CompatDescriptor;
 
-// Options contained in a configuration file.
-typedef struct ConfigFileOptions ConfigFileOptions;
-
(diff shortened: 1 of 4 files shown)


robgjansen (approved):
Always nice to see more code removed than added ;)

PR #1654, opened 2021-09-20
Moved the process descriptor functions to rust

stevenengler:
Moved the functions for manipulating a process' descriptor table from C to rust.


src/main/host/process.rs
```diff
@@ -63,6 +64,54 @@ impl Process {
         unsafe { &*self.memory_manager_ptr() }
     }
 
+    /// Register a descriptor and return its fd handle.
+    pub fn register_descriptor(&mut self, desc: CompatDescriptor) -> u32 {
```
sporksmith, line 68: Maybe not in this MR, but it seems probably worth a type-safe wrapper for descriptor-numbers.

sporksmith (approved):

src/main/host/process.rs
stevenengler: I'll leave this for another PR, but maybe a [`RawFd`](https://doc.rust-lang.org/std/os/unix/io/type.RawFd.html), which is an `i32`, might be our best option.

PR #1660, opened 2021-09-21
Remove file from epoll interest list when closed

### LINKED ISSUE #1101: Epoll doesn't remove closed descriptors/files from the interest list
When a descriptor/file that an epoll object is watching closes, the epoll object doesn't remove it from its `watching` table. If the descriptor isn't removed using `EPOLL_CTL_DEL`, then the epoll object will never unref that descriptor/file and it will remain in the `watching` table until the epoll descriptor itself is closed.

> Will closing a file descriptor cause it to be removed from all epoll interest lists?
>
> Yes, but be aware of the following point. [...] A file descriptor is removed from an interest list only after all the file descriptors referring to the underlying open file description have been closed.

I think when `_epoll_descriptorStatusChanged()` runs, it should check if the descriptor/file has been closed (the `EpollWatch` flag `EWF_CLOSED` is set) and if so remove it from the `watching` table.

stevenengler:
Closes #1101.

I don't see a good way to add an integration test for this, but once we have this in rust, a unit test would be easier.

### DIFF
--- a/src/main/host/descriptor/epoll.c
+++ b/src/main/host/descriptor/epoll.c
@@ -725,6 +725,14 @@ static void _epoll_descriptorStatusChanged(Epoll* epoll, const EpollKey* key) {
                 /* this calls unref on the watch if its in the table */
                 g_hash_table_remove(epoll->ready, key);
             }
+
+            /* if it's closed, then remove it from the watching list */
+            if (watch->flags & EWF_CLOSED) {
+                /* unref gets called on the watch when it is removed from these tables */
+                g_hash_table_remove(epoll->watching, key);
+                /* we should have removed it from the ready list earlier */
+                utility_assert(!g_hash_table_contains(epoll->ready, key));
+            }
         }
     }
 


src/main/host/descriptor/epoll.c
```diff
@@ -725,6 +725,12 @@ static void _epoll_descriptorStatusChanged(Epoll* epoll, const EpollKey* key) {
                 /* this calls unref on the watch if its in the table */
                 g_hash_table_remove(epoll->ready, key);
             }
+
+            /* if it's closed, then remove it from the watching list */
+            if (watch->flags & EWF_CLOSED) {
+                /* unref gets called on the watch when it is removed from these tables */
+                g_hash_table_remove(epoll->watching, &key);
```
robgjansen, line 732: Do we ever want the watch to exist in the ready table but not the watching table? I'm not sure if that ever happens, but wondering if we need a
```
g_hash_table_remove(epoll->ready, key);
```
here to make sure that if we're not watching it, it will never be in the ready table either? (Or maybe an assert?)

robgjansen (approved):

src/main/host/descriptor/epoll.c
stevenengler: If it's closed, the `_epollwatch_isReady(watch)` check above should always return `false`, and then it will run `g_hash_table_remove(epoll->ready, key)`. So it should already be removed from the ready table. But I'll add an assert here to make sure it's removed from the ready table.

stevenengler (commented):
I noticed a bug after reviewing, so I fixed it as well:

Original:
```
g_hash_table_remove(epoll->watching, &key);
```

Fixed:
```
g_hash_table_remove(epoll->watching, key);
```

PR #1666, opened 2021-09-22
Added dup2() and dup3() support for rust descriptors

stevenengler:


src/main/host/syscall/unistd.rs
```diff
@@ -55,23 +55,86 @@ pub fn dup(ctx: &mut ThreadContext, args: &SysCallArgs) -> SyscallResult {
...
+    let old_fd = libc::c_int::from(args.get(0));
+    let new_fd = libc::c_int::from(args.get(1));
 
-    if let Some(flags) = flags {
-        new_desc.set_flags(flags);
+    // get the descriptor, or return early if it doesn't exist
+    let desc = match syscall::get_descriptor(ctx.process, old_fd)? {
+        CompatDescriptor::New(desc) => desc,
+        // we don't support dup2 for legacy descriptors
+        CompatDescriptor::Legacy(_) => return Err(nix::errno::Errno::ENOSYS.into()),
```
sporksmith, line 73: Log a warning?

src/test/dup/test_dup.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    tests
+}
+
+fn test_dup() -> Result<(), String> {
+    let (read_fd, write_fd) = nix::unistd::pipe().unwrap();
+
+    test_utils::run_and_close_fds(&[write_fd, read_fd], || {
+        let write_fd_dup =
+            test_utils::check_system_call!(|| { unsafe { libc::dup(write_fd) } }, &[])?;
```
sporksmith, line 51: Seems worth validating the end state a bit more. e.g. write to both the old and new descriptor and validate that data from both shows up on the read end? Same for other tests below.

sporksmith (approved):

src/main/host/syscall/unistd.rs
stevenengler: Good idea, added.

src/test/dup/test_dup.rs
stevenengler: The pipe tests already had this same test, but I moved it to the dup tests and run it for all dup(), dup2(), and dup3().

PR #1669, opened 2021-09-22
Handle file access and creation flags in fcntl()

stevenengler:


src/main/host/descriptor/mod.rs
```diff
@@ -32,6 +32,17 @@ bitflags::bitflags! {
     }
 }
 
+impl FileStatus {
+    pub fn as_o_flags(&self) -> libc::c_int {
+        self.bits() as i32
+    }
+
+    pub fn from_o_flags(flags: libc::c_int) -> (Self, libc::c_int) {
```
sporksmith, line 40: Could use a doc-comment explaining the second return value

src/main/host/descriptor/mod.rs
```diff
@@ -32,6 +32,17 @@ bitflags::bitflags! {
     }
 }
 
+impl FileStatus {
+    pub fn as_o_flags(&self) -> libc::c_int {
```
sporksmith, line 36: Maybe worth using `nix::fcntl::OFlag` instead of bare ints, throughout? Or too much conversion boilerplate when we do need to get to a bare int?

sporksmith (approved):

src/main/host/descriptor/mod.rs
stevenengler: Added more comments.

src/main/host/descriptor/mod.rs
stevenengler: Thanks, I didn't know about that. I added it in and it seems like an improvement.

stevenengler:
@sporksmith I'm re-marking you as reviewer to make sure I didn't mess up my bitwise operations in the conversion to the nix type.

src/main/host/descriptor/mod.rs
```diff
@@ -44,6 +58,53 @@ bitflags::bitflags! {
...
+
+    /// Returns a tuple of the `FileMode` and any remaining flags, or an empty `Err` if
+    /// the flags aren't valid (for example specifying both `O_RDWR` and `O_RDONLY`).
+    pub fn from_o_flags(flags: OFlag) -> Result<(Self, OFlag), ()> {
+        // O_RDONLY is 0, so some of the following checks are redundant
+        let has_rdwr = flags.contains(OFlag::O_RDWR);
+        let has_rdonly = flags.contains(OFlag::O_RDONLY);
+        let has_wronly = flags.contains(OFlag::O_WRONLY);
+
+        if has_rdwr {
```
sporksmith, line 86: Alternatively:

```
match (has_rdwr, has_rdonly, has_wronly) {
  (true, false, false) => Ok(...),
  (false, true, false) => Ok(...),
  (false, fasle, true) => Ok(...),
  _ => Err(())
}
```

sporksmith (approved):

src/main/host/descriptor/mod.rs
stevenengler: I like this format, but I found an error in my existing code. Since `O_RDONLY` is 0, `flags.contains(OFlag::O_RDONLY)` is always true, which is not what we want. So I fixed the logic in this function, and also added unit tests to make sure it's handled properly. For example `O_RDONLY | O_WRONLY == FileMode::WRITE` since `O_RDONLY` is 0.

sporksmith: Whoops, good catch!

PR #1673, opened 2021-09-23
Improved error message when assigning IP more than once

stevenengler:
I think this is an improvement.

### DIFF
--- a/src/main/routing/network_graph.rs
+++ b/src/main/routing/network_graph.rs
@@ -356,6 +356,8 @@ impl std::convert::From<&ShadowEdge> for PathProperties {
     }
 }
 
+pub struct IpPreviouslyAssignedError;
+
 /// Tool for assigning IP addresses to graph nodes.
 #[derive(Debug)]
 pub struct IpAssignment<T: Copy + Eq + Hash + std::fmt::Display> {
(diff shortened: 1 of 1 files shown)


src/main/routing/network_graph.rs
```diff
@@ -742,8 +744,13 @@ mod export {
 
         match ip_assignment.assign_ip(node_id, ip_addr) {
             Ok(()) => return 0,
-            Err(e) => {
-                error!("{}", e);
+            Err(IpPreviouslyAssignedError) => {
+                error!(
+                    "IP {} was assigned to multiple hosts, one on node {} and another on node {}",
```
sporksmith, line 749: Can we reasonably look up the host names here? I think it'd be a bit friendlier to give the host names rather than the node names; I think we only get such collisions when the ip address was specified at the host level, right?

stevenengler: We don't store the hostnames in any of the routing data, only in the host objects themselves which we're in the middle of building/constructing during this code, so there's nowhere to look up the conflicting hostname. The log should show the error `"Could not register host %s"` immediately after, so the user should see what at least one of the hostnames are.

sporksmith (approved):

PR #1687, opened 2021-09-28
Filter flags in fcntl() F_SETFL

stevenengler:


src/main/host/syscall/fcntl.rs
```diff
@@ -51,14 +51,29 @@ fn fcntl(ctx: &mut ThreadContext, args: &SysCallArgs) -> SyscallResult {
                     | OFlag::O_TMPFILE
                     | OFlag::O_TRUNC,
             );
+
+            let mut file = desc.get_file().borrow_mut();
+            let old_flags = file.get_status().as_o_flags();
+
+            // fcntl(2): "On Linux, this command can change only the O_APPEND, O_ASYNC, O_DIRECT,
```
sporksmith, line 58: Was going to ask if we ought to return an error, but under `BUGS` the man page has `It  is  not  possible  to  use  F_SETFL  to change the state of the O_DSYNC and O_SYNC flags.  Attempts to change the state of these flags are silently ignored.`. Probably worth a comment here that we're intentionally reproducing the 'silently ignore' behavior of the kernel

sporksmith (approved):

src/main/host/syscall/fcntl.rs
stevenengler: The docs don't really say, but I think in general we should ignore any flags that we can't change since those flags may have come from a previous `fcntl(fd, F_GETFL)`. For example `int flags = fcntl(fd, F_GETFL); fcntl(fd, F_SETFL, flags);` should always be valid.

I added some comments.

PR #1689, opened 2021-09-29
Some refactoring in the shim

robgjansen:
This is mostly refactoring, i.e., moving code around to different files and renaming things.

The general goal is to stop preloading the shim, and instead move the code that we use to preload functions from the managed processes into libraries that we inject using `LD_PRELOAD`. Those preload libraries are linked to the shim and will cause the shim to be loaded correctly for each managed process. I added a dummy preload library that links to the shim to make sure that even if we are not preloading libc or openssl, we will still inject the shim into the managed process.

During the created of the new libc preload library, I expanded the list of syscall wrappers that we intercept so that we can take advantage of the slightly faster preload method of interception rather than the slightly slower seccomp method.

Apologize in advance for so much renaming :/


src/lib/dummy_preload/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+# Normally, the shim is linked to the openssl or libc preload libs and the shim
+# will be injected when those libs are preloaded. However, if the preloading of
+# both of those libs is disabled, then we inject the dummy lib instead (which is
+# also linked to the shim) in order to ensure that the shim is still loaded into
+# the managed process space as required for Shadow to work properly.
+add_library(shadow_dummy SHARED dummy.c)
```
robgjansen, line 6: I'm not sure that we want the dummy preload lib. We could instead just always require that the libc preload lib is always preloaded.

src/main/core/manager.c
```diff
@@ -28,9 +28,14 @@
...
 
-#define PRELOAD_SHIM_LIB_STR "libshadow-shim.so"
+#define PRELOAD_DUMMY_LIB_STR "libshadow_dummy.so"
+#define PRELOAD_LIBC_LIB_STR "libshadow_libc.so"
 #define PRELOAD_OPENSSL_RNG_LIB_STR "libshadow_openssl_rng.so"
 
+// Allow turning off libc preloading at run-time.
+static bool _use_libc_preload = true;
+ADD_CONFIG_HANDLER(config_getUseLibcPreload, _use_libc_preload)
+
```
robgjansen, line 38: I don't think that disabling libc preload is actually an important use-case, and so maybe we don't want to support that option?

src/lib/dummy_preload/CMakeLists.txt
stevenengler: One reason we might want to disable the libc preload could be if we think the libc call behaves differently to the direct syscall and we want to test shadow with the real libc call. But maybe the solution to this would be to just comment out everything in `syscall_wrappers.c` instead.

src/main/core/manager.c
```diff
@@ -473,23 +474,28 @@ static gchar** _manager_generateEnvv(Manager* manager, InterposeMethod interpose
...
     } else if (manager->preloadOpensslRngPath != NULL) {
         debug("adding Shadow preload lib path %s", manager->preloadOpensslRngPath);
         g_ptr_array_add(ldPreloadArray, g_strdup(manager->preloadOpensslRngPath));
+        need_dummy_preload = false;
+    }
+
+    if(need_dummy_preload) {
+        debug("adding Shadow preload lib path %s", manager->preloadDummyPath);
+        g_ptr_array_add(ldPreloadArray, g_strdup(manager->preloadDummyPath));
     }
```
stevenengler: `preloadDummyPath` may be NULL here.

src/main/core/manager.c
```diff
@@ -212,31 +213,31 @@ Manager* manager_new(Controller* controller, const ConfigOptions* config, Simula
...
-             manager->preloadShimPath);
+    manager->preloadDummyPath = _manager_scanRPathForLib(PRELOAD_DUMMY_LIB_STR);
+    if (manager->preloadDummyPath != NULL) {
+        info("found preload library %s at path %s", PRELOAD_DUMMY_LIB_STR,
+             manager->preloadDummyPath);
     } else {
-        // The shim is required, so panic if we can't find it.
-        utility_panic("could not find required preload library %s in rpath", PRELOAD_SHIM_LIB_STR);
+        // The dummy preload is only used if we have no other preloads configured.
+        warning("could not find preload library %s in rpath", PRELOAD_DUMMY_LIB_STR);
```
stevenengler: Why did this change to a warning? It seems like this should be required in case the libc or openssl preloads aren't found or enabled.

src/main/core/manager.c
```diff
@@ -473,23 +474,28 @@ static gchar** _manager_generateEnvv(Manager* manager, InterposeMethod interpose
      *   - preload values from LD_PRELOAD entries in the environment process option */
     GPtrArray* ldPreloadArray = g_ptr_array_new();
 
-    if (manager->preloadShimPath != NULL) {
-        debug("adding Shadow preload shim path %s", manager->preloadShimPath);
-        g_ptr_array_add(ldPreloadArray, g_strdup(manager->preloadShimPath));
-    }
+    // Track if we need to preload the dummy lib.
+    bool need_dummy_preload = true;
```
stevenengler, line 477: Is there any harm in always preloading the dummy lib?

robgjansen:
Capturing out-of-band discussion here:

Do we want to support disabling libc preload? (Most but not all tests pass without it.)
  - -> We like the option to disable libc preloading, so let's keep it
  - -> Keep it as 'experimental' for now
  - -> Once we implement the remaining issues (vdso rewriting, in-shadow dns server), we can move as an officially supported option

src/lib/dummy_preload/CMakeLists.txt
robgjansen: We decided that it is useful to be able to disable the libc preload lib, but we will keep it as experimental for now. So we'll need a dummy lib to inject the shim in case no other preload is used.

src/main/core/manager.c
robgjansen: We decided that it is useful to be able to disable the libc preload lib, but we will keep it as experimental for now.

src/main/core/manager.c
```diff
@@ -214,6 +221,15 @@ Manager* manager_new(Controller* controller, const ConfigOptions* config, Simula
...
     }
 
+    manager->preloadSyscallsPath = _manager_scanRPathForLib(PRELOAD_SYSCALLS_LIB_STR);
+    if (manager->preloadSyscallsPath != NULL) {
+        info("found optional preload library %s at path %s", PRELOAD_SYSCALLS_LIB_STR,
+             manager->preloadSyscallsPath);
+    } else {
+        // The syscalls preload is optional and may not be used by and managed processes.
+        warning("could not find optional preload library %s in rpath", PRELOAD_SYSCALLS_LIB_STR);
+    }
```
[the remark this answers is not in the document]
robgjansen: I refactored the path scanning logic in 8cf7d9f and always panic when required preload libs cannot be found.

src/main/core/manager.c
robgjansen: We decided to always preload the dummy lib, and I changed to doing that now in 8cf7d9f.

src/main/core/manager.c
robgjansen: Sorry, that was copied and didn't get updated properly on my rebase. In any case, the path checking logic has been updated now to always panic if required libs are not found.

src/main/core/manager.c
robgjansen: Fixed in my refactoring in 8cf7d9f.

robgjansen:
It's ready for another look now @stevenengler.

src/lib/libc_preload/gen_syscalls_wrappers_c.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # ignore the x32-specific abi, since shadow only supports x86_64
+    if 'x32' in abi:
+        continue
+
+    syscalls[str(name)] = [num, str(entry)]
+
+header = \
+'''// This file is generated with the 'generate_syscalls_c.py' script and in
+// general SHOULD NOT be edited manually.
```
stevenengler, line 36: This file was renamed to `gen_syscalls_wrappers_c.py`.

stevenengler (approved):

src/lib/injector_preload/injector.c
```diff
@@ -0,0 +1,27 @@
...
+
+/// The purpose of the injector library is to inject the shim into the process
+/// space of each managed process, since the shim is needed for interaction with
+/// Shadow. (Technically, the shim will already be injected if there are other
+/// preloaded libraries that link to it, but this library provides a minimally
+/// invasive way to inject the shim without preloading other symbols.)
+///
+/// One goal is to ensure the shim is injected, so we link to it and call one
+/// shim function in a constructor. The other goal is to be minimally invasive,
+/// so let's not define any other symbols here.
```
stevenengler, line 19: Might be useful to add a comment here that we preload the injector lib rather than the preload the shim directly so that we inject the shim without intercepting any functions from the managed process.

src/lib/libc_preload/gen_syscalls_wrappers_c.py
robgjansen: Good catch! Fixed.

src/lib/injector_preload/injector.c
robgjansen: I updated and expanded the comment.

PR #1698, opened 2021-10-04
Fix some clippy lints

stevenengler:
Fixes for some of the smaller clippy lints. There are still about 60 warnings left, but this is a start.


src/main/host/memory_manager/mod.rs
```diff
@@ -163,7 +163,7 @@ where
 
     fn deref(&self) -> &Self::Target {
         match &self.0 {
-            CopiedOrMapped::Copied(v) => &v,
+            CopiedOrMapped::Copied(v) => v,
```
sporksmith, line 166: I'm confused how this works, since `v` is a `Vec`, not a `&Vec`. If it compiles I guess it's right, though...?

src/main/host/syscall/fcntl.rs
```diff
@@ -127,10 +127,10 @@ fn fcntl(ctx: &mut ThreadContext, args: &SysCallArgs) -> SyscallResult {
             if let PosixFile::Pipe(pipe) = desc.get_file() {
                 SysCallReg::from(i32::try_from(pipe.borrow().max_size()).unwrap())
             } else {
-                Err(Errno::EINVAL)?
+                return Err(Errno::EINVAL.into());
```
sporksmith, line 130: I'm not sure this transformation is a win, but I guess I'm willing to along with clippy :)

sporksmith (approved):

src/main/host/memory_manager/mod.rs
stevenengler: I think that's because the reference in `&self.0` causes the pattern to be in ref binding mode, so `v` is bound to a reference to the inner type `&Vec` instead of `Vec`.

src/main/host/syscall/fcntl.rs
stevenengler: I wasn't sure originally, but I think I agree with clippy now. The explicit `return` shows that we intend to always return an error, and imo makes it easier to skip over that branch when looking through the code.

Issue #1699, opened 2021-10-05
Investigate performance regression


robgjansen:
We believe we have discovered a performance regression between shadow v2.0.0-pre.4 and 154a11d8bc2a84eb09f884cf27d5186b0890ec2b.

- In v2.0.0-pre.4: a 10% Tor network took about 10 real hours to complete 1 simulated hour. We believe CPU utilization to be high (>90%) across all cores during the simulation.
- In 154a11d8bc2a84eb09f884cf27d5186b0890ec2b, using the same machine, a 10% Tor network took about 25 real hours to complete 1 simulated hour. CPU utilization hovers between 40-50%.

We should investigate.

robgjansen:
One hypothesis is that the difference is due to a change in the behavior of the Shadow event scheduler and how we are calculating the run-ahead time (the time that each worker thread is allowed to execute into the future) for each scheduling round.

We use a conservative scheduling algorithm, so we must always guarantee that time only moves forward. The run-ahead time represents the earliest time from "now" at which any host could cause a new event to occur at any other host. Generally, hosts are only sending "packet" events to other hosts, and so we use network latency as the run-ahead time.

More precisely, this is how we compute the run-ahead time:
- In v2.0.0-pre.4, we compute it as the min over the inter-host latencies _only for those hosts that are communicating with one another_. I.e., we start at a large value, and then whenever a packet is sent between a pair of hosts, we lower the run-ahead time if the latency between those hosts is smaller. (This was the intended behavior, though see #1611)
- In 154a11d8bc2a84eb09f884cf27d5186b0890ec2b, we have a new network graph implementation in which we calculate at startup the min run-ahead between _all pairs of hosts in the config file_ whether or not they ever communicate.
 
The run-ahead calculated in the v2.0.0-pre.4 version of the code is an upper bound on what will be calculated in the 154a11d8bc2a84eb09f884cf27d5186b0890ec2b version. It's possible that, due to a smaller run-ahead time, Shadow is not able to parallelize work as effectively, leading to lower CPU utilization and longer real times to execute simulations. It's also possible that the run-ahead time is smaller now primarily due to the fixes in #1611 and not due to computing the run-ahead at startup. We have an experiment planned to investigate.

robgjansen:
[...]

PR #1700, opened 2021-10-05
Experimental option for dynamically updating the minimum time jump

stevenengler:
Added an experimental option to dynamically update the minimum time jump as packets are sent on new edges between graph nodes. This should model the behaviour of Shadow before the graph code was moved to rust (one difference is that the initial time jump is now the min network latency rather than 10 ms). If the first packet between two nodes has a latency less than the current time jump, that packet will arrive late.

### DIFF
--- a/src/main/core/controller.c
+++ b/src/main/core/controller.c
@@ -93,6 +93,9 @@ Controller* controller_new(const ConfigOptions* config) {
 
     controller->minJumpTimeConfig = config_getRunahead(config);
 
+    controller->endTime = config_getStopTime(config);
+    controller->bootstrapEndTime = config_getBootstrapEndTime(config);
+
     /* these are only avail in glib >= 2.30
      * setup signal handlers for gracefully handling shutdowns */
     //  TODO
@@ -187,14 +190,9 @@ static gboolean _controller_loadNetworkGraph(Controller* controller) {
 static void _controller_initializeTimeWindows(Controller* controller) {
     MAGIC_ASSERT(controller);
 
-    /* set simulation end time */
-    controller->endTime = config_getStopTime(controller->config);
-
     controller->executeWindowStart = 0;
     SimulationTime jump = _controller_getMinTimeJump(controller);
     controller->executeWindowEnd = jump;
-
-    controller->bootstrapEndTime = config_getBootstrapEndTime(controller->config);
 }
 
 static void _controller_registerArgCallback(const char* arg, void* _argArray) {
--- a/src/main/core/controller.c
+++ b/src/main/core/controller.c
@@ -48,6 +48,10 @@ struct _Controller {
     /* minimum allowed time jump when sending events between nodes */
     SimulationTime minJumpTimeConfig;
     SimulationTime minJumpTime;
+    bool isJumpTimeDynamic;
+
+    /* the next min jump time is updated by workers, so needs to be locked */
+    GRWLock nextMinJumpTimeLock;
     SimulationTime nextMinJumpTime;
 
     /* start of current window of execution */
@@ -92,9 +96,12 @@ Controller* controller_new(const ConfigOptions* config) {
     controller->random = random_new(config_getSeed(config));
 
     controller->minJumpTimeConfig = config_getRunahead(config);
+    controller->isJumpTimeDynamic = config_getUseDynamicMinJumpTime(config);
 
     controller->endTime = config_getStopTime(config);
     controller->bootstrapEndTime = config_getBootstrapEndTime(config);
+
+    g_rw_lock_init(&controller->nextMinJumpTimeLock);
 
     /* these are only avail in glib >= 2.30
      * setup signal handlers for gracefully handling shutdowns */
@@ -133,6 +140,8 @@ void controller_free(Controller* controller) {
         controller->random = NULL;
     }
 
+    g_rw_lock_clear(&controller->nextMinJumpTimeLock);
+
     MAGIC_CLEAR(controller);
     g_free(controller);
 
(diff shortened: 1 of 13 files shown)


robgjansen:
See also #1699

src/main/core/controller.c
```diff
@@ -538,13 +530,21 @@ gboolean controller_managerFinishedCurrentRound(Controller* controller,
...
+gdouble controller_getLatency(Controller* controller, Address* srcAddress, Address* dstAddress,
+                              bool updateMinTimeJump) {
     MAGIC_ASSERT(controller);
+
+    uint64_t latencyNs = routinginfo_getLatencyNs(controller->routingInfo, controller->ipAssignment,
+                                                  htonl(address_toHostIP(srcAddress)),
+                                                  htonl(address_toHostIP(dstAddress)));
+
+    if (controller->dynamicJumpTime && updateMinTimeJump) {
+        _controller_updateMinTimeJumpNs(controller, latencyNs);
```
robgjansen, line 542: This can be called by multiple worker threads at once I think, so don't we need some sort of lock inside the `_controller_updateMinTimeJumpNs` function so two threads don't simultaneously write the new jump time?

src/main/core/worker.c
```diff
@@ -551,7 +551,7 @@ void worker_sendPacket(Host* srcHost, Packet* packet) {
      * control has problems responding to packet loss */
     if (bootstrapping || chance <= reliability || packet_getPayloadLength(packet) == 0) {
         /* the sender's packet will make it through, find latency */
-        gfloat latency = worker_getLatencyForAddresses(srcAddress, dstAddress);
+        gfloat latency = worker_getLatencyForAddresses(srcAddress, dstAddress, true);
```
robgjansen, line 554: 1. Is it worth adding the extra `bool updateMinTimeJump` argument to all of the `getLatency*` functions, rather than just calling one extra new function call here to explicitly update the min jump time? (I think this is the only place we pass `true`.)
2. If we want to avoid too much locking in the controller when trying to write the new jump time, each worker could keep it's own local `minJumpTime` state and only ask the controller to update when it finds a smaller local jump time.

robgjansen (changes requested):

robgjansen:
Also, a better term to use for the "jump time" is `runahead` since that's the term generally used to describe this in the "parallel and distributed simulation" community. You might consider renaming the `jump` variables to `runahead`.

src/main/core/controller.c
```diff
@@ -472,13 +461,16 @@ gint controller_run(Controller* controller) {
         return 1;
     }
 
-    controller_updateMinTimeJumpNs(
-        controller, routinginfo_smallestLatencyNs(controller->routingInfo));
+    controller->minJumpTime =
+        routinginfo_smallestLatencyNs(controller->routingInfo) * SIMTIME_ONE_NANOSECOND;
+    info("using an initial minimum time jump of %" G_GUINT64_FORMAT " ns", controller->minJumpTime);
```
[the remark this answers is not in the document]
stevenengler: > Don't we want to initialize `minJumpTime` to something else if `isJumpTimeDynamic` is true? Otherwise we would use the smaller of the dynamic jump time and the `routinginfo_smallestLatencyNs` jump time, which is not what we want.

When calculating `nextMinJumpTime`, it doesn't look at `minJumpTime` at all. It uses the minimum of `routinginfo_smallestLatencyNs()` and `nextMinJumpTime`, not `minJumpTime`, so the initial value of `minJumpTime` doesn't matter.

> I was going to suggest that we want to init to 0 if if `isJumpTimeDynamic` is true. But actually, I'm not sure what happens in that case for the first scheduling round. Maybe we instead want to initialize it to the max latency between any 2 pairs of nodes in the graph?

Setting to 0 will break the scheduler, it needs to be non-zero. Initializing it to the max latency would work, but would provide weird behaviour if the graph has a latency outlier. For example if one graph node has an edge with a latency of 3 seconds, I don't think we'd want to start the simulation with a scheduling interval of 3 seconds. Starting with the minimum latency makes sure that we're using reasonable scheduling intervals until the first packet is sent and `minJumpTime` is set.

src/main/core/controller.c
stevenengler: Good catch, we didn't have any locking before, but it worked out since the controller only has one manager which did use its own locking.

src/main/core/worker.c
stevenengler: >     1. Is it worth adding the extra `bool updateMinTimeJump` argument to all of the `getLatency*` functions, rather than just calling one extra new function call here to explicitly update the min jump time? (I think this is the only place we pass `true`.)

My reason against doing it this way is that then we need another group of functions `worker_updateMinJumpTime()` -> `manager_updateMinJumpTime()` -> `controller_updateMinJumpTime()`, and we have to change the return type of all of our latency functions from `gdouble` to `SimulationTime`, but I added these now.

>     2. If we want to avoid too much locking in the controller when trying to write the new jump time, each worker could keep it's own local `minJumpTime` state and only ask the controller to update when it finds a smaller local jump time.

Added.

src/main/core/controller.c
```diff
@@ -49,6 +49,7 @@ struct _Controller {
     SimulationTime minJumpTimeConfig;
     SimulationTime minJumpTime;
     SimulationTime nextMinJumpTime;
+    bool dynamicJumpTime;
```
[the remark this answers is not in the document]
stevenengler: Changed.

robgjansen (approved):
Thanks for the updates!

robgjansen (approved):
I think the new locking process works!

PR #1704, opened 2021-10-06
Handle exit while scheduled

sporksmith:
Add regression test and bug fix for a bug encountered while running arti:

* A syscallcondition is triggered, scheduling a thread to run
* The process exits, cleaning up the thread, canceling the condition's listeners, etc.
* The scheduled task runs, but since the syscallcondition has been cancelled, it no longer has references to the process, thread, etc.


src/test/regression/test_exit_after_signal_sched.rs
```diff
@@ -0,0 +1,46 @@
...
+    let child_finished = std::sync::Arc::new(std::sync::atomic::AtomicBool::new(false));
+    {
+        let child_finished = child_finished.clone();
+        std::thread::spawn(move || {
+            // This should initially block.
+            assert_eq!(1, nix::unistd::write(write_fd, &[0]).unwrap());
+            // Signal to the parent that the `write` returned.
+            child_finished.store(true, std::sync::atomic::Ordering::SeqCst);
+        })
+    };
```
stevenengler, line 33: The `child_finished` doesn't seem to be read anywhere?

sporksmith: Fixed; thanks!

PR #1710, opened 2021-10-08
Fix hostname to addr lookups

robgjansen:
Fix `getaddrinfo` so that looking up hostnames through Shadow's custom `SYS_shadow_hostname_to_addr_ipv4` syscall works again in preload mode and enables us to avoid scans over the `/etc/hosts` file.

Also enabled the `getaddrinfo` tests so they run in Shadow, and did some refactoring.

I believe this is a second performance fix for #1699 (in addition to setting a min runahead of 1 millisecond), but need an experiment to verify.

### DIFF
--- a/src/main/host/syscall/shadow.c
+++ b/src/main/host/syscall/shadow.c
@@ -6,8 +6,10 @@
 #include "main/host/syscall/shadow.h"
 
 #include <errno.h>
+#include <netdb.h>
 #include <stdarg.h>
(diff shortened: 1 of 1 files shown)


robgjansen:
This is definitely a great fix to have merged, but an experiment indicates that it was not affecting performance in a 10% Tor network nearly as much as I thought it was.

run_time_name_to_addr

PR #1711, opened 2021-10-10
Set default min runahead to 1ms

robgjansen:
Larger values for the `--runahead` option improves our ability to parallelize work across workers, leading to faster run-times. 1 millisecond was the lowest possible minimum in Shadow 1.x, and I think a reasonable default since most latencies for our target networks are in terms of milliseconds.

refs #1699

### DIFF
--- a/docs/shadow_config_spec.md
+++ b/docs/shadow_config_spec.md
@@ -337,7 +337,7 @@ Show the simulation progress at the bottom of the terminal.
 
 #### `experimental.runahead`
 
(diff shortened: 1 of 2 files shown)


robgjansen:
Effect on Shadow performance in a 10% Tor network when adjusting `--runahead`:
run_time_runahead

We might also consider increasing to larger than 1ms - I do not notice Tor performance changes that could be attributed to changing the runahead to 1ms or 2ms.

PR #1734, opened 2021-10-19
Update fedora 34 to 35

stevenengler:
I don't think Fedora 35 is technically released yet, but should be sometime this week and the docker image is already available.

Also fixes some gcc 11 warnings in tests.


stevenengler:
Added in #1896.

PR #1773, opened 2021-11-10
Fix eventfd syscall wrapper

stevenengler:
The libc preload library was creating a C function `eventfd()` which called the `SYS_eventfd` syscall, but `eventfd()` should be calling `SYS_eventfd2`.

### DIFF
--- a/src/lib/libc_preload/syscall_wrappers.c
+++ b/src/lib/libc_preload/syscall_wrappers.c
@@ -138,11 +138,9 @@ INTERPOSE(epoll_pwait2);
 INTERPOSE(epoll_wait);
 #endif
 #ifdef SYS_eventfd // kernel entry: num=284 func=sys_eventfd
-INTERPOSE(eventfd);
+INTERPOSE_REMAP(eventfd, eventfd2);
 #endif
-#ifdef SYS_eventfd2 // kernel entry: num=290 func=sys_eventfd2
-INTERPOSE(eventfd2);
-#endif
+// Skipping SYS_eventfd2
 #ifdef SYS_execve // kernel entry: num=59 func=sys_execve
 INTERPOSE(execve);
 #endif
--- a/src/lib/libc_preload/gen_syscall_wrappers_c.py
+++ b/src/lib/libc_preload/gen_syscall_wrappers_c.py
@@ -11,6 +11,14 @@ than using seccomp to interpose, so we should use preloading whenever possible.
 # See also https://github.com/torvalds/linux/tree/master/arch/x86/entry/syscalls
 syscall_tbl = 'https://raw.githubusercontent.com/torvalds/linux/master/arch/x86/entry/syscalls/syscall_64.tbl'
 
+# libc wrappers which use a different syscall
+remap = {}
+remap['eventfd'] = 'eventfd2' # libc eventfd() calls SYS_eventfd2
+
+# syscalls we should not generate C wrappers for
+skip = []
+skip.append('eventfd2') # libc doesn't have an eventfd2() wrapper
+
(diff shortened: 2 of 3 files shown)


src/lib/libc_preload/syscall_wrappers.c
```diff
@@ -137,12 +137,8 @@ INTERPOSE(epoll_pwait2);
 #ifdef SYS_epoll_wait // kernel entry: num=232 func=sys_epoll_wait
 INTERPOSE(epoll_wait);
 #endif
-#ifdef SYS_eventfd // kernel entry: num=284 func=sys_eventfd
-INTERPOSE(eventfd);
-#endif
-#ifdef SYS_eventfd2 // kernel entry: num=290 func=sys_eventfd2
-INTERPOSE(eventfd2);
-#endif
+// Skipping SYS_eventfd
```
robgjansen, line 140: Should we instead do:

```C
#ifdef SYS_eventfd // kernel entry: num=284 func=sys_eventfd
INTERPOSE_REMAP(eventfd, eventfd2);
#endif
```

That way we still get the benefit of skipping the seccomp interception overhead whenever eventfd is called.

src/lib/libc_preload/syscall_wrappers.c
```diff
@@ -137,12 +137,8 @@ INTERPOSE(epoll_pwait2);
...
+// Skipping SYS_eventfd2
```
robgjansen, line 141: Has this function ever been defined in libc? Leaving it in wouldn't hurt much unless someone defines their own eventfd2 function and then we would end intercepting it.

Ultimately, I think it comes down to: Do we really want to maintain this special case? Maybe we don't want to maintain the `gen_syscall_wrappers_c.py` script at all and we should just delete it. Then we could just maintain the raw `syscall_wrappers.c` file directly without trying to make sure the script will regenerate it correctly.

(The reason I committed the script was that I thought it would be useful for us later to detect when new syscalls are added, but I don't think we necessarily need to use it to keep `syscall_wrappers.c` exactly up to date.)

robgjansen (changes requested):
We may want to discuss whether or not we want to maintain `gen_syscall_wrappers_c.py`.

src/lib/libc_preload/syscall_wrappers.c
stevenengler: Changed this to remap it instead.

src/lib/libc_preload/syscall_wrappers.c
stevenengler: > Has this function ever been defined in libc? Leaving it in wouldn't hurt much unless someone defines their own eventfd2 function and then we would end intercepting it.

Glibc has never had an `eventfd2` function. I think it would be better to avoid generating wrappers that don't exist in glibc so that we don't accidentally cause unexpected behaviour in the plugin, but maybe we don't need to worry about it seriously.

> Ultimately, I think it comes down to: Do we really want to maintain this special case? Maybe we don't want to maintain the `gen_syscall_wrappers_c.py` script at all and we should just delete it. Then we could just maintain the raw `syscall_wrappers.c` file directly without trying to make sure the script will regenerate it correctly.
> 
> (The reason I committed the script was that I thought it would be useful for us later to detect when new syscalls are added, but I don't think we necessarily need to use it to keep `syscall_wrappers.c` exactly up to date.)

I don't think it's that bad to have special cases. We can remove the `gen_syscall_wrappers_c.py` script if you want, I don't have an opinion either way.

robgjansen: OK, let's keep it for now. Thanks for the updates!

robgjansen (approved):

PR #1774, opened 2021-11-10
Add ioctl() to Rust descriptors, and convert eventfd from C to Rust

stevenengler:


src/main/host/descriptor/eventfd.rs
```diff
@@ -0,0 +1,224 @@
...
+        self.copy_state(
+            FileState::CLOSED | FileState::ACTIVE | FileState::READABLE | FileState::WRITABLE,
+            FileState::CLOSED,
+            event_queue,
+        );
+
+        Ok(0.into())
+    }
+
+    pub fn read<W>(
```
robgjansen, line 55: Nit: Does rust let us use more than one letter for the type name? Would something like `WriteSeek` or `WrSk` be more descriptive?

src/main/host/descriptor/eventfd.rs
```diff
@@ -0,0 +1,224 @@
...
+            bytes.write_all(&to_write)?;
+            self.counter = 0;
+        }
+
+        self.update_state(event_queue);
+
+        return Ok(NUM_BYTES.into());
+    }
+
+    pub fn write<R>(
```
robgjansen, line 103: Nit: Does rust let us use more than one letter for the type name? Would something like `ReadSeek` or `RdSk` be more descriptive?

src/main/host/descriptor/eventfd.rs
```diff
@@ -0,0 +1,224 @@
...
+            return Err(Errno::EINVAL.into());
+        }
+
+        if self.counter == 0 {
+            log::trace!("Eventfd counter is 0 and cannot be read right now");
+            return Err(Errno::EWOULDBLOCK.into());
+        }
+
+        // behavior defined in `man 2 eventfd`
+        if self.semaphore_mode {
```
robgjansen, line 88: I was expecting an enum or something because of the name, but since it's a boolean can we change the name so that is clear? `is_semphore_mode`?

src/main/host/descriptor/eventfd.rs
```diff
@@ -0,0 +1,224 @@
...
+        const MAX_ALLOWED: u64 = u64::MAX - 1;
+        if value > MAX_ALLOWED - self.counter {
+            log::trace!("The write value does not currently fit into the counter");
+            return Err(Errno::EWOULDBLOCK.into());
+        }
+
+        self.counter += value;
+        self.update_state(event_queue);
+
+        return Ok(NUM_BYTES.into());
```
robgjansen, line 149: Do we want the `return` statements (punctuated with `;`) throughout the read and write functions, or is that considered non-standard rust style?

src/main/host/syscall/ioctl.c
```diff
@@ -227,41 +226,6 @@ static int _syscallhandler_ioctlUDPHelper(SysCallHandler* sys, UDP* udp, int fd,
     return result;
 }
 
-static int _syscallhandler_ioctlEventDHelper(SysCallHandler* sys, EventD* eventd, int fd,
-                                             unsigned long request, PluginPtr argPtr) {
-    int result = -EINVAL;
-
-    switch (request) {
-        case FIONBIO: {
```
robgjansen, line 235: It doesn't look like the `FIONBIO` case was copied over to the new `eventfd.ioctl` function. Is that a mistake?

robgjansen (approved):
My comments on the rust code are only surface-level. If there are tricky parts here or you want better feedback, consider asking Jim to glance at it too.

src/main/host/descriptor/eventfd.rs
stevenengler: Rust does let you use more than one letter for the generic type name, but I don't think I've ever seen anyone do that. One advantage is that it makes it clear that it's a generic and not a real type name. In this case, the `W` name is only used in the function parameters and not the function body, so I don't think we really gain any readability by making it longer.

src/main/host/descriptor/eventfd.rs
stevenengler: Changed.

src/main/host/descriptor/eventfd.rs
stevenengler: I'm not sure what you mean here. Since this is the last return statement of the function, I did change the `return Ok(NUM_BYTES.into());` here into just `Ok(NUM_BYTES.into()`.

src/main/host/syscall/ioctl.c
stevenengler: I put this in the syscall handler (`syscall/ioctl.rs`) directly since it should apply to all file types, but maybe for consistency it would be better to copy it into each individual file (`descriptor/pipe.rs` and `descriptor/eventfd.rs`).

robgjansen: Oh, I think I just missed it then. Sorry!

I think we would only want to copy to individual files if the implementation changes depending on the type?

src/main/host/syscall_types.rs
```diff
@@ -90,6 +90,18 @@ impl From<SysCallReg> for u64 {
...
 
+impl From<u32> for SysCallReg {
+    fn from(v: u32) -> Self {
+        Self { as_u64: v as u64 }
+    }
+}
+
+impl From<SysCallReg> for u32 {
+    fn from(v: SysCallReg) -> u32 {
+        (unsafe { v.as_u64 }) as u32
```
sporksmith, line 101: This should probably be a  TryFrom...?

sporksmith (approved):

src/main/host/syscall_types.rs
stevenengler: I think so, but the existing conversions from other types like `i32` already use `From` rather than `TryFrom`, so I just made it consistent with the others for now. I'll make a note to fix all of these in a future PR.

src/main/host/syscall/ioctl.c
stevenengler: Agreed, I'll leave this in the syscall handler.

PR #1786, opened 2021-11-23
Improved socket tests and small bug fixes

stevenengler:
Test improvements include improved tcp/udp tests, and also add unix socket tests for `getpeername()`, `getsockname()`, `sendto()`, and `recvfrom()`. The unix socket tests are currently disabled when running in shadow.

### DIFF
--- a/src/test/socket_utils.rs
+++ b/src/test/socket_utils.rs
@@ -1,0 +1,303 @@
+use crate::get_errno;
+
+use std::convert::TryInto;
+
+/// A container for different types of socket addresses.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum SockAddr {
+    Generic(libc::sockaddr_storage),
+    Inet(libc::sockaddr_in),
+    Unix(libc::sockaddr_un),
+}
+
+impl SockAddr {
+    /// Initialize the inet socket address struct with invalid/dummy data.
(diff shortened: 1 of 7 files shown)


src/test/socket_utils.rs
```diff
@@ -0,0 +1,303 @@
...
+    };
+
+    // connect to the server address
+    {
+        let rv = unsafe { libc::connect(fd_client, addr_ptr, addr_len) };
+        assert!(rv == 0 || (rv == -1 && get_errno() == libc::EINPROGRESS));
+    }
+
+    // shadow needs to run events, otherwise the accept call won't know it
+    // has an incoming connection (SYN packet)
```
sporksmith, line 222: If we didn't do this, wouldn't the `accept4` just block until the packet arrives? If the `usleep` is only needed for `SOCK_NONBLOCK`, maybe worth mentioning that explicitly. (And maybe only doing it if `SOCK_NONBLOCK` is set?)

sporksmith (approved):

src/test/socket_utils.rs
stevenengler: Yeah it's only needed for the non-blocking cases. I added a `if (flags & libc::SOCK_NONBLOCK) != 0` condition here.

PR #1787, opened 2021-11-24
Add unnamed unix sockets created with socketpair()

stevenengler:
This replaces the existing C version, which was limited to only stream sockets.


src/main/host/descriptor/socket/unix.rs
```diff
@@ -0,0 +1,400 @@
...
+    pub fn read<W>(
+        &mut self,
+        mut _bytes: W,
+        _offset: libc::off_t,
+        _event_queue: &mut EventQueue,
+    ) -> SyscallResult
+    where
+        W: std::io::Write + std::io::Seek,
+    {
+        // we could call recvfrom() here, but for now we expect that there are no code paths that
```
sporksmith, line 170: Would application-code calling `read` on the socket get here? If so, seems like we ought to handle it more gracefully (though OK if not in this PR)

src/main/host/descriptor/socket/unix.rs
```diff
@@ -0,0 +1,400 @@
...
+    pub fn write<R>(
+        &mut self,
+        mut _bytes: R,
+        _offset: libc::off_t,
+        _event_queue: &mut EventQueue,
+    ) -> SyscallResult
+    where
+        R: std::io::Read + std::io::Seek,
+    {
+        // we could call sendto() here, but for now we expect that there are no code paths that
```
sporksmith, line 184: Would application-code calling `write` on the socket get here? If so, seems like we ought to handle it more gracefully (though OK if not in this PR)

src/main/host/descriptor/socket/unix.rs
```diff
@@ -0,0 +1,400 @@
...
+            return Err(nix::errno::Errno::EBADF.into());
+        }
+
+        let send_buffer = match (&self.send_buffer, addr) {
+            // already connected but a destination address was given
+            (Some(send_buffer), Some(_addr)) => match self.socket_type {
+                UnixSocketType::Stream => return Err(Errno::EISCONN.into()),
+                UnixSocketType::SeqPacket => send_buffer,
+                UnixSocketType::Dgram => {
+                    // TODO: need to properly look up the socket once we support named sockets
```
sporksmith, line 209: Maybe worth logging a warning, since this could be the point where a simulation starts diverging from "real" behavior?

src/main/host/descriptor/socket/unix.rs
```diff
@@ -0,0 +1,400 @@
...
+                    return Err(Errno::ECONNREFUSED.into());
+                }
+            },
+            // already connected and no destination address was given
+            (Some(send_buffer), None) => send_buffer,
+            // not connected but a destination address was given
+            (None, Some(_addr)) => {
+                // TODO: need to properly look up the socket once we support named sockets
```
sporksmith, line 217: Log a warning?

src/main/host/syscall/socket.rs
```diff
@@ -0,0 +1,578 @@
...
+    flags: libc::c_int,
+    addr_ptr: PluginPtr,
+    addr_len: libc::socklen_t,
+) -> SyscallResult {
+    // get the send flags
+    let flags = match MsgFlags::from_bits(flags) {
+        Some(x) => x,
+        None => {
+            warn!("Invalid sendto flags: {}", flags);
+            MsgFlags::from_bits_truncate(flags)
```
sporksmith, line 70: Is there a reason not to return EINVAL for this case?

src/main/host/syscall/socket.rs
```diff
@@ -0,0 +1,578 @@
...
+    buf_len: libc::size_t,
+    flags: libc::c_int,
+    addr_ptr: PluginPtr,
+    addr_len_ptr: PluginPtr,
+) -> SyscallResult {
+    // get the recv flags
+    let flags = match MsgFlags::from_bits(flags) {
+        Some(x) => x,
+        None => {
+            warn!("Invalid recvfrom flags: {}", flags);
```
sporksmith, line 158: EINVAL?

src/main/host/syscall/socket.rs
```diff
@@ -0,0 +1,578 @@
...
+/// length pointers are NULL. The plugin's address length will be updated to store the size of the
+/// socket address, even if greater than the provided buffer size. If the address is `None`, the
+/// plugin's address length will be set to 0.
+fn write_sockaddr(
+    process: &mut Process,
+    addr: Option<nix::sys::socket::SockAddr>,
+    plugin_addr: PluginPtr,
+    plugin_addr_len: TypedPluginPtr<libc::socklen_t>,
+) -> Result<(), SyscallError> {
+    if plugin_addr.is_null() || plugin_addr_len.is_null() {
```
sporksmith, line 425: Redundant; the MemoryManager will return EFAULT for NULL pointers (assuming the usual case that they're actually inaccessible)

src/main/host/syscall/socket.rs
```diff
@@ -0,0 +1,578 @@
...
+    }
+
+    let plugin_addr = TypedPluginPtr::new::<u8>(plugin_addr, (plugin_addr_len).try_into().unwrap());
+    let mut plugin_addr = process.memory_mut().memory_ref_mut_uninit(plugin_addr)?;
+
+    // make sure that we have a real libc sockaddr before converting to a u8 pointer
+    let from_addr: &libc::sockaddr = from_addr;
+
+    let len_to_copy = std::cmp::min(from_len.try_into().unwrap(), plugin_addr.len());
+    unsafe {
```
sporksmith, line 464: I think `process.memory_mut().copy_to_ptr` would be a bit simpler here. It also potentially saves an allocation and copy through an intermediate buffer in the case the address isn't mapped into shadow.

sporksmith (changes requested):

src/main/host/descriptor/socket/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+use crate::host::descriptor::{FileMode, FileState, FileStatus, SyscallResult};
+use crate::host::memory_manager::MemoryManager;
+use crate::host::syscall_types::{PluginPtr, SysCallReg, SyscallError};
+use crate::utility::event_queue::EventQueue;
+
+use unix::UnixSocketFile;
+
+use nix::sys::socket::SockAddr;
+
+//pub mod buffer;
```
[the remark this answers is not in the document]
stevenengler: Fixed.

src/main/host/descriptor/socket/unix.rs
stevenengler: Added a warning.

src/main/host/descriptor/socket/unix.rs
stevenengler: Added a warning.

src/main/host/syscall/socket.rs
stevenengler: Updated.

src/main/host/syscall/socket.rs
stevenengler: This is here to avoid unexpected behaviour when one plugin ptr is null and the other isn't. For example if `plugin_addr` is null and `plugin_addr_len` isn't null, this function would modify the value at `plugin_addr_len` before trying to write to `plugin_addr` and returning `EFAULT`.

I added a comment here to explain the purpose of this check.

src/main/host/descriptor/socket/unix.rs
stevenengler: The `read` syscall handler automatically calls the `recvfrom` syscall handler if the fd is a socket, so calling `read()` on a socket fd should never reach here.

src/main/host/descriptor/socket/unix.rs
stevenengler: The `write` syscall handler automatically calls the `sendto` syscall handler if the fd is a socket, so calling `write()` on a socket fd should never reach here.

src/main/host/syscall/socket.rs
sporksmith: Could you add a comment explaining that?

sporksmith (changes requested):

src/main/host/descriptor/socket/unix.rs
stevenengler: Expanded the comment.

src/main/host/descriptor/socket/unix.rs
stevenengler: Expanded the comment.

src/main/host/syscall/socket.rs
stevenengler: Added.

src/main/host/syscall/socket.rs
stevenengler: Removed the early exit.

sporksmith (approved):

src/main/host/syscall/socket.rs
stevenengler: Added a comment.

PR #1791, opened 2021-11-29
Replaced PRNG with a rust version

stevenengler:
This allows us to access the prng from rust. Since we're changing the prng, this may have performance implications.

Edit: Preliminary testing seems to show a performance improvement in a small tor simulation.

Edit 2: In a longer tor experiment (1% network), there is no significant wall-clock difference when using this PR (they both took 01:31:12 to complete).


src/main/utility/random.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    use rand::{Rng, RngCore, SeedableRng};
+
+    #[no_mangle]
+    pub unsafe extern "C" fn random_new(seed: u64) -> *mut Random {
+        let mut rng = Xoshiro256PlusPlus::seed_from_u64(seed);
+
+        // From https://docs.rs/rand/0.8.4/rand/rngs/struct.SmallRng.html:
+        //
+        // > SmallRng is not a good choice when: [...] Seeds with many zeros are provided. In such
```
sporksmith, line 20: Yikes. Glad you pre-emptively caught this!

sporksmith (approved):

PR #1795, opened 2021-12-02
Support creating/binding abstract-named unix sockets

stevenengler:


src/main/host/descriptor/socket/abstract_unix_ns.rs
```diff
@@ -0,0 +1,273 @@
...
+/// generators become stable.
+fn incremental_name<const L: usize>(index: usize) -> [u8; L] {
+    // there are a limited number of valid names
+    assert!(index < CHARSET.len().pow(L as u32));
+
+    let mut name = [0u8; L];
+
+    // set each character of the name
+    for x in 0..L {
+        let charset_index = (index / CHARSET.len().pow(x as u32)) % CHARSET.len();
```
sporksmith, line 203: This is a bit tough to follow. Maybe this can be broken up into multiple statements and/or commented a bit to help explain how this works?

src/main/host/descriptor/socket/abstract_unix_ns.rs
```diff
@@ -0,0 +1,273 @@
...
+            Self::NameInUse => write!(f, "Name is already in use"),
+            Self::NoNamesAvailable => {
+                write!(f, "Names in the ephemeral name range are all in use")
+            }
+            Self::NameNotFound => write!(f, "Name was not found in the address map"),
+        }
+    }
+}
+
+/// The characters that are valid in auto-generated names.
```
sporksmith, line 177: A comment on how these constants are chosen would be helpful, particularly if there's a man page or something to cite.

src/main/host/descriptor/socket/unix.rs
```diff
@@ -158,6 +162,53 @@ impl UnixSocketFile {
...
+            }
+        } else if addr.path_len() == 0 {
+            // if given an "unnamed" address
+            let namespace = Arc::clone(&socket.borrow().namespace);
+            if AbstractUnixNamespace::autobind(&namespace, socket, rng).is_err() {
+                // no autobind addresses remaining
+                return Err(Errno::EADDRINUSE.into());
+            }
+        } else {
+            log::warn!("Only abstract names are supported for unix sockets");
```
sporksmith, line 203: Nit - probably worth inserting 'currently'

src/main/host/syscall/socket.c
```diff
@@ -725,6 +725,12 @@ SysCallReturn syscallhandler_bind(SysCallHandler* sys,
         return (SysCallReturn){.state = SYSCALL_DONE, .retval.as_i64 = -EINVAL};
     }
 
+    /* Make sure the addr PluginPtr is not NULL. */
+    if (!addrPtr.val) {
```
sporksmith, line 729: I think we don't need this check, since the MemoryManager will detect if the pointer isn't accessible?

src/main/host/syscall/socket.c
```diff
@@ -733,12 +739,6 @@ SysCallReturn syscallhandler_bind(SysCallHandler* sys,
...
     }
 
-    /* Make sure the addr PluginPtr is not NULL. */
-    if (!addrPtr.val) {
-        debug("binding to a NULL address is invalid");
-        return (SysCallReturn){.state = SYSCALL_DONE, .retval.as_i64 = -EINVAL};
-    }
-
     const struct sockaddr* addr = process_getReadablePtr(sys->process, addrPtr, addrlen);
     utility_assert(addr);
```
sporksmith, line 743: *this* should return -EFAULT on NULL instead of crashing. i.e. check whether accessing the pointer succeeded here, instead of whether we think it'll succeed above

sporksmith (approved):

src/main/host/descriptor/socket/abstract_unix_ns.rs
```diff
@@ -0,0 +1,273 @@
...
+
+use atomic_refcell::AtomicRefCell;
+use rand::seq::SliceRandom;
+
+use crate::host::descriptor::socket::UnixSocketFile;
+use crate::host::descriptor::FileState;
+use crate::host::descriptor::StateListenerFilter;
+use crate::utility::event_queue::Handle;
+
+type NamespaceEntry = (
```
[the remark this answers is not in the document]
sporksmith: Oops, I means to circle back to this comment after reading the rest of the code. No strong feeling either way, really.

src/main/host/syscall/socket.c
stevenengler: Removed this check since it isn't required with the change from your following comment.

src/main/host/syscall/socket.c
stevenengler: Yeah I think this is a common issue throughout the C syscall handlers. I fixed this for this case.

src/main/host/descriptor/socket/unix.rs
stevenengler: Added.

src/main/host/descriptor/socket/abstract_unix_ns.rs
stevenengler: Added.

src/main/host/descriptor/socket/abstract_unix_ns.rs
stevenengler: I think I improved this, it's essentially just changing the index from base 10 to the digits as base `CHARSET.len()` (16).

src/main/host/descriptor/socket/abstract_unix_ns.rs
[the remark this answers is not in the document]
stevenengler: Made this a struct instead.

PR #1800, opened 2021-12-06
Fix test regexes which caused incorrect tests to pass

stevenengler:
We need to ignore the return values of managed processes in some tests, and one of the sendto/recvfrom tests was passing when it shouldn't have been.

### DIFF
--- a/src/test/tor/minimal/CMakeLists.txt
+++ b/src/test/tor/minimal/CMakeLists.txt
@@ -22,6 +22,9 @@ add_custom_target(tor-minimal-shadow-data-template ALL
 
 add_shadow_tests(BASENAME tor-minimal
                  LOGLEVEL info
+                 # don't check tor's return value since it will be killed by shadow at the end of
+                 # the sim
+                 CHECK_RETVAL false
                  ARGS
                    --use-cpu-pinning true
                    --parallelism 2
--- a/src/test/CMakeLists.txt
+++ b/src/test/CMakeLists.txt
@@ -125,14 +125,16 @@ macro(add_shadow_tests)
          )
       endif()
 
+      # check that we didn't leak any reference-counted objects
+      set(FAIL_REGEX "Memory leak detected")
+
       # need to check the test's return code, not just shadow's (see shadow/shadow#902)
       if(SHADOW_TEST_CHECK_RETVAL)
-         set_property(TEST ${SHADOW_TEST_NAME} PROPERTY FAIL_REGULAR_EXPRESSION "main error code '.*' for process")
+         list(APPEND FAIL_REGEX "main error code '.*' for process")
       endif()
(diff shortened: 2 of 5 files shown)


sporksmith (approved):

src/test/tor/minimal/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ add_custom_target(tor-minimal-shadow-data-template ALL
 
 add_shadow_tests(BASENAME tor-minimal
                  LOGLEVEL info
+                 CHECK_RETVAL false
```
sporksmith, line 25: Uh oh, we're not *supposed* to need this, right? If this is a temporary workaround to get the fix of the test framework in, then an issue and a comment with a link to the issue here would be good...

stevenengler: Since we don't shut down tor properly, shadow will eventually send it a sigkill and tor will exit with a non-zero return value. We don't want the test to fail in this case, so I think we need to set this. If tor exits with a non-zero return value before the simulation ends, then shadow itself will return a non-zero exit code.

sporksmith: Maybe I'm misremembering, but I thought shadow still exited with code 0 if managed processes were killed via SIGKILL

sporksmith: (And if I'm wrong, would be good to leave a comment here with this explanation)

stevenengler: I think that's the case, but only if it's at the end of the simulation.

sporksmith: Here's where I think we ignore SIGKILL for purpose of determining if shadow should exit with an error: https://github.com/shadow/shadow/blob/main/src/main/host/process.c#L349

stevenengler: Okay it looks like you're right. If a managed process is killed by SIGKILL, then shadow exits with an exit code of 0, and if a managed process is killed by any other signal or exits with a non-zero exit code, then shadow exists with a non-zero exit code.

But I think we still need to ignore the log message since it is still logged even when tor is killed by a SIGKILL.

sporksmith: Oh, yes I was confused and thinking this was checking shadow's return code.

Maybe there's some way we could make this a bit clearer, but I think it's mostly outside the scope of this PR. (Though a one line comment here that we expect tor to still be running at the exit wouldn't hurt)

stevenengler: Yeah, we have two similarly named variables `CHECK_RETVAL` and `EXPECT_ERROR`, so we should probably rename these.

> (Though a one line comment here that we expect tor to still be running at the exit wouldn't hurt)

Added a comment.

Issue #1794, opened 2021-12-01
Livelock with curl on Debian 10


cdoepmann:
I am trying to get shadow to work (ultimately, for Tor research, but staying with a toy example for now), but experience the following issue: Whenever the shadow simulation is started, it appears to go to some livelock, using full CPU and not getting the actual simulation done. I have tried to investigate the problem, but digging into the shadow internals is somewhat cumbersome due to my lack of knowledge of the code base.

**Describe the issue**
Shadow goes to livelock, takes full CPU resources and simulation does not continue. Both with the toy example from the user guide (curl) ~~as well as with the Tor example~~ (edit: The Tor example failed due to issue #1788 instead).

**To Reproduce**
command line:
```
shadow shadow.yaml > shadow.log
```

`shadow.yaml`:
```
general:
  # stop after 10 simulated seconds
  stop_time: 10s
  log_level: debug

network:
  graph:
    # use a built-in network graph containing
    # a single vertex with a bandwidth of 1 Gbit
    type: 1_gbit_switch

hosts:
  # a host with the hostname 'server'
  server:
    network_node_id: 0
    processes:
    - path: /usr/bin/python3
      args: -m http.server 80
      start_time: 3s
  # one host
  client:
    network_node_id: 0
    quantity: 1
    processes:
    - path: /usr/bin/curl
      args: -s server
      start_time: 5s
```

**Operating System (please complete the following information):**
`lsb_release -d`:
```
Description:    Debian GNU/Linux 10 (buster)
```
`uname -a`:
```
Linux cdoepmann-debian 4.19.0-18-amd64 #1 SMP Debian 4.19.208-1 (2021-09-29) x86_64 GNU/Linux
```

**Shadow (please complete the following information):**
`shadow --version`:
```
Shadow v2.0.0-88-g0a686c9b 2021-11-29--19:57:10
```
(same for original v2.0.0)

**Additional context**

[shadow.log](https://github.com/shadow/shadow/files/7635145/shadow.log)

backtrace when interupting the livelock (shadow process):
```
#0  syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
#1  0x00007ffff7cc813f in g_cond_wait () from /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#2  0x00005555567bae82 in countdownlatch_await (latch=0x555556cc3620)
    at /home/christoph/tools/shadow/src/main/utility/count_down_latch.c:39
[...]

PR #1804, opened 2021-12-08
Add note to docs about cURL spinloop

stevenengler:
Adds a note to the docs about #1794. We should maybe consider removing this example, but I think we should have this in the docs in the meantime.


robgjansen (approved):
I agree it's a good idea to add a blurb to the example in the docs in case those running the example hit the bug.

PR #1805, opened 2021-12-08
Fix rust unread struct field warning

stevenengler:
```text
warning: field is never read: `memory_copier`
  --> host/memory_manager/memory_mapper.rs:67:5
   |
67 |     memory_copier: MemoryCopier,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default
```

### DIFF
--- a/src/main/host/memory_manager/memory_mapper.rs
+++ b/src/main/host/memory_manager/memory_mapper.rs
@@ -1,4 +1,4 @@
-use crate::host::memory_manager::{page_size, MemoryCopier, MemoryManager};
+use crate::host::memory_manager::{page_size, MemoryManager};
 use crate::host::syscall_types::{PluginPtr, SyscallResult, TypedPluginPtr};
 use crate::host::thread::Thread;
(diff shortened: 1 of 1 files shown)


src/main/host/memory_manager/memory_mapper.rs
```diff
@@ -64,7 +64,7 @@ fn log_regions<It: Iterator<Item = (Interval, Region)>>(level: log::Level, regio
 /// MemoryMapper via its `handle_*` methods.
 #[derive(Debug)]
 pub struct MemoryMapper {
-    memory_copier: MemoryCopier,
+    _memory_copier: MemoryCopier,
```
sporksmith, line 67: I think we can just delete this field.

sporksmith (approved):

src/main/host/memory_manager/memory_mapper.rs
stevenengler: Removed the field, and also the now unused `let memory_copier = MemoryCopier::new(thread.system_pid())`.

PR #1810, opened 2021-12-10
Upgrade clap to 3.0.0-rc.3

stevenengler:


src/main/Cargo.toml
```diff
@@ -16,7 +16,7 @@ bitflags = "1.3"
 # issue: https://github.com/tokio-rs/bytes/issues/287
 # pr: https://github.com/tokio-rs/bytes/pull/513
 bytes = { git = "https://github.com/stevenengler/bytes", rev = "871c7a18482186c91dbe2a2d9be20cc2c636fb02" }
-clap = { version = "3.0.0-beta.5", features = ["wrap_help"] }
+clap = { version = "3.0.0-beta.5", features = ["derive", "wrap_help"] }
```
sporksmith, line 19: Should this beta.5 also be rc.3?

sporksmith (approved):

src/main/Cargo.toml
stevenengler: Oops good catch, I'm not sure how the lockfile got updated but not the toml.

PR #1828, opened 2021-12-27
Allows TCPS_LISTEN to be a valid listening state

RWails:
Allows TCPS_LISTEN to be a valid listening state; enables socket double-listen test in Shadow.

### DIFF
--- a/src/main/host/descriptor/tcp.c
+++ b/src/main/host/descriptor/tcp.c
@@ -1431,7 +1431,7 @@ gboolean tcp_isValidListener(TCP* tcp) {
  */
 gboolean tcp_isListeningAllowed(TCP* tcp) {
     MAGIC_ASSERT(tcp);
-    if(tcp->state == TCPCS_NONE && tcp->flags == TCPF_NONE) {
+    if ((tcp->state == TCPS_CLOSED || tcp->state == TCPS_LISTEN) && tcp->flags == TCPF_NONE) {
         return TRUE;
     } else {
         return FALSE;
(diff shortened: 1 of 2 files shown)


src/main/host/descriptor/tcp.c
```diff
@@ -1431,7 +1431,7 @@ gboolean tcp_isValidListener(TCP* tcp) {
  */
 gboolean tcp_isListeningAllowed(TCP* tcp) {
     MAGIC_ASSERT(tcp);
-    if(tcp->state == TCPCS_NONE && tcp->flags == TCPF_NONE) {
+    if ((tcp->state == TCPS_CLOSED || tcp->state == TCPS_LISTEN) && tcp->flags == TCPF_NONE) {
```
RWails, line 1434: In addition to adding `TCPS_LISTEN` to the set of allowed listening states, I also make the change of `TCPCS_NONE` -> `TCPS_CLOSED`.

`TCPCS_NONE` is a TCP child state, not a TCP state. I think the correct state to check for is `TCPS_CLOSED`, which just happens to alias to the same value as `TCPCS_NONE(=0)`.

stevenengler (approved):
Looks good to me! (Just making a note here that this change allows multiple `listen()` calls, but doesn't yet change the backlog size.)

https://github.com/shadow/shadow/blob/5353405e136fb9d7a630fffc4bbd7f3bec426b03/src/main/host/syscall/socket.c#L1073-L1079

stevenengler:
> Do we want to change the backlog here, or merge as-is? Also, is there a test we could/should write for the Linux backlog setting behavior?

I think it's fine to merge as-is, but if we don't update the backlog then we should keep the issue open. For a test, I don't think we have any normal backlog tests. So we'd probably want a test that checks what happens when the backlog fills up normally, and another two that test the behaviour if the backlog grows or shrinks (does it discard existing pending connections, or does it only update a high water mark?).

robgjansen:
@stevenengler could you take over this PR and do the rebase and merge when you get a chance? Thanks!

PR #1838, opened 2022-01-04
Small updates to the documentation and help text

stevenengler:


src/main/core/support/configuration.rs
```diff
@@ -16,15 +16,18 @@ use super::units::{self, Unit};
 use crate::cshadow as c;
 use log_bindings as c_log;
 
+const START_HELP_TEXT: &str = "\
+    Run real applications over simulated networks.\n\n\
```
robgjansen, line 20: Why 2 newlines instead of 1?

robgjansen (approved):

src/main/core/support/configuration.rs
stevenengler: I just thought it looked better as a separate paragraph.

PR #1848, opened 2022-01-12
Show the progress when stderr is not a tty

stevenengler:
Previously if stderr was not a tty, Shadow would ignore the "progress" option. Now Shadow will periodically print the progress to stderr.

### DIFF
--- a/src/main/utility/status_bar.rs
+++ b/src/main/utility/status_bar.rs
@@ -6,6 +6,7 @@ use std::io::Write;
 use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 use std::sync::RwLock;
+use std::time::Duration;
 
 const SAVE_CURSOR: &str = "\u{1B}[s";
 const RESTORE_CURSOR: &str = "\u{1B}[u";
@@ -26,7 +27,7 @@ pub struct StatusBar<T: StatusBarState> {
 
 impl<T: 'static + StatusBarState> StatusBar<T> {
     /// Create and start drawing the status bar.
-    pub fn new(state: T, redraw_interval: std::time::Duration) -> Self {
+    pub fn new(state: T, redraw_interval: Duration) -> Self {
         let state = Arc::new(RwLock::new(state));
         let stop_flag = Arc::new(AtomicBool::new(false));
 
@@ -39,11 +40,7 @@ impl<T: 'static + StatusBarState> StatusBar<T> {
         }
     }
 
-    fn redraw_loop(
-        state: Arc<RwLock<T>>,
-        stop_flag: Arc<AtomicBool>,
-        redraw_interval: std::time::Duration,
-    ) {
+    fn redraw_loop(state: Arc<RwLock<T>>, stop_flag: Arc<AtomicBool>, redraw_interval: Duration) {
         // we re-draw the status bar every interval, even if the state hasn't changed, since the
         // terminal might have been resized and the scroll region might have been reset
         while !stop_flag.load(std::sync::atomic::Ordering::Acquire) {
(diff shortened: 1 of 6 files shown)


stevenengler:
@sporksmith Can you take an extra look at the condvar usage in this PR? As far as I can tell there aren't any race conditions there, but this is the first time I've used a condition variable so it could use an extra look.

src/main/utility/status_bar.rs
```diff
@@ -118,6 +114,79 @@ impl<T: 'static + StatusBarState> StatusBar<T> {
...
+
+impl<T: 'static + StatusBarState> StatusPrinter<T> {
+    /// Create and start drawing the status bar.
+    pub fn new(state: T) -> Self {
+        let state = Arc::new(RwLock::new(state));
+        let stop_cond = Arc::new((Mutex::new(false), Condvar::new()));
+
+        Self {
+            state: Arc::clone(&state),
+            stop_cond: Arc::clone(&stop_cond),
```
sporksmith, line 131: Optional nit - a "one shot channel" (`std::sync::mpsc::channel<()>`) might be a little easier to maintain than these lower level sync primitives. The print loop can use `recv_timeout` with a 30s timeout. On timeout print the status. When something is received, exit.

src/main/utility/status_bar.rs
```diff
@@ -118,6 +114,79 @@ impl<T: 'static + StatusBarState> StatusBar<T> {
...
+            std::io::stderr()
+                .write_all(format!("Progress: {}\n", *state.read().unwrap()).as_bytes())
+                .unwrap();
+            let _ = std::io::stderr().flush();
+
+            // interval is limited to 2 hours
+            const MAX_INTERVAL: Duration = Duration::from_secs(2 * 60 * 60);
+
+            // exponential backoff so that long simulations don't flood stderr
+            print_interval = std::cmp::min(print_interval.saturating_mul(2), MAX_INTERVAL);
```
sporksmith, line 168: This is clever, but potentially confusing. If the last update took an hour, and it's been an hour and a half since that one, I might reasonably think the simulation is stuck.

Maybe the progress line could include something like "next update in {}"?

sporksmith: I'd also lean towards an upper limit of more like 30m... maybe even 10m.

src/main/utility/status_bar.rs
```diff
@@ -118,6 +114,79 @@ impl<T: 'static + StatusBarState> StatusBar<T> {
...
+        }
+    }
+
+    /// Stop printing the status.
+    pub fn stop(&mut self) {
+        *self.stop_cond.0.lock().unwrap() = true;
+        self.stop_cond.1.notify_all();
+        if let Some(handle) = self.thread.take() {
+            handle.thread().unpark();
```
sporksmith, line 177: Is this needed? I don't see any call to `park`? This seems lower level than we should need to mess with here.

sporksmith (approved):

src/main/utility/status_bar.rs
stevenengler: Good idea, I've replaced the mutex/condvar with a channel.

src/main/utility/status_bar.rs
stevenengler: Oops good catch, that shouldn't have been there.

src/main/utility/status_bar.rs
stevenengler: I didn't think of that. I added a "next update" message for now, and if this ends up being confusing we can just stick with a constant interval or something.

> I'd also lean towards an upper limit of more like 30m... maybe even 10m.

I didn't want to set the interval too low since it would generate a lot of messages when people run long simulations (for example 14 day sims). I've changed this to 30 minutes for now.

sporksmith (approved):

Issue #1370, opened 2021-05-20
Decide if preloading AES encryption is important


robgjansen:
For our Tor experiments, we used to [preloading openssl encryption functions](https://github.com/shadow/shadow-plugin-tor/blob/064c46041510bd2bcbcbc6426bbc5687bf372f20/src/tor/shadowtor-preload.c#L70) in Shadow version 1.x. This used to lead to a big performance boost (see Figure 2 in the [original Shadow paper](https://www.robgjansen.com/publications/shadow-ndss2012.pdf)).

Since that, modern CPUs enable optimizations that can perform AES operations with very low latency. We are wondering if it still makes sense to maintain the code needed to preload the AES functions going forward.

To help us decide, we should run an experiment in Shadow 2.x both with and without preloading the AES functions, and test how performance is affected.

robgjansen:
Initial results in a 5% Tor network indicate no noticeable performance change when performing the crypto or preloading the crypto. Note that this only tests skipping the AES symmetric encryption, not public key operations (which were never supported by the tor preload lib).

[run_time.pdf](https://github.com/shadow/shadow/files/6517706/run_time.pdf)

Note that we never counted how often the functions are called when preloading, to better understand how many cryptographic operations we are able to skip. Not sure that we need that, but it would help us be a bit more confident about the results.

robgjansen:
We decided that we want to at the very least double check that the preloaded cipher functions are getting intercepted on the specific machine that was used to run this experiment.

We may also consider counting the frequency with which those functions are called to help us guess how much performance overhead we should be expecting from skipping those calls.

robgjansen:
I ran a 10% Tor network with shadow/shadow@154a11d8bc2a84eb09f884cf27d5186b0890ec2b with and without the following code (taken from shadow-tor-preload) preloaded into the managed processes:

```c
/*
 * The Shadow Simulator
 * Copyright (c) 2010-2011, Rob Jansen
 * See LICENSE for licensing information
 */

#include <stdio.h>
#include <string.h>

static unsigned long aes_e_cnt = 0;
static unsigned long aes_d_cnt = 0;
static unsigned long aes_ce_cnt = 0;
static unsigned long aes_cd_cnt = 0;
static unsigned long aes_evp_cnt = 0;

static void _aes_print_counters() {
    fprintf(stderr,
            "Counters: {AES_encrypt:%lu, AES_decrypt:%lu, AES_ctr128_encrypt:%lu, "
[...]

PR #1853, opened 2022-01-14
New lib for optionally preloading crypto functions

robgjansen:
Adds a new lib that preloads some AES functions so we can optionally skip some CPU crypto operations for faster simulation on older hardware (e.g. CPUs that do not support AES-NI).

When enabled (it's off by default), the preload lib will be injected into the LD_PRELOAD environment for all nodes, and we skip some AES symmetric encryption and decryption operations, including the ones used by Tor. (Disabling encryption is safe in private, synthetic Shadow nets.)

refs #1370


stevenengler:
Adding benchmark runtime graph:

run_time

robgjansen:
I verified that the crypto preload lib is correctly appearing in the nodes' `LD_PRELOAD` env variable, but the AES crypto functions are not getting intercepted properly:

```
Counters: {'AES_encrypt':0, 'AES_decrypt':0, 'AES_ctr128_encrypt':0, 'CRYPTO_ctr128_encrypt':0, 'CRYPTO_ctr128_encrypt_ctr32':0, 'EVP_Cipher':0}
```

Which is why we see no significant change in performance. A grab from `perf top` shows us that indeed we are spending a lot of time in `AES_encrypt`:

perf_top

robgjansen:
It appears that tor is being linked against `libcrypto.so`, which I thought would be enough to make `AES_encrypt` preloadable via the `LD_PRELOAD` method. This case does appear to be working in my dev VM, but not on the gh runner. Note that the `libshadow_openssl_rng.so` preload works the same way too, and is on by default.

I don't think we should have to build openssl manually [like we used to do](https://github.com/shadow/shadow-plugin-tor/blob/a1be39535c2bec04826ce70b99d43f80c082c0d0/setup#L232-L241) (and then link tor to it using its `--with-openssl-dir` build option). But clearly something is not quite right here...

stevenengler:
It looks like I can reproduce this in an ubuntu 20.04 container on an ubuntu 20.04 host. I see the `Loading the preloaded crypto interception lib` in the plugin's stderr, but no counters (I removed the `(cnt % 1000) == 0` condition). I see this behaviour with tor 0.4.5.9, 0.4.6.9, and 0.3.5.17.

stevenengler:
The openssl crypto functions are not intercepted in an ubuntu 20.04 container, but they are in a debian 11 container. The openssl rand functions are intercepted in both the ubuntu 20.04 and debian 11 containers.

robgjansen:
> The openssl crypto functions are not intercepted in an ubuntu 20.04 container, but they are intercepted in a debian 11 container. The openssl rand functions are intercepted in both the ubuntu 20.04 and debian 11 containers.

Explanation of the problem is at https://github.com/shadow/shadow/issues/1370#issuecomment-1022491092, work-around is to conditionally intercept `EVP_EncryptUpdate`.

Now waiting on a performance benchmark test to see if it works.

robgjansen:
Latest findings: doesn't seem to work on Ubuntu 20.04 because there is a fair amount of AES triggered internally in openssl. It does seem to work as expected in Debian 11 though:
run_time

src/lib/openssl_preload/crypto.c
```diff
@@ -0,0 +1,219 @@
...
+// Caches whether or not a memory address is from libssl.so
+typedef struct {
+    void* addr;
+    bool is_libssl;
+} bt_cache_entry_t;
+
+// We use a small cache size: in ad-hoc experiments with tor-0.4.6.9, we observed
+// at most three callers of EVP_EncryptUpdate.
+#define EVP_BACKTRACE_CACHE_LEN 10
+static bt_cache_entry_t evp_backtrace_cache[EVP_BACKTRACE_CACHE_LEN];
```
sporksmith, line 24: This cache doesn't appear to be thread safe; it probably ought to be a protected with e.g. a `pthread_mutex_t`.

Technically we shouldn't have true parallel access if it's running under Shadow, as long as no blocking syscalls are made in the middle of accessing it. (And we don't introduce true preemption in Shadow in the future). I'd lean towards just putting in and using a mutex to be  safe, but I'd be ok with just documenting that this code isn't thread-safe except when running under Shadow.

src/lib/openssl_preload/crypto.c
```diff
@@ -0,0 +1,219 @@
...
+
+// For storing a pointer to the original EVP_EncryptUpdate function.
+typedef int EVP_EncryptUpdate_func(void*, unsigned char*, int*, const unsigned char*, int);
+static void* evp_eu_funcptr = NULL;
+
+#ifdef DEBUG
+// Counters for verifying that interception is happening correctly.
+static unsigned long aes_e_cnt = 0;
```
sporksmith, line 32: These also aren't thread-safe.

We could either:
* Protect them with a mutex (maybe a single global mutex that also protects the cache)
* Document that this code is thread  unsafe except under Shadow
* Make these individually atomics  and use the corresponding operators like [atomic_fetch_add](https://en.cppreference.com/w/c/atomic/atomic_fetch_add). If we're going to make this code thread-safe though, I think the first option (single global mutex) would be more efficient assuming no contention in practice.

src/lib/openssl_preload/crypto.c
```diff
@@ -0,0 +1,219 @@
...
+#endif
+}
+
+#ifdef DEBUG
+__attribute__((destructor)) void _crypto_unload() {
+    fprintf(stderr, "Unloading the preloaded crypto interception lib\n");
+    _print_counters();
+}
+
+#define _maybe_print_counters(cnt)                                                                 \
```
sporksmith, line 71: Is there a reason to make this a macro instead of a function?

src/lib/openssl_preload/crypto.c
```diff
@@ -0,0 +1,219 @@
...
+    {                                                                                              \
+        if ((cnt % 1000) == 0) {                                                                   \
+            _print_counters();                                                                     \
+        }                                                                                          \
+    }
+#endif
+
+void AES_encrypt(const unsigned char* in, unsigned char* out, const void* key) {
+#ifdef DEBUG
```
sporksmith, line 80: Optional - it looks like you could move these `#ifdef`s into or around the macro/function definition.

src/lib/openssl_preload/crypto.c
```diff
@@ -0,0 +1,219 @@
...
+typedef int EVP_EncryptUpdate_func(void*, unsigned char*, int*, const unsigned char*, int);
+static void* evp_eu_funcptr = NULL;
+
+#ifdef DEBUG
+// Counters for verifying that interception is happening correctly.
+static unsigned long aes_e_cnt = 0;
+static unsigned long aes_d_cnt = 0;
+static unsigned long aes_ce_cnt = 0;
+static unsigned long cry_ce_cnt = 0;
+static unsigned long cry_cec_cnt = 0;
```
sporksmith, line 36: s/cry/cnt/ ?

sporksmith (approved):

src/lib/openssl_preload/crypto.c
robgjansen: The counters do not exist when DEBUG is defined, so if I use a function then the nonexistent counter variables break compilation (when building in release mode).

robgjansen: Oh I guess the debug mode version could be a function, and then the release mode version could be an empty macro. :)

src/lib/openssl_preload/crypto.c
robgjansen: Done.

src/lib/openssl_preload/crypto.c
robgjansen: Done.

robgjansen:
@sporksmith What do you think about just removing the whole cache and `EVP_EncryptUpdate` code? I don't think we need it on Debian 11 anyway (or when using an openssl whose functions go through the PLT), and it would be much simpler.

sporksmith:
> @sporksmith What do you think about just removing the whole cache and EVP_EncryptUpdate code? I don't think we need it on Debian 11 anyway (or when using an openssl whose functions go through the PLT), and it would be much simpler.

The current approach might be nice to keep around if we ever want to pursue it further - e.g. we care about getting it to work on other platforms, or it stops working on Debian. I could go either way though.

src/lib/openssl_preload/crypto.c
```diff
@@ -0,0 +1,232 @@
...
+             "'EVP_Cipher':%lu, 'EVP_EncryptUpdate':%lu}\n",
+             aes_e_cnt, aes_d_cnt, aes_ce_cnt, crypto_ce_cnt, crypto_cec_cnt, evp_c_cnt,
+             evp_eu_cnt);
+}
+
+__attribute__((constructor)) void _crypto_load() {
+    debuglog("Loading the preloaded crypto interception lib\n");
+
+    int result = pthread_mutex_lock(&global_state_lock);
+    assert(result == 0);
```
sporksmith, line 63: asserts are disabled in debug builds.

I think we want something like `if (result != 0) { abort(); }`.

Or if we want to pull in the logging library here (I don't think there's a reason not to...)  `if (result != 0) { panic("pthread_mutex_lock: %s", strerror(result));}`

src/lib/openssl_preload/crypto.c
```diff
@@ -0,0 +1,232 @@
...
+            }
+        }
+
+        result = pthread_mutex_unlock(&global_state_lock);
+        assert(result == 0);
+    }
+
+    if (caller_addr != NULL && !caller_is_libssl) {
+        // Skip the crypto in calls made from the application, e.g. tor.
+        _increment(&evp_eu_cnt);
```
sporksmith, line 223: Optional - right now the lock will be taken and released twice in each call to `EVP_EncryptUpdate` - once to access the cache and once to access the counter. We could require that the caller  has already taken the lock before calling _increment, and then here continue holding the lock until after its called.

src/lib/openssl_preload/crypto.c
```diff
@@ -0,0 +1,232 @@
...
+
+// Caches whether or not a memory address is from libssl.so
+typedef struct {
+    void* addr;
+    bool is_libssl;
+} bt_cache_entry_t;
+
+// Lock for safely accessing this lib's global state from multiple threads.
+// TODO: We currently do no error checking when operating on this lock.
+pthread_mutex_t global_state_lock = PTHREAD_MUTEX_INITIALIZER;
```
sporksmith, line 31: static?

robgjansen: Oops, thanks!

src/lib/openssl_preload/crypto.c
robgjansen: Done.

src/lib/openssl_preload/crypto.c
robgjansen: Done.

src/lib/openssl_preload/crypto.c
robgjansen: Added a global mutex to protect global state.

src/lib/openssl_preload/crypto.c
robgjansen: Added a global mutex to protect global state.

robgjansen:
Full benchmark results:
https://github.com/shadow/benchmark-results/tree/master/tor/2022-02-11-T22-29-57

Issue #1846, opened 2022-01-11
Remove stack switch in shim_emulated_syscallv


sporksmith:
This was meant to be a short term way to get the memory manager's stack remapping working, and should be cleaned up.

What happens now is that every time the shim initiates a syscall with Shadow, it runs the handler on a thread-local statically allocated stack. On the Shadow side, the MemoryManager lazily remaps the stack into shared memory the first time it handles a syscall. This involve several control transfers between Shadow and the shim; having the shim-side code using a different stack ensures *its* stack doesn't get clobbered out from underneath it.

Once the MemoryManager has remapped the stack, there is no longer a need to switch stacks. Since there isn't currently a clear signal to the shim side that the stack has been remapped though, we currently switch over to it every time.

This shouldn't hurt performance, but confuses gdb backtraces.

We should probably get this out of the main syscall handling path, and instead have a dedicated special syscall we make from the shim side during startup to request Shadow to remap the stack (if it's going to). That one syscall would need to be made on a separate stack, but subsequent syscalls would no longer need to swap stacks.

We could probably also use a temporary allocation for this special-purpose stack instead of allocating it statically. e.g. we could allocate it with an anonymous `mmap` initially, and then `munmap`ing it when we're done with it.

PR #1854, opened 2022-01-14
Explicitly initialize MemoryManager

sporksmith:
Initialize Memory Manager explicitly from shim    
    
Instead of assuming any syscall *might* initialize the Memory Manager,    
and hence switching to a custom stack on every syscall, explicitly    
request Shadow to initialize the Memory Manager at startup.    
    
This allows us to free the stack used for this purpose, and should    
simplify debugging.    
    
Because it's no longer part of the syscall path itself, we can now use    
the higher level functions `getcontext` and `swapcontext` instead of raw    
assembly. Previously syscalls made by those functions would trap,    
resulting in infinite recursion.    
    
Fixes https://github.com/shadow/shadow/issues/1846


sporksmith:
This includes / depends-on https://github.com/shadow/shadow/pull/1852. Probably makes sense to review and merge that one first.

PR #1859, opened 2022-01-19
Add per-process shared memory and clean up

sporksmith:
* Adds per-process shared memory, which we need to hold per-process signal state.
* Moves per-thread shared memory from ThreadPtrace to Thread, since we will need it for per-thread signal state.
* Moves the simtime from per-thread state to per-process state.
* Migrates thread_preload to use the simtime in shared memory instead of explicitly sending it in IPC messages.

### DIFF
--- a/src/lib/shim/shim.c
+++ b/src/lib/shim/shim.c
@@ -85,7 +85,7 @@ static int* _shim_allowNativeSyscallsFlag() {
     return shimtlsvar_ptr(&v, sizeof(bool));
 }
 
-static void _shim_set_allow_native_syscalls(bool is_allowed) {
+static void _shim_ptrace_set_allow_native_syscalls(bool is_allowed) {
     if (_shim_thread_shared_mem()) {
         _shim_thread_shared_mem()->ptrace_allow_native_syscalls = is_allowed;
         trace("%s native-syscalls via shmem %p", is_allowed ? "allowing" : "disallowing",
@@ -146,7 +146,7 @@ bool shim_swapAllowNativeSyscalls(bool new) {
     bool old = *_shim_allowNativeSyscallsFlag();
     *_shim_allowNativeSyscallsFlag() = new;
     if (_using_interpose_ptrace && (new != old)) {
-        _shim_set_allow_native_syscalls(new);
+        _shim_ptrace_set_allow_native_syscalls(new);
     }
     return old;
 }
--- a/src/lib/shim/shim.c
+++ b/src/lib/shim/shim.c
@@ -60,7 +60,7 @@ static ShMemBlock* _shim_shared_mem_blk() {
     static ShimTlsVar v = {0};
     return shimtlsvar_ptr(&v, sizeof(ShMemBlock));
 }
-static ShimSharedMem* _shim_shared_mem() {
+static ShimThreadSharedMem* _shim_shared_mem() {
     if (_shim_shared_mem_blk() == NULL) {
         return NULL;
     }
--- a/src/lib/shim/shim.c
+++ b/src/lib/shim/shim.c
@@ -444,7 +444,6 @@ static void _shim_preload_only_child_ipc_wait_for_start_event() {
 
     shimevent_recvEventFromShadow(ipc, &event, /* spin= */ true);
     assert(event.event_id == SHD_SHIM_EVENT_START);
-    shim_sys_set_simtime_nanos(event.event_data.start.simulation_nanos);
 }
 
 static void _shim_ipc_wait_for_start_event() {
@@ -455,7 +454,6 @@ static void _shim_ipc_wait_for_start_event() {
     trace("waiting for start event on %p", shim_thisThreadEventIPC);
     shimevent_recvEventFromShadow(shim_thisThreadEventIPC(), &event, /* spin= */ true);
     assert(event.event_id == SHD_SHIM_EVENT_START);
-    shim_sys_set_simtime_nanos(event.event_data.start.simulation_nanos);
 }
 
 static void _shim_parent_init_ptrace() {
--- a/src/lib/shim/shim.c
+++ b/src/lib/shim/shim.c
@@ -566,9 +566,9 @@ __attribute__((constructor)) void _shim_load() {
 void shim_ensure_init() { _shim_load(); }
 
 struct timespec* shim_get_shared_time_location() {
-    if (_shim_thread_shared_mem() == NULL) {
+    if (_shim_process_shared_mem() == NULL) {
         return NULL;
     } else {
-        return &_shim_thread_shared_mem()->sim_time;
+        return &_shim_process_shared_mem()->sim_time;
     }
 }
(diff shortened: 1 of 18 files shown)


stevenengler (approved):
It's not clear to me how synchronization is achieved, especially since the IPC locking isn't as straightforward as it used to be, and ptrace mode uses more of the shim than it used to. I think it would be good to expand the  comments to explain where the synchronization comes from in both modes, and more specifically when this shared memory is allowed to be read from and/or written to, both in the shim and in shadow. Is it guaranteed that there will be some kind of memory barrier or synchronization point between any shared memory accesses by the shim or shadow?

sporksmith:
The high level idea is that the memory is safe to access when no thread in the managed process is executing, which we mostly get automatically since whenever a managed thread is executing, the Shadow worker that owns a given process is waiting for it to stop executing. Roughly.

As you point out it gets a bit more subtle with the IPC spin lock, especially since syscalls made in the spin lock machinery (I think just an eventual call to `futex` when it gives up on spinning) go through the shim's syscall machinery, which in turn may indeed access the shared memory e.g. to get the current time for logging.

I'm going to try out adding some synchronization - probably making the individual struct members atomics. I don't think it'll add substantial overhead, but maybe it'll be worth running a performance regression test to be sure.

sporksmith:
Ok, ready for another look.

After changing the shared time to a 64 bit EmulatedTime and poking around in [godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:17,endLineNumber:52,positionColumn:17,positionLineNumber:52,selectionStartColumn:17,selectionStartLineNumber:52,startColumn:17,startLineNumber:52),source:'%23include+%3Cstdatomic.h%3E%0A%23include+%3Cstdint.h%3E%0A%23include+%3Csys/time.h%3E%0A%0Aextern+_Atomic+uint64_t+*x%3B%0Aextern+uint64_t+*y%3B%0A%0Avoid+nonatomic()+%7B%0A++++*y+%3D+1234%3B%0A%7D%0A%0A//+Compared+to+%60nonatomic%60,+this+%22just%22+uses+the+xchg%0A//+instruction+instead+of+a+mov.+I+think+this+does+imply%0A//+some+additional+low-level+locking/sync,+though.%0Avoid+explicit_atomic()+%7B%0A++++atomic_store(x,+1234)%3B%0A%7D%0A%0A//+We+don!'t+actually+need+to+explicitly+use+atomic_store+etc.%0A//+The+compiler+generates+the+same+code+when+dereferencing+normally.%0Avoid+implicit_atomic()+%7B%0A++++*x+%3D+1234%3B%0A%7D%0A%0Aextern+_Atomic+struct+timespec+*tx%3B%0Aextern+struct+timespec+*ty%3B%0A%0Avoid+nonatomic_t()+%7B%0A++++*ty+%3D+(struct+timespec)%7B%0A++++++++.tv_sec+%3D+12,%0A++++++++.tv_nsec+%3D+34,%0A++++%7D%3B%0A%7D%0A%0A//+struct+timespec+doesn!'t+fit+in+64+bits,%0A//+making+the+underlying+atomic+operations+a+bit+more+complex.%0A//+It+probably+wouldn!'t+end+up+mattering+that+much+in+practice%0A//+since+there+should+be+little/no+contention+on+the+atomic,+but+better+to+avoid+all+things+being+equal.%0A//%0A//+This+is+why+I+changed+the+shared++struct+timespec+to+a+64-bit+EmulatedTime.%0Avoid+explicit_atomic_t()+%7B%0A++++struct+timespec+ts+%3D+%7B%0A++++++++.tv_sec+%3D+12,%0A++++++++.tv_nsec+%3D+34,%0A++++%7D%3B%0A++++atomic_store(tx,+ts)%3B%0A%7D%0A%0A//+As+above+we+can+dereference+directly.%0A//+Accessing+individual+fields+through+the+pointer%0A//+is+also+possible,+but+generates+compiler+warnings,%0A//+as+it+should.%0Avoid+implicit_atomic_t()+%7B%0A++++*tx+%3D+(struct+timespec)%7B%0A++++++++.tv_sec+%3D+12,%0A++++++++.tv_nsec+%3D+34,%0A++++%7D%3B%0A%7D%0A%0A'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:51.79881821629181,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:cg112,filters:(b:'0',binary:'1',commentOnly:'1',demangle:'1',directives:'0',execute:'1',intel:'0',libraryCode:'1',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'-O2',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,tree:'1'),l:'5',n:'0',o:'x86-64+gcc+11.2+(C,+Editor+%231,+Compiler+%231)',t:'0')),k:48.20118178370819,l:'4',m:100,n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4), I think this is pretty safe performance-wise. If it's easy to fire off a performance test I guess  we might as well do so and let it run over the weekend. (Do you currently need to do it, or is there a way for me to do it?)

stevenengler:
> If it's easy to fire off a performance test I guess we might as well do so and let it run over the weekend. (Do you currently need to do it, or is there a way for me to do it?)

Anyone can do it from the benchmark repo in the "Actions" tab. I've started two, the first which will benchmark the latest main, and another which will benchmark the PR and plot against main.

sporksmith:
> Anyone can do it from the benchmark repo in the "Actions" tab.

Ah, very nice

> I've started two, the first which will benchmark the latest main, and another which will benchmark this PR and plot against main.

Cool, thanks!

stevenengler:
> Anyone can do it from the benchmark repo in the "Actions" tab. I've started two, the first which will benchmark the latest main, and another which will benchmark this PR and plot against main.

Here's the runtime graph. The [lines on the other graphs](https://github.com/shadow/benchmark-results/tree/master/tor/2022-01-22-T12-51-37) don't exactly match, but I think those are just determinism issues in shadow.

run_time

sporksmith:
Thanks! Agreed that these look fine. LMK if the changes look good

src/main/bindings/rust/wrapper.rs
```diff
@@ -718,6 +718,7 @@ pub struct _Thread {
     _unused: [u8; 0],
 }
 pub type Thread = _Thread;
+pub type atomic_bool = u8;
```
stevenengler, line 721: Maybe it would be best to avoid exporting these types to Rust, since it doesn't look like bindgen handles the atomic correctly.

stevenengler (approved):
Thanks for the comments, and looks good to me.

PR #1881, opened 2022-01-29
Implement basic signal emulation

sporksmith:
Implements most of MS2 from https://github.com/shadow/shadow/discussions/1851.

* Adds an explicit locking mechanism to the shim-shared-memory structures. The "protected" bits of these structures are now protected by a host-wide lock. While we could have used atomics again here, there are many operations that touch multiple parts of this state; it's safer to use a lock to ensure such operations get a consistent view.
* Adds signal state to the thread and process  shim-shared-memory structures, including configured masks, pending signals, etc.
* Adds basic signal handling to preload mode including configuring signal handlers and masks, sending signals to threads and processes, and interrupting blocked syscalls with signals.
* Adds tests for the new signal functionality and enables both the old and new tests for preload mode. I left the old signal test in place to help ensure this doesn't introduce regressions in ptrace mode's signal handling. ptrace mode doesn't pass the new tests (in particular it doesn't support interrupting blocked syscalls).

Notable pieces still missing:
* SA_RESTART: automatically restart interrupted syscalls.
* SA_ONSTACK: run syscall handlers in stack previously provided in `sigaltstack`.

Neither of these are very difficult, but seemed worth leaving out of this already-large-PR, and maybe deferring to MS3.

### DIFF
--- a/src/lib/shim/shim_syscall.c
+++ b/src/lib/shim/shim_syscall.c
@@ -89,9 +89,88 @@ long shim_native_syscall(long n, ...) {
     return rv;
 }
 
-// Only called from asm, so need to tell compiler not to discard.
-__attribute__((used)) static SysCallReg
-_shim_emulated_syscall_event(const ShimEvent* syscall_event) {
+static void _shim_process_signals(ShimShmemHostLock* host_lock) {
+    int signo;
+    siginfo_t siginfo;
+    while ((signo = shimshmem_takePendingUnblockedSignal(
+                host_lock, shim_processSharedMem(), shim_threadSharedMem(), &siginfo)) != 0) {
+        shd_kernel_sigset_t blocked_signals =
+            shimshmem_getBlockedSignals(host_lock, shim_threadSharedMem());
+
+        struct shd_kernel_sigaction action =
+            shimshmem_getSignalAction(host_lock, shim_processSharedMem(), signo);
+
+        if (action.ksa_handler == SIG_IGN) {
+            continue;
+        }
+
+        if (action.ksa_handler == SIG_DFL) {
+            switch (shd_defaultAction(signo)) {
+                case SHD_DEFAULT_ACTION_IGN:
+                    // Ignore
+                    continue;
+                case SHD_DEFAULT_ACTION_CORE:
+                case SHD_DEFAULT_ACTION_TERM: {
+                    // Deliver natively to terminate/drop core.
+                    if (sigaction(signo, &(struct sigaction){.sa_handler = SIG_DFL}, NULL) != 0) {
+                        panic("sigaction: %s", strerror(errno));
+                    }
+                    shimshmemhost_unlock(shim_hostSharedMem(), &host_lock);
+                    raise(signo);
+                    panic("Unreachable");
+                }
+                case SHD_DEFAULT_ACTION_STOP: panic("Stop via signal unimplemented.");
+                case SHD_DEFAULT_ACTION_CONT: panic("Continue via signal unimplemented.");
+            };
+            panic("Unreachable");
+        }
+
+        trace("Handling signo %d", signo);
+
+        shd_kernel_sigset_t handler_mask = shd_sigorset(&blocked_signals, &action.ksa_mask);
+        if (!(action.ksa_flags & SA_NODEFER)) {
+            // Block another instance of the same signal.
+            shd_sigaddset(&handler_mask, signo);
+        }
+        if (action.ksa_flags & SA_RESETHAND) {
+            shimshmem_setSignalAction(host_lock, shim_processSharedMem(), signo,
+                                      &(struct shd_kernel_sigaction){.ksa_handler = SIG_DFL});
+        }
+        if (action.ksa_flags & SA_ONSTACK) {
+            error("SA_ONSTACK unimplemented; ignoring.");
+        }
+        if (action.ksa_flags & SA_RESTART) {
+            error("SA_RESTART unimplemented; ignoring.");
+        }
+
+        // Call signal handler with host lock released, and native syscalls disabled.
+        shimshmemhost_unlock(shim_hostSharedMem(), &host_lock);
+        bool oldAllowNativeSyscalls = shim_swapAllowNativeSyscalls(false);
+        if (action.ksa_flags & SA_SIGINFO) {
+            // ucontext is unimplemented for now. Most signal handlers don't use
+            // it, and passing NULL and getting a loud crash if they try to use
+            // it is probably preferable over passing in a subtly-wrong fake.
+            //
+            // In cases where we intercepted the original syscall via seccomp,
+            // it might be reasonable to pass the ucontext_t that *our* signal
+            // handler received. Otherwise it's much murkier how such a context
+            // ought to be constructed.
+            ucontext_t* ucontext = NULL;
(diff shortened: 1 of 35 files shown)


sporksmith:
Sorry this PR grew into a bit of a monster. I didn't want to merge the new data structures without the code that used those data structure, nor that code without tests to exercise it. I might be able to split this PR into some smaller individual commits along those lines if that'd make things easier for you, though I'm not sure they'd build or make a lot of sense on their own.

sporksmith:
Fixes #1455

src/main/host/syscall_handler.c
```diff
@@ -488,6 +492,21 @@ SysCallReturn syscallhandler_make_syscall(SysCallHandler* sys,
...
+    // on a slow device has already transferred some data by the time it is interrupted by a signal
+    // handler, then the call will return a success  status  (normally,  the  number of bytes
+    // transferred)." from signal(7).
+    if (scr.state == SYSCALL_BLOCK &&
+        thread_unblockedSignalPending(sys->thread, sys->shimShmemHostLock)) {
+        SysCallCondition* condition = scr.cond;
+        utility_assert(condition);
+        syscallcondition_unref(condition);
+        scr = (SysCallReturn){.state = SYSCALL_DONE, .retval = -EINTR};
+    }
```
stevenengler, line 508: To clarify, if thread A is blocked on a syscall, and thread B sends a signal to thread A, shadow immediately interrupts thread A to process the signal? And after processing the signal, thread A's syscall will return EINTR and shadow will discard the blocking syscall condition?

stevenengler (approved):
Cool! I learned a lot about signals reading it :)

Also, I was testing the PR and I ran into the error "Lock is already held. This is probably a deadlock".

Here's the contrived simulation config if you want to look at it:

```yaml
general:
  stop_time: 10s

network:
  graph:
    type: 1_gbit_switch

hosts:
  server:
    network_node_id: 0
    processes:
    - path: /bin/python3
      args: -m http.server 80
      start_time: 3s
    - path: /bin/kill
      args: '1000'
      start_time: 4s
    - path: /bin/true
      start_time: 5s
```

Edit:

I also get a different error with the following config:

```yaml
general:
  stop_time: 10s

network:
  graph:
    type: 1_gbit_switch

hosts:
  server:
    network_node_id: 0
    processes:
    - path: /bin/sleep
      args: '5'
      start_time: 3s
    - path: /bin/kill
      args: '1000'
      start_time: 4s
    - path: /bin/true
      start_time: 5s
```

```
**ERROR ENCOUNTERED**
	At process: 581136 (parent 565393)
	At file: src/main/host/syscall/time.c
	At line: 68
	At function: _syscallhandler_nanosleep_helper
	Message: nanosleep unblocked but a timer is still pending.
**BEGIN BACKTRACE**
Obtained 16 stack frames:
    build/src/main/shadow(+0x128efd5) [0x5634e1752fd5]
    build/src/main/shadow(utility_handleError+0xeb) [0x5634e17531af]
    build/src/main/shadow(+0x12b57ab) [0x5634e17797ab]
    build/src/main/shadow(syscallhandler_nanosleep+0x67) [0x5634e177987e]
    build/src/main/shadow(syscallhandler_make_syscall+0x204a) [0x5634e176cced]
    build/src/main/shadow(threadpreload_resume+0x25e) [0x5634e1743634]
    build/src/main/shadow(thread_resume+0xcb) [0x5634e173e346]
    build/src/main/shadow(process_continue+0x14c) [0x5634e1736638]
    build/src/main/shadow(+0x127e241) [0x5634e1742241]
    build/src/main/shadow(task_execute+0x78) [0x5634e17598fb]
    build/src/main/shadow(event_execute+0x16b) [0x5634e1759384]
    build/src/main/shadow(worker_runEvent+0x30) [0x5634e172f94f]
    build/src/main/shadow(+0x129018d) [0x5634e175418d]
    build/src/main/shadow(+0x126b7fb) [0x5634e172f7fb]
    /lib/x86_64-linux-gnu/libpthread.so.0(+0x9609) [0x7f078dab0609]
    /lib/x86_64-linux-gnu/libc.so.6(clone+0x43) [0x7f078d51c293]
**END BACKTRACE**
**ABORTING**
```

src/main/host/syscall_handler.c
sporksmith: Not quite. I tried to clarify the comment. In this particular scenario:

* thread A is blocked on a syscall
* thread B sends a signal to A, which sets a pending bit on A and schedules it to run
* thread B continues running until it is blocked or exits
* thread A runs, first retrying the original syscall it was blocked on.
* If the syscall completes then its value is allowed to be returned. In the shim, the return code will be saved on the stack, the signal handler will run, and then the syscall's return value is returned. Effectively we're emulating that the signal arrived just after the syscall completed.
* If the syscall is still blocked, then instead of blocking the thread again we return EINTR. As in the other case, the signal handler runs, and then EINTR is returned for the original syscall.

src/lib/shim/shim_syscall.c
```diff
@@ -89,9 +89,88 @@ long shim_native_syscall(long n, ...) {
...
+                    // Deliver natively to terminate/drop core.
+                    if (sigaction(signo, &(struct sigaction){.sa_handler = SIG_DFL}, NULL) != 0) {
+                        panic("sigaction: %s", strerror(errno));
+                    }
+                    raise(signo);
+                    panic("Unreachable");
+                }
+                case SHD_DEFAULT_ACTION_STOP:
+                    panic("Stop via signal unimplemented.") case SHD_DEFAULT_ACTION_CONT
+                        : panic("Continue via signal unimplemented.")
```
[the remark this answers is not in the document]
sporksmith: Whoops! Looks like I forgot the `;` after the `panic` call. Somehow it still compiles, but clang-format did... this... to it. Need  to pay more attention to my clang-format diffs...

sporksmith:
> Also, I was testing the PR and I ran into the error "Lock is already held. This is probably a deadlock".

Ah nice. I failed to release the hostlock before raising a fatal signal natively in the shim-side handler. Fixed: https://github.com/shadow/shadow/pull/1881/commits/ee7dc2b9602c256ca32aeea56e8824a08247bc00

sporksmith:
> I also get a different error with the following config:

I actually do have some fixes for how nanosleep handles getting interrupted. It was originally in this PR but I broke it out to slim this one down. I'll resurrect that one after getting this merged.

We'll probably need to fix up some other syscalls as well; e.g. I think there are others with timeouts that assume the timeout expired if `_syscallhandler_wasBlocked` was  true. 

Opened https://github.com/shadow/shadow/issues/1889

sporksmith:
Btw while debugging your first example, I learned that the python simplehttp server *doesn't* install a handler to gracefully shut down on SIGTERM. It *does* for SIGINT. With this config the process and shadow exit gracefully:

```
general:
  stop_time: 10s

network:
  graph:
    type: 1_gbit_switch

hosts:
  server:
    network_node_id: 0
    processes:
    - path: /bin/python3
      args: -m http.server 80
      start_time: 3s
    - path: /bin/kill
      args: -SIGINT '1000'
      start_time: 4s
    - path: /bin/true
      start_time: 5s
```

It looks like nginx also shuts down on SIGINT (in addition to SIGTERM): http://nginx.org/en/docs/control.html

Maybe SIGINT will be a better default "graceful shutdown" signal, though we'll probably want a way to override it as well.

PR #1886, opened 2022-02-03
Add option to log syscalls

stevenengler:
Adds the experimental "use_strace_logging" option to log syscalls. The syscalls are logged to `hosts/hostname/hostname.procname.pid.strace` files (one for each process). The goals are:

1. strace-inspired formatting of arguments/values
2. easy to log syscalls without much per-syscall boilerplate

The first goal is achieved using an abuse of generics, and the second goal is achieved using a proc macro.

For example, to add logging for a Rust syscall, annotate it with the `log_syscall` macro.

```rust
#[log_syscall(/* rv */ libc::ssize_t, /* fd */ libc::c_int, /* buf */ *const libc::c_void,
              /* count */ libc::size_t)]
pub fn read(ctx: &mut ThreadContext, args: &SysCallArgs) -> SyscallResult { ... }

#[log_syscall(/* rv */ libc::c_int, /* oldfd */ libc::c_int, /* newfd */ libc::c_int,
              /* flags */ nix::fcntl::OFlag)]
pub fn dup3(ctx: &mut ThreadContext, args: &SysCallArgs) -> SyscallResult { ... }
```

Example of the `strace` file output for tor:

```
00:01:00.000 pipe2([-1, -1] (0x5555558d0448), O_CLOEXEC | O_NDELAY | O_NONBLOCK) = 0
00:07:56.000 socket(Inet, 526337, 6) = 20
00:07:56.000 write(1, "Jan 01 00:07:56.000 [notice] Circuit han"..., 89) = 89
00:07:56.000 write(1, "Jan 01 00:07:56.000 [notice] Since start"..., 261) = 261
00:07:56.000 write(1, "Jan 01 00:07:56.000 [notice] Heartbeat: "..., 250) = 250
00:07:56.000 epoll_ctl(...) = 0
00:07:56.100 epoll_wait(...) = 1
00:07:56.100 getsockopt(...) = 0
00:07:56.100 sendto(20, "GET /tor/status-vote/current/consensus/B"..., 266, (empty), 0x0, 0) = 266
00:07:56.100 epoll_ctl(...) = 0
00:07:56.200 epoll_wait(...) = 1
00:07:56.200 recvfrom(20, 0x55555597b30c, 7968, (empty), 0x0, 0x0) = 1167
00:07:56.301 epoll_wait(...) = 1
00:07:56.301 read(16, 0x5555559a5913, 5) = 5
00:07:56.301 read(16, 0x5555559a5918, 4065) = 1429
00:07:56.301 read(16, 0x5555559a5ead, 2636) = -11 (EAGAIN: Try again)
00:07:56.301 close(20) = 0
00:07:56.301 rename(...) = <native>
00:07:56.301 madvise(...) = <native>
00:07:56.301 connect(...) = -115 (EINPROGRESS: Operation now in progress)
```

Limitations:
- syscall handlers implemented in C only show `...` for arguments
- syscalls handled entirely by the shim won't be logged
- there is a performance penalty when using it, which is why it's off by default
- the return values for native syscalls won't be logged


docs/shadow_config_spec.md
```diff
@@ -472,6 +473,13 @@ Type: Bool
 
 Use seccomp to trap syscalls.
 
+#### `experimental.use_strace_logging`
+
+Default: false  
+Type: Bool
+
+Log syscalls for each process.
```
sporksmith, line 481: Probably worth mentioning the file name here

src/lib/syscall-logger/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+/// will become,
+///
+/// ```
+/// pub fn close(ctx: &mut ThreadContext, args: &SysCallArgs) -> SyscallResult {
+///     ...
+///     let rv = close_original(ctx, args);
+///     ...
+///     rv
+/// }
+/// pub fn close_original(ctx: &mut ThreadContext, args: &SysCallArgs) -> SyscallResult {
```
sporksmith, line 26: Should this be non-pub?

src/main/host/process.c
```diff
@@ -500,6 +516,13 @@ static void _process_start(Process* proc) {
     descriptor_ref((LegacyDescriptor*)proc->stderrFile);
     g_free(stderrFileName);
 
+    if (proc->straceLoggingEnabled) {
+        gchar* straceFileName = _process_outputFileName(proc, "strace");
+        proc->straceFd = open(
+            straceFileName, O_CREAT | O_TRUNC | O_WRONLY, S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH);
```
sporksmith, line 522: I *think* O_CLOEXEC would also be preferable here to avoid leaking the file descriptors into the managed processes. I see we're not using it for other files like stdout and stderr though and am trying to think whether there's a reason why.

I think even if we did want the shim to access these files directly, ideally we'd just clear the O_CLOEXEC flag for *that* process's files in between fork and exec, so that every process doesn't get every file descriptor from the managed processes that started before it.

Maybe it'd be better  to change all the files to use O_CLOEXEC in a separate PR though.

src/main/host/syscall/format.rs
```diff
@@ -0,0 +1,520 @@
...
+
+impl TryFromSyscallReg for nix::sys::socket::MsgFlags {
+    fn try_from_reg(reg: SysCallReg) -> Option<Self> {
+        Self::from_bits(reg.into())
+    }
+}
+
+// implement display formatting
+
+simple_display_impl!(i8, i16, i32, i64, isize);
```
sporksmith, line 163: I think we could reduce the macro usage here while keeping the boilerplate low with an extra layer of Traits. IMO this would make the code easier to read  and debug, though maybe I just need to get more comfortable with Rust macros :).

I think it'd go something like:

```
trait SimpleSyscallDataDisplay : Display {};

impl SyscallDataDisplay for SimpleSyscallDataDisplay {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "{}", self)
    }
};

impl SimpleSyscallDataDisplay for i8;
impl SimpleSyscallDataDisplay for i16;
impl SimpleSyscallDataDisplay for i32;
...
```

sporksmith (approved):
Very cool!

docs/shadow_config_spec.md
stevenengler: Added.

src/lib/syscall-logger/src/lib.rs
stevenengler: Good idea, added.

src/main/host/process.c
stevenengler: Changed this file to be CLOEXEC.

src/main/host/syscall/format.rs
stevenengler: I'll merge now, and then if there's a better way to do it I can make a new PR later.

Issue #1889, opened 2022-02-04
Update blocking syscall handlers to handle being interrupted by a signal


sporksmith:
A good starting point is probably auditing usage of `_syscallhandler_wasBlocked` - some handlers assume that if they're running again after being blocked, then it must be the case that either the  original blocking condition is somehow satisfied or a  timeout expired. This is no longer true.

sporksmith:
> A good starting point is probably auditing usage of _syscallhandler_wasBlocked

There are a few places to check:

```
$ git grep _syscallhandler_wasBlocked
src/main/host/syscall/epoll.c:        if (timeout_ms == 0 || _syscallhandler_wasBlocked(sys)) {
src/main/host/syscall/futex.c:    if (!_syscallhandler_wasBlocked(sys) && *futexVal != (uint32_t)expectedVal) {
src/main/host/syscall/futex.c:    if (_syscallhandler_wasBlocked(sys)) {
src/main/host/syscall/poll.c:        if (dont_block || _syscallhandler_wasBlocked(sys)) {
src/main/host/syscall/protected.c:bool _syscallhandler_wasBlocked(const SysCallHandler* sys) { return sys->blockedSyscallNR >= 0; }
src/main/host/syscall/protected.h:bool _syscallhandler_wasBlocked(const SysCallHandler* sys);
src/main/host/syscall/socket.c:        } else if (_syscallhandler_wasBlocked(sys) && errcode == -EISCONN) {
src/main/host/syscall/time.c:    int wasBlocked = _syscallhandler_wasBlocked(sys);
src/main/host/syscall_handler.c:          _syscallhandler_wasBlocked(sys) ? " (previously BLOCKed)" : "");
src/main/host/syscall_handler.c:    if (sys->syscall_counter && !_syscallhandler_wasBlocked(sys)) {
src/main/host/syscall_handler.c:          _syscallhandler_wasBlocked(sys) ? "BLOCK->" : "",
```

I started looking at `epoll`, which I'm pretty sure will currently incorrectly return 0 instead of `EINTR` when interrupted by a signal.

I started writing an `epoll` test case for this, which requires arranging for a signal to be delivered while blocked in `epoll`. I think the only way to do this that's currently supported is to start a thread that sleeps until the designated time, then raises the signal. This is doable, but a bit cumbersome, especially for tests in C.

I'm thinking about implementing something like `alarm` or `setitimer`, which would streamline this a bit. Implementing `setitimer` would additionally address #1958.

[...]

PR #1890, opened 2022-02-04
Fix and test nanosleep signal interruption

sporksmith:
Progress on #1889

### DIFF
--- a/src/lib/libc_preload/syscall_wrappers.c
+++ b/src/lib/libc_preload/syscall_wrappers.c
@@ -1,8 +1,3 @@
-/*
- * The Shadow Simulator
- * See LICENSE for licensing information
- */
-
 /// This file is generated with the 'gen_syscall_wrappers_c.py' script and in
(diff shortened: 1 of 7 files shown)


stevenengler (changes requested):

stevenengler (approved):

src/lib/libc_preload/syscall_wrappers.c
```diff
@@ -240,6 +235,9 @@ INTERPOSE(ftruncate);
 #ifdef SYS_futex // kernel entry: num=202 func=sys_futex
 INTERPOSE(futex);
 #endif
+#ifdef SYS_futex_waitv // kernel entry: num=449 func=sys_futex_waitv
+INTERPOSE(futex_waitv);
+#endif
```
stevenengler, line 240: I don't see a `futex_waitv` libc function, so maybe this shouldn't be included here. (And the same for `set_mempolicy_home_node` below).

PR #1891, opened 2022-02-04
Improved flag handling for File* descriptors, and make their OS fds CLOEXEC

stevenengler:


src/main/host/descriptor/file.c
```diff
@@ -46,24 +48,33 @@ struct _File {
...
+    /* A subset of EMULATED_FLAG_MASK */
+    int emulatedFlags;
     /* Info related to our OS-backed file. */
     struct {
         int fd;
-        int flags;
-        mode_t mode;
-        char* abspath;
+        /* A combination of access mode, file creation, and file status flags. */
+        int originalFlags;
```
sporksmith, line 57: If these represent the flags, mode, and path of the native object, i think `native` would be a more helpful prefix. It's a term we already use elsewhere and emphasizes the contrast with emulatedFlags, etc.

I'm also not sure these prefixes are necessary here, since the field being in the `osfile` or not already imply whether they're the native or emulated versions. I'm ok with or without them, though.

src/main/host/descriptor/file.h
```diff
@@ -19,6 +19,9 @@
 #include "main/core/support/definitions.h"
 #include "main/host/syscall/kernel_types.h"
 
+/* Shadow emulates only specific flags. */
```
sporksmith, line 22: I'm not quite following this comment. Is this the set of flags that we emulate in Shadow instead of just passing them through to the native file?

src/main/host/descriptor/file.h
```diff
@@ -50,10 +53,13 @@ int file_openat(File* file, File* dir, const char* pathname, int flags, mode_t m
...
 
 /* Returns the flags that were used to open the file. */
-int file_getFlags(File* file);
+int file_getOriginalFlags(File* file);
 
 /* Returns the mode that was used to open the file. */
-mode_t file_getMode(File* file);
+mode_t file_getOriginalMode(File* file);
+
+/* Get only the flags that shadow emulates. Will be a subset of EMULATED_FLAG_MASK. */
```
sporksmith, line 61: This comment is also a little unclear, though maybe expanding the earlier comment to describe what you mean by 'emulate' will help address it.

src/main/host/syscall/mman.c
```diff
@@ -149,13 +149,33 @@ static int _syscallhandler_openPluginFile(SysCallHandler* sys, File* file) {
...
+    /* Get original flags that were used to open the file. */
+    int flags = file_getOriginalFlags(file);
+    /* Use only the file creation flags, except O_CLOEXEC. */
+    flags &= (creationFlags & ~O_CLOEXEC);
+    /* Add any file access mode and file status flags. */
+    flags |= fcntl(file_getOSBackedFD(file), F_GETFL);
+    /* Be careful not to try re-creating or truncating it. */
+    flags &= ~(O_CREAT|O_EXCL|O_TMPFILE|O_TRUNC);
+
+    /* Add the O_CLOEXEC creation flag if it's set for the descriptor. */
```
sporksmith, line 169: I think this currently doesn't matter, since we have the managed thread close the file again after the mmap is complete. Maybe just as well to have to have it though in case e.g. we later decide to keep the file open.

Could probably use a comment explaining that rationale (or yours, if different) for doing it, though.

src/main/host/descriptor/file.c
```diff
@@ -286,6 +297,15 @@ int file_openat(File* file, File* dir, const char* pathname, int flags, mode_t m
         file->type = FILE_TYPE_REGULAR;
     }
 
+    int originalFlags = flags;
+
+    // move any emulated flags from 'flags' to 'emulatedFlags'
```
sporksmith, line 302: I think this is part of my confusion. I thought that `emulatedFlags` were the flags set on the file descriptor from the managed process's point of view, and that `originalFlags` was the set that Shadow had set on the native file descriptor. IIUC though with this change, the set of flags from the managed process's point of view is actually `originalFlags | emulatedFlags`.

This could be addressed with some explanation in the struct definition and/or header, but I think the way I *thought* it worked would actually be clearer. Is there a reason not do it that way?

sporksmith (changes requested):
Cool, definitely a bigger change than I anticipated when I mentioned using CLOEXEC, but good to do it :)

src/main/host/descriptor/file.c
stevenengler: Shadow doesn't track the file flags for the `File*` objects. It only stores the original flags that were used when the file was first opened. If the plugin uses `fcntl` to change the file flags, Shadow calls `fcntl` directly on the native file descriptor and the stored flags are no longer up to date. I renamed `flags` to `originalFlags` to try to make it clearer that these flags are historical and shouldn't be used as the file's actual flags.

The idea of `emulatedFlags` is separate from this. Rather than passing every flag from the plugin to the native file descriptor as Shadow does now, there are some flags that shouldn't be passed through (for example `O_CLOEXEC`). The `emulatedFlags` field was meant to keep track of these flags that don't get passed through from the plugin to the native file descriptor.

I agree that I should improve the documentation and naming here.

> IIUC though with this change, the set of flags from the managed process's point of view is actually originalFlags | emulatedFlags.

The flags from the managed process' point of view is `(fcntl(nativeFd, FD_GETFL) & EMULATED_FLAG_MASK) | emulatedFlags`. Shadow doesn't know anything about the actual flags.

src/main/host/descriptor/file.c
stevenengler: These represent the flags, mode, and path of the native object, but only at time of creation. They're not kept up to date as the plugin changes the file flags, mode, and path change, which is why I added the "original" prefix.

src/main/host/descriptor/file.c
```diff
@@ -46,24 +48,33 @@ struct _File {
     /* File is a sub-type of a descriptor. */
     LegacyDescriptor super;
     FileType type;
+    /* A subset of EMULATED_FLAG_MASK */
+    int emulatedFlags;
```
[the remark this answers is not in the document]
stevenengler: Expanded the comment.

src/main/host/descriptor/file.h
stevenengler: Right, they're the flags the are handled within Shadow and aren't passed to the native file. Expanded the comment.

src/main/host/descriptor/file.h
stevenengler: Expanded the comment.

src/main/host/syscall/mman.c
stevenengler: My goal was just to open the file with the same flags as what the plugin had set, since before it only used the flags that the file had when it was opened and not the current flags. I changed this and added a comment.

sporksmith (approved):
Much clearer now; thanks!

PR #1892, opened 2022-02-07
Add ubuntu 21.10 to test matrix

sporksmith:


sporksmith:
@stevenengler This is a pretty surprising failure: https://github.com/shadow/shadow/runs/5100489526?check_suite_focus=true#step:8:269

The test is writing until it would be forced to block, then reading until it would be forced to block. In the failure, not as much is read as was written.

I wonder if because this is a socket, it still goes through the network stack even though both ends are on localhost, and some data could still be on the sender side when the reader has finished reading everything...? If so maybe the loop should keep trying until `n_written` have been read (or the test times out), even if there are some `EWOULDBLOCK` errors...?

stevenengler (approved):

sporksmith:
> Also, I think we should only test 21.10 in the CI until 22.04 is released, and then we can switch to testing 22.04 in the CI instead.

21.10 is scheduled to be EOL in July 2022, so yeah we won't support it for long :). Probably makes sense to support the latest Ubuntu, though.

sporksmith:
@robgjansen could you make the ubuntu:21.10 tests required?

stevenengler:
> @robgjansen could you make the ubuntu:21.10 tests required?

@robgjansen And also fedora:35

robgjansen:
> > @robgjansen could you make the ubuntu:21.10 tests required?
> 
> @robgjansen And also fedora:35

Done and done!

PR #1896, opened 2022-02-08
Add Fedora 35 to supported releases and CI tests

stevenengler:
Replaces #1734. Also update the EOL dates for Ubuntu to show the standard EOL dates rather than the enterprise / paid support EOL dates.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -40,6 +40,7 @@ jobs:
           - 'debian:10-slim'
           - 'debian:11-slim'
           - 'fedora:34'
+          - 'fedora:35'
           # EOL May 2024 https://www.centos.org/centos-stream/
           - 'quay.io/centos/centos:stream8'
         cc: ['gcc', 'clang']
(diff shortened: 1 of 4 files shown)


.github/workflows/run_tests.yml
```diff
@@ -30,16 +30,17 @@ jobs:
     strategy:
       matrix:
         container:
-          # EOL April 2028 https://wiki.ubuntu.com/Releases
+          # EOL April 2023 https://wiki.ubuntu.com/Releases
```
sporksmith, line 33: It's not clear to me whether we should stop supporting at "end of life" or "end of standard support", but I suppose the earlier "end of standard support" date is a good time to reconsider whether it's worth continuing to support.

sporksmith, line 33: "EOL" -> "End of Standard Support" ?

sporksmith (approved):

.github/workflows/run_tests.yml
stevenengler: I think most people using Shadow won't be subscribed to Ubuntu's support program, and it would be a pain to test that version anyways since we'd have to somehow subscribe our CI to the program (I don't think we'd be eligible for a free subscription anyways). So I don't think we should support longer than the standard support date.

stevenengler: I'll update this.

PR #1903, opened 2022-02-10
Print fewer non-graphic characters in strace string

stevenengler:
If it's printing a lot of non-graphic bytes, then it's probably not a string. This PR limits the number of non-graphic bytes shown.

Example, rather than printing 40 copies of `\x01`, it now only prints 12.

```
00:00:01.030 [tid 1000] sendto(3, "\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01"..., 65508, (empty), 0x0, 0) = -90 (EMSGSIZE: Message too long)
```

### DIFF
--- a/src/main/host/syscall/format.rs
+++ b/src/main/host/syscall/format.rs
@@ -193,16 +193,28 @@ impl SyscallPtrDisplay for SyscallPtr<*const i8> {
         // much at once (for example bytes past the null byte), we may try to read from an invalid
         // address
 
-        // get bytes until either null byte or a read error
+        // to avoid printing too many escaped bytes, limit the number of non-graphic and non-ascii
+        // characters
+        let mut non_graphic_remaining = DISPLAY_LEN / 3;
+
         let mut s: Vec<NonZeroU8> = reader
             .bytes()
+            // get bytes until either null byte or a read error
             .map_while(|x| x.ok().and_then(|y| NonZeroU8::new(y)))
+            // stop after a certain number of non-graphic characters
+            .map_while(|x| {
+                if !x.get().is_ascii_graphic() {
+                    non_graphic_remaining = non_graphic_remaining.saturating_sub(1);
+                }
+                (non_graphic_remaining > 0).then(|| x)
+            })
             .collect();
+
         let len = s.len();
         s.truncate(DISPLAY_LEN);
         let s: std::ffi::CString = s.into();
 
-        if len > DISPLAY_LEN {
+        if len > DISPLAY_LEN || non_graphic_remaining <= 0 {
             write!(f, "{:?}...", s)
         } else {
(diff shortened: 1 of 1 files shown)


src/main/host/syscall/format.rs
```diff
@@ -193,16 +193,28 @@ impl SyscallPtrDisplay for SyscallPtr<*const i8> {
         // much at once (for example bytes past the null byte), we may try to read from an invalid
```
sporksmith, line 193: Orthogonal to this PR, but [memory_ref_prefix](https://shadow.github.io/docs/rust/shadow_rs/host/memory_manager/struct.MemoryManager.html#method.memory_ref_prefix) or [copy_prefix_from_ptr](https://shadow.github.io/docs/rust/shadow_rs/host/memory_manager/struct.MemoryManager.html#method.copy_prefix_from_ptr) might be useful here.

src/main/host/syscall/format.rs
```diff
@@ -193,16 +193,28 @@ impl SyscallPtrDisplay for SyscallPtr<*const i8> {
...
         let mut s: Vec<NonZeroU8> = reader
             .bytes()
+            // get bytes until either null byte or a read error
             .map_while(|x| x.ok().and_then(|y| NonZeroU8::new(y)))
+            // stop after a certain number of non-graphic characters
+            .map_while(|x| {
+                if !x.get().is_ascii_graphic() {
+                    non_graphic_remaining = non_graphic_remaining.saturating_sub(1);
+                }
+                (non_graphic_remaining > 0).then(|| x)
```
sporksmith, line 209: [TIL](https://doc.rust-lang.org/std/primitive.bool.html#method.then)

sporksmith (approved):

src/main/host/syscall/format.rs
sporksmith: > will the entire function return an error, or will it return something with access to bytes 0-30?

It's meant to do the latter.

`readv_ptrs` calls `readv_iovecs`, which calls `process_vm_readv`. `process_vm_readv` supports partial reads.

stevenengler: Okay I think I see, so `memory_ref_prefix` will return an error if it could not copy any memory, otherwise it will return the memory read up until a page boundary where the copy read fails.

src/main/host/syscall/format.rs
stevenengler: Yeah I don't know if this is too weird, but I think it's readable. (And it looks like `bool::then_some()` will be stabilized soon.)

src/main/host/syscall/format.rs
stevenengler: I'll make a new PR for this.

PR #1905, opened 2022-02-10
Improve error message when parsing yaml process arguments

stevenengler:
Example config:

```yaml
hosts:
  testnode:
    network_node_id: 0
    processes:
    - path: /bin/sleep
      args: 4
```

Old error:

```
Could not parse configuration file "/tmp/shadow.yaml"

Caused by:
    hosts.testnode.processes[0]: data did not match any variant of untagged enum ProcessArgs at line 5 column 11
```

New error:

```
Could not parse configuration file "/tmp/shadow.yaml"

Caused by:
    hosts.testnode.processes[0]: invalid type: integer `4`, expected a string or a sequence of strings at line 5 column 11
```

### DIFF
--- a/src/main/core/support/configuration.rs
+++ b/src/main/core/support/configuration.rs
@@ -713,11 +713,53 @@ impl std::ops::Deref for Quantity {
     }
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[derive(Debug, Clone, Serialize, JsonSchema)]
 #[serde(untagged)]
 pub enum ProcessArgs {
     List(Vec<String>),
     Str(String),
+}
+
+/// Serde doesn't provide good deserialization error messages for untagged enums, so we implement
+/// our own. For example, if serde finds a yaml value such as 4 for the process arguments, it won't
+/// deserialize it to the string "4" and the yaml parsing will fail. The serde-generated error
+/// message will say something like "data did not match any variant of untagged enum ProcessArgs at
+/// line X column Y" which isn't very helpful to the user, so here we try to give a better error
+/// message.
+impl<'de> serde::Deserialize<'de> for ProcessArgs {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        struct ProcessArgsVisitor;
+
+        impl<'de> serde::de::Visitor<'de> for ProcessArgsVisitor {
+            type Value = ProcessArgs;
+
+            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+                formatter.write_str("a string or a sequence of strings")
+            }
+
+            fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+            where
+                E: serde::de::Error,
+            {
+                Ok(Self::Value::Str(v.to_owned()))
+            }
+
+            fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+            where
+                A: serde::de::SeqAccess<'de>,
+            {
+                let mut v = vec![];
+
+                while let Some(val) = seq.next_element()? {
+                    v.push(val);
+                }
+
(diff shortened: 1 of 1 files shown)


src/main/core/support/configuration.rs
```diff
@@ -707,13 +707,58 @@ impl std::ops::Deref for Quantity {
...
+            }
+            serde_yaml::Value::Number(n) if n.is_f64() => {
+                serde::de::Unexpected::Float(n.as_f64().unwrap())
+            }
+            serde_yaml::Value::Number(_) => {
+                unimplemented!("Did serde_yaml add a new number type?")
+            }
+            serde_yaml::Value::String(ref s) => serde::de::Unexpected::Str(s),
+            serde_yaml::Value::Sequence(_) => serde::de::Unexpected::Seq,
+            serde_yaml::Value::Mapping(_) => serde::de::Unexpected::Map,
```
sporksmith, line 752: I *think* this looks fine, but I'm confused how we get here for anything other than a `Mapping`; any other valid yaml value should be able to be parsed as a `String` or `Vec<String>`, shouldn't it? e.g. in your example, shouldn't the value `4` be successfully parsed as a string "4"?  Or do bare integers need to be quoted to be considered a valid string?

sporksmith (approved):

src/main/core/support/configuration.rs
stevenengler: With Serde you get one chance to deserialize something. If you try to deserialize to a `Vec<String>` and it's unsuccessful, you can't retry again with a `String`. To get this functionality, you need to first deserialize to an intermediate type like [`serde_yaml::Value`](https://docs.serde.rs/serde_yaml/enum.Value.html). Since this type implements `Clone`, you can keep cloning it and trying to deserialize it to different types. But once you've deserialized to `serde_yaml::Value`, you've lost the original representation. For example if you deserialize to a `serde_yaml::Value::Bool`, there's no 1:1 correlation with a `String`. Should `serde_yaml::Value::Bool(true)` be `"true"`, `"TRUE"`, or `"True"` (which are all valid yaml booleans)?

stevenengler:
@sporksmith I changed the deserialization method. It should have the same behaviour as the previous commit, but I think it's simpler and more standard. Sorry for making you review it again :)

src/main/core/support/configuration.rs
```diff
@@ -707,13 +707,55 @@ impl std::ops::Deref for Quantity {
...
+        struct ProcessArgsVisitor;
+
+        impl<'de> serde::de::Visitor<'de> for ProcessArgsVisitor {
+            type Value = ProcessArgs;
+
+            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+                formatter.write_str("a string or a sequence of strings")
+            }
+
+            fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
```
sporksmith, line 734: Possibly orthogonal to this PR, but would it be simpler to force `ProcessArgs` to always be a list? e.g. a single string value would become a list of size 1 ...?

sporksmith (approved):

src/main/core/support/configuration.rs
stevenengler: The main reason we support a list is to make it easier to distinguish between argv values as passed during exec. If we put the entire string in a single list item, then we would call the plugin with an argv of length 2 (process name and single argument). Right now, we use `parse_string_as_args()` later to split the string into a list following the usual shell parsing rules. But we could add the `parse_string_as_args()` call directly into this `Deserialize` impl.

sporksmith: Ah I see. Yeah, moving the string-splitting here might make sense, but up to you.

I'd also be a little tempted to only support a list in the yaml to avoid having our own implementation of shell parsing rules, but it'd definitely be a bit more verbose in the config files, and maybe not worth changing...

sporksmith: Sounds good. No strong opinion on whether it's better to split here or later - totally up to you

Issue #1549, opened 2021-07-30
snowflake simulation broken


stevenengler:
Edited to top-post blocking issues:

- [x] #1948
- [x] #1968
- [x] #1974
- [x] #1975

---

When running the following simulation, the simulation completes successfully in ptrace mode, but fails in preload mode.

```
(cd go-server && go build main.go)
rm -rf shadow.data && build/src/main/shadow --log-level info --interpose-method preload go-minimal.yaml
```

```
00:00:00.003525 [worker-1] 00:00:03.000000000 [INFO] [server:11.0.0.4] [process.c:497] [_process_start] starting process 'server.main.1000'
00:00:00.003603 [worker-1] 00:00:03.000000000 [INFO] [server:11.0.0.4] [thread_preload.c:199] [threadpreload_run] forking new thread with environment 'SHADOW_SPAWNED=TRUE SHADOW_LOG_START_TIME=837172211482 SHADOW_INTERPOSE_METHOD=PRELOAD <snip>'
00:00:00.003668 [worker-1] 00:00:03.000000000 [INFO] [server:11.0.0.4] [process.c:519] [_process_start] process 'server.main.1000' started
00:00:00.007286 [worker-1] 00:00:03.000000000 [WARN] [server:11.0.0.4] [syscall_handler.c:503] [syscallhandler_make_syscall] Detected unsupported syscall 204 called from thread 1000 in process server.main.1000 on host server
00:00:00.007287 [worker-1] 00:00:03.000000000 [ERROR] [server:11.0.0.4] [syscall_handler.c:507] [syscallhandler_make_syscall] Returning error 38 (ENOSYS) for unsupported syscall 204, which may result in unusual behavior
signal 6 received on thread with no signal stack
fatal error: non-Go code disabled sigaltstack
...
```

I'm not sure under what conditions they abort, but if it works in ptrace mode, we probably want it to also work in preload mode.

-------

The following is a slightly modified version of https://github.com/shadow/shadow/issues/1540#issuecomment-887044728 (uses 3 clients rather than 1).

```go
package main

import (
	"net/http"
)

type SimpleHandler struct{}

func (h *SimpleHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	data := []byte("Hi, I'm a simple web server\n")
	w.Write(data)
}

func main() {
	handler := &SimpleHandler{}
	http.ListenAndServe(":80", handler)
}
```

```yaml
general:
  stop_time: 10s

network:
  graph:
[...]

PR #1911, opened 2022-02-14
Implement sigaltstack

sporksmith:
The golang runtime uses sigaltstack, and detects when its handlers aren't running on the configured stack.

Helps with some of the errors in #1549 .

### DIFF
--- a/src/lib/shim/shadow_signals.h
+++ b/src/lib/shim/shadow_signals.h
@@ -17,6 +17,13 @@
 #define SHD_SIGRT_MIN 32
 #define SHD_SIGRT_MAX 64
 
+// Definition is sometimes missing in the userspace headers. We could include
+// the kernel signal header, but it has definitions that conflict with the
+// userspace headers.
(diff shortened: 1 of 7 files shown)


src/test/file/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 include_directories(${GLIB_INCLUDES})
 link_libraries(${GLIB_LIBRARIES})
 add_executable(test-file test_file.c)
+add_compile_options(test-file )
```
stevenengler, line 4: What's this line for? Is it to clear some default options?

stevenengler (approved):

src/test/file/CMakeLists.txt
sporksmith: Oops, good catch. I think that must have been a stale unsaved buffer from when I was fixing the file test. (I'd started to add compiler flags to ignore warnings, but switched to using pragmas instead)

PR #1919, opened 2022-02-18
Refactor rust shared memory cleanup

robgjansen:


src/main/shmem/cleanup.rs
```diff
@@ -192,7 +139,7 @@ mod tests {
         let expired = Path::new(s);
 
         touch(&expired).unwrap();
-        try_shm_cleanup();
+        assert!(try_shm_cleanup().is_ok());
```
stevenengler, line 142: `try_shm_cleanup().unwrap()` is nicer since it will print the error message.

stevenengler (approved):

PR #1920, opened 2022-02-18
If strace string logging pointer is invalid, print address

stevenengler:
When the buffer address is invalid, print the address rather than an empty string.

Old:
```
sendto(3, "", 0, (empty), 0x0, 0) = 0
```

New:
```
sendto(3, 0x1, 0, (empty), 0x0, 0) = 0
```

### DIFF
--- a/src/main/host/syscall/format.rs
+++ b/src/main/host/syscall/format.rs
@@ -180,35 +180,33 @@ simple_debug_impl!(nix::sys::socket::MsgFlags);
 
 impl SyscallPtrDisplay for SyscallPtr<*const i8> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>, mem: &MemoryManager) -> std::fmt::Result {
-        if self.ptr.is_null() {
-            return write!(f, "{:p}", self.ptr);
-        }
         const DISPLAY_LEN: usize = 40;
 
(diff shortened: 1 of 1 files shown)


stevenengler:
> Maybe _always_ print the pointer value? I think that's what strace does.

Oops, the auto-merge merged it before I could do anything :)

On my computer strace doesn't print the pointer, but I think the strace format is different depending on what strace version you use. This would be easy to add though.

```
$ strace curl google.com
...
sendto(5, "GET / HTTP/1.1\r\nHost: google.com"..., 74, MSG_NOSIGNAL, NULL, 0)
```

> Btw I noticed this sometimes prints past the end of the actual buffer, especially e.g. in golang, which doesn't use null-terminated strings. I suppose it'd be a bit more complicated to fix it though, since we have to get the length from a different syscall parameter.

Yeah it doesn't work nicely for non-C-style strings or byte buffers. I'm not sure of a nice way to improve this though without making it too complicated.

sporksmith:
> Oops, the auto-merge merged it before I could do anything :)

Oops, yeah I forgot to check for auto-merge :)

> On my computer strace doesn't print the pointer, but I think the strace format is different depending on what strace version you use. This would be easy to add though.

Maybe I misremembered the strace behavior. I'm fine with either way

PR #1928, opened 2022-02-22
Fix shm cleanup tests

stevenengler:
~Fixes a race condition by making the unit tests run single-threaded, and makes sure that there are no pre-existing shm files.~

Fixes a race condition by making the unit tests use temporary directories.

### DIFF
--- a/src/main/Cargo.lock
+++ b/src/main/Cargo.lock
@@ -264,6 +264,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e78d4f1cc4ae33bbfc157ed5d5a5ef3bc29227303d595861deb238fcec4e9457"
 
 [[package]]
+name = "fastrand"
+version = "1.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c3fcf0cee53519c866c09b5de1f6c56ff9d647101f81c1964fa632e148896cdf"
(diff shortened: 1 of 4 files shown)


sporksmith (changes requested):
It's a little unfortunate to force all the rust unit tests to be single-threaded, but probably doesn't make much difference in practice since we don't have that many.

Having the unit  tests mess with the real contents of `/dev/shm` seems fairly undesirable though. e.g. it looks like tests would still fail if there was a still-live shadow simulation running in the background (granted it's a bit unlikely).

It'd probably be better if we added a way to override the root path from `/dev/shm`, and have the tests each create and use their own unique temp dir so that they don't interfere with eachother. Wdyt?

stevenengler:
> It's a little unfortunate to force all the rust unit tests to be single-threaded, but probably doesn't make much difference in practice since we don't have that many.
> 
> Having the unit tests mess with the real contents of `/dev/shm` seems fairly undesirable though. e.g. it looks like tests would still fail if there was a still-live shadow simulation running in the background (granted it's a bit unlikely).
> 
> It'd probably be better if we added a way to override the root path from `/dev/shm`, and have the tests each create and use their own unique temp dir so that they don't interfere with eachother. Wdyt?

Changed the PR to use a temporary directory instead.

sporksmith (approved):

PR #1935, opened 2022-02-24
Cmake affinity

sporksmith:
This currently doesn't work. Waiting to hear back from cmake support https://discourse.cmake.org/t/property-processor-affinity-has-no-effect/5106


sporksmith:
Doesn't seem likely the cmake issue will be fixed soon

PR #1936, opened 2022-02-25
Improve behaviour when TCP_NODELAY is set

stevenengler:
Shadow doesn't support nagle's algorithm, so shadow always behaves as if `TCP_NODELAY` is enabled. Some programs will fail if `setsockopt(fd, SOL_TCP, TCP_NODELAY, &1, sizeof(int))` returns an error, so we treat this as a no-op for compatibility.

Three examples I can think of that fail when they're unable to set `TCP_NODELAY` are nginx (unless using `tcp_nodelay off`), python's urllib library, and salticidae-based programs.

### DIFF
--- a/src/main/host/syscall/socket.c
+++ b/src/main/host/syscall/socket.c
@@ -301,6 +301,17 @@ static int _syscallhandler_getTCPOptHelper(SysCallHandler* sys, TCP* tcp, int op
 
             return 0;
         }
+        case TCP_NODELAY: {
+            /* Shadow doesn't support nagle's algorithm, so shadow always behaves
+             * as if TCP_NODELAY is enabled.
+             */
+            int val = 1;
+            int num_bytes = MIN(*optlen, sizeof(int));
+            memcpy(optval, &val, num_bytes);
+            *optlen = num_bytes;
+
+            return 0;
+        }
         default: {
             warning(
                 "getsockopt at level SOL_TCP called with unsupported option %i",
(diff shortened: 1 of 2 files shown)


src/main/host/syscall/socket.c
```diff
@@ -1145,6 +1156,38 @@ SysCallReturn syscallhandler_setsockopt(SysCallHandler* sys,
...
+                int enable = 0;
+                errcode = process_readPtr(sys->process, &enable, optvalPtr, sizeof(int));
+                if (errcode == 0) {
+                    if (enable) {
+                        // wants to enable TCP_NODELAY
+                        debug("Ignoring TCP_NODELAY");
+                    } else {
+                        // wants to disable TCP_NODELAY
+                        warning("Cannot disable TCP_NODELAY");
+                        errcode = -ENOPROTOOPT;
```
robgjansen, line 1184: Do we want this to be an error? Or should we just print the warning but then pretend like it worked so we don't cause other programs to fail?

src/main/host/syscall/socket.c
```diff
@@ -1145,6 +1156,38 @@ SysCallReturn syscallhandler_setsockopt(SysCallHandler* sys,
...
+
+                int enable = 0;
+                errcode = process_readPtr(sys->process, &enable, optvalPtr, sizeof(int));
+                if (errcode == 0) {
+                    if (enable) {
+                        // wants to enable TCP_NODELAY
+                        debug("Ignoring TCP_NODELAY");
+                    } else {
+                        // wants to disable TCP_NODELAY
+                        warning("Cannot disable TCP_NODELAY");
```
robgjansen, line 1183: I would add a few more details to the warning message, like "Shadow does not implement Nagle's algorithm so setting `TCP_NODELAY` to 0 will have no effect."

robgjansen (approved):

src/main/host/syscall/socket.c
stevenengler: > Do we want this to be an error?

Do you mean to have shadow exit with an error message? Or just bump this from a warning message to error message? The other syscalls that return a linux error only print warning-level log messages.

> Or should we just print the warning but then pretend like it worked so we don't cause other programs to fail?

I haven't seen any programs try to disable `TCP_NODELAY` (probably because it's disabled by default on Linux). For enabling `TCP_NODELAY` I think a no-op is the correct behaviour in shadow, but for disabling `TCP_NODELAY` I think a no-op would be incorrect behaviour, and we should wait to see if any applications actually fail when `setsockopt` returns an error and then decide if we should add the incorrect behaviour.

robgjansen: OK, you convinced me. I'm fine with keeping as-is. I think it does make more sense to return an error when they are _explicitly_ trying to disable in a way we don't support. We can always keep a look out if returning an error here causes any apps to fail.

src/main/host/syscall/socket.c
stevenengler: Added more context to the warning message.

PR #1938, opened 2022-02-26
Organize rust syscall handler in a module

robgjansen:
Creates a new rust syscall handler to replace the current C handler. Reduces a bit of boilerplate when calling rust syscalls from the current C handler because we don't have to export every rust syscall to C (at the cost of an extra rust match statement). Generally is a bit better organized and will make it easier to transition more syscalls over to rust.


stevenengler:
I'm happy to see the duplicate rust syscall handler wrappers gone and replaced by a match :)

An issue with this design is that it will cause issues as we move the `Thread` object to Rust. The main issue is that the way it's written now, the `SyscallHandler` is a member of the `Thread` object, but we're also passing the `Thread` object to every syscall handler as part of the `ThreadContext` object. If this was written entirely in Rust, this would require mutably borrowing the thread object twice when calling a syscall handler, which wouldn't be allowed. If we want the syscall handler functions to be written as methods of the `SyscallHandler` struct, then the `SyscallHandler` will have to live outside of the `Thread` object at the time that the syscall handler is run.

I think there are 3 ways of dealing with this:

1. The thread object holds an `Option<SyscallHandler>`. Before each syscall handler runs, the thread object calls `syscall_handler_rs.take()` to remove the `SyscallHandler` object from the `Thread` object (and `syscall_handler_rs` becomes `None`). When the syscall handler finishes, the thread object re-adds the `SyscallHandler` object back to the thread object.
2. The syscall handlers are stored outside of the thread object. For example we could have a `HashMap<(host_id, proc_id, thread_id), SyscallHandler>` which stores all of the `SyscallHandler` objects.
3. The thread object continues to hold the `SyscallHandler`, but we don't implement the syscall handler functions as methods on the `SyscallHandler` type. When a syscall handler function runs, it can access the `SyscallHandler` data using `ctx.thread.syscall_handler_rs`.

My order of preference would be 3, 1, 2. Maybe @sporksmith has thoughts about this as well.

robgjansen:
The `SyscallHandler` contains a set of methods that share state and helpers and to me it makes more sense to structure this as a module with the `struct`+`impl`+`method` pattern. I do wonder whether we'll end up at your option 2 eventually anyway, since the `Host`, `Process`, and `Thread` classes are so important to the rest of the codebase.

sporksmith:
I think any of the proposed approaches would work out ok, and it's mostly a matter of taste.

We *could* still get encapsulation with approach 3 - IIUC as long as the handlers are in sub-modules of the module that defines the `SyscallHandler` type, they'd have access to its private members.

OTOH it does seem a little unfortunate for the handlers themselves to have to know how to get the SyscallHandler out of a Thread, and for many (most? all?) of them to have to do that.

I'd lean towards keeping them as methods of SyscallHandler for now, and we can address the issue with 1) or 2) once the calling code is in Rust. It can either extract the handler from the Thread (approach 1, and which is the current plan for how we're going down the call hierarchy Host -> Process -> Thread), or get it from a ~Host-owned object store (approach 2).

src/main/host/syscall/handler/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+        }
+        unsafe {
+            Box::from_raw(handler_ptr);
+        }
+    }
+
+    #[no_mangle]
+    pub extern "C" fn rustsyscallhandler_syscall(
+        sys: *mut SyscallHandler,
+        csys: *mut c::SysCallHandler,
```
sporksmith, line 76: It'd be a little nicer to have Thread store a pointer to (Rust) `SyscallHandler`, and have that store the pointer to `c::SysCallHandler`, but also fine to leave that for a future MR.

sporksmith (approved):

src/main/host/syscall_handler.c
```diff
@@ -229,12 +233,11 @@ static void _syscallhandler_post_syscall(SysCallHandler* sys, long number,
...
-#define HANDLE_RUST(s)                                                         \
-    case SYS_##s:                                                              \
-        _syscallhandler_pre_syscall(sys, args->number, #s);                    \
-        scr = rustsyscallhandler_##s(sys, args);                               \
-        _syscallhandler_post_syscall(sys, args->number, #s, &scr);             \
+#define HANDLE_RUST(s)                                                                             \
+    case SYS_##s:                                                                                  \
+        _syscallhandler_pre_syscall(sys, args->number, #s);                                        \
+        scr = rustsyscallhandler_syscall(sys->syscall_handler_rs, sys, args);                      \
+        _syscallhandler_post_syscall(sys, args->number, #s, &scr);                                 \
```
stevenengler, line 240: This should change to the following so that we don't accidentally use `syscall_handler_rs` twice:

```c
{
    _syscallhandler_pre_syscall(sys, args->number, #s);
    SyscallHandler* handler = sys->syscall_handler_rs;
    sys->syscall_handler_rs = NULL;
    scr = rustsyscallhandler_syscall(handler, sys, args);
    sys->syscall_handler_rs = handler;
    _syscallhandler_post_syscall(sys, args->number, #s, &scr); 
}
```

(and below for `SYS_fcntl64`)

src/main/host/syscall/handler/mod.rs
```diff
@@ -0,0 +1,85 @@
+use crate::cshadow as c;
+use crate::host::context::{ThreadContext, ThreadContextObjs};
+use crate::host::syscall_types::SysCallArgs;
+use crate::host::syscall_types::{SyscallError, SyscallResult};
```
stevenengler, line 4: What's the reason for making a new `handler` module? What files/types will go in the `syscall` module and not the `syscall::handler` module?

Also I think the existing functions in the syscall module (`src/main/host/syscall/mod.rs`) should also move into this `handler` module since they're helpers specifically for the handler functions.

stevenengler (approved):
> OTOH it does seem a little unfortunate for the handlers themselves to have to know how to get the SyscallHandler out of a Thread, and for many (most? all?) of them to have to do that.

I don't think most handlers will use the `SyscallHandler` data. It only currently contains the Epoll and ShimShmemHostLock objects, blockedSyscallNR, and some performance counters. The `Host*`. `Process*` and `Thread*` pointers won't be able to move over to the Rust version.

src/main/host/syscall/handler/mod.rs
robgjansen: > What's the reason for making a new handler module? What files/types will go in the syscall module and not the syscall::handler module?

I was thinking we would place the syscall-related types and functions that are not specific to the handler itself in `src/main/host/syscall` and then isolate the handlers `impl`s in `src/main/host/syscall/handler`. So, for example, `syscall_condition`, `syscall_types`, and maybe `kernel_types` would go in `src/main/host/syscall`.

> Also I think the existing functions in the syscall module (src/main/host/syscall/mod.rs) should also move into this handler module since they're helpers specifically for the handler functions.

OK, thanks! I'm also planning to continue iterating on organization etc. as I move more handlers to Rust but this PR was already getting pretty big.

src/main/host/syscall/handler/mod.rs
robgjansen: 👍  I'll save the re-org for a future PR.

src/main/host/syscall_handler.c
robgjansen: Done!

PR #1940, opened 2022-03-01
Support retransmission of SYN packets

stevenengler:
Since Shadow doesn't give SYN packets a sequence number, they aren't retransmitted if the packet gets dropped. This PR gives them a sequence number. This is required to support finite-sized syn/accept queues in the listener.

As an example, here are the pcaps for the following shadow simulation of a client connecting to a server. In both cases, `_tcp_processPacket()` was modified to drop all SYN packets when in the `TCPS_LISTEN` state:

```c
case TCPS_LISTEN: {
    /* receive SYN, send SYNACK, move to SYNRECEIVED */
    if(header->flags & PTCP_SYN) {
        MAGIC_ASSERT(tcp->server);
        flags |= TCP_PF_PROCESSED;

        packet_addDeliveryStatus(packet, PDS_RCV_SOCKET_DROPPED);
        return;
```

Without PR (packet is not retransmitted):
1646103355_grim

With PR:
1646103411_grim

```yaml
general:
  stop_time: 120s

network:
  graph:
    type: 1_gbit_switch

hosts:
  server:
    network_node_id: 0
    processes:
    - path: /bin/python3
      args: -m http.server 80
      start_time: 3s
  client:
    network_node_id: 0
    processes:
    - path: /bin/curl
      args: -s server
      start_time: 5s
```

### DIFF
--- a/src/main/host/descriptor/tcp.c
+++ b/src/main/host/descriptor/tcp.c
@@ -843,9 +843,10 @@ static Packet* _tcp_createPacketWithoutPayload(TCP* tcp, Host* host, enum Protoc
     _tcp_updateReceiveWindow(tcp);
 
     /* control packets have no sequence number
-     * (except FIN, so we close after sending everything) */
+     * (except SYN and FIN, so we close after sending everything) */
+    /* TODO: all FIN packets (including FIN,ACK) should increment the sequence number */
     gboolean isFinNotAck = ((flags & PTCP_FIN) && !(flags & PTCP_ACK));
-    guint sequence = !isEmpty || isFinNotAck ? tcp->send.next : 0;
+    guint sequence = !isEmpty || isFinNotAck || (flags & PTCP_SYN) ? tcp->send.next : 0;
 
     /* create the TCP packet. the ack, window, and timestamps will be set in _tcp_flush */
     Packet* packet = packet_new(host);
@@ -1177,7 +1178,7 @@ void tcp_networkInterfaceIsAboutToSendPacket(TCP* tcp, Host* host, Packet* packe
         tcp->send.delayedACKCounter = 0;
     }
 
-    if(header->sequence > 0 || (header->flags & PTCP_SYN)) {
+    if(header->sequence > 0) {
         /* store in retransmission buffer */
         _tcp_addRetransmit(tcp, packet);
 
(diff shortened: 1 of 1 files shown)


robgjansen (approved):
I'm happy to merge this as long as we can verify with a Tor benchmark test that traffic levels etc. are consistent with the prior version.

stevenengler:
@robgjansen The change in network behaviour seems to be that TCP connections where the application does a large write at the beginning before the handshake takes place, there may be 2 packets worth of data ready to send by the time that the handshake completes. In current Shadow, only one packet gets sent after the handshake and the second isn't sent until the first data packet is ACKed. With this patch, the first two data packets are both sent at the same time. I believe this is why we see shorter transfer times for tgen, since we're essentially skipping one round-trip. I can replicate the old behaviour where Shadow sends a single data packet by initializing `tcp->send.unacked` to 2 instead of 1.

The issue is that I can't figure out why the current Shadow is sending only a single data packet. The initial `tcp->send.window` is set to 10, so I think we should be able to initially send more than a single packet? Do you have any ideas here? Does the new behaviour where Shadow sends multiple packets at the same time sound okay to you?

### Current Shadow:

1651800782_grim

### With PR:

1651800766_grim

stevenengler:
This PR should wait until #2130. I don't think this PR should cause any further performance changes after that one.

robgjansen:
You solved this in #2130, right? (I.e., once #2130 is merged, then this PR will no longer have results we don't understand.)

stevenengler:
> You solved this in #2130, right? (I.e., once #2130 is merged, then this PR will no longer have results we don't understand.)

Right, #2130 changed `tcp->send.unacked` from 1 to 0 since the first SYN packet has a sequence number of 0 (that PR caused the network results to change). This PR will add retransmission and change `tcp->send.unacked` back from 0 to 1 since the first SYN packet will now have a sequence number of 1 (this PR should not cause the network results to change).

I'll run one last benchmark before merging this PR to make sure the performance doesn't change from anything in this PR.

stevenengler:
https://github.com/shadow/benchmark-results/tree/master/tor/2022-05-12-T11-13-06

There are still differences in the graphs, for example the top 5% of 1 MiB transfers are faster while the 5 MiB transfers are slower, but I'm not sure if this is really a network effect rather than just the simulations diverging.

In the tor minimal test, both PRs generate identical strace files and pcap files (except that this PR has sequence numbers one greater than main), even when increasing the number of tor clients from 1 to 15. My guess is that larger simulations like the benchmarking simulation diverge once the first SYN packet is retransmitted.

transfer_time_5242880 exit

transfer_time_1048576 exit

1652383403_grim

robgjansen:
Retransmitting SYN packets when they are dropped is only more accurate than letting the SYN get dropped and then never retransmitting, independent of any second-order effects that might occur in application layer protocols (e.g., Tor). So I think we should be confident to merge this PR.

PR #1944, opened 2022-03-01
Enable simple-http golang test

sporksmith:


src/test/golang/simple_http.yaml
```diff
@@ -15,6 +15,9 @@ hosts:
     network_node_id: 0
     quantity: 3
     processes:
-    - path: /bin/curl
-      args: -s server
+    # Would be more idiomatic to use curl, but that fails on debian 10 due to:
+    # - Different path to curl (/bin/curl vs /usr/bin/curl)
+    # - Deadlock in this version of curl. https://github.com/shadow/shadow/issues/1794
+    - path: /usr/bin/python3
+      args: -c "import urllib.request; urllib.request.urlopen('http://server').read()"
```
stevenengler, line 22: nit: maybe instead use `import urllib.request; print(urllib.request.urlopen('http://server').read().decode('utf-8'))` so that it prints the request to stdout (just makes it easier to manually verify that it's working correctly).

stevenengler (approved):

src/test/golang/simple_http.yaml
sporksmith: Thanks, done

PR #1949, opened 2022-03-03
Fixed crash when attempting to capture UDP packets in pcap

stevenengler:
Logging on every UDP packet when pcap logging is enabled might be overkill, but on the other hand we should expect debug logging to generate a lot of noise.

### DIFF
--- a/src/main/host/network_interface.c
+++ b/src/main/host/network_interface.c
@@ -339,6 +339,12 @@ void networkinterface_disassociate(NetworkInterface* interface, const CompatSock
 }
 
 static void _networkinterface_capturePacket(NetworkInterface* interface, Packet* packet) {
+    if (packet_getProtocol(packet) != PTCP) {
+        /* we don't support capturing UDP packets */
+        debug("Network interface skipped capturing a UDP packet");
+        return;
+    }
+
     PCapPacket* pcapPacket = g_new0(PCapPacket, 1);
 
     pcapPacket->headerSize = packet_getHeaderSize(packet);


src/main/host/network_interface.c
```diff
@@ -339,6 +339,12 @@ void networkinterface_disassociate(NetworkInterface* interface, const CompatSock
 }
 
 static void _networkinterface_capturePacket(NetworkInterface* interface, Packet* packet) {
+    if (packet_getProtocol(packet) != PTCP) {
+        /* we don't support capturing UDP packets */
+        debug("Network interface skipped capturing a UDP packet");
+        return;
+    }
+
```
robgjansen, line 347: Don't we want something like this instead? Then we don't have to spam the log with the debug messages for apps that use UDP.

```diff
--- a/src/main/host/network_interface.c
+++ b/src/main/host/network_interface.c
@@ -349,22 +349,26 @@ static void _networkinterface_capturePacket(NetworkInterface* interface, Packet*
         packet_copyPayloadShadow(packet, 0, pcapPacket->payload, pcapPacket->payloadLength);
     }
 
-    PacketTCPHeader* tcpHeader = packet_getTCPHeader(packet);
-
-    pcapPacket->srcIP = tcpHeader->sourceIP;
-    pcapPacket->dstIP = tcpHeader->destinationIP;
-    pcapPacket->srcPort = tcpHeader->sourcePort;
-    pcapPacket->dstPort = tcpHeader->destinationPort;
-
-    if(tcpHeader->flags & PTCP_RST) pcapPacket->rstFlag = TRUE;
-    if(tcpHeader->flags & PTCP_SYN) pcapPacket->synFlag = TRUE;
-    if(tcpHeader->flags & PTCP_ACK) pcapPacket->ackFlag = TRUE;
-    if(tcpHeader->flags & PTCP_FIN) pcapPacket->finFlag = TRUE;
-
-    pcapPacket->seq = (guint32)tcpHeader->sequence;
-    pcapPacket->win = (guint16)tcpHeader->window;
-    if(tcpHeader->flags & PTCP_ACK) {
-        pcapPacket->ack = (guint32)tcpHeader->acknowledgment;
+    pcapPacket->srcIP = packet_getSourceIP(packet);
+    pcapPacket->dstIP = packet_getDestinationIP(packet);
+    pcapPacket->srcPort = packet_getSourcePort(packet);
+    pcapPacket->dstPort = packet_getDestinationPort(packet);
+
+    if(packet_getProtocol(packet) == PTCP) {
+        PacketTCPHeader* tcpHeader = packet_getTCPHeader(packet);
+
+        if(tcpHeader) {
+            if(tcpHeader->flags & PTCP_RST) pcapPacket->rstFlag = TRUE;
+            if(tcpHeader->flags & PTCP_SYN) pcapPacket->synFlag = TRUE;
+            if(tcpHeader->flags & PTCP_ACK) pcapPacket->ackFlag = TRUE;
+            if(tcpHeader->flags & PTCP_FIN) pcapPacket->finFlag = TRUE;
+
+            pcapPacket->seq = (guint32)tcpHeader->sequence;
+            pcapPacket->win = (guint16)tcpHeader->window;
+            if(tcpHeader->flags & PTCP_ACK) {
+                pcapPacket->ack = (guint32)tcpHeader->acknowledgment;
+            }
+        }
     }
 
     pcapwriter_writePacket(interface->pcap, pcapPacket);
```

robgjansen (changes requested):

src/main/host/network_interface.c
stevenengler: Wouldn't the pcap writer still write this packet as if it was a TCP packet? For example, it would write a TCP header for a UDP packet, which we wouldn't want. I think this needs to be changed in both the `PCapPacket` and `PCapWriter`. I've started rewriting the pcap writer in rust, which I think should make this easier.

PR #1952, opened 2022-03-04
Add rust handler for sysinfo

robgjansen:


src/main/host/syscall/handler/sysinfo.rs
```diff
@@ -0,0 +1,53 @@
...
+
+impl SyscallHandler {
+    // Using libc::c_void here because the libc::sysinfo struct is a return value only.
+    #[log_syscall(/* rv */ libc::c_int, /* info */ *const libc::sysinfo)]
+    pub fn sysinfo(&self, ctx: &mut ThreadContext, args: &SysCallArgs) -> SyscallResult {
+        // Pointer to the plugin memory where we write the result.
+        let info_ptr = TypedPluginPtr::new::<libc::sysinfo>(PluginPtr::from(args.get(0)), 1);
+        if info_ptr.is_null() {
+            return Err(Errno::EFAULT.into());
+        }
```
stevenengler, line 19: This isn't needed since the `copy_to_ptr()` below will return `EFAULT`.

src/main/host/syscall/handler/sysinfo.rs
```diff
@@ -0,0 +1,53 @@
...
+
+        // Seconds are needed for uptime.
+        let time = Worker::current_time().unwrap_or(emulated_time::SIMULATION_START);
```
stevenengler, line 22: I think it would be better to `unwrap()` since if the worker is not initialized, then something bad has happened and we probably want to catch it early rather than continue and return the wrong value.

src/main/host/syscall/handler/sysinfo.rs
```diff
@@ -0,0 +1,53 @@
...
+use crate::host::syscall::handler::SyscallHandler;
+use crate::host::syscall_types::{PluginPtr, SysCallArgs, SyscallResult, TypedPluginPtr};
+use crate::utility::pod;
+
+use nix::errno::Errno;
+use syscall_logger::log_syscall;
+
+impl SyscallHandler {
+    // Using libc::c_void here because the libc::sysinfo struct is a return value only.
+    #[log_syscall(/* rv */ libc::c_int, /* info */ *const libc::sysinfo)]
```
stevenengler, line 13: Remove comment?

src/main/host/syscall/handler/sysinfo.rs
```diff
@@ -0,0 +1,53 @@
...
+            .as_secs();
+
+        // Get a zeroed struct to make sure we init all fields.
+        let mut info = pod::zeroed::<libc::sysinfo>();
+
+        // These values are chosen arbitrarily; we don't think it matters too
+        // much, except to maintain determinism. For example, Tor make decisions
+        // about how many circuits to allow to be open (and other OOM settings)
+        // based on available memory.
+        info.uptime = i64::try_from(seconds).unwrap_or(0);
```
stevenengler, line 34: Maybe change the `unwrap_or(0)` to `unwrap_or(i64::MAX)`?

src/test/sysinfo/test_sysinfo.rs
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+fn main() {
+    let mut info = unsafe { std::mem::zeroed::<libc::sysinfo>() };
+    let info_ptr = &mut info as *mut libc::sysinfo;
+    let rv = unsafe { libc::sysinfo(info_ptr) };
+
+    println!("Found return value {:?}.", rv);
+    assert_eq!(rv, 0);
+    println!("Found uptime {:?}.", rv);
```
stevenengler, line 13: Should this be `info.uptime` instead of `rv`?

stevenengler (approved):

src/main/host/syscall/handler/sysinfo.rs
robgjansen: I was going to us `i64::MAX` but wasn't sure if that would be the only reason `try_from` would fail. But I think it's the most likely reason so I think it makes sense.

src/test/sysinfo/test_sysinfo.rs
robgjansen: Oops good catch!!

src/main/host/syscall/handler/sysinfo.rs
robgjansen: Maybe we don't want `Worker::current_time()` to return `Option` then, if we're going to always unwrap it anyway?

stevenengler: I'm not sure what the original reason for returning an `Option` was, but since the `Worker` is a global that can be accessed from anywhere within Shadow, I think it makes sense to keep it an `Option` since there might be cases where we don't know if the global worker is set or not and we want to handle both cases. For example the logger gets the current time, and there might not be a worker active at the time. In this case, it's specifically an error if the worker isn't set during a syscall handler.

PR #1962, opened 2022-03-10
Rearranged a few syscall helper functions

stevenengler:
`get_descriptor()` and `get_descriptor_mut()` - these tied the lifetime of the descriptor refs to the lifetime of the syscall handler ref, but we don't need this restriction.

`empty_sockaddr()`, `write_sockaddr()`, and `read_sockaddr()` - these are standalone helpers that should be independent of a syscall handler object.

Also simplified a few socket plugin memory helpers (only needed the process memory, not the process itself).

### DIFF
--- a/src/main/host/syscall/handler/fcntl.rs
+++ b/src/main/host/syscall/handler/fcntl.rs
@@ -16,7 +16,7 @@ impl SyscallHandler {
         let cmd: i32 = args.args[1].into();
 
         // get the descriptor, or return early if it doesn't exist
-        let desc = match self.get_descriptor_mut(ctx.process, fd)? {
+        let desc = match Self::get_descriptor_mut(ctx.process, fd)? {
             CompatDescriptor::New(d) => d,
             // if it's a legacy descriptor, use the C syscall handler instead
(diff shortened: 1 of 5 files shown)


robgjansen (approved):
My thinking for making the helpers part of the handler object was that they might become useful when migrating other syscalls to rust, but I suppose the functions can be called whether or not they are defined inside the `impl` block so 🤷

stevenengler:
> My thinking for making the helpers part of the handler object was that they might become useful when migrating other syscalls to rust, but I suppose the functions can be called whether or not they are defined inside the `impl` block so shrug

One downside of leaving them as part of the handler is that you can't access them from outside of a handler. For example I plan on using `empty_sockaddr()` from outside of the syscall handler code, so I need to move it anyways in order to call it.

In general my reason for moving them was mostly following the ["How Non-Member Functions Improve Encapsulation"](https://www.drdobbs.com/cpp/how-non-member-functions-improve-encapsu/184401197) advise, but I'm not hard-set on it either way. Maybe making them associated functions instead would meet the same goal.

PR #1971, opened 2022-03-17
Fix a bug when converting to a nix 'SockAddr'

stevenengler:
The nix documentation for [`sockaddr_storage_to_addr()`](https://docs.rs/nix/latest/nix/sys/socket/fn.sockaddr_storage_to_addr.html) says "[The length] must be *at least as large* as all the useful parts of the structure" (emphasis added), but for unix pathname sockets, it can panic if the length is not exactly equal to `offsetof(addr.path) + strlen(addr.path)` (or that +1). This PR adds a workaround for that issue.

### DIFF
--- a/src/main/host/syscall/handler/socket.rs
+++ b/src/main/host/syscall/handler/socket.rs
@@ -665,7 +665,218 @@ fn read_sockaddr(
         )?;
     }
 
+    // apply a nix bug workaround that may shorten the addr length
+    let corrected_addr_len = sockaddr_storage_len_workaround(&addr, addr_len);
+    assert!(corrected_addr_len <= addr_len);
+
     Ok(Some(nix::sys::socket::sockaddr_storage_to_addr(
-        &addr, addr_len,
+        &addr,
+        corrected_addr_len,
     )?))
+}
+
+/// Workaround for a nix bug. It's unclear if the bug is that the documentation for
+/// `sockaddr_storage_to_addr()` is incorrect, or if the implemention of
+/// `sockaddr_storage_to_addr()` is incorrect.
+///
+/// https://github.com/nix-rust/nix/issues/1680
+///
+/// At the time of writing, the nix main branch has changed the structure of nix's sockaddr types
+/// and calling `sockaddr_storage_to_addr()` no longer panics. But the documentation is still
+/// inconsistent, so it's unclear if the "working" behaviour in nix is still considered a bug.
+///
+/// A function called from within 'sockaddr_storage_to_addr()' assumes that the given length is the
+/// exact length of the address, whereas the documentation for 'sockaddr_storage_to_addr()' only
+/// specifies "[The len] must be at least as large as all the useful parts of the structure".  For
+/// example if a unix address containing 'sockaddr_un { family: AF_UNIX, path: b"test\0" }' is given
+/// to 'sockaddr_storage_to_addr()' with a length of 'sizeof(sockaddr_un)', it may panic due to an
+/// out-of-bounds read.
+fn sockaddr_storage_len_workaround(
+    addr: &nix::sys::socket::sockaddr_storage,
+    addr_len: usize,
+) -> usize {
+    // if not an AF_UNIX sockaddr, no change is required
+    if addr_len < std::mem::size_of_val(&addr.ss_family) || addr.ss_family != libc::AF_UNIX as u16 {
+        return addr_len;
+    }
+
+    // copy the bytes from the 'sockaddr_storage' into a 'sockaddr_un'
+    let mut tmp_addr: libc::sockaddr_un = unsafe { std::mem::zeroed() };
+    unsafe {
+        std::ptr::copy(
+            &addr as *const _ as *const u8,
+            &mut tmp_addr as *mut _ as *mut u8,
+            std::mem::size_of_val(&tmp_addr),
+        )
+    };
+
+    // find the position of the first null byte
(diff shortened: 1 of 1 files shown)


src/main/host/syscall/handler/socket.rs
```diff
@@ -665,7 +665,164 @@ fn read_sockaddr(
...
+    // 'sizeof(sockaddr_un)', it may panic due to an out-of-bounds read. This also applies to
+    // abstract addresses.
+    if addr.ss_family == libc::AF_UNIX as u16 {
+        // copy the bytes from the 'sockaddr_storage' into a 'sockaddr_un'
+        let mut tmp_addr: libc::sockaddr_un = unsafe { std::mem::zeroed() };
+        unsafe {
+            std::ptr::copy(
+                &addr as *const _ as *const u8,
+                &mut tmp_addr as *mut _ as *mut u8,
+                std::mem::size_of_val(&tmp_addr),
```
sporksmith, line 689: Should this be `addr_len`? (Or maybe should there be a check somewhere that `addr_len >= std::mem::size_of_val(&tmp_addr)`?

In particular, a provided `addr_len` that's too small probably ought to result in EINVAL even if reading past that length gives a valid address...?

And maybe providing a pointer to something smaller than `sizeof(struct sockaddr_un)` should succeed, as long as the first `addr_len` bytes contains a valid address? (I'm less sure about this one)

src/main/host/syscall/handler/socket.rs
```diff
@@ -665,7 +665,164 @@ fn read_sockaddr(
...
+
+    nix::sys::socket::sockaddr_storage_to_addr(addr, addr_len)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_unix_sockaddr_to_nix_1() {
```
sporksmith, line 717: Could these names be a little more descriptive about what property their testing? e.g. replace the _1, _2, and _3 suffixes with something about what distinguishes these from eachother?

sporksmith (approved):

src/main/host/syscall/handler/socket.rs
stevenengler: > Should this be `addr_len`? (Or maybe should there be a check somewhere that `addr_len >= std::mem::size_of_val(&tmp_addr)`?

I'm not sure what you mean here. We could use `std::cmp::min(addr_len, std::mem::size_of_val(&tmp_addr))`, but I don't think it hurts to copy some extra bytes since we're guaranteed that they're initialized. And we're guaranteed that the size of `sockaddr_storage` should always be large enough to fit any of the sockaddr types (ex: `sockaddr_in`, `sockaddr_un`, etc).

> In particular, a provided addr_len that's too small probably ought to result in EINVAL even if reading past that length gives a valid address...?

In general I want to leave this type of error handling to the nix `sockaddr_storage_to_addr()` function, but I agree we should make sure it's at least 2 bytes before we read the `ss_family`,

> And maybe providing a pointer to something smaller than sizeof(struct sockaddr_un) should succeed, as long as the first addr_len bytes contains a valid address? (I'm less sure about this one)

I'm not sure what you mean here. We're copying from a `sockaddr_storage` struct which is guaranteed to have ~118 bytes.

sporksmith: Yeah, sorry, I was confused. I think the concerns I had here are addressed in the caller.

IMO the API would be a little clearer if `addr_len` were returned instead of mutated - I didn't catch before that it's output only. Up to you, though.

stevenengler: > IMO the API would be a little clearer if addr_len were returned instead of mutated - I didn't catch before that it's output only. Up to you, though.

It's only mutated within the function and the caller doesn't see the change (`mut addr_len: usize` vs `&mut addr_len: usize`), is that what you mean?

sporksmith: Oh wait, it's *not* output-only.

Maybe this could be clarified a bit by documenting and/or `assert`ing what properties are expected to hold (that should have already been ensured by the caller)

sporksmith: Oof, I missed that too.

Sometimes I put the `mut` in the body of the function in cases like this since it's not really part of the API, though it does mean a little boilerplate. Up to you, though.
```
fn sockaddr_storage_to_nix(
    addr: &nix::sys::socket::sockaddr_storage,
    addr_len: usize,
) -> Result<nix::sys::socket::SockAddr, nix::Error> {
    let mut addr_len = addr_len;
```

stevenengler: > Yeah, sorry I keep getting confused here :).
> 
> I think what you have looks correct, but it'd be nice if it were a little easier to follow.
> 
> I still think moving the `mut` out of the function signature would help a little bit.
> 
> _Maybe_ it'd be an improvement to make this a helper that _just_ finds and returns the "right" `addr_len`, and letting the caller still call `nix::sys::socket::sockaddr_storage_to_addr` itself would be a bit clearer, but I'm not sure.

I've made this into a function that just tweaks the length and returns a new length.

> It'd be nice if you could file an issue and/or send a PR to fix this in nix :) (and perhaps add a reference in the comment to help track when this is no longer needed)

I had made an issue, but the behaviour is different between the latest nix master branch and the latest nix release.

https://github.com/nix-rust/nix/issues/1680

PR #1988, opened 2022-03-24
When resuming blocked syscall handlers, use a cached/saved file reference

stevenengler:
When a syscall handler restarts after being blocked, it will look up the fd in the descriptor table again. But if the fd was closed while the syscall handler was blocked, this lookup will fail and the blocked syscall handler will return `EBADF`. Instead, the reference-counted file object will be saved/cached in the syscall condition object so that it can be used when resuming the syscall handler. This better models Linux's behaviour.

### DIFF
--- a/src/main/host/syscall/handler/socket.rs
+++ b/src/main/host/syscall/handler/socket.rs
@@ -143,23 +143,35 @@ impl SyscallHandler {
         let addr_ptr: PluginPtr = args.get(4).into();
         let addr_len: libc::socklen_t = args.get(5).into();
 
-        // get the descriptor, or return early if it doesn't exist
-        let desc = match Self::get_descriptor(ctx.process, fd)? {
-            CompatDescriptor::New(desc) => desc,
-            // if it's a legacy descriptor, use the C syscall handler instead
-            CompatDescriptor::Legacy(_) => {
-                return unsafe {
-                    c::syscallhandler_sendto(
-                        ctx.thread.csyscallhandler(),
-                        args as *const c::SysCallArgs,
-                    )
-                    .into()
+        // if we were previously blocked, get the active file from the last syscall handler
+        // invocation since it may no longer exist in the descriptor table
+        let file = ctx
+            .thread
+            .syscall_condition()
+            // if this was for a C descriptor, then there won't be an active file object
+            .map(|x| x.active_file().cloned())
+            .flatten();
+
+        let file = match file {
+            // we were previously blocked, so re-use the file from the previous syscall invocation
+            Some(x) => x,
+            // get the file from the descriptor table, or return early if it doesn't exist
+            None => match Self::get_descriptor(ctx.process, fd)? {
+                CompatDescriptor::New(desc) => desc.open_file().clone(),
+                // if it's a legacy descriptor, use the C syscall handler instead
+                CompatDescriptor::Legacy(_) => {
+                    return unsafe {
+                        c::syscallhandler_sendto(
+                            ctx.thread.csyscallhandler(),
+                            args as *const SysCallArgs,
+                        )
+                        .into()
+                    };
                 }
-            }
+            },
         };
 
-        let open_file = desc.open_file().clone();
-        self.sendto_helper(ctx, open_file, buf_ptr, buf_len, flags, addr_ptr, addr_len)
+        self.sendto_helper(ctx, file, buf_ptr, buf_len, flags, addr_ptr, addr_len)
     }
 
     pub fn sendto_helper(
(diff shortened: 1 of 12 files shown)


src/main/host/syscall/handler/socket.rs
```diff
@@ -234,31 +248,35 @@ impl SyscallHandler {
...
+        // if we were previously blocked, get the active file from the last syscall handler
+        // invocation since it may no longer exist in the descriptor table
+        let file = ctx
+            .thread
+            .syscall_condition()
+            // if this was for a C descriptor, then there won't be an active file object
+            .map(|x| x.active_file().cloned())
+            .flatten();
+
+        let file = match file {
```
sporksmith, line 260: optional nit - maybe slightly more compact and idiomatic to do something like:

```
let file = file.unwrap_or_else(|| {
...
});
```

src/main/host/syscall/handler/unistd.rs
```diff
@@ -172,18 +172,35 @@ impl SyscallHandler {
...
+        // if we were previously blocked, get the active file from the last syscall handler
+        // invocation since it may no longer exist in the descriptor table
+        let file = ctx
+            .thread
+            .syscall_condition()
+            // if this was for a C descriptor, then there won't be an active file object
+            .map(|x| x.active_file().cloned())
+            .flatten();
+
+        let file = match file {
```
sporksmith, line 184: Maybe `unwrap_or_else`

src/main/host/syscall/handler/unistd.rs
```diff
@@ -194,21 +211,35 @@ impl SyscallHandler {
...
```
sporksmith, line 223: Maybe `unwrap_or_else`

src/main/host/syscall/handler/unistd.rs
```diff
@@ -269,18 +302,35 @@ impl SyscallHandler {
...
```
sporksmith, line 314: Maybe `unwrap_or_else`

src/main/host/syscall/handler/unistd.rs
```diff
@@ -291,18 +341,35 @@ impl SyscallHandler {
...
+            // we were previously blocked, so re-use the file from the previous syscall invocation
```
sporksmith, line 354: Maybe `unwrap_or_else`

src/main/host/syscall_condition.rs
```diff
@@ -1,43 +1,129 @@
...
+    _phantom: PhantomData<&'a ()>,
+}
+
+// do not define any mutable methods for this type
+impl<'a> SysCallConditionRef<'a> {
+    /// Borrows from a C pointer. i.e. doesn't increase the ref count, nor decrease the ref count
+    /// when dropped.
+    /// `ptr` must point to a valid object that will not be accessed by other threads
+    /// for the lifetime of this object.
+    pub fn borrow_from_c(ptr: *mut cshadow::SysCallCondition) -> Self {
```
sporksmith, line 20: This seems like ought to be `unsafe`, since violating the contract above would be UB

src/main/host/syscall_condition.rs
```diff
@@ -1,43 +1,129 @@
...
+    condition: SysCallConditionRef<'a>,
+}
+
+// any immutable methods should be implemented on SysCallConditionRef instead
+impl<'a> SysCallConditionRefMut<'a> {
+    /// Borrows from a C pointer. i.e. doesn't increase the ref count, nor decrease the ref count
+    /// when dropped.
+    /// `ptr` must point to a valid object that will not be accessed by other threads
+    /// for the lifetime of this object.
+    pub fn borrow_from_c(ptr: *mut cshadow::SysCallCondition) -> Self {
```
sporksmith, line 50: `unsafe`?

sporksmith (changes requested):

src/main/host/syscall/handler/socket.rs
stevenengler: The issue with this is that there's a short-circuit path (`return unsafe { c::syscallhandler_sendto(...) }`) if it's a legacy descriptor, so if you put this in a closure, then you only short-circuit the closure and not the syscall handler.

sporksmith: Ah, gotcha. Nevermind then

src/main/host/syscall_condition.rs
stevenengler: Made these functions `unsafe`. I left the `Thread::syscall_condition()` and `Thread::syscall_condition_mut()` functions as safe since I think if we have a `Thread` object, then we can already assume that their internal syscall condition isn't being modified from other threads.

sporksmith (approved):

src/main/host/syscall_condition.rs
sporksmith: :+1:

Issue #1792, opened 2021-11-30
Support spinning / spinlocks / busy-loops


sporksmith:
This is a tracking issue for supporting "spin" waiting with `sched_yield`. Today it doesn't work at all - executing a `sched_yield` is effectively a no-op inside of a Shadow simulation. If it's used in a loop that's waiting for another thread on the same host to do something (like release a spin lock), it'll never happen.

*Typically* such spin loops are waiting for a condition that will be satisfied by another thread running on the same host - e.g. releasing a spin lock. It's also possible to end up in a spin loop that's doing non-blocking IO calls, but that's usually a bug (such as https://github.com/smol-rs/async-io/pull/73)

## Spinlock fallback types

Spinlocks often have some type of fallback to a less aggressive waiting mechanism if the condition they're waiting for isn't met for a while.

* No fallback: Without some fallback, currently under Shadow A spinlock that has no never falls back to something less aggressive currently deadlocks, since other threads never get a chance to run.

* Wallclock fallback: A spinlock that spins for some maximum wallclock duration also currently deadlocks, since `sched_yield` neither moves time forward nor gives other threads a chance to run.

* Iteration fallback: A spinlock that spins some maximum number of *iterations* will eventually work. Arguably a "good" spinlock implementation will have such a limit.

## Potential solutions

* Reschedule thread: Implement `sched_yield` to allow other ready threads to run, but not move time forward: This would minimize the change to "simulation-visible" behavior, and could allow Shadow to handle spin-loops where another thread is ready to satisfy the condition (e.g. do some processing and release a spinlock). This is a little tricky though - it'd require adding some way of scheduling a "run this thread again" event at the same sim-time, but after other alread-scheduled events.

* Nanosleep: Implement `sched_yield` as effectively a `nanosleep` of `n` ns. With a very small value (e.g. 1 ns) this would solve at least as many spin-loops as "reschedule thread", but would be much easier to implement. It'd slightly change simulation results, but since even the cheapest syscalls take ~[20 ns](https://stackoverflow.com/questions/23599074/system-calls-overhead), `n` up to that amount would only make the simulation *more* accurate (well, until that minimum syscall latency gets smaller on future systems). This could also fix some spin loops that are stuck waiting for a wallclock fallback, but with a very small `n` could take many iterations to exit the loop. 
 
[...]

PR #1989, opened 2022-03-24
ShimSharedMem locking refactor

sporksmith:
Hold the HostShimSharedMemLock in a wider scope on the Shadow side. This is in preparation to add and use more protected state, which will be accessed outside of the SyscallHandler.

This should be roughly the same amount of locking and unlocking as before in practice.

Progress on #1792

### DIFF
--- a/src/main/host/thread_preload.c
+++ b/src/main/host/thread_preload.c
@@ -53,6 +53,14 @@ static ThreadPreload* _threadToThreadPreload(Thread* thread) {
 
 static Thread* _threadPreloadToThread(ThreadPreload* thread) { return (Thread*)thread; }
 
+static void _threadpreload_continuePlugin(ThreadPreload* thread, const ShimEvent* event) {
+    // We're about to let managed thread execute, so need to release the shared memory lock.
+    // Reacquired in _threadpreload_waitForNextEvent.
+    host_unlockShimShmemLock(thread->base.host);
+
+    shimevent_sendEventToPlugin(thread->ipc_data, event);
+}
+
 void threadpreload_free(Thread* base) {
     ThreadPreload* thread = _threadToThreadPreload(base);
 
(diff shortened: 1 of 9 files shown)


src/main/host/thread_preload.c
```diff
@@ -52,6 +52,11 @@ static ThreadPreload* _threadToThreadPreload(Thread* thread) {
 
 static Thread* _threadPreloadToThread(ThreadPreload* thread) { return (Thread*)thread; }
 
+static void _threadpreload_continuePlugin(ThreadPreload* thread, const ShimEvent* event) {
+    host_unlockShimShmemLock(thread->base.host);
+    shimevent_sendEventToPlugin(thread->ipc_data, event);
+}
```
stevenengler, line 58: Maybe add comments here about where this was expected to have been locked, since it can be difficult to follow the locked/unlocked state of a lock when the locking and unlocking are in different parts of the code.

src/main/host/thread_preload.c
```diff
@@ -223,6 +228,7 @@ static inline void _threadpreload_waitForNextEvent(ThreadPreload* thread) {
     MAGIC_ASSERT(_threadPreloadToThread(thread));
     utility_assert(thread->ipc_data);
     shimevent_recvEventFromPlugin(thread->ipc_data, &thread->currentEvent);
+    host_lockShimShmemLock(thread->base.host);
```
stevenengler, line 231: Maybe add comments here about where this is expected to be unlocked since there's no nearby `host_unlockShimShmemLock()` call.

stevenengler (approved):

src/main/host/thread_preload.c
sporksmith: Done

src/main/host/thread_preload.c
sporksmith: Done

Issue #1873, opened 2022-01-26
Support tracing managed processes in gdb


sporksmith:
The `--gdb` flag is intended to help with this in general, but it pauses at a point before Shadow has started its managed processes, so it's not particularly useful for attaching to one of those.

We can trace shadow itself either by starting it under  gdb or attaching to it externally (with or without the `--gdb` flag), but tracing a managed process from its beginning is a little tricky. If we start from tracing shadow, we need to set `follow-fork-mode` to `child` to attach to the managed process, but if we set that from the beginning, then we'll follow the *first* fork, which is actually a utility process for getting CPU info.  

I've been working around this by setting a breakpoint at `_threadpreload_fork_exec` in shadow, and then changing `follow-fork-mode` to `child` just before the target process is started. It looks something like this:

```
gdb --args shadow ....
break _threadpreload_fork_exec
run
<check arguments to see which managed process is about to be forked>
set follow-fork-mode child  # if its the target process
continue
handle SIGSYS nostop
<set any other breakpoints etc>
continue
```

This isn't ideal as the supported way to do this though, since it involves setting a breakpoint on an internal function.

We could use something like `catch syscall vfork`. At least that would decouple us from the name of the internal function, but the user still needs to inspect the internals to see if the fork that's about to happen is going to exec the target process.

Something like the `--gdb` flag is probably a better solution, but I think we'll want to be able to specify a target process or simulation time to pause at.

sporksmith:
We do have some documentation here: https://shadow.github.io/docs/guide/developer_guide.html#debugging-shadow-using-gdb. For managed processes it recommends using a core dump, but that's not always sufficient.

stevenengler:
Todo:
- [ ] add a new cli option to help users debug specific managed processes
- support the workflows:
  - [ ] attach a new gdb session directly to the managed process
  - [ ] attach a new gdb session to shadow, and use the `follow-fork-mode` to switch from debugging shadow to debugging the managed process
- [ ] document the new option and the above workflows that use this option

sporksmith:
Something we'll want to address and/or document:

On some systems - particularly Ubuntu - non-root users can only ptrace-attach to child processes of the current process. This is why `gdb -p` doesn't work on these systems without `sudo`. https://wiki.ubuntu.com/SecurityTeam/Roadmap/KernelHardening#ptrace_Protection

[...]

PR #1997, opened 2022-03-31
Add '--debug-hosts' option to help debug managed processes

stevenengler:
This is mostly to add the `--debug-hosts` option and to pipe the option through to the process objects. We will likely want to change the exact behaviour of this option. The option is included in the experimental configuration options, but in the future we'll make it a cli-only option. This is part of #1873.


sporksmith (commented):
Played with this a bit. I think this is a good start, though I think it might be worth later extending a bit too.

On some systems - particularly Ubuntu - non-root users can only ptrace-attach to child processes of the current process. This is why `gdb -p` doesn't work on these systems without `sudo`. https://wiki.ubuntu.com/SecurityTeam/Roadmap/KernelHardening#ptrace_Protection

I think this feature as-is is still a good start, but users on such systems will have to run gdb as root, or disable this protection globally (which also requires root).

It looks like `CAP_SYS_PTRACE` rather than full root should also be adequate. This still requires access to the root user to get the capability in the first place, but at least lets you drop the rest of the root privileges and not disable the ptrace protection globally. `sudo capsh --caps=cap_sys_ptrace=ep --print -- -c "gdb -p 12129"` seems to work, but we should probably check a bit more carefully to get this right before documenting such a thing.

I think we could work out an extension to support the `gdb --args shadow ...` use-case, where we pause just before the `fork`, at which point the user can use `set follow-fork-mode child` in gdb to switch over to the child that's about to be forked. That could be a later extension, though. I think we could even do it using the same config option. Since gdb stops on any signal, even if it's handled, we could install a no-op handler for e.g. SIGUSR1 and raise that signal just before forking processes in the given hosts. In gdb that'd cause it to pause. Outside of gdb it should be a no-op.

stevenengler:
@sporksmith Do you want to copy your comment over to the issue?

src/main/core/support/configuration.rs
```diff
@@ -1,8 +1,9 @@
-use std::collections::BTreeMap;
```
sporksmith, line 1: It'd be nice if we caught somewhere if any of the specified host names aren't host names in the simulation. Maybe we can do it somewhere here? (Also fine to consider that as something to add later)

src/main/host/process.c
```diff
@@ -581,6 +584,12 @@ static void _process_start(Process* proc) {
     worker_setActiveProcess(NULL);
     worker_setActiveThread(NULL);
 
+    if (proc->pause_for_debugging) {
```
sporksmith, line 587: Optional - moving this between `vfork` and `exec` in `thread_preload.c` would let the user attach before early initialization. We'd need to be careful here to use only "async signal safe" functionality; probably `write` and `read` rather than the higher level libc functions.

If so, probably unnecessary to also implement in thread_ptrace.c, since we won't support ptrace-attaching externally anyway. (Maybe even better to explicitly error if this feature is enabled with --interpose-method=ptrace, but maybe not even worth bothering with since it'll be removed soonish...)

sporksmith (approved):

sporksmith:
> @sporksmith Do you want to copy your comment over to the issue?

Done

src/main/core/support/configuration.rs
stevenengler: Yeah, I think it would be good but I'll leave it as something to do later. It would be much easier to do when the worker is in rust, since we need to wait until the hostnames are expanded before we check. For example, a host named 'client' in the config may get renamed to 'client1' and 'client2' if `quantity` is 2.

src/main/host/process.c
stevenengler: I'll leave something like this for a future PR. I'm still hesitant about adding additional code after the `vfork()`.

PR #1998, opened 2022-03-31
Fix a case where getaddrinfo() incorrectly returns 0

stevenengler:
This causes iperf3 to segfault in [`netannounce()`](https://github.com/esnet/iperf/blob/dfcea9f6a09ead01089a3c9d20c7032f2c0af2c1/src/net.c#L251). (iperf3 still doesn't run correctly under shadow, but this fixes the segfault.)

### DIFF
--- a/src/lib/shim/shim_api_addrinfo.c
+++ b/src/lib/shim/shim_api_addrinfo.c
@@ -368,7 +368,11 @@ int shim_api_getaddrinfo(const char* node, const char* service, const struct add
             }
         }
         // We've finished adding all relevant addresses.
-        return 0;
+        if (*res != NULL) {
+            return 0;
+        } else {
+            return EAI_NONAME;
+        }
     }
 
     // "`node` specifies either a numerical network address..."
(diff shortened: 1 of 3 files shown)


src/lib/shim/shim_api_addrinfo.c
```diff
@@ -368,7 +368,11 @@ int shim_api_getaddrinfo(const char* node, const char* service, const struct add
             }
         }
         // We've finished adding all relevant addresses.
-        return 0;
+        if (*res != NULL) {
+            return 0;
+        } else {
+            return EAI_NONAME;
```
sporksmith, line 374: Could you add a test for this case to src/test/resolver/test_getaddrinfo.c?

sporksmith (approved):

src/lib/shim/shim_api_addrinfo.c
stevenengler: Added an IPv6 test

PR #2008, opened 2022-04-06
Time jump bug fixes

sporksmith:


src/main/host/thread_preload.c
```diff
@@ -347,7 +347,7 @@ SysCallCondition* threadpreload_resume(Thread* base) {
                 break;
             }
             default: {
-                utility_panic("unknown event type");
+                utility_panic("unknown event type: %d");
```
stevenengler, line 350: Is this supposed to include the type?

stevenengler (approved):

src/main/host/thread_preload.c
sporksmith: :facepalm: done

PR #2010, opened 2022-04-06
Add support for the `TCP_CONGESTION` sockopt

stevenengler:
Shadow only supports reno so `setsockopt()` only allows `reno`. We also don't intercept `/proc/sys/net/ipv4/tcp_available_congestion_control` or similar file paths, so applications may still try to set unsupported congestion control algorithms thinking that they actually are supported.

This feature is required for iperf3, but you may need to use the `--congestion` flag to specifically set "reno" rather than allowing iperf to choose the algorithm.

### DIFF
--- a/src/main/host/syscall/socket.c
+++ b/src/main/host/syscall/socket.c
@@ -18,6 +18,8 @@
 #include "main/host/descriptor/descriptor.h"
 #include "main/host/descriptor/socket.h"
 #include "main/host/descriptor/tcp.h"
+#include "main/host/descriptor/tcp_cong.h"
+#include "main/host/descriptor/tcp_cong_reno.h"
 #include "main/host/descriptor/udp.h"
 #include "main/host/host.h"
 #include "main/host/process.h"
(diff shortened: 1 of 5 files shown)


src/main/host/syscall/socket.c
```diff
@@ -361,6 +384,67 @@ static int _syscallhandler_getSocketOptHelper(SysCallHandler* sys, Socket* sock,
...
+            }
+
+            if (optlen < 4 || strncmp(name, "reno", optlen) != 0) {
+                warning("Shadow sockets only support 'reno' for TCP_CONGESTION");
+                return -ENOENT;
+            }
+
+            // shadow doesn't support other congestion types, so do nothing
+            const char* current_name = tcp_cong(tcp)->hooks->tcp_cong_name_str();
+            utility_assert(current_name != NULL && strcmp(current_name, "reno") == 0);
```
sporksmith, line 436: Could we expose this constant in `tcp_cong.h` rather than repeating it here? e.g. by exposing `tcp_cong_reno_name_str_()`?

sporksmith (approved):

src/main/host/syscall/socket.c
stevenengler: Added this as a constant so the string is only in one place.

PR #2014, opened 2022-04-07
Support goroutine preemption

sporksmith:


src/test/golang/test_go_preempt.go
```diff
@@ -0,0 +1,31 @@
...
+	t_finish := t0.Add(100 * time.Millisecond)
+	for time.Now().Before(t_finish) {
+		// Spin.
+
+		// For this test to pass, time must eventually move forward in such a
+		// spin loop.
+
+		// Really, though, this test is intended to test compatibility with
+		// golang's preemption mechanism, which will eventually interrupt with
+		// a signal to allow other goroutines to run.
```
stevenengler, line 19: I don't know much about Go, but I don't see any other goroutines in this program. Does this test expect that the go runtime itself will be running its own goroutines, or should this test add a second goroutine to test this preemption?

src/lib/shim/shim_syscall.c
```diff
@@ -209,9 +219,30 @@ static void _shim_process_signals(ShimShmemHostLock* host_lock) {
...
+            //
+            // If we want the context from the point at which the syscall was
+            // made, we could use the one passed to our sigsys handler for
+            // syscalls intercepted via seccomp. It's unclear what we'd do for
+            // syscalls that were intercepted via LD_PRELOAD instead, though.
+            //
+            // For now, pass NULL, and when we encounter a use-case tries to use
+            // the context it should crash or error, at which point we can
+            // revisit.
+            ucontext_t* ctx = NULL;
```
stevenengler, line 240: Is it worth adding a warning message, or does this path happen often?

stevenengler (approved):

src/test/golang/test_go_preempt.go
sporksmith: `main` starts 2 instances of this goroutine

src/lib/shim/shim_syscall.c
```diff
@@ -209,9 +219,30 @@ static void _shim_process_signals(ShimShmemHostLock* host_lock) {
             // Restore the signal-stack configuration.
             shimshmem_setSigAltStack(host_lock, shim_threadSharedMem(), ss_original);
         } else {
-            // Call signal handler with host lock released, and native syscalls disabled.
+            // It's not clear what ucontext_t would be appropriate to pass here.
+            // A signal handler that ultimately swaps to the provided the
+            // context should have arranged for the signal handler to use an
```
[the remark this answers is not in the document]
sporksmith: Done

src/lib/shim/shim_syscall.c
sporksmith: Hmm, good question. I was thinking it may happen often, but right now our only real-world use case with signal handlers is the golang runtime, which so far only exercises the sigaltstack path.

Went ahead and added the warning for now. We can always reduce the severity if it's too noisy.

PR #2017, opened 2022-04-08
Add support for SA_RESTART

### LINKED ISSUE #1974: Implement SA_RESTART
We currently ignore the `SA_RESTART` flag when handling signals.

From `signals(7)`:

> If a blocked call to one of the following interfaces is interrupted  by
a  signal  handler,  then the call is automatically restarted after the
signal handler returns if the SA_RESTART flag was used;  otherwise  the
call fails with the error EINTR:

This is fairly straightforward to implement. In the shim, after handling all pending signals, if all of the handed signals' actions has `SA_RESTART` set, and the syscall originally returned `EINTR`, and the syscall is one of the ones eligible to be restarted, we can just re-execute the syscall.

The main potentially tricky parts are:
* Validating that the syscall is one of those eligible to be restarted. This is documented in `signals(7)`, but sometimes depends on parameters to the syscall, not just the syscall number. It might be though that we can delegate such logic to the syscall handlers themselves, such that they don't return `EINTR` in the first place in such cases.
* There are a few restartable syscalls that have relative timeouts, which should be updated appropriately to account for elapsed time when restarted. From `restart_syscall(2)` these are `poll`, `nanosleep`, `clock_nanosleep`, and `futex`  with `FUTEX_WAIT`.

sporksmith:
Fixes #1974

### DIFF
--- a/src/lib/shim/shim_syscall.c
+++ b/src/lib/shim/shim_syscall.c
@@ -119,9 +119,12 @@ static void _call_signal_handler(const struct shd_kernel_sigaction* action, int 
     shim_swapAllowNativeSyscalls(true);
 }
 
-static void _shim_process_signals(ShimShmemHostLock* host_lock) {
+// Handle pending unblocked signals, and return whether *all* corresponding
+// signal actions had the SA_RESTART flag set.
+static bool _shim_process_signals(ShimShmemHostLock* host_lock) {
     int signo;
     siginfo_t siginfo;
+    bool restartable = true;
     while ((signo = shimshmem_takePendingUnblockedSignal(
                 host_lock, shim_processSharedMem(), shim_threadSharedMem(), &siginfo)) != 0) {
         shd_kernel_sigset_t blocked_signals =
@@ -166,8 +169,8 @@ static void _shim_process_signals(ShimShmemHostLock* host_lock) {
             shimshmem_setSignalAction(host_lock, shim_processSharedMem(), signo,
                                       &(struct shd_kernel_sigaction){.ksa_handler = SIG_DFL});
         }
-        if (action.ksa_flags & SA_RESTART) {
-            error("SA_RESTART unimplemented; ignoring.");
+        if (!(action.ksa_flags & SA_RESTART)) {
+            restartable = false;
         }
 
         const stack_t ss_original = shimshmem_getSigAltStack(host_lock, shim_threadSharedMem());
@@ -250,6 +253,7 @@ static void _shim_process_signals(ShimShmemHostLock* host_lock) {
         // Restore mask
         shimshmem_setBlockedSignals(host_lock, shim_threadSharedMem(), blocked_signals);
     }
+    return restartable;
 }
 
 static SysCallReg _shim_emulated_syscall_event(const ShimEvent* syscall_event) {
@@ -295,8 +299,25 @@ static SysCallReg _shim_emulated_syscall_event(const ShimEvent* syscall_event) {
                 // (e.g. `kill(getpid(), signo)`), or may have been sent by another thread
                 // while this one was blocked in a syscall.
                 ShimShmemHostLock* host_lock = shimshmemhost_lock(shim_hostSharedMem());
-                _shim_process_signals(host_lock);
+                const bool allSigactionsHadSaRestart = _shim_process_signals(host_lock);
                 shimshmemhost_unlock(shim_hostSharedMem(), &host_lock);
+
+                // Check whether a blocking syscall was interrupted by a signal.
+                // Note that handlers don't usually return -EINTR directly;
+                // instead `syscall_handler_make_syscall` converts "blocked"
+                // results to -EINTR when an unblocked signal is pending.
+                if (rv.as_i64 == -EINTR) {
+                    // Syscall was interrupted by a signal. Consider restarting. See signal(7).
+                    const bool syscallSupportsSaRestart =
+                        res.event_data.syscall_complete.restartable;
+                    trace("Syscall interrupted by signals. allSigactionsHadSaRestart:%d "
+                          "syscallSupportsSaRestart:%d",
+                          allSigactionsHadSaRestart, syscallSupportsSaRestart);
+                    if (allSigactionsHadSaRestart && syscallSupportsSaRestart) {
+                        shimevent_sendEventToShadow(ipc, syscall_event);
+                        continue;
+                    }
+                }
 
                 return rv;
             }
--- a/src/main/host/syscall/format.rs
+++ b/src/main/host/syscall/format.rs
@@ -1,3 +1,4 @@
+use nix::errno::Errno;
 use std::any::TypeId;
 use std::fmt::Display;
 use std::marker::PhantomData;
@@ -501,7 +502,7 @@ where
     pub fn new(rv: &'a SyscallResult, options: FmtOptions, mem: &'a MemoryManager) -> Option<Self> {
         match &rv {
             SyscallResult::Ok(_)
-            | SyscallResult::Err(SyscallError::Errno(_))
+            | SyscallResult::Err(SyscallError::Failed(_))
             | SyscallResult::Err(SyscallError::Native) => Some(Self {
                 rv,
                 options,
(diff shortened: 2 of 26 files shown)


src/main/host/syscall_types.rs
```diff
@@ -328,10 +328,16 @@ impl<T> TypedPluginPtr<T> {
 #[derive(Debug, PartialEq, Eq)]
 pub enum SyscallError {
     Errno(nix::errno::Errno),
-    Cond(SysCallCondition),
+    Blocked(Blocked),
     Native,
 }
```
stevenengler, line 333: How does this handle the case in the C code where you have a `(SysCallReturn){.state = SYSCALL_DONE, .retval = -EINTR, .restartable = scr.restartable}` (equivalent to a `SyscallError::Errno` with a `restartable` value)? I think this might be less confusing with a new `SyscallError::Restart` variant instead of making a special case out of `-EINTR`.

Also, I think that converting between types will lose information. For example if you have a variable `ret` which is a `(SysCallReturn){.state = SYSCALL_DONE, .retval = -EINTR, .restartable = true}`, I don't think that `assert_eq!(ret, SyscallResult::from(ret).into())` would hold. There is the `test_syscallreturn_roundtrip` test in `descriptor/mod.rs`, but it doesn't test this case. There is some code such as [`log_syscall`](https://github.com/shadow/shadow/blob/49fb5c3e0f522b082dfca9fa71d583ca906bac23/src/main/host/syscall/format.rs#L572) which relies on this conversion not losing information (it takes an owned `c::SysCallReturn` and returns a round-trip owned `c::SysCallReturn` since the rust code has no way of working with a reference/ptr to a `c::SysCallReturn`, only an owned `c::SysCallReturn`).

src/test/signal/test_signals.rs
```diff
@@ -501,6 +501,37 @@ fn test_handled_kill_interrupts_syscall() -> Result<(), Box<dyn Error>> {
...
 }
 
+fn test_restart() -> Result<(), Box<dyn Error>> {
+    let signal = Signal::SIGUSR1;
+    unsafe {
+        signal::sigaction(
+            signal,
+            &signal::SigAction::new(
+                signal::SigHandler::Handler(nop_signal_handler),
+                signal::SaFlags::SA_RESTART,
```
stevenengler, line 511: What happens if there are two signal handlers, and only one `SA_RESTART`? It looks like shadow will restart if at least one signal handler is `SA_RESTART`, so maybe there should be a test for this?

src/main/host/syscall_handler.c
```diff
@@ -562,7 +562,8 @@ SysCallReturn syscallhandler_make_syscall(SysCallHandler* sys,
         SysCallCondition* condition = scr.cond;
         utility_assert(condition);
         syscallcondition_unref(condition);
-        scr = (SysCallReturn){.state = SYSCALL_DONE, .retval = -EINTR};
+        scr = (SysCallReturn){
+            .state = SYSCALL_DONE, .retval = -EINTR, .restartable = scr.restartable};
```
stevenengler, line 566: The comment above says "The shim-side code will run the signal handlers before returning EINTR to the original syscall". This should be updated to include comments about what happens if it's restartable.

src/lib/shim/shim_syscall.c
```diff
@@ -263,8 +267,20 @@ static SysCallReg _shim_emulated_syscall_event(const ShimEvent* syscall_event) {
                 // (e.g. `kill(getpid(), signo)`), or may have been sent by another thread
                 // while this one was blocked in a syscall.
                 ShimShmemHostLock* host_lock = shimshmemhost_lock(shim_hostSharedMem());
-                _shim_process_signals(host_lock);
+                const bool sigactionsHadSaRestart = _shim_process_signals(host_lock);
                 shimshmemhost_unlock(shim_hostSharedMem(), &host_lock);
+                if (rv.as_i64 == -EINTR) {
+                    // Syscall was interrupted by a signal. Consider restarting. See signal(7).
+                    const bool syscallSupportsSaRestart =
+                        res.event_data.syscall_complete.restartable;
```
stevenengler, line 275: I think there should be a comment somewhere around here explaining that `syscallhandler_make_syscall()` checks for pending signals ahead of time to convert blocking syscall results to `EINTR` syscall results. It's a little confusing how the syscall result changes throughout shadow. For example, the syscall handler may decide to block, and then shadow may decide to change that block into a `EINTR`, then the shim may decide to change that `EINTR` into a new syscall event for the same syscall (restarts the syscall).

stevenengler (changes requested):

src/main/host/syscall_types.rs
sporksmith: > How does this handle the case in the C code where you have a `(SysCallReturn){.state = SYSCALL_DONE, .retval = -EINTR, .restartable = scr.restartable}` (equivalent to a `SyscallError::Errno` with a `restartable` value)?

Hmm, this was somewhat intentional but I see the confusion and how this might lead to problems down the road. Right now handlers themselves never return `(SysCallReturn){.state = SYSCALL_DONE, .retval = -EINTR, .restartable = true}`; they only set the `restartable` flag on `SYSCALL_BLOCKED` results, and the `syscall_handler` code rewrites the `SYSCALL_BLOCKED` into `SYSCALL_DONE` if there's an unblocked syscall pending.

But yeah, this is made trickier because handlers *can* themselves check if unblocked signals are pending and return `-EINTR` themselves, and some need to (e.g. nanosleep). Currently the only ones that do this are also not restartable, but I'm not sure that'll always be the case.

Still working on this, but I'm leaning towards having the `restartable` flag in both the `Blocked` and `Failed` case:

```
 #[derive(Debug, PartialEq, Eq)]
 pub enum SyscallError {
-    Errno(nix::errno::Errno),
+    Failed(Failed),
     Blocked(Blocked),
     Native,
 }
...
+#[derive(Debug, PartialEq, Eq)]
+pub struct Failed {
+    pub errno: nix::errno::Errno,
+    pub restartable: bool,
+}
```

>  I think this might be less confusing with a new `SyscallError::Restart` variant instead of making a special case out of `-EINTR`.

We would still need a `restartable` flag in the `Blocked` case, unless we want to force every restartable signal handler that might otherwise return `Blocked` to also check whether there's an unblocked signal pending. It seems better to have that logic in one place.

> Also, I think that converting between types will lose information. For example if you have a variable `ret` which is a `(SysCallReturn){.state = SYSCALL_DONE, .retval = -EINTR, .restartable = true}`, I don't think that `assert_eq!(ret, SyscallResult::from(ret).into())` would hold. There is the `test_syscallreturn_roundtrip` test in `descriptor/mod.rs`, but it doesn't test this case. There is some code such as [`log_syscall`](https://github.com/shadow/shadow/blob/49fb5c3e0f522b082dfca9fa71d583ca906bac23/src/main/host/syscall/format.rs#L572) which relies on this conversion not losing information (it takes an owned `c::SysCallReturn` and returns a round-trip owned `c::SysCallReturn` since the rust code has no way of working with a reference/ptr to a `c::SysCallReturn`, only an owned `c::SysCallReturn`).

Ok, I'll be mroe careful about this. I think my current plan to add the `restartable` flag to both the `Blocked` and `Failed` cases will address it. I'll look at adding some more round trip tests as well...

sporksmith: Finished making this change. I admit it's still ugly, but it's the least ugly option I can think of at the moment...

src/main/host/syscall/format.rs
```diff
@@ -525,12 +526,17 @@ where
...
-                    as_i64: -(*e as i64),
+                    as_i64: -(errno as i64),
                 };
                 let rv = SyscallVal::<'_, RV>::from_mem(rv, self.options, self.mem);
-                write!(f, "{} ({})", rv, e)
+                if errno == Errno::EINTR {
+                    write!(f, "{} ({}, restartable:{})", rv, errno, failed.restartable)
+                } else {
+                    write!(f, "{} ({})", rv, errno)
+                }
```
stevenengler, line 539: I think for the strace syscall formatting we don't need to print the `restartable` flag.

stevenengler (approved):

src/main/host/syscall_types.rs
stevenengler: Sounds good to me. I don't see a nicer way of handling this either.

src/main/host/syscall/format.rs
sporksmith: Yeah, agreed. Done.

PR #2020, opened 2022-04-11
Fixed formatting of syscall logging macro calls

stevenengler:
Just something that's bothered me for a little while :)


sporksmith (approved):
Ok with me if it's ok with the rustfmt lint check, though I'll be surprised if it is. It seems like it should allow either the new formatting or the old but not both. Unless maybe it ignores function attributes for some reason...?

stevenengler:
> Ok with me if it's ok with the rustfmt lint check, though I'll be surprised if it is. It seems like it should allow either the new formatting or the old but not both. Unless maybe it ignores function attributes for some reason...?

I think it ignores macro invocations in attributes.

Issue #2021, opened 2022-04-11
Fail to patch vdso_time on ubuntu 20.04.2


sporksmith:
When running a tor simulation on an Ubuntu 20.04.2 machine, I get this warning:

```
00:00:00.000001 warning [patch_vdso.c:169] [_inject_trampoline] Symbol '__vdso_time' not large enough to inject trampoline
```

The code currently fails gracefully here - it doesn't patch that vdso function and allows the simulation to continue. That should be ok for most purposes, but may cause us to fail to intercept time, e.g. in golang.

sporksmith:
Here's the disassembly of `__vdso_time` on that machine:

```
(gdb) disassemble __vdso_time                                                                                                               
Dump of assembler code for function time:                                                                                                   
   0x00007ffc443fa8d0 <+0>:     mov    -0x3837(%rip),%rax        # 0x7ffc443f70a0                                                           
   0x00007ffc443fa8d7 <+7>:     test   %rdi,%rdi                                                                                            
   0x00007ffc443fa8da <+10>:    je     0x7ffc443fa8df <time+15>                                                                             
   0x00007ffc443fa8dc <+12>:    mov    %rax,(%rdi)                                                                                          
   0x00007ffc443fa8df <+15>:    retq                                                                                                        
End of assembler dump. 
```

This is 16 bytes (0x00007ffc443fa8df-0x00007ffc443fa8d0+1).

In [_inject_trampoline](https://github.com/shadow/shadow/blob/97eb2fd86a6898ac48fea0777d5cc54e89b9b4a2/src/lib/shim/patch_vdso.c#L160), we check for a loose upper bound of 20 bytes, but actually only need 13 bytes. So we just need to use a tighter bound here. The loose bound was to avoid the risk of a future edit making the trampoline larger without correctly updating the bound. An alternative way to address this is to calculate the size needed dynamically.

PR #2022, opened 2022-04-11
Use precise vdso trampoline size

sporksmith:
Fixes https://github.com/shadow/shadow/issues/2021

### DIFF
--- a/src/lib/shim/patch_vdso.c
+++ b/src/lib/shim/patch_vdso.c
@@ -164,13 +164,17 @@ static void _inject_trampoline(struct ParsedElf* parsedElf, const char* vdsoFnNa
         warning("Couldn't find symbol '%s' to override", vdsoFnName);
         return;
     }
-    // Loose upper bound on size of instructions we're writing, below.
-    if (symbol->st_size < 20) {
+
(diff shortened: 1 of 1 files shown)


src/lib/shim/patch_vdso.c
```diff
@@ -164,13 +164,17 @@ static void _inject_trampoline(struct ParsedElf* parsedElf, const char* vdsoFnNa
...
-    // Loose upper bound on size of instructions we're writing, below.
-    if (symbol->st_size < 20) {
+
+    // Manually calculated trampoline size. Later validated in assertion.
+    const size_t trampolineSize = 13;
+
+    if (symbol->st_size < trampolineSize) {
         warning("Symbol '%s' not large enough to inject trampoline", vdsoFnName);
         return;
     }
```
stevenengler, line 174: Should this be a failure instead? I feel like shim warnings will rarely be noticed, and this could cause the sim to produce unexpected results.

stevenengler (approved):

src/lib/shim/patch_vdso.c
sporksmith: Good point. Done.

PR #2026, opened 2022-04-12
Add a socket test for autobound socket names

stevenengler:
This should have been included in #2024.

### DIFF
--- a/src/test/socket/getsockname/test_getsockname.rs
+++ b/src/test/socket/getsockname/test_getsockname.rs
@@ -99,6 +99,11 @@ fn get_tests() -> Vec<test_utils::ShadowTest<(), String>> {
                     passing.clone(),
                 ),
                 test_utils::ShadowTest::new(
+                    &append_args("test_autobound_socket"),
+                    move || test_autobound_socket(domain, sock_type),
+                    set![TestEnv::Libc, TestEnv::Shadow],
+                ),
+                test_utils::ShadowTest::new(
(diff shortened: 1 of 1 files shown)


src/test/socket/getsockname/test_getsockname.rs
```diff
@@ -461,6 +466,91 @@ fn test_bound_socket(domain: libc::c_int, sock_type: libc::c_int) -> Result<(),
...
+        fd: fd,
+        addr: Some(addr),
+        addr_len: Some(addr.ptr_size()),
+    };
+
+    test_utils::run_and_close_fds(&[fd], || check_getsockname_call(&mut args, None))?;
+
+    let expected_addr_len = match domain {
+        libc::AF_INET => std::mem::size_of::<libc::sockaddr_in>() as u32,
+        // 2 + 1 + 5
```
sporksmith, line 519: Not clear to me where these come from. Possibly overkill, but one way to spell it out is something like: 

```
libc:: AF_UNIX => {
  let foo_len = 2;
  let bar_len = 1;
  let baz_len = 5;
  foo_len + bar_len + baz_len
}
```

sporksmith (approved):

src/test/socket/getsockname/test_getsockname.rs
stevenengler: I added a comment instead explaining where they came from.

PR #2032, opened 2022-04-13
Add more documentation for 'strace_logging_mode' option

stevenengler:
Was missing a "limitations" section.


sporksmith (approved):
Hmm, yeah I hadn't realized some of the caveats around signal handling and SA_RESTART. Maybe we can think about either splitting the strace log so that the result is logged after the common "finalization" code, or finding a nice way to invoke the "finalization" code at the end of every syscall handler.

stevenengler:
> Hmm, yeah I hadn't realized some of the caveats around signal handling and SA_RESTART. Maybe we can think about either splitting the strace log so that the result is logged after the common "finalization" code, or finding a nice way to invoke the "finalization" code at the end of every syscall handler.

Yeah, I'm thinking maybe it would be better if we had a global table mapping syscall numbers to syscall formatting functions, and then `syscallhandler_make_syscall` could do the formatting before running the syscall handler, and then later decide whether to log it or not. But I haven't thought about it much yet.

PR #2042, opened 2022-04-13
Test if listening sockets are readable/writable

stevenengler:
Wanted to check if listening unix sockets with incoming connection queue space available were writable or not, so just made it into a generalized test instead.

### DIFF
--- a/src/test/socket/listen/test_listen.rs
+++ b/src/test/socket/listen/test_listen.rs
@@ -4,7 +4,10 @@
  */
 
 use test_utils::set;
+use test_utils::socket_utils;
 use test_utils::TestEnvironment as TestEnv;
+
+use nix::poll::PollFlags;
 
 struct ListenArguments {
     fd: libc::c_int,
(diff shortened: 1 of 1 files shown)


src/test/socket/listen/test_listen.rs
```diff
@@ -122,6 +125,40 @@ fn get_tests() -> Vec<test_utils::ShadowTest<(), String>> {
...
+                    |s| format!("{} <domain={},type={},flag={}>", s, domain, sock_type, flag);
+
+                // skip tests that use SOCK_SEQPACKET with INET sockets
+                if domain == libc::AF_INET && sock_type == libc::SOCK_SEQPACKET {
+                    continue;
+                }
+
+                tests.extend(vec![
+                    test_utils::ShadowTest::new(
+                        &append_args("test_readable"),
```
sporksmith, line 148: Nit - these could use a slightly more descriptive label and name. e.g. `test_listeningSocketNotReadable`

sporksmith (approved):

src/test/socket/listen/test_listen.rs
stevenengler: Changed the function names.

stevenengler:
I made a small additional change since it could lead to an accidental bug in the future.

```rust
let poll_fd = nix::poll::PollFd::new(fd, PollFlags::POLLOUT);
let count = nix::poll::poll(&mut [poll_fd], 50).unwrap();
```

It's not immediately obvious that `poll_fd` won't have any `revents` set, since `poll_fd` is copied into the `&mut [poll_fd]` slice.

So instead I changed it to the following so that the `PollFd` will be updated with the correct `revents`, even though we don't currently read them:

```rust
let mut poll_fds = [nix::poll::PollFd::new(fd, PollFlags::POLLOUT)];
let count = nix::poll::poll(&mut poll_fds, 50).unwrap();
```

PR #2048, opened 2022-04-13
Increment time locally when able

sporksmith:
* Add a scheduler API `scheduler_nextHostEventTime` to peek at the time of the next scheduled event for a given Host. I only implemented this for the couple schedulers we use in practice - the others panic if this API is called. I could implement this API for the other schedulers, but it doesn't seem worth it since we don't currently use them, and plan to remove them.
* Allow time to be moved forward during the execution of an event. Use this to account for unblocked syscall latency without having to schedule a new event, or without even having to context switch from the shim to Shadow, as long as the maximum runahead time isn't exceeded. The max runahead time is the first of: the time of the next event scheduled for the current host, and the end of the current scheduler round.
* Update tests so that all tests pass when the default `--unblocked-syscall-limit` is 1 instead of its current default of 0; i.e. when moving time forward on every unblocked syscall is enabled by default. This PR doesn't actually change the default, though.

### DIFF
--- a/src/main/core/scheduler/scheduler_policy_host_single.c
+++ b/src/main/core/scheduler/scheduler_policy_host_single.c
@@ -286,6 +286,40 @@ static Event* _schedulerpolicyhostsingle_pop(SchedulerPolicy* policy, Simulation
     return NULL;
 }
 
+static EmulatedTime _schedulerpolicyhostsingle_nextHostEventTime(SchedulerPolicy* policy, Host* host) {
+    MAGIC_ASSERT(policy);
+    HostSinglePolicyData* data = policy->data;
+
+    /* figure out which hosts we should be checking */
+    HostSingleThreadData* tdata =
+        g_hash_table_lookup(data->threadToThreadDataMap, GUINT_TO_POINTER(pthread_self()));
+    utility_assert(tdata);
+
+    HostSingleQueueData* qdata = g_hash_table_lookup(data->hostToQueueDataMap, host);
+    utility_assert(qdata);
+
+    /* FIXME split  this timer? */
+#ifdef USE_PERF_TIMERS
+    /* tracking idle time spent waiting for the host queue lock */
+    g_timer_continue(tdata->popIdleTime);
+#endif
+    g_mutex_lock(&(qdata->lock));
+#ifdef USE_PERF_TIMERS
+    g_timer_stop(tdata->popIdleTime);
+#endif
+
+    EmulatedTime nextEventTime = 0;
+    Event* nextEvent = priorityqueue_peek(qdata->pq);
+    if (nextEvent) {
+        utility_assert(event_getHost(nextEvent) == host);
+        nextEventTime = event_getTime(nextEvent) + EMULATED_TIME_OFFSET;
+    }
+
+    g_mutex_unlock(&(qdata->lock));
+
+    return nextEventTime;
+}
+
 static void _schedulerpolicyhostsingle_findMinTime(Host* host, HostSingleSearchState* state) {
     HostSingleQueueData* qdata = g_hash_table_lookup(state->data->hostToQueueDataMap, host);
     utility_assert(qdata);
@@ -345,6 +379,7 @@ SchedulerPolicy* schedulerpolicyhostsingle_new() {
     policy->getAssignedHosts = _schedulerpolicyhostsingle_getHosts;
     policy->push = _schedulerpolicyhostsingle_push;
     policy->pop = _schedulerpolicyhostsingle_pop;
+    policy->nextHostEventTime = _schedulerpolicyhostsingle_nextHostEventTime;
     policy->getNextTime = _schedulerpolicyhostsingle_getNextTime;
     policy->free = _schedulerpolicyhostsingle_free;
 
--- a/src/main/core/scheduler/scheduler_policy_thread_single.c
+++ b/src/main/core/scheduler/scheduler_policy_thread_single.c
@@ -67,7 +67,9 @@ static void _schedulerpolicythreadsingle_addHost(SchedulerPolicy* policy, Host* 
     }
     g_queue_push_tail(tdata->assignedHosts2, host);
 
-    /* finally, store the host-to-thread mapping */
+    /* finally, store the host-to-thread mapping. TODO: simplify, since this
+     * each scheduler instance of this type only ever has a single host. */
+    utility_assert(g_hash_table_size(data->hostToThreadMap) == 0);
     g_hash_table_replace(data->hostToThreadMap, host, GUINT_TO_POINTER(assignedThread));
 }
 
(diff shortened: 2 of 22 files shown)


sporksmith:
Still some cleanup needed, as well as fixing the interactions with signal handling. Doing some testing though to see how much this helps and if incrementing time at a finer granularity fixes the weird results we've seen in some tor sims.

The current version *always* moves time forward, but only reschedules if time would go past the next scheduled event on that host, or the round end time.

In the tor minimal test there is about 3.5% overhead. We should be able to reduce this further by reducing the "cost" of VDSO functions.

On `main`:
```
$ ./setup build --werror --test --extra && hyperfine "./setup test --extra tor-minimal-shadow-preload"
Benchmark #1: ./setup test --extra tor-minimal-shadow-preload
  Time (mean ± σ):     14.775 s ±  0.122 s    [User: 17.619 s, System: 4.392 s]
  Range (min … max):   14.505 s … 14.901 s    10 runs
```

On this branch:
```
$ ./setup build --werror --test --extra && hyperfine "./setup test --extra tor-minimal-shadow-preload"
Benchmark #1: ./setup test --extra tor-minimal-shadow-preload
  Time (mean ± σ):     15.296 s ±  0.391 s    [User: 18.250 s, System: 4.403 s]
  Range (min … max):   14.783 s … 16.155 s    10 runs
```

I kicked off a larger sim again on the Tor gitlab runner. Should have a result tomorrow. https://gitlab.torproject.org/jnewsome/sponsor-61-sims/-/jobs/123711

sporksmith:
> Still some cleanup needed, as well as fixing the interactions with signal handling. Doing some testing though to see how much this helps and if incrementing time at a finer granularity fixes the weird results we've seen in some tor sims.
> 
> The current version _always_ moves time forward, but only reschedules if time would go past the next scheduled event on that host, or the round end time.
> 
> In the tor minimal test there is about 3.5% overhead. We should be able to reduce this further by reducing the "cost" of VDSO functions.
> 
> On `main`:
> 
> ```
> $ ./setup build --werror --test --extra && hyperfine "./setup test --extra tor-minimal-shadow-preload"
> Benchmark #1: ./setup test --extra tor-minimal-shadow-preload
>   Time (mean ± σ):     14.775 s ±  0.122 s    [User: 17.619 s, System: 4.392 s]
>   Range (min … max):   14.505 s … 14.901 s    10 runs
> ```
> 
> On this branch:
> 
> ```
> $ ./setup build --werror --test --extra && hyperfine "./setup test --extra tor-minimal-shadow-preload"
> Benchmark #1: ./setup test --extra tor-minimal-shadow-preload
>   Time (mean ± σ):     15.296 s ±  0.391 s    [User: 18.250 s, System: 4.403 s]
>   Range (min … max):   14.783 s … 16.155 s    10 runs
> ```
> 
> I kicked off a larger sim again on the Tor gitlab runner. Should have a result tomorrow. https://gitlab.torproject.org/jnewsome/sponsor-61-sims/-/jobs/123711

The simulation completed. It ran very slowly, and has significantly different behavior. More work needed before accounting for syscall latency is enabled by default. 
[tornet.plot.pages.pdf](https://github.com/shadow/shadow/files/8492335/tornet.plot.pages.pdf)

src/main/core/worker.h
```diff
@@ -76,8 +76,8 @@ const ConfigOptions* worker_getConfig();
 gboolean worker_scheduleTask(Task* task, Host* host, SimulationTime nanoDelay);
 void worker_sendPacket(Host* src, Packet* packet);
 bool worker_isAlive(void);
-// Maximum time that the current thread may run ahead to.
-EmulatedTime worker_maxThreadTime(Host* host);
+// Maximum time that the current event may run ahead to.
+EmulatedTime worker_maxEventRunaheadTime(Host* host);
```
stevenengler, line 80: There is also the `worker_updateMinRunahead()` function, so maybe one of these functions should change names, or they should have documentation explaining that they refer to different runaheads.

stevenengler (changes requested):

src/main/core/scheduler/scheduler_policy_host_single.c
```diff
@@ -286,6 +286,35 @@ static Event* _schedulerpolicyhostsingle_pop(SchedulerPolicy* policy, Simulation
...
+    g_mutex_lock(&(qdata->lock));
+#ifdef USE_PERF_TIMERS
+    g_timer_stop(tdata->popIdleTime);
+#endif
+
+    Event* nextEvent = priorityqueue_peek(qdata->pq);
+
+    g_mutex_unlock(&(qdata->lock));
+
+    return nextEvent;
```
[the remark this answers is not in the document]
sporksmith: Hmm, my thinking was that `qdata->lock` here is only needed to protect the hash table itself, not the individual host queue or events in it; those are protected by the Host lock.

That is perhaps a bit subtle, though. Since we really only need the time of the next event I'll change this interface to just return that.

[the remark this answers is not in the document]
sporksmith: Changed to return the time instead of the Event.

src/main/core/scheduler/scheduler_policy_thread_single.c
```diff
@@ -108,6 +110,23 @@ static void _schedulerpolicythreadsingle_push(SchedulerPolicy* policy, Event* ev
...
+    ThreadSingleThreadData* tdata =
+        g_hash_table_lookup(data->threadToThreadDataMap, GUINT_TO_POINTER(pthread_self()));
+    utility_assert(tdata);
+
+    g_mutex_lock(&(tdata->lock));
+    Event* nextEvent = priorityqueue_peek(tdata->pq);
+    utility_assert(event_getHost(nextEvent) == host);
+    g_mutex_unlock(&(tdata->lock));
+
+    return nextEvent;
```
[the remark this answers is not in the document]
sporksmith: Changed to return the time instead of the Event.

src/main/core/worker.h
sporksmith: Changed `worker_updateMinRunahead` to `worker_updateMinHostRunahead`

src/main/core/worker.c
```diff
@@ -510,14 +510,14 @@ gboolean worker_scheduleTask(Task* task, Host* host, SimulationTime nanoDelay) {
...
-    SimulationTime max = _worker_getRoundEndTime();
+    EmulatedTime max = _worker_getRoundEndTime() + EMULATED_TIME_OFFSET;
 
-    Event* event = scheduler_peekHost(_worker_pool()->scheduler, host);
-    if (event) {
-        max = MIN(max, event_getTime(event));
+    EmulatedTime nextEventTime = scheduler_nextHostEventTime(_worker_pool()->scheduler, host);
+    if (nextEventTime) {
+        max = MIN(max, nextEventTime);
     }
```
stevenengler, line 518: Maybe add a comment here that 0 is not expected to be a valid `nextEventTime` emulated time since the offset `EMULATED_TIME_OFFSET` is greater than 0 (there will never be an event at an emulated time of 0)?

stevenengler (approved):

src/main/core/worker.c
sporksmith: Added a comment to `scheduler_nextHostEventTime` and made the comparison to 0 here explicit

PR #2049, opened 2022-04-14
Unify host locks

sporksmith:
It's simpler, and probably more efficient, to take the Host's shared memory lock the whole time an event is executing. By itself this is a performance win in cases where we'd otherwise need to take and release the lock multiple times while executing a single event.

We make this strictly a performance improvement, and a further simplification, by replacing the existing Host lock with the Host shared memory lock.

### DIFF
--- a/src/lib/shim/shim_shmem.c
+++ b/src/lib/shim/shim_shmem.c
@@ -96,8 +96,9 @@ struct _ShimShmemThread {
 size_t shimshmemhost_size() { return sizeof(ShimShmemHost); }
 
 void shimshmemhost_init(ShimShmemHost* hostMem, Host* host, uint32_t unblockedSyscallLimit) {
-    // We use `memcpy` here to allow us to initialize the const members of
-    // `hostMem`.
+    assert(hostMem);
+    // We use `memcpy` instead of struct assignment here to allow us to
+    // initialize the const members of `hostMem`.
     memcpy(hostMem,
            &(ShimShmemHost){
                .host_id = host_getID(host),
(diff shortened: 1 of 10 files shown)


stevenengler:
> We make this strictly a performance improvement, and a further simplification, by replacing the existing Host lock with the Host shared memory lock.

I'm not entirely clear on how this works. It seems like the new process is generally:

1. Take host lock when executing an event.
2. Before we switch to the plugin, release the host lock.
3. After switching from the plugin back to shadow, re-acquire the host lock.
4. Release the host lock when done executing the event.

My understanding of the host lock is that it protects everything within the host, including all process and thread objects. In this case I don't think we can release the host lock in `_threadpreload_continuePlugin()` while still holding a reference to the `Thread`.

sporksmith:
> > We make this strictly a performance improvement, and a further simplification, by replacing the existing Host lock with the Host shared memory lock.
> 
> I'm not entirely clear on how this works. It seems like the new process is generally:
> 
>     1. Take host lock when executing an event.
> 
>     2. Before we switch to the plugin, release the host lock.
> 
>     3. After switching from the plugin back to shadow, re-acquire the host lock.
> 
>     4. Release the host lock when done executing the event.
> 

Right.

> My understanding of the host lock is that it protects everything within the host, including all process and thread objects. In this case I don't think we can release the host lock in `_threadpreload_continuePlugin()` while still holding a reference to the `Thread`.

Hmm, darn I think you're right. Reverted that part and just moved where we take and release the shared memory lock.

stevenengler (approved):

PR #2051, opened 2022-04-15
Model VDSO calls as taking less time than regular syscalls

sporksmith:


src/test/examples/basic-file-transfer/basic-file-transfer-example.yaml
```diff
@@ -5,7 +5,7 @@ general:
 experimental:
   # needed for https://github.com/shadow/shadow/issues/1794
   # this is only included in the test, and not in the documentation example
-  unblocked_syscall_limit: 500
+  max_unapplied_cpu_latency: 1microsecond
```
stevenengler, line 8: The `docs/iperf3.md` and `docs/curl.md` should also be updated.

src/lib/shim/shim_sys.c
```diff
@@ -30,6 +30,20 @@ static EmulatedTime _shim_sys_get_time() {
 
 uint64_t shim_sys_get_simtime_nanos() { return _shim_sys_get_time() / SIMTIME_ONE_NANOSECOND; }
 
+static SimulationTime _shim_sys_latency_for_syscall(long n) {
+    switch (n) {
+        case SYS_clock_gettime:
+        case SYS_time:
+        case SYS_gettimeofday:
```
stevenengler, line 37: This doesn't make the distinction between syscalls and the libc VDSO wrappers. For example, my understanding is that if you call `syscall(SYS_clock_gettime)`, it won't be handled by the VDSO and should incur the full syscall latency penalty. I don't think Shadow needs to handle that correctly, but I think there should be a comment here explaining how shadow's behaviour differs from linux.

Also, shadow still has a `syscallhandler_clock_gettime()` implementation. Is there still any path where this function gets called? If not should we remove it? And if there is, should we also apply the VDSO latency from the shadow side?

stevenengler (changes requested):

docs/shadow_config_spec.md
```diff
@@ -337,6 +337,20 @@ Type: "ptrace" OR "preload"
...
+Type: String
+
+Max amount of execution-time latency allowed to accumulate before the clock is
+moved forward. Moving the clock forward is a potentially expensive operation, so
+larger values reduce simulation overhead, at the cost of coarser time jumps.
+
+Note also that accumulated-but-unapplied latency is discarded when a thread is
+blocked on a syscall.
+
+0 to never account for CPU latency.
```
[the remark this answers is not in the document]
sporksmith: Got rid of the 0 special-case here, after rebasing to include #2069

src/test/examples/basic-file-transfer/basic-file-transfer-example.yaml
sporksmith: Done in #2069

src/lib/shim/shim_sys.c
sporksmith: > I think there should be a comment here explaining how shadow's behaviour differs from linux.

Done

> Also, shadow still has a syscallhandler_clock_gettime() implementation. Is there still any path where this function gets called? 

It'll still get called in ptrace mode, if the managed code didn't call the `clock_gettime` libc function. We can probably safely delete it when we remove ptrace mode.

> And if there is, should we also apply the VDSO latency from the shadow side?

In the cases where it'd happen, charging the syscall latency is actually more-correct :)

src/main/core/support/configuration.rs
```diff
@@ -415,6 +415,13 @@ pub struct ExperimentalOptions {
     #[clap(help = EXP_HELP.get("unblocked_syscall_latency").unwrap().as_str())]
     pub unblocked_syscall_latency: Option<units::Time<units::TimePrefix>>,
 
+    /// Simulated latency of a vdso "syscall". For efficiency Shadow only
+    /// actually adds this latency if and when `unblocked_syscall_limit` is
+    /// reached.
+    #[clap(long, value_name = "seconds")]
+    #[clap(help = EXP_HELP.get("unblocked_vdso_latency").unwrap().as_str())]
+    pub unblocked_vdso_latency: Option<units::Time<units::TimePrefix>>,
```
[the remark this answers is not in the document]
sporksmith: Done

sporksmith:
I ended up rebasing since there were quite a lot of conflicts to sort out, and doing a merge commit for review purposes would have meant having to resolve them all again in the final rebase (though in hindsight I suppose I could have just not done a final rebase and used a merge commit, leaving in a merge from main to this branch...)

PR #2054, opened 2022-04-18
Changed `SharedBuf` to use a new `BufferState` rather than `FileState`

stevenengler:
Previously the `SharedBuf` used the `FileState::READABLE` state to indicate that the buffer was readable or that it had no writers (and a similar behaviour for the `FileState::WRITABLE` state). This is the behaviour expected by pipes (ex: a pipe should always be readable when there are no writers), but is not the behaviour expected by unix sockets (ex: a unix socket should *not* always be readable when there are no connected sockets). With the new state type `BufferState`, there are two new state flags `NO_READERS` and `NO_WRITERS`. So the object (a pipe or unix socket) using the buffer can listen for the `READABLE`, `WRITABLE`, `NO_READERS`, and `NO_WRITERS` state flags individually (or not at all). For example, unix sockets would ignore these two new flags.

### DIFF
--- a/src/main/host/descriptor/pipe.rs
+++ b/src/main/host/descriptor/pipe.rs
@@ -3,7 +3,7 @@ use nix::errno::Errno;
 use std::sync::Arc;
 
 use crate::cshadow as c;
-use crate::host::descriptor::shared_buf::SharedBuf;
+use crate::host::descriptor::shared_buf::{BufferHandle, BufferState, SharedBuf};
 use crate::host::descriptor::{
     FileMode, FileState, FileStatus, StateEventSource, StateListenerFilter,
 };
@@ -19,7 +19,7 @@ pub struct PipeFile {
     mode: FileMode,
     status: FileStatus,
     write_mode: WriteMode,
-    buffer_event_handle: Option<Handle<(FileState, FileState)>>,
+    buffer_event_handle: Option<BufferHandle>,
     // should only be used by `OpenFile` to make sure there is only ever one `OpenFile` instance for
     // this file
     has_open_file: bool,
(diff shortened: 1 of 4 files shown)


src/main/host/descriptor/pipe.rs
```diff
@@ -301,26 +310,43 @@ impl PipeFile {
...
 
-        // if not open for writing, remove the writable flag
-        if !self.mode.contains(FileMode::WRITE) {
-            state.remove(FileState::WRITABLE);
+        // only update the readable state if the file is open for reading
+        if self.mode.contains(FileMode::READ) {
+            mask.insert(FileState::READABLE);
+            // file is readable if the buffer is readable or there are no writers
+            if buffer_state.intersects(BufferState::READABLE | BufferState::NO_WRITERS) {
+                file_state.insert(FileState::READABLE);
```
sporksmith, line 330: What happens when a read operation is performed on a pipe with no writers? Successful read of 0 bytes? Do we have a test case for this?

src/main/host/descriptor/pipe.rs
```diff
@@ -301,26 +310,43 @@ impl PipeFile {
...
+            }
         }
 
-        state
+        // only update the writable state if the file is open for writing
+        if self.mode.contains(FileMode::WRITE) {
+            mask.insert(FileState::WRITABLE);
+            // file is writable if the buffer is writable or there are no readers
+            if buffer_state.intersects(BufferState::WRITABLE | BufferState::NO_READERS) {
+                file_state.insert(FileState::WRITABLE);
```
sporksmith, line 339: What happens when a write operation is performed on a pipe with no readers? Is the data discarded? Do we have a test case for this?

sporksmith (changes requested):

src/main/host/descriptor/pipe.rs
stevenengler: Yeah it returns 0 (EOF). This should be covered by the two tests `test_read_after_write_close_with_empty_buffer` and `test_read_after_write_close_with_nonempty_buffer`.

src/main/host/descriptor/pipe.rs
stevenengler: The write will return `EPIPE` (and I think also signal `SIGPIPE`, but we don't handle `SIGPIPE` yet in shadow and I think CMake ignores `SIGPIPE` in all tests like the non-shadow/linux tests). This should be covered by the two tests `test_write_after_read_close_with_full_buffer` and `test_write_after_read_close_with_nonfull_buffer`.

src/main/host/descriptor/pipe.rs
sporksmith: How did these tests pass before?

More generally if this change is fixing behavior (and not a pure refactor) then ideally there'd be a new test for the previously-broken behavior. (Or some existing test modified to catch why the previous behavior was wrong)

stevenengler: The pipe behaviour was fine, but the unix socket behaviour was wrong. There aren't currently any tests for sending/receiving on connected unix sockets since shadow doesn't yet support `connect()` for unix sockets, and we can't add tests for this until we support `connect()`. This was just a preemptive bug fix, and should hopefully be tested by the unix socket tests when we add them.

sporksmith: Ah, ok gotcha

sporksmith (approved):

PR #2064, opened 2022-04-20
Allow "null" values for some cli options

stevenengler:
There is currently no way to set a "null" state for cli options with clap, even when that same option supports "null" values in the configuration file. This change allows you to specify "null" for any cli option that uses the `NullableOption` wrapper in the configuration code.

In the shadow configuration, "null" is generally used to mean "disabled". For example, the "template_directory" (`Option<String>`) option is disabled by setting it to "null" (and is set to "null" by default since shadow doesn't use a template directory by default), which corresponds to a `None` value in the rust code.


stevenengler:
@sporksmith Any opinions about whether this behaviour is needed, and/or if it's logical?

sporksmith:
I don't have an immediate need for it for the syscall latency options, since I'll be adding a non-nullable boolean flag to control the feature.

In general I think it makes sense, though, e.g. for the existing options you're converting here

PR #2070, opened 2022-04-20
ci/run.sh: Use --security-opt=seccomp=unconfined

sporksmith:
Docker's seccomp policy disables the `personality` syscall, which causes shadow to be unable to disable ASLR, which causes some tests to fail.

https://docs.docker.com/engine/security/seccomp/#significant-syscalls-blocked-by-the-default-profile

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -148,9 +148,16 @@ EOF
     DOCKER_CREATE_FLAGS=()
     # Shadow needs some space allocated to /dev/shm.
     DOCKER_CREATE_FLAGS+=("--shm-size=1g")
-    # We don't actually need the native CAPS_SYS_PTRACE capability, but when this capability
-    # is added, Docker allows syscalls that shadow needs, such as `ptrace` and `process_vm_readv`.
-    DOCKER_CREATE_FLAGS+=("--cap-add=SYS_PTRACE")
+    # Docker's default seccomp policy disables the `personality` syscall, which
+    # shadow uses to disable ASLR. This causes shadow's determinism tests to fail.
+    # https://github.com/moby/moby/issues/43011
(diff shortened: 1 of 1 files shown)


stevenengler:
Why do the tests fail if it can't disable ASLR? I think shadow should only log a warning if the personality syscall fails, so is something else failing?

sporksmith:
The determinism tests fail because mmap calls return different addresses.

It also appears that when the syscall is blocked, execution stops for ~3s, causing the whole test suite to take much longer, and a few tests to timeout.

Before:

```
$ sudo ci/run.sh -o "debian:10-slim;gcc;debug"
....
Total Test time (real) = 141.46 sec

The following tests FAILED:
	 26 - determinism1-compare-shadow-ptrace (Failed)
	 29 - determinism1-compare-shadow-preload (Failed)
	 32 - determinism2-shadow-ptrace-compare (Failed)
	 35 - determinism2-shadow-preload-compare (Failed)
	 49 - basic-file-transfer-example-shadow-ptrace (Timeout)
	110 - busy_wait-shadow-ptrace (Timeout)
	111 - busy_wait-shadow-preload (Timeout)
Errors while running CTest
```

After:

```
100% tests passed, 0 tests failed out of 220

Label Time Summary:
golang    =  14.32 sec*proc (12 tests)

Total Test time (real) =  31.99 sec
```

sporksmith:
In my other PR, disabling the trace-level logging for these tests was enough to make them *usually* pass, but `basic-file-transfer-example-shadow-ptrace` was still failing sometimes until I disabled seccomp.

sporksmith:
Clarified the comment

stevenengler (approved):

PR #2073, opened 2022-04-21
Add a few empty Rust socket syscall handlers

stevenengler:
This adds Rust syscall handlers for `accept()`, `accept4()`, `connect()`, and `listen()`. If they are called for C descriptors then they forward the arguments to their respective C syscall handler. Otherwise if they are rust descriptors, then they just return `ENOSYS`. This is to prevent the `A descriptor exists for fd=..., but it is not a legacy descriptor. Returning NULL` warning when socket-related syscalls are called on rust sockets that aren't yet completely supported (in this case unix sockets).


src/main/host/syscall/handler/socket.rs
```diff
@@ -448,6 +448,182 @@ impl SyscallHandler {
...
+        };
+
+        // get the socket for the descriptor
+        let _socket = match desc.open_file().inner_file() {
+            GenericFile::Socket(x) => x,
+            _ => return Err(Errno::ENOTSOCK.into()),
+        };
+
+        // TODO: support rust sockets
+        log::debug!(
```
sporksmith, line 620: warning?

src/main/host/syscall/handler/socket.rs
```diff
@@ -585,6 +761,76 @@ impl SyscallHandler {
...
```
sporksmith, line 793: warning?

sporksmith (approved):

src/main/host/syscall/handler/socket.rs
```diff
@@ -448,6 +448,182 @@ impl SyscallHandler {
...
+            "Returning error ENOSYS for listen() syscall on Rust descriptor fd {}",
```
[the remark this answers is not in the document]
stevenengler: Changed to a warning, and changed the text.

Issue #2046, opened 2022-04-13
Find reasonable defaults for modeling time taken by syscalls


sporksmith:
From https://github.com/shadow/shadow/discussions/2038

We currently have features for modeling the time taken by syscalls. However, this is disabled by default. We should do some evaluation to understand how to correctly tune the parameters for this feature, and ideally find reasonable parameters that can be enabled by default.

sporksmith:
We're getting mixed results in the current version of this feature.

In our "benchmark" workflow, enabling this modeling appears to make the simulation slightly more efficient, and doesn't seem to change the simulated results much. The latter is less clear though, since the benchmark is really designed more to evaluate simulation performance, and the simulation results themselves are quite noisy. https://github.com/shadow/benchmark-results/tree/master/tor/2022-04-12-T18-47-32

In a larger simulation using Tor's gitlab infrastructure, this feature appears to make the simulation *much* more expensive, and to very significantly change the simulation results. [pipeline](https://gitlab.torproject.org/jnewsome/sponsor-61-sims/-/pipelines/35774); [tornet.plot.pages.pdf](https://github.com/shadow/shadow/files/8483767/tornet.plot.pages.pdf) 

Not clear what's going on, but maybe it makes more sense to circle back to this after updating this feature to allow more accurate and granular time modeling https://github.com/shadow/shadow/issues/2047.

sporksmith:
vdso-latency=100ns and max-unapplied-cpu-latency=100micros unblocked-syscall-latency=2us looks pretty good: 
[tornet.plot.pages.pdf](https://github.com/shadow/shadow/files/8525611/tornet.plot.pages.pdf).

There are some higher rates and wider confidence intervals in the onion service numbers, but the onion service results are already questionable...

We should probably do another run once final version of the relevant code is actually merged, and maybe try further reducing these latencies, which should further reduce the effect on tor sims.

sporksmith:
We have 3 parameters to potentially tune here, and they are probably not independent.

The unblocked syscall and vdso latencies seem like pretty good starting points, since they're in the ballpark of what they'd be on real systems.

Let's take a closer look at `max-unapplied-cpu-latency`, starting from the snowflake sim, since that was the primary motivation for this feature in the first place.

`max-unapplied-cpu-latency=100micros` turns out to be a bit high here. 0-9us appears to give ~optimal simulation time, with it climbing after that.

```
[...]

PR #2077, opened 2022-04-21
Update syscall latency modeling default parameters

sporksmith:
Update defaults for syscall latency modeling    
    
These defaults significantly improve simulation performance of our test snowflake simulation. On my machine it brings the simulation time down from 55s to 33s. See https://github.com/shadow/shadow/issues/2046#issuecomment-1105765653 through https://github.com/shadow/shadow/issues/2046#issuecomment-1105821624                                
    
This decrease in max-unapplied-cpu-latency does appear to make tor *simulated* performance worse, but not tor *simulation* performance.  i.e. it doesn't make the simulation take longer, but it results in higher transfer latencies and lower goodput. This could use further investigation, especially before defaulting latency modeling on as a whole, but doesn't seem necessarily incorrect. See https://github.com/shadow/shadow/issues/2046#issuecomment-1110414532

### DIFF
--- a/docs/shadow_config_spec.md
+++ b/docs/shadow_config_spec.md
@@ -350,7 +350,7 @@ Which interposition method to use.
 
 #### `experimental.max_unapplied_cpu_latency`
 
-Default: "10 microseconds"  
+Default: "1 microsecond"  
 Type: String
 
 Max amount of execution-time latency allowed to accumulate before the clock is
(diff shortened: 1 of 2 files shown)


robgjansen (approved):
I understand that we might continue to tweak the defaults, but the option name and the improvements for snowflake seem like enough to merge this in the meantime.

robgjansen:
Oh, I think we forgot to update the corresponding default values listed in the doc entries in `docs/shadow_config_spec.md`? Disabling auto-merge until this is addressed.

sporksmith:
> Oh, I think we forgot to update the corresponding default values listed in the doc entries in docs/shadow_config_spec.md?

Oops, done

PR #2082, opened 2022-04-22
nix 0.23.1 -> 0.24.0 with fixes

sporksmith:


sporksmith:
@stevenengler There are some build warnings about deprecated socket APIs, but leaving them for now since it looks like it's touching some of the code you're actively working on. e.g.:

```
warning: use of deprecated enum `nix::sys::socket::SockAddr`: use SockaddrLike or SockaddrStorage instead
  --> main/host/descriptor/socket/mod.rs:13:23
   |
13 | use nix::sys::socket::SockAddr;
   |                       ^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default
```

sporksmith:
Looks like `basic-file-transfer` is failing. Investigating...

src/main/host/memory_manager/memory_copier.rs
```diff
@@ -178,8 +176,8 @@ impl MemoryCopier {
 
         let towrite = src.len();
         trace!("write_ptr writing {} bytes", towrite);
-        let local = [nix::sys::uio::IoVec::from_slice(src)];
-        let remote = [nix::sys::uio::RemoteIoVec {
+        let local: [std::io::IoSlice; 1] = [std::io::IoSlice::new(src)];
+        let remote: [nix::sys::uio::RemoteIoVec; 1] = [nix::sys::uio::RemoteIoVec {
             base: usize::from(dst.ptr()),
             len: towrite,
```
stevenengler, line 182: Why are the types specified here? (I don't mind having the type specified, I'm just wondering if there's a reason for it.)

src/test/ifaddrs/test_ifaddrs.rs
```diff
@@ -15,7 +15,7 @@ fn main() {
 
     let mut ip_vec = vec![];
     for ifaddr in addrs {
-        if let Some(nix::sys::socket::SockAddr::Inet(address)) = ifaddr.address {
+        if let Some(address) = ifaddr.address {
             let address_str = address.to_string();
             let ip = address_str.split(":").collect::<Vec<&str>>()[0];
             println!(
```
stevenengler, line 21: I think this code will have weird behaviour if it's an IPv6 address (for example where colons in an IPv6 address like `fe80::1ff:fe23::` aren't used to split the port). Maybe this should have an extra `if let Some(address) = address.as_sockaddr_in() {...}` check?

stevenengler (approved):
> @stevenengler There are some build warnings about deprecated socket APIs, but leaving them for now since it looks like it's touching some of the code you're actively working on. e.g.:

I'll make a follow up PR to "allow" deprecation warnings everywhere that we currently use the old socket address types. I'm not sure yet how to switch to the new types, and I'd like to do that after unix sockets are done. And there are a lot of warnings generated by this.

src/main/host/memory_manager/memory_copier.rs
sporksmith: It's probably overkill; removed.

I have VSC set up to show me the types in places like this, but leaving them implicit sometimes makes the code less readable outside of VSC. I think I experimented with the shortcut to make the types explicit while I was editing this, but it's not really needed in this case since they're already explicit on the RHS

src/main/host/syscall/handler/socket.rs
```diff
@@ -199,7 +199,12 @@ impl SyscallHandler {
             }
         };
 
-        let supported_flags = MsgFlags::MSG_DONTWAIT;
+        // MSG_DONTWAIT is currently a no-op, since we haven't implemented the behavior
+        // it's meant to disable.
+        // TODO: Once we've implemented generating a SIGPIPE when the peer on a
+        // stream-oriented socket has closed the connection, MSG_NOSIGNAL should
+        // disable it.
+        let supported_flags = MsgFlags::MSG_DONTWAIT | MsgFlags::MSG_NOSIGNAL;
```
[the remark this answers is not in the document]
sporksmith: Good catch! Fixed

src/test/ifaddrs/test_ifaddrs.rs
sporksmith: Done. Rearranged a bit to avoid excessive nesting.

src/main/host/syscall/handler/socket.rs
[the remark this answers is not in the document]
sporksmith: > Is there a reason we're adding this flag now? Does something break if we don't allow and ignore it?

Forgot to respond to this part -

Apparently nix considered this an invalid flag before, which we only `warn` for. Now that `nix` recognizes the flag, we started returning an error instead.

PR #2085, opened 2022-04-25
Calculate host RNG seeds more consistently

### LINKED ISSUE #2074: Derive host RNG seed from the hostname
Shadow currently derives the host RNG seed from the global RNG by iterating over all hosts in alphanumeric order, and for each host getting a new value from the global RNG. This behaviour can lead to unexpected simulation changes when hosts are renamed, added, or removed. For example, adding a new host can change the seed of other hosts that are ordered after it.

Instead we should get a single value from the global RNG and for each host, xor that value with a hash of the hostname since hostnames are required to be unique. Changing a host's hostname will still change that host's seed, but adding or removing other hosts would no longer affect the host's seed.

This is not technically a determinism issue since we don't try to have any determinism guarantees between two different shadow configs, but it's related to determinism, so it's going in the determinism milestone.

The documentation at https://shadow.github.io/docs/guide/shadow_config_spec.html#hosts would also need to be updated.

stevenengler:
Host RNG seeds now vary only based on their hostname (and the global seed) and not the order of hosts. This means you can add or remove hosts without affecting the seeds of other hosts.

@robgjansen You told me not to work on this for 2.1, but I wrote it on a plane so I don't think it counts :)

(I did check that each host was being given a different seed.)

Closes #2074.

### DIFF
--- a/docs/shadow_config_spec.md
+++ b/docs/shadow_config_spec.md
@@ -634,13 +634,10 @@ configuration, with the field name being used as the network hostname.
 
 Shadow assigns each host to a network node in the [network graph](network_graph_overview.md).
 
-When Shadow iterates over these hosts, Shadow iterates over them in the order
-of their sorted hostnames. While this implementation detail is generally
-unimportant, it's important to note that Shadow assigns each host its own RNG.
-These host-specific RNGs are seeded from a global RNG, so the seed used for
-each host depends on the order that Shadow iterates over the hosts. This means
-that changing a hostname or adding a new host may change the seed used for the
-hosts' RNGs, and may subtly affect the simulation results.
+In Shadow, each host is given an RNG whose seed is derived from the global seed
+([`general.seed`](#generalseed)) and the hostname. This means that changing a
+host's name will change that host's RNG seed, subtly affecting the simulation
+results.
 
 #### `hosts.<hostname>.bandwidth_down`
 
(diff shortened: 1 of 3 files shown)


src/main/utility/hash.rs
```diff
@@ -0,0 +1,18 @@
+mod export {
+    use std::collections::hash_map::DefaultHasher;
+    use std::ffi::CStr;
+    use std::hash::Hasher;
+
+    /// An easy way for C code to hash a string.
```
robgjansen, line 6: I think we could also just use `g_str_hash` from glib in `c` code for now if we wanted to avoid exporting another function.

robgjansen (approved):

docs/shadow_config_spec.md
```diff
@@ -634,13 +634,10 @@ configuration, with the field name being used as the network hostname.
...
 Shadow assigns each host to a network node in the [network graph](network_graph_overview.md).
 
-When Shadow iterates over these hosts, Shadow iterates over them in the order
-of their sorted hostnames. While this implementation detail is generally
-unimportant, it's important to note that Shadow assigns each host its own RNG.
-These host-specific RNGs are seeded from a global RNG, so the seed used for
-each host depends on the order that Shadow iterates over the hosts. This means
-that changing a hostname or adding a new host may change the seed used for the
-hosts' RNGs, and may subtly affect the simulation results.
+In Shadow, each host is given an RNG whoose seed is derived from the global
```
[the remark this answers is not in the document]
stevenengler: Fixed.

src/main/utility/hash.rs
stevenengler: Good idea, I didn't know about that function.

Issue #1958, opened 2022-03-09
Support for iPerf2


stevenengler:
iPerf2 fails to run in Shadow due to missing support for the [`setitimer`](https://linux.die.net/man/2/setitimer) syscall. A workaround is to set `#define HAVE_SETITIMER 0` before building.

stevenengler:
We now support this syscall, and modern versions of iPerf2 seem to work in Shadow. We also have a section in our compatibility notes: https://shadow.github.io/docs/guide/compatibility_notes.html#iperf-2

PR #2089, opened 2022-04-26
Add test_itimer and ensure_ord

sporksmith:
This sketches out tests for interval timers as manipulated by getitimer
and setitimer, and enables those in Linux.

Also adds a helper macro `ensure_ord` to succinctly create anyhow
"assertions" (ensure-tions?) with useful error messages.

Progress on #1889 and #1958


src/test/itimer/test_itimer.rs
```diff
@@ -0,0 +1,330 @@
...
+
+// Neither `libc` nor `nix` wrap `setitimer`.
+fn setitimer(which: i32, new_value: &libc::itimerval) -> nix::Result<ITimer> {
+    let mut old_value: libc::itimerval = unsafe { MaybeUninit::zeroed().assume_init() };
+    if unsafe {
+        libc::syscall(
+            libc::SYS_setitimer,
+            which,
+            new_value as *const _,
+            &mut old_value as *mut _,
```
stevenengler, line 45: nit: I think newer versions of rust don't require the `*const _` and `*mut _`, so removing these might make it fit on one line.

src/test/itimer/test_itimer.rs
```diff
@@ -0,0 +1,330 @@
...
+        libc::ITIMER_REAL,
+        &libc::itimerval {
+            it_value,
+            it_interval,
+        },
+    )?;
+    // The man page isn't clear about what should happen here,
+    // but experimentally the timer is completely disabled, even though we
+    // specified a non-zero interval.
+    ensure_ord!(getitimer(libc::ITIMER_REAL)?, ==, ITimer{value: TimeVal::zero(), interval: TimeVal::zero()});
```
stevenengler, line 271: Maybe check that `SIGNAL_CTR` is still zero (I would expect it to be, but just to make sure)?

src/test/itimer/test_itimer.rs
sporksmith: Oh nice, done

src/test/itimer/test_itimer.rs
sporksmith: Done

PR #2092, opened 2022-04-26
Add "#[allow(deprecated)]" wherever we use deprecated nix socket address types/fns

stevenengler:
I don't know of a nicer way to do this, since I don't see a way to allow *specific* deprecated types/fns globally.

Fixing the deprecation issues is in #2093.

### DIFF
--- a/src/main/host/descriptor/socket/mod.rs
+++ b/src/main/host/descriptor/socket/mod.rs
@@ -10,6 +10,8 @@ use crate::utility::event_queue::EventQueue;
 
 use unix::UnixSocketFile;
 
+// https://github.com/shadow/shadow/issues/2093
+#[allow(deprecated)]
 use nix::sys::socket::SockAddr;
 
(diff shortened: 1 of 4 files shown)


src/main/host/descriptor/socket/mod.rs
```diff
@@ -10,6 +10,7 @@ use crate::utility::event_queue::EventQueue;
 
 use unix::UnixSocketFile;
 
+#[allow(deprecated)]
```
sporksmith, line 13: It's probably worth having a comment with the link to #2093 in all of these places. Just something like `TODO: https://github.com/shadow/shadow/issues/2093`

sporksmith (approved):

src/main/host/descriptor/socket/mod.rs
stevenengler: Added.

Issue #2091, opened 2022-04-26
Delivery of "synchronous" signals in preload mode (SIGSEGV, SIGILL)


sporksmith:
In preload mode, we emulate `sigaction`, which prevents handlers from being installed natively. While we implement delivery of signals to those handlers for signals arriving via syscalls, we do not arrange to propagate signals that the host OS raises. e.g. if the managed thread executes an illegal instruction, the host Linux kernel will raise a native SIGILL, which will be delivered natively, killing the process.  This is *usually* what would happen anyway, but a signal handler installed by the managed process to catch and "swallow" the SIGILL won't be invoked.

This is probably more common for SIGSEGV, where the signal handler might allocate memory or call `mprotect` to make the original access legal, and then return instead of crashing. This case is also slightly further complicated because in preload mode we do install our own SIGSEGV handler, but it only handles SIGSEGV's that were raised from executing RDTSC, and crashes on any other SIGSEGV without delegating to a signal handler that the managed process may have installed.

This issue appears to affect OpenJDK. https://github.com/shadow/shadow/discussions/2084

sporksmith:
Partly working in #1917, but seems to be breaking thread local storage after a sigsegv is handled, which generally results in a crash not long after, particularly if new threads are created afterwards. Investigating.

sporksmith:
Build the jdk with debug symbols and poked around a little. It looks like in #2084 the SIGSEGV we see is a NULL pointer dereference, which the signal handler transforms into a NullPointerException.

https://github.com/openjdk/jdk/blob/739769c8fc4b496f08a92225a12d07414537b6c0/src/hotspot/os_cpu/linux_x86/os_linux_x86.cpp

PR #2096, opened 2022-04-26
Add test for handling sigsegv

sporksmith:
Progress on https://github.com/shadow/shadow/issues/2091

### DIFF
--- a/src/test/signal/test_signals.rs
+++ b/src/test/signal/test_signals.rs
@@ -4,6 +4,7 @@ use nix::sys::signal::Signal;
 use nix::unistd;
 use once_cell::sync::OnceCell;
 use signal_hook::low_level::channel::Channel as SignalSafeChannel;
+use std::arch::asm;
 use std::error::Error;
 use std::iter::Iterator;
 use std::os::unix::io::RawFd;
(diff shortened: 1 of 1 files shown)


src/test/signal/test_signals.rs
```diff
@@ -888,6 +889,61 @@ fn test_sigaltstack_autodisarm() -> Result<(), Box<dyn Error>> {
...
+
+    // Install our SIGSEGV handler, which is going to mutate registers in the
+    // caller to fix the null pointer dereference below.
+    // This may seem esoteric, but this is sometimes done to implement custom
+    // memory management. Observed in OpenJDK in particular - see
+    // https://github.com/shadow/shadow/issues/2094.
+    unsafe {
+        signal::sigaction(
+            signal::SIGSEGV,
+            &signal::SigAction::new(
```
stevenengler, line 925: Unregister this handler at the end of the test to prevent it from causing weird behaviour in later tests?

src/test/signal/test_signals.rs
```diff
@@ -888,6 +889,61 @@ fn test_sigaltstack_autodisarm() -> Result<(), Box<dyn Error>> {
...
 
+static GLOBAL_STATIC: u32 = 0xdeadbeef;
+extern "C" fn change_rax_from_null_to_global_static(
+    signal: i32,
+    info: *mut libc::siginfo_t,
+    voidctx: *mut std::ffi::c_void,
+) {
+    assert_eq!(signal, signal::SIGSEGV as i32);
+
+    let info = unsafe { info.as_mut().unwrap() };
```
stevenengler, line 900: I don't think this reference needs to be mutable?

stevenengler (approved):

src/test/signal/test_signals.rs
sporksmith: Done

src/test/signal/test_signals.rs
sporksmith: Good point; done

PR #2097, opened 2022-04-27
Support listen() for unix sockets

stevenengler:
Added support for the listen() syscall, updated all of the existing listen() tests to test unix sockets, and added a (currently disabled in shadow) additional listen() test.

### DIFF
--- a/src/main/host/descriptor/socket/unix.rs
+++ b/src/main/host/descriptor/socket/unix.rs
@@ -218,6 +218,15 @@ impl UnixSocketFile {
             .ioctl(&mut self.common, request, arg_ptr, memory_manager)
     }
 
+    pub fn listen(
+        &mut self,
+        backlog: i32,
+        event_queue: &mut EventQueue,
+    ) -> Result<(), SyscallError> {
+        self.protocol_state
+            .listen(&mut self.common, backlog, event_queue)
+    }
+
     pub fn connect(
         socket: &Arc<AtomicRefCell<Self>>,
         addr: nix::sys::socket::UnixAddr,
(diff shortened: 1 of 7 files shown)


src/main/host/descriptor/socket/unix.rs
```diff
@@ -1320,6 +1444,17 @@ impl UnixSocketCommon {
     }
 }
 
+fn backlog_to_queue_size(backlog: i32) -> u32 {
+    // linux also makes this cast, so negative backlogs wrap around to large positive backlogs
+    // https://elixir.free-electrons.com/linux/v5.11.22/source/net/unix/af_unix.c#L628
```
sporksmith, line 1449: Is this a bug or a feature in Linux? If it's a bug, maybe worth logging a warning when backlog is negative?

src/main/host/descriptor/socket/unix.rs
```diff
@@ -1320,6 +1444,17 @@ impl UnixSocketCommon {
...
+    let backlog = backlog as u32;
+
+    // the linux '__sys_listen()' applies the somaxconn max to all protocols, including unix sockets
+    let queue_limit = std::cmp::min(backlog, unsafe { c::SHADOW_SOMAXCONN }.try_into().unwrap());
+
+    queue_limit.saturating_add(1)
```
sporksmith, line 1455: Why +1? (Probably worth a comment)

sporksmith (approved):

src/main/host/descriptor/socket/unix.rs
stevenengler: I don't think it's a bug or a feature, just a legacy artifact that probably can't be changed without breaking some userland programs. I don't think a warning is needed as it's not a bug or issue in shadow.

src/main/host/descriptor/socket/unix.rs
stevenengler: Just because that's what Linux does :) But I'll add a comment here.

PR #2111, opened 2022-05-02
Clean up and extend time types

sporksmith:
Replaced `process_readTimeSpec` with helpers in Rust for converting to and from `libc::timespec`, and exposed those back to C. Also added conversions for `libc::timeval`, which we'll need for the itimer syscalls.

While writing the unit tests I decided that having the Rust SimulationTime support a wider range of values than the C SimulationTime is too likely to cause confusion. I changed it to wrap a C SimulationTime, and likewise for the Rust EmulationTime.

I think it's probably worth using SimulationTime broadly in Rust code after all rather than using Duration. e.g. here it was useful because we can implement the `From` trait between `SimulationTime` and `libc::timeval` and `libc::timespec`. We can't do the same for `Duration` since we don't own the type.

### DIFF
--- a/src/main/core/support/emulated_time.rs
+++ b/src/main/core/support/emulated_time.rs
@@ -5,16 +5,12 @@ Deals with instances of time in a Shadow simulation.
 use crate::core::support::simulation_time;
 use crate::core::support::simulation_time::SimulationTime;
 use crate::cshadow as c;
-use std::time::Duration;
 
 /// An instant in time (analagous to std::time::Instant) in the Shadow
 /// simulation.
 // Internally represented as Duration since the Unix Epoch.
 #[derive(Copy, Clone, Eq, PartialEq, Debug, PartialOrd, Ord)]
-pub struct EmulatedTime(Duration);
-
-/// The  Unix epoch (00:00:00 UTC on 1 January 1970)
-pub const UNIX_EPOCH: EmulatedTime = EmulatedTime(Duration::from_secs(0));
+pub struct EmulatedTime(c::EmulatedTime);
 
 // Duplicated from the EMULATED_TIME_OFFSET macro in definitions.h.
 pub(super) const SIMULATION_START_SEC: u64 = 946684800;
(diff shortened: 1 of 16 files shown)


src/main/host/syscall/futex.c
```diff
@@ -93,15 +98,15 @@ static SysCallReturn _syscallhandler_futexWaitHelper(SysCallHandler* sys, Plugin
...
     SysCallCondition* cond = syscallcondition_new(trigger);
-    if (timeoutVPtr.val) {
-        SimulationTime timeoutSimulationTime = timeout.tv_sec * SIMTIME_ONE_SECOND + timeout.tv_nsec * SIMTIME_ONE_NANOSECOND;
-        EmulatedTime timeoutEmulatedTime =
-            (type == TIMEOUT_RELATIVE) ? timeoutSimulationTime + worker_getCurrentEmulatedTime()
-                                       : timeoutSimulationTime;
+    if (timeoutSimTime != SIMTIME_INVALID) {
+        EmulatedTime timeoutEmulatedTime = (type == TIMEOUT_RELATIVE)
+                                               ? timeoutSimTime + worker_getCurrentEmulatedTime()
+                                               : timeoutSimTime;
```
stevenengler, line 109: Use the conversion macro here?

src/main/core/support/emulated_time.rs
```diff
@@ -5,74 +5,99 @@ Deals with instances of time in a Shadow simulation.
...
 pub const EMUTIME_MAX: c::EmulatedTime = u64::MAX - 1;
 pub const EMUTIME_MIN: c::EmulatedTime = 0;
 
-/// The start time of the simulation - 00:00:00 UTC on 1 January, 2000.
-pub const SIMULATION_START: EmulatedTime = EmulatedTime(Duration::from_secs(SIMULATION_START_SEC));
-
 impl EmulatedTime {
+    /// The start time of the simulation - 00:00:00 UTC on 1 January, 2000.
+    pub const SIMULATION_START: Self =
+        Self(SIMULATION_START_SEC * simulation_time::SIMTIME_ONE_SECOND);
```
stevenengler, line 24: I never knew that you could put constants in type impls.

src/main/core/support/emulated_time.rs
```diff
@@ -5,74 +5,99 @@ Deals with instances of time in a Shadow simulation.
...
     pub fn to_abs_simtime(self) -> SimulationTime {
-        SimulationTime::from(self.duration_since(&SIMULATION_START))
+        SimulationTime::from(self.duration_since(&Self::SIMULATION_START))
     }
 
-    /// Returns the duration since `earlier`, or panics if `earlier` is after `self`.
-    pub fn duration_since(&self, earlier: &EmulatedTime) -> Duration {
+    /// Returns the duration since `earlier`, or panics if `earlier` is after `self`, or
+    /// if the difference can't be represented as SimulationTime.
+    pub fn duration_since(&self, earlier: &EmulatedTime) -> SimulationTime {
```
stevenengler, line 61: It still seems useful to be able to get durations from two `SimulationTime`s or two `EmulatedTime`s, and to have a distinction between durations and instances in time. But maybe that can be improved in the future when we don't need to worry about C types.

stevenengler (approved):

src/main/core/support/emulated_time.rs
sporksmith: > and to have a distinction between durations and instances in time. 

Yeah, that's the idea - `SimulationTime` is always a duration (except for some "legacy" uses in the C code), and `EmulatedTime` is always an instant. It might make sense to rename them something like `SimulatedDuration` and `SimulatedInstant` respectively, but probably not without likewise renaming the C types, which we probably shouldn't do before getting rid of all the remaining instances of `SimulationTime` being used to represent an instant in the C code.

> It still seems useful to be able to get durations from two SimulationTimes or two EmulatedTimes,

We can get a duration between two EmulatedTimes by using `EmulatedTime::duration_since` and `From`/`Into`. e.g. `Duration d = myEmulatedTime.duration_since(EmulatedTime::UNIX_EPOCH).into()`

PR #2116, opened 2022-05-03
Changed the process of registering as a reader/writer for a buffer

stevenengler:
The `SharedBuf` type allows objects to register as a reader or writer with `add_reader(event_queue)` and `add_writer(event_queue)` and deregister with `remove_reader(event_queue)` and `remove_writer(event_queue)`, but this can be problematic if an object forgets to deregister as a reader/writer before it's dropped. Registering as a reader/writer is useful for types like Pipe, where each end of the pipe needs to know whether the other end has a reader/writer.

Instead, the `add_reader(event_queue)` and `add_writer(event_queue)` now return handles that must be returned to the buffer with `remove_reader(handle, event_queue)` and `remove_writer(handle, event_queue)`. Otherwise if they don't and those handles are dropped, they'll log a warning (and panic in debug mode). The `SharedBuf` will also warn (and panic in debug mode) if the buffer is dropped while it still has a non-zero number of readers or writers.

Also fixes two instances where a unix socket was being dropped without being closed properly.

In the future maybe there's a better design where a buffer is split into two types for the read and write ends, but for now I think this is an improvement.

### DIFF
--- a/src/main/host/descriptor/socket/unix.rs
+++ b/src/main/host/descriptor/socket/unix.rs
@@ -336,8 +336,8 @@ struct ConnOrientedConnected {
     peer_addr: nix::sys::socket::UnixAddr,
     send_buffer: Arc<AtomicRefCell<SharedBuf>>,
     writer_handle: WriterHandle,
-    #[allow(dead_code)]
-    send_buffer_handle: BufferHandle,
+    // this handle is never accessed, but we store it because of its drop impl
+    _send_buffer_handle: BufferHandle,
 }
 struct ConnOrientedClosed {}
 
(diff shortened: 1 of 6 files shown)


src/main/host/descriptor/socket/unix.rs
```diff
@@ -330,16 +336,19 @@ struct ConnOrientedConnected {
...
 }
 struct ConnOrientedClosed {}
 
 struct ConnLessInitial {
     bound_addr: Option<nix::sys::socket::UnixAddr>,
     peer_addr: Option<nix::sys::socket::UnixAddr>,
     send_buffer: Option<Arc<AtomicRefCell<SharedBuf>>>,
-    send_buffer_event_handle: Option<BufferHandle>,
+    writer_handle: Option<WriterHandle>,
+    #[allow(dead_code)]
```
sporksmith, line 350: Needed? Or left over from an intermediate state? If it is needed, a brief comment about why and/or when to remove it would be helpful

sporksmith (approved):

src/main/host/descriptor/socket/unix.rs
stevenengler: It seems like it's recommended to use an underscore in this case rather than a `dead_code` warning, so I did that instead. https://github.com/rust-lang/rust/issues/21775

PR #2121, opened 2022-05-04
Split Timer into Timer and TimerFd

sporksmith:
This is a refactor we've considered for a while, and makes sense to do before adding more internal Timers and/or migrating Timer to Rust.

Timer was previously used to represent a timer file descriptor as returned by `timerfd_create`. In places where we didn't need the file descriptor but did need a timer with similar behavior, we used Timer with a fake descriptor number.

This change:

* Renames `Timer` -> `TimerFd`.
* Extracts the core timer functionality from `TimerFd` into `Timer`.
* Changes the `Timer` APIs to use Shadow's EmulatedTime and SimulatedTime types instead of libc's `struct itimerspec`.
* Replaces the `TimerFd` in `SyscallCondition` with a `Timer`.

Reviewing the diff as a whole in the GH UI is probably a bit tricky, since it doesn't handle moving code from one file to another very well. I'd recommend either going through the individual commits, or at least chunking together the commits before and after "Move Timer into its own module".


src/main/host/descriptor/timerfd.c
```diff
@@ -61,28 +100,50 @@ static void _timerfd_close(LegacyDescriptor* descriptor, Host* host) {
...
+    MAGIC_INIT(timerfd);
 
-    descriptor_init(&(timer->super), DT_TIMER, &_timerfdFunctions);
-    descriptor_adjustStatus(&(timer->super), STATUS_DESCRIPTOR_ACTIVE, TRUE);
+    descriptor_init(&(timerfd->super), DT_TIMER, &_timerfdFunctions);
+    descriptor_adjustStatus(&(timerfd->super), STATUS_DESCRIPTOR_ACTIVE, TRUE);
+
+    descriptor_refWeak(timerfd);
+    Task* task = task_new(_timerfd_expire, timerfd, NULL, descriptor_unrefWeak, NULL);
+    timerfd->timer = timer_new(task);
```
stevenengler, line 84: Why is this task needed when it wasn't before? (Before this PR there was only a single `task_new()` call in `timer.c`, but this PR has a `task_new()` call in both `timerfd.c` and `timer.c`).

src/main/host/timer.c
```diff
@@ -0,0 +1,187 @@
...
+    }
+}
+
+Timer* timer_new(Task* task) {
+    Timer* rv = g_new(Timer, 1);
+    *rv = (Timer){.task = task,
+                  .referenceCount = 1,
+                  .nextExpireTime = EMUTIME_INVALID,
+
+                  MAGIC_INITIALIZER};
```
stevenengler, line 62: nit: Is the extra line intended?

stevenengler (approved):
I don't really follow most of the changes in this PR, but I'm a fan of splitting up the `Timer` into a separate `TimerFd` so looks good to me :)

src/main/host/descriptor/timerfd.c
sporksmith: Before there was a single task object that both did the timer stuff (check whether it had been cancelled, whether it should fire now, whether it should reschedule, etc) and the timerfd stuff (notify descriptor listeners).

Now there's a Timer task that does the timer stuff (and whatever extra task you hand it), and a TimerFd task that notifies descriptor listeners

src/main/host/timer.c
sporksmith: Done

PR #2128, opened 2022-05-06
Move quickstart, add context to quicklinks

robgjansen:


stevenengler (approved):
Should the quickstart also have a link to the dependencies page before the code block? Otherwise anyone following the quickstart will see a bunch of build failures due to missing dependencies, and it will turn into more of a "slowstart" :)

PR #2137, opened 2022-05-09
Partially implement rseq and enable Ubuntu 22.04

sporksmith:
Implements enough of rseq to get tests to pass in Ubuntu 22.04, and a little bit of input validation.

Most of the potentially observable missing functionality is returning 0 when we ought to return an error, such as if the caller tries to unregister an rseq without having already registered one.

The "core" functionality of rseq isn't applicable in Shadow - this is keeping an address up to date with which CPU the thread is running on (we can just always report 0), and taking action if we preempt the thread in the middle of a restartable sequence. The latter never happens since we don't reschedule threads other than when a syscall is made, and syscalls are not allowed inside restartable sequences.

### DIFF
--- a/docs/supported_platforms.md
+++ b/docs/supported_platforms.md
@@ -2,7 +2,7 @@
 
 ## Officially supported platforms
 
-  + Ubuntu 18.04 and 20.04
+  + Ubuntu 18.04, 20.04, 22.04
   + Debian 10 and 11
   + Fedora 34 and 35
   + CentOS Stream 8
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -36,6 +36,8 @@ jobs:
           - 'ubuntu:20.04'
           # End of standard support: July 2022 https://wiki.ubuntu.com/Releases
           - 'ubuntu:21.10'
+          # End of standard support: April 2027 https://wiki.ubuntu.com/Releases
+          - 'ubuntu:22.04'
           # EOL ~August 2022 https://wiki.debian.org/DebianReleases
           - 'debian:10-slim'
           - 'debian:11-slim'
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -10,6 +10,7 @@ CONTAINERS=(
     ubuntu:18.04
     ubuntu:20.04
     ubuntu:21.10
+    ubuntu:22.04
     debian:10-slim
     debian:11-slim
     fedora:34


src/main/host/syscall/handler/sched.rs
```diff
@@ -0,0 +1,89 @@
...
+        }
+        let mem = ctx.process.memory_mut();
+        let mut rseq = mem.memory_ref_mut(rseq_ptr)?;
+
+        // rseq is mostly unimplemented, but also mostly unneeded in Shadow.
+        // We'd only need to implement the "real" functionality if we ever implement
+        // true preemption, in which case we'd need to do something if we ever pre-empted
+        // while the user code was in a restartable sequence. As it is, Shadow only
+        // reschedules threads at system calls, and system calls are disallowed inside
+        // restartable sequences.
```
stevenengler, line 80: I think the thread can be pre-empted for another reason though, for example in Shadow I think this can happen for signals generated by hardware errors? For example, a critical section dereferences an invalid pointer, the sigsegv handler runs immediately, and then we should jump to the abort instruction once the handler completes.

This probably rarely happens in practice so I don't think we need to handle it, but recovering from signals during a restartable sequence is something Linux does support. Is there a way we can warn on the incorrect behaviour anyways? For example, can we store the `struct rseq` pointer in the `SysCallHandler`, and before running a signal handler read the plugin's `struct rseq_cs` to make sure we're not in a restartable sequence? Is it even straightforward for the shim to know the instruction address before the signal was invoked (is it in one of the signal context objects)?

I'm also fine with not handling this at all, but maybe add a comment that we don't handle this and that weird things might happen.

docs/supported_platforms.md
```diff
@@ -2,7 +2,7 @@
 
 ## Officially supported platforms
 
-  + Ubuntu 18.04 and 20.04
+  + Ubuntu 18.04, 20.04, 21.10, 22.04
```
stevenengler, line 5: I think we had decided to not officially support 21.10 since its lifespan is so short and we only wanted to support LTS releases. I think we decided that stopping support for a platform would be considered a breaking change, so I think it would be better to leave it as unsupported (plus it leaves more room in the CI for future LTS platforms that we'll also need to support).

stevenengler (approved):

src/main/host/syscall/handler/sched.rs
sporksmith: Good point; I was thinking triggering a hardware signal inside an rseq might be undefined behavior the same way making a syscall is, but I'm not really sure.

Conceptually it wouldn't be too difficult to implement more fully, but it's a bit of a rabbit hole right now. I think we'd want to store the `struct rseq` in shared memory in `ShimThreadProtectedSharedMem`, so that the shim-side signal handling code can get to it. We haven't set up bindings for that or any of the other `shadow-shim-helper` yet. I *think* we can probably just rewrite that whole library in Rust and expose back to C rather than investing in safe Rust wrappers, but again it's a bit of a rabbit hole.

Opened https://github.com/shadow/shadow/issues/2139 to track

PR #2143, opened 2022-05-11
Task rs no shared

sporksmith:


sporksmith:
I think we might want shared ownership after all. In particular `Timer` currently executes its "callback" task itself, but it'd be safer to run it as its own Event in case the callback end up manipulating the Timer itself. For repeating Timers, this requires cloning the Task.

stevenengler:
> In particular Timer currently executes its "callback" task itself, but it'd be safer to run it as its own Event in case the callback end up manipulating the Timer itself.

The `EventQueue` should help with these recursive callbacks, and would avoid the overhead of adding a new event to the scheduler.

sporksmith:
I think we'd have the  same issue there - the `EventQueue` takes ownership of the callback. For `Timer` to schedule repeating events it needs to be able to clone its callback (`TaskRef`)

PR #2147, opened 2022-05-11
SyncSendPointer: be more explicit about safety

sporksmith:
Make SyncSendPointer unsafe and add HostObjectPointer    
    
Constructing a `SyncSendPointer` implies that the object pointed to by it as actually `Sync` and `Send`. Updated docs and marked unsafe to reflect that.
    
In most places we were using `SyncSendPointer`, the pointer is *not* actually fully thread-safe, and we instead rely on the Host lock being held at access time. Created `HostObjectPointer` for such pointers.

### DIFF
--- a/src/main/core/work/task.rs
+++ b/src/main/core/work/task.rs
@@ -45,7 +45,7 @@ pub mod export {
     use crate::{
         cshadow,
         host::host::{Host, HostId},
-        utility::SyncSendPointer,
+        utility::{notnull::notnull_mut, HostTreePointer},
     };
 
     pub type TaskCallbackFunc =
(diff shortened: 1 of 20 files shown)


sporksmith:
I considered pushing the SyncSendPointer and/or `unsafe` markers further up the call stack but didn't want to get too carried away...

Thinking about it some more though, maybe we should have a type like `HostTreeObjPtr` that plays a similar role as `SyncSendPointer`, and maybe replaces it, which is explicitly associated with a Host and accessed only with the Host lock held. Wdyt?

sporksmith:
As with the checks in C-constructed `TaskRef`s, we could have runtime checks that the lock is held, but of course can't guard against illegal aliasing / recursive access

sporksmith:
Updated PR and description per discussion. I ended up keeping SyncSendPointer, which we currently only need for ChildPidWatcher, which executes callbacks on its own thread but only ever uses IPCData*, which actually is thread safe.

If we like the approach for HostObjectPointer, I think we could potentially use a similar construction to make e.g. `HostRc` and `HostRefcell` that are marked `Send` and `Sync`, but instead of using atomics validate that the lock of the Host associated with them is held.

src/main/core/work/task.rs
```diff
@@ -197,20 +172,50 @@ pub mod export {
...
         Box::into_raw(Box::new(task.clone()))
     }
 
     /// Destroys this reference to the `Task`, dropping the `Task` if no references remain.
+    ///
+    /// SAFETY: Host lock for the TaskRef's `object` and `argument` must be held.
     #[no_mangle]
     pub unsafe extern "C" fn taskref_drop(task: *mut TaskRef) {
         unsafe { Box::from_raw(task) };
     }
```
stevenengler, line 214: There should be a null pointer check somewhere I think. (Or we could specify in "SAFETY" that the ptr must be valid, but I think a check would be better.)

src/main/utility/mod.rs
```diff
@@ -19,23 +19,75 @@ pub mod stream_len;
...
+    ///
+    /// SAFETY: Pointer must only be dereferenced with the Host lock held, in
+    /// addition to the normal safety requirements for dereferencing a pointer.
+    pub unsafe fn ptr(&self) -> *mut T {
+        #[cfg(debug_assertions)]
+        Worker::with_active_host_info(|i| {
+            debug_assert_eq!(self.host_id, i.id);
+        });
+        self.ptr
     }
```
stevenengler, line 91: Why make this a debug-only check? Rust types that check safety always run in release mode as well, so I don't see a reason not to here. It would also allow us to make `ptr()` safe. Or what about making it return a `Result<*mut T, ()>` that the calling code needs to `unwrap()`?

stevenengler, line 91: It would be nice if this had a fallback if a worker hasn't been initialized. Otherwise any unit tests that somewhere internally use a `HostTreePointer` will unexpectedly crash if the test didn't set up this global worker beforehand.

src/main/utility/mod.rs
```diff
@@ -19,23 +19,75 @@ pub mod stream_len;
...
+            #[cfg(debug_assertions)]
+            host_id,
+            ptr,
+        }
+    }
+
+    /// Create a pointer for the current host.
+    ///
+    /// SAFETY: see `new_for_host`.
+    pub unsafe fn new(ptr: *mut T) -> Self {
```
stevenengler, line 76: Does this need to be `unsafe`? The `Worker::with_active_host_info()` below is already checking for the host lock and binding this to that specific host, so there aren't any other guarantees that the calling code needs to uphold.

sporksmith: My thinking *was* that the runtime checks would only be enabled in debug builds, as for the Magic checks. The debug checks are a best effort to catch mistakes, but it's still the user's responsibility to avoid mistakes.

On further thought and after having worked out the more complete design of this idea, I think I agree though that it's better to think of these checks more like those in a RefCell and keep them on in release builds. Taking another pass...

src/main/core/work/task.rs
sporksmith: Done

src/main/utility/mod.rs
sporksmith: Done - this is no longer `unsafe`

src/main/utility/mod.rs
sporksmith: No longer debug-only.

sporksmith: I don't think adding an escape hatch here is the way to go.

Maybe we could add a test-only interface to `Worker` that allows setting the host-info, faking that the lock is held? I'd be inclined to hold of on that until we need it, though.

src/main/utility/mod.rs
```diff
@@ -19,23 +19,72 @@ pub mod stream_len;
...
+
     pub fn ptr(&self) -> *mut T {
         self.0
     }
+}
+
+/// A type that allows us to make a pointer Send + Sync since there is no way
+/// to add these traits to the pointer itself.
+#[derive(Clone, Copy, Debug)]
+pub struct HostTreePointer<T> {
```
stevenengler, line 47: Maybe a comment about what makes this different from `SyncSendPointer`?

src/main/utility/mod.rs
```diff
@@ -19,23 +19,72 @@ pub mod stream_len;
...
+
+unsafe impl<T> Send for HostTreePointer<T> {}
+unsafe impl<T> Sync for HostTreePointer<T> {}
+
+impl<T> HostTreePointer<T> {
+    /// Create a pointer that may only be accessed when the lock for `host_id`
+    /// is held.
+    pub fn new_for_host(host_id: HostId, ptr: *mut T) -> Self {
+        // Suppress "never read" warning in release builds.
+        Self { host_id, ptr }
```
stevenengler, line 60: Stale comment?

src/main/utility/mod.rs
stevenengler: I'm good with holding off on that. I'm just always hesitant of having internal code polluted by global state, even though it's useful here.

stevenengler (approved):

src/main/utility/mod.rs
stevenengler: I'm good with holding off on that. I'm just always hesitant of having internal code polluted by global state, even though it's useful here.

src/main/utility/mod.rs
sporksmith: Done

src/main/utility/mod.rs
sporksmith: Done

sporksmith:
I ended up needing to fix https://github.com/shadow/shadow/issues/2152 for the tor tests to pass; PTAL

sporksmith:
I'm very confused. The CI builds are failing because Worker is imported in both line 1 and 24 in utility/mod.rs:

```
 error[E0252]: the name `Worker` is defined multiple times
  --> main/utility/mod.rs:24:13
   |
1  | use crate::core::worker::Worker;
   |     --------------------------- previous import of the type `Worker` here
...
24 | use crate::{core::worker::Worker, host::host::HostId};
   |             ^^^^^^^^^^^^^^^^^^^^--
   |             |
   |             `Worker` reimported here
```

... but it's *not imported on line 1: https://github.com/shadow/shadow/pull/2147/files#diff-15036aa6d9d2165aeaab07d0f84223193504fd97b4b18ab9c85074847e550f44

stevenengler:
That's a weird one, I can't replicate it.

stevenengler (approved):

stevenengler:
> I wonder if GitHub is trying to run the workflow on an attempted merged PR or something?

This seems to be exactly what's happening: https://github.community/t/gh-actions-checking-out-the-wrong-commit/227559/2

I think I've read this before, and I think we discussed this a long time ago when we were first adding CI support to Shadow. But I completely forgot about it, and I think it's unintuitive.

We should be able to stop this behaviour by having the checkout action checkout `github.event.pull_request.head.sha` instead.

sporksmith:
Wow, thanks for tracking that down. For the moment I just merged main into my branch, after which I got the same compilation error locally, then fixed it.

The [GH doc](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) says

> Note: Workflows will not run on pull_request activity if the pull request has a merge conflict. The merge conflict must be resolved first.

> Conversely, workflows with the pull_request_target event will run even if the pull request has a merge conflict. Before using the pull_request_target trigger, you should be aware of the security risks. For more information, see [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target).

I'm almost certain I've seen the action still run when there's a merge conflict... weird.

> We should be able to stop this behaviour by having the checkout action checkout github.event.pull_request.head.sha instead.

I wonder now if GH allows you to merge a PR that's out of date with `main`, so long as the checks passed. In that case running the actions on the PR head instead of the merge result would potentially end up allowing you to merge code with the kind of breakage we saw here...

Issue #2166, opened 2022-05-18
Pull request actions run on merge result instead of PR head


sporksmith:
This can result in confusing errors in the case that all of

* The PR merges cleanly
* The action (tests) fails
* The corresponding tests pass on the PR head

This has been rare, but is very confusing when it happens. See https://github.com/shadow/shadow/pull/2147#issuecomment-1129404865

sporksmith:
One solution, proposed by @stevenengler , is to have our checkout actions explicitly check out `github.event.pull_request.head.sha`.

Another possibility is to use the [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) event instead of the [pull_request](github.event.pull_request.head.sha) event. The latter has some security warnings that I don't fully comprehend at the moment.

Before doing either we should be sure that GH won't let us merge a PR where the actions pass on the PR head, but the PR head is out of date with the main branch, meaning we don't actually know whether they'd pass on the merge result.

It looks like this is an option for protected  branches: [Require branches to be up to date before merging](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule). I think we do have this enabled in this repo, but we should verify before making one of the above changes.

sporksmith:
I tried changing `pull_request` to `pull_request_target`, but now the actions aren't running at all.

I checked the settings for my fork of shadow, on which the branch is based. I have "Allow all actions and reusable workflows" checked, so I'm not sure what's happening there...

stevenengler:
> Another possibility is to use the pull_request_target event instead of the pull_request event. The latter has some security warnings that I don't fully comprehend at the moment.

I don't think we should take this approach. The `pull_request_target` runs the PR code in the context of the shadow repository (shadow/shadow) rather than the context of the PR author's repository (ex: stevenengler/shadow). This means that anyone can run arbitrary code in the context of the shadow repository, including accessing secrets for the shadow repository or shadow organization, has full read/write permissions by default, and can access and poison the shadow repository's [cache](https://github.com/actions/cache). The `pull_request_target` is designed to only be used for workflows that have no dependence on user-supplied code (the workflow doesn't run any code from the repository, including bash scripts). 

> I tried changing `pull_request` to `pull_request_target`, but now the actions aren't running at all.

[...]

PR #2167, opened 2022-05-18
Run actions on pull_request_target instead of pull_request

sporksmith:
This causes the actions to run on the head of the pull request instead
of on the merge result. This is generally what we want, particularly
since we require PRs to be up to date with the target branch before
merging anyway.

This also causes code to run in the context of the pull request branch.
If our actions used secrets this would cause them not to get them, or if
they tried to manipulate the repo they wouldn't be able to, etc. This
should be ok since these don't use secrets, and we probably don't *want*
a PR to be able to inject malicious code into a PR that can use our
permissions. (We already generally drop permissions, but even better if
the action never has them in the first place.)

Fixes https://github.com/shadow/shadow/issues/2166

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -11,7 +11,7 @@ defaults:
 on:
   push:
     branches: [main]
-  pull_request:
+  pull_request_target:
     types: [opened, synchronize]
 
 env:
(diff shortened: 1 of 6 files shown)


sporksmith:
This isn't what we want. Using pull-request-target causes the action to run in the target repo instead of the source repo. https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

Issue #2151, opened 2022-05-13
Shadow will return success even when plugins are killed by Linux OOM killer


stevenengler:
In general Shadow will return a non-zero error code if any of the plugins failed. But since Shadow doesn't have a proper shutdown sequence, Shadow sends a SIGKILL to all plugin processes at the end of the simulation and does not consider those killed processes to be an error.

https://github.com/shadow/shadow/blob/cc0ec729d5ab40d9347373a79f36c82c4b14b8df/src/main/host/process.c#L392-L400

This means that plugins killed by a SIGKILL for any reason will not be considered as an error. If Linux decides to kill plugin processes due to memory exhaustion (or any other reason), Shadow will not report that as an error and the simulation will have invalid results while reporting that the simulation was successful.

I have not actually tested this since I don't want to invoke the oom killer on my machine, but I'm pretty confident that it will break in this case.

Possibly related to #2134.

stevenengler:
Adding to 2.1 since I think this is important.

sporksmith:
Nicely tracked down; yuck.

More generally it's probably also worth the `exitcode` file distinguishing between the process being killed by shadow at simulation end and other ways it might receive a SIGKILL. Maybe there shouldn't be a file at all in this case, or it should be empty.

PR #2171, opened 2022-05-18
Distinguish between sources of death by SIGKILL

sporksmith:
* Add a regression test for https://github.com/shadow/shadow/issues/2151

* Treat death by signal 9 as an error if Shadow didn't    
  intentionally use it to stop the process.    
    
* Write an empty exitcode file when Shadow does kill the process (via    
  signal 9), which is a more intuitive result for a process that didn't    
  exit within the simulation itself.

Fixes https://github.com/shadow/shadow/issues/2151

### DIFF
--- a/src/main/host/process.c
+++ b/src/main/host/process.c
@@ -127,6 +127,7 @@ struct _Process {
 
     gint returnCode;
     gboolean didLogReturnCode;
+    gboolean killedByShadow;
 
     // int thread_id -> Thread*.
     GHashTable* threads;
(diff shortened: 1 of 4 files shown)


src/main/host/process.c
```diff
@@ -690,6 +705,8 @@ void process_stop(Process* proc) {
 
     info("terminating process '%s'", process_getName(proc));
 
+    proc->killedByShadow = true;
```
stevenengler, line 708: Are we sure this is only run when shadow kills the process, and not if the process exits unexpectedly and shadow tries to stop the process again later (at the stop time). The below `_process_terminate_threads()` has a `process_isRunning()` check before sending the `SIGKILL`, and I think I remember seeing the "terminating process" info log message twice for some processes. I'll run some tests to see if this is working as expected.

stevenengler: So `process_stop()` is run and `proc->killedByShadow` is set even for processes that unexpectedly died (like the 'kill' process in the below test). I think the `proc->killedByShadow = true` should be moved somewhere else, maybe directly before the `kill()` call.

stevenengler (approved):
Oops, meant to approve.

src/main/host/process.c
sporksmith: Oof, good catch. Done!

PR #2172, opened 2022-05-19
Use 'ubuntu-18.04' environment in CI tests

stevenengler:
The 'ubuntu-18.04' environment currently uses kernel 5.4.

https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-Readme.md

This only changes the workflows that test Shadow. Others like the lint checks still use 'ubuntu-latest'.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -19,7 +19,10 @@ env:
 
 jobs:
   codecov:
-    runs-on: ubuntu-latest
+    # use the oldest kernel supported by github's CI (make sure to update the
+    # minimum supported kernel version in documentation when changing)
+    # https://github.com/actions/virtual-environments
+    runs-on: ubuntu-18.04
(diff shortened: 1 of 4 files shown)


.github/workflows/coverage.yml
```diff
@@ -19,7 +19,9 @@ env:
 
 jobs:
   codecov:
-    runs-on: ubuntu-latest
+    # use the oldest kernel supported by github's CI
+    # https://github.com/actions/virtual-environments
```
sporksmith, line 23: Probably worth adding a reminder in these comments to update the minimum supported kernel version when changing (even if we haven't documented it yet)

sporksmith (approved):

.github/workflows/coverage.yml
stevenengler: Added.

PR #2174, opened 2022-05-19
Code of conduct

### LINKED ISSUE #2146: Add code of conduct
While it hasn't been an issue for us, I think it would be good to have a code of conduct document anyways.

Two possible options:
- adapt Tor's code of conduct (remove sections about the "Community Council", etc)
    - https://gitweb.torproject.org/community/policies.git/plain/code_of_conduct.txt
- adapt GitHub's default code of conduct (probably remove most of the "Enforcement Guidelines" section)
    - https://www.contributor-covenant.org/version/2/1/code_of_conduct/

We'd also need to figure out a private contact method.

stevenengler:
This uses the ["Contributor Covenant"](https://www.contributor-covenant.org/) template, with the following changes:

1. Shortened the 'Enforcement Guidelines' section of code of conduct. — As a small project with few people, I think it would be better to handle code of conduct violations on a case-by-case basis rather than codifying our enforcement guidelines.
2. Added to the 'Scope' section of code of conduct. — I think this paragraph from Tor's code of conduct is good to have.
3. Added a contact email address.

Closes #2146. Feedback welcome.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,133 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, caste, color, religion, or sexual
+identity and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
+* Being respectful of differing opinions, viewpoints, and experiences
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -68,49 +68,10 @@ reporter of any incident.
...
-
-### 4. Permanent Ban
-
-**Community Impact**: Demonstrating a pattern of violation of community
-standards, including sustained inappropriate behavior, harassment of an
-individual, or aggression toward or disparagement of classes of individuals.
-
-**Consequence**: A permanent ban from any sort of public interaction within the
-community.
+Community leaders will determine the consequences for any action they deem in
```
sporksmith, line 77: I have a slight preference to keeping the original enforcement guidelines. It's good to set expectations of roughly what consequences to expect/enforce for a given behavior, and they're still just "guidelines", so I think in practice we could still decide on a case by case basis.

I'm fine with either way, though.

sporksmith (approved):

CODE_OF_CONDUCT.md
stevenengler: Okay talked with rob and I removed this commit. So the old/original list of actions/consequences remains.

PR #2176, opened 2022-05-19
Convert Timer to Rust

sporksmith:


src/main/host/timer.rs
```diff
@@ -0,0 +1,237 @@
...
+    expire_interval: SimulationTime,
+    expiration_count: u64,
+    next_expire_id: u64,
+    min_valid_expire_id: u64,
+    on_expire: Box<dyn Fn(&mut Host) + Send + Sync>,
+}
+
+impl Timer {
+    /// Create a new Timer that synchronously executes `on_expire` on
+    /// expiration.  `on_expire` will cause a panic if it calls mutable methods
```
stevenengler, line 36: What does "synchronously" mean in this context?

src/main/host/timer.rs
```diff
@@ -0,0 +1,237 @@
...
+        internal.expiration_count = 0;
+        e
+    }
+
+    pub fn remaining_time(&self) -> SimulationTime {
+        self.magic.debug_check();
+        let t = if let Some(t) = self.internal.borrow().next_expire_time {
+            t
+        } else {
+            return SimulationTime::ZERO;
```
stevenengler, line 77: Function could use comments explaining that the remaining time is 0 for a timer with no expire time. Or should this return an `Option` instead?

src/main/host/timer.rs
```diff
@@ -0,0 +1,237 @@
...
+        internal_brw.expiration_count += 1;
+        if internal_brw.expire_interval > SimulationTime::ZERO {
+            internal_brw.next_expire_time =
+                Some(internal_brw.next_expire_time.unwrap() + internal_brw.expire_interval);
+            Self::schedule_new_expire_event(&mut *internal_brw, internal_weak.clone(), host);
+        }
+
+        // Re-borrow as an immutable reference while executing the callback.
+        drop(internal_brw);
+        let internal_brw = internal.borrow();
```
stevenengler, line 135: I don't think the `drop` is needed here?

src/main/host/timer.rs
```diff
@@ -0,0 +1,237 @@
...
+        &mut self,
+        host: &mut Host,
+        expire_time: EmulatedTime,
+        expire_interval: SimulationTime,
+    ) {
+        self.magic.debug_check();
+
+        let mut internal = self.internal.borrow_mut();
+        debug_assert!(expire_time >= Worker::current_time().unwrap());
+        internal.next_expire_time = Some(expire_time);
```
stevenengler, line 165: What happens in release builds if the expire time is less? Does the scheduler handle this reasonably?

stevenengler (approved):

src/main/host/timer.rs
sporksmith: I meant as opposed to pushing it to the scheduler queue. Changed to 'directly'; I think that's a little clearer.

src/main/host/timer.rs
sporksmith: Done. Added comments and used an Option here. Pushed the "zero" special handling into the C interface.

src/main/host/timer.rs
sporksmith: This borrow fails without the drop above. IIUC for let bindings that are overwritten/shadowed this way, the "orphaned" object isn't actually dropped until the end of the current scope.

src/main/host/timer.rs
sporksmith: Yes, the scheduler allows it and things generally work.

PR #2187, opened 2022-05-27
Update iPerf 2 docs now that itimer is supported

stevenengler:
The iPerf 2 version on ubuntu 20.04 is `2.0.13 (21 Jan 2019)`. The next release was `2.1.1 (22 June 2021)`, which has been tested to work correctly in Shadow. (There may have been a `2.1.0` release, but it's not tagged or "branched" in git so I didn't test it.) The most recent version is `2.1.7 (11 April 2022)` which also works.

### DIFF
--- a/docs/iperf2.md
+++ b/docs/iperf2.md
@@ -14,13 +14,13 @@ hosts:
   server:
     network_node_id: 0
     processes:
-    - path: iperf2-code/src/iperf
+    - path: /usr/bin/iperf
       args: -s
       start_time: 0s
   client:
     network_node_id: 0
     processes:
-    - path: iperf2-code/src/iperf
+    - path: /usr/bin/iperf
       args: -c server -t 5
       start_time: 2s
 ```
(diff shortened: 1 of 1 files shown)


docs/iperf2.md
```diff
@@ -31,5 +31,7 @@ rm -rf shadow.data; shadow shadow.yaml > shadow.log
 
 ## Notes
 
-1. Shadow doesn't support `setitimer()` so you must build iPerf 2 with `#define
-HAVE_SETITIMER 0` set.
+1. You must use an iPerf 2 version >= `2.1.1`. Older versions of iPerf have a
+[busy loop][busy-loop] that is incompatible with Shadow.
```
sporksmith, line 35: Maybe be a little more specific here so that it doesn't sound like all busy loops are incompatible with shadow? e.g. "no-syscall busy-loop", or "cpu-only busy-loop"?

sporksmith (approved):

sporksmith:
Thanks for checking up on this and updating!

docs/iperf2.md
stevenengler: Added the "no-syscall".

PR #2194, opened 2022-05-31
epoll_wait: return EINTR when interrupted

sporksmith:
Progress on https://github.com/shadow/shadow/issues/1889

### DIFF
--- a/docs/compatibility_notes.md
+++ b/docs/compatibility_notes.md
@@ -1,5 +1,12 @@
 # Compatibility Notes
 
+- [libopenblas](#libopenblas)
+- [cURL](#curl)
+- [Nginx](#nginx)
+- [iPerf 2](#iperf-2)
+- [iPerf 3](#iperf-3)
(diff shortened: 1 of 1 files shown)


src/test/test_common.c
```diff
@@ -127,3 +127,12 @@ bool running_in_shadow() {
     // There is the same function in the Rust tests utils code
     return getenv("SHADOW_SPAWNED") != NULL;
 }
+
+bool running_in_shadow_ptrace() {
+    // There is the same function in the Rust tests utils code
+    const char* method = getenv("SHADOW_INTERPOSE_METHOD");
+    if (method == NULL) {
+        return false;
+    }
```
stevenengler, line 136: nit: maybe add a `running_in_shadow()` check so that its behaviour exactly matches the rust version?

stevenengler (approved):

src/test/test_common.c
sporksmith: Done

PR #2198, opened 2022-05-31
When debugging a host, stop Shadow with SIGTSTP

stevenengler:
Rather than waiting for the user to provide input on stdin, shadow stops itself and waits for something to send it a SIGCONT.

### DIFF
--- a/src/main/host/process.c
+++ b/src/main/host/process.c
@@ -622,13 +622,22 @@ static void _process_start(Process* proc) {
 
     if (proc->pause_for_debugging) {
         // will block until logger output has been flushed
-        // there is a race condition where other threads may log between the fprintf() and getchar()
+        // there is a race condition where other threads may log between the fprintf() and raise()
         // below, but it should be rare
         logger_flush(logger_getDefault());
 
-        fprintf(stderr, "Managed process '%s' has started with PID %d. Press ENTER to continue...",
+        // print with hopefully a single syscall to avoid splitting these messages (fprintf should
+        // not buffer stderr output)
+        fprintf(stderr,
+                "** Pausing with SIGTSTP to enable debugger attachment to managed process '%s' "
+                "(pid %d)\n"
+                "** If running Shadow under Bash, resume Shadow by pressing Ctrl-Z to background "
+                "this task and then typing \"fg\".\n"
+                "** (If you wish to kill Shadow, type \"kill %%%%\" instead.)\n"
+                "** If running Shadow under GDB, resume Shadow by typing \"signal SIGCONT\".\n",
                 process_getName(proc), proc->nativePid);
-        getchar();
+
+        raise(SIGTSTP);
     }
 
     /* call main and run until blocked */
--- a/src/main/host/process.c
+++ b/src/main/host/process.c
@@ -621,6 +621,11 @@ static void _process_start(Process* proc) {
     worker_setActiveThread(NULL);
 
     if (proc->pause_for_debugging) {
+        // will block until logger output has been flushed
+        // there is a race condition where other threads may log between the fprintf() and getchar()
+        // below, but it should be rare
+        logger_flush(logger_getDefault());
+
         fprintf(stderr, "Managed process '%s' has started with PID %d. Press ENTER to continue...",
                 process_getName(proc), proc->nativePid);
         getchar();


src/main/host/process.c
```diff
@@ -606,9 +606,23 @@ static void _process_start(Process* proc) {
...
+                "** Pausing with SIGTSTP to enable debugger attachment to managed process '%s' "
+                "(pid %d)\n"
+                "** If running under Bash, resume Shadow by pressing Ctrl-Z to background this "
+                "task and then typing \"fg\".\n"
+                "** (If you wish to stop Shadow, type \"kill %%%%\" instead.)\n"
+                "** If running under GDB, resume Shadow by typing \"signal SIGCONT\".\n",
                 process_getName(proc), proc->nativePid);
-        getchar();
+
+        raise(SIGTSTP);
```
sporksmith, line 625: I just tested this out.

Flushing the logger is definitely an improvement.

I'm on the fence whether using a signal instead of stdin is an improvement or not. I think my main thinking when I suggested it was that it'd be nicer in cases where you couldn't access stdin as easily (e.g. if shadow is being called from a script that doesn't pass it's own stdin through to shadow), but I don't think I've actually hit that use-case in practice. e.g. when I want to run one of our tests under gdb it still feels easier to grab the command-line from the test output and run it directly than to e.g. edit the shadow_test macro to add the debug_hosts I want. OTOH maybe we could add an option to `setup test` and wire that through cmake to make it easier, in which case the signal-based approach would be needed I think.

Anyway, I'm fine with either way

src/main/host/process.c
```diff
@@ -606,9 +606,23 @@ static void _process_start(Process* proc) {
...
+        // there is a race condition where other threads may log between the fprintf() and raise()
+        // below, but it should be rare
+        logger_flush(logger_getDefault());
+
+        // print with hopefully a single syscall to avoid splitting these messages (fprintf should
+        // not buffer stderr output)
+        fprintf(stderr,
+                "** Pausing with SIGTSTP to enable debugger attachment to managed process '%s' "
+                "(pid %d)\n"
+                "** If running under Bash, resume Shadow by pressing Ctrl-Z to background this "
```
sporksmith, line 619: Ah darn, I was hoping bash would already treat the child process getting SIGSTOP the way it does ctrl-z. Oh well.

sporksmith (approved):

src/main/host/process.c
```diff
@@ -606,9 +606,23 @@ static void _process_start(Process* proc) {
...
+                "task and then typing \"fg\".\n"
+                "** (If you wish to stop Shadow, type \"kill %%%%\" instead.)\n"
+                "** If running under GDB, resume Shadow by typing \"signal SIGCONT\".\n",
```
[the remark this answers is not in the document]
stevenengler: Added

src/main/host/process.c
```diff
@@ -606,9 +606,23 @@ static void _process_start(Process* proc) {
...
+        logger_flush(logger_getDefault());
+
+        // print with hopefully a single syscall to avoid splitting these messages (fprintf should
+        // not buffer stderr output)
+        fprintf(stderr,
+                "** Pausing with SIGTSTP to enable debugger attachment to managed process '%s' "
+                "(pid %d)\n"
+                "** If running under Bash, resume Shadow by pressing Ctrl-Z to background this "
+                "task and then typing \"fg\".\n"
+                "** (If you wish to stop Shadow, type \"kill %%%%\" instead.)\n"
```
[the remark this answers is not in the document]
stevenengler: Changed.

src/main/host/process.c
stevenengler: It's nice when using Shadow with multiple threads since all threads get stopped (including the logging thread). Without stopping Shadow, another thread may log stuff, or another thread I think could also reach this "host debug" code, and we'd end up with shadow paused for two hosts at once.

src/main/host/process.c
stevenengler: Yeah I don't think there's a way to background the process from the process itself.

src/main/host/process.c
sporksmith: Ah right, yeah I'd forgotten about that motivation. sg!

PR #2199, opened 2022-05-31
shadow tests: check retval files instead of shadow log messages

sporksmith:
I accidentally broke the regex in
ba800b837fa0b77961a3907c7f6719fc05d11411.

Checking the exitcode files should be less prone to getting broken.

### DIFF
--- a/src/main/host/process.c
+++ b/src/main/host/process.c
@@ -620,9 +620,23 @@ static void _process_start(Process* proc) {
     worker_setActiveThread(NULL);
 
     if (proc->pause_for_debugging) {
-        fprintf(stderr, "Managed process '%s' has started with PID %d. Press ENTER to continue...",
+        // will block until logger output has been flushed
+        // there is a race condition where other threads may log between the fprintf() and raise()
+        // below, but it should be rare
+        logger_flush(logger_getDefault());
+
+        // print with hopefully a single syscall to avoid splitting these messages (fprintf should
+        // not buffer stderr output)
+        fprintf(stderr,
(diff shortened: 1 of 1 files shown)


src/test/CMakeLists.txt
```diff
@@ -126,6 +126,28 @@ macro(add_shadow_tests)
...
+         # Validate that all processes exited with code 0.  Processes that don't
+         # exit have an empty exitcode file, which will cause this check to
+         # fail.
+         # We *don't* fail in the case that nothing matched the glob, which is
+         # why we need to explicitly check whether the file exists.
+         set(CHECK_RETVAL_CMD "
+             for f in ${SHADOW_TEST_NAME}.data/hosts/*/*.exitcode
+             do
+               # Check if glob actually matched
+               if [ ! -f $f ]\; then continue\; fi
```
stevenengler, line 139: I think you can just set `shopt -s nullglob` before the `for`, and you don't need to worry about this condition.

stevenengler (approved):
Also remove the "don't change the formatting of this string since some integration tests depend on it" comment since I don't think it's needed anymore?

src/test/CMakeLists.txt
sporksmith: `nullglob` is bash-specific, and it looks like CMake is using /usr/bin/sh. Maybe there's a way to force it to use bash, but I don't think it's worth forcing the dependency

sporksmith:
> Also remove the "don't change the formatting of this string since some integration tests depend on it" comment since I don't think it's needed anymore?

Ah, the comment I blatantly ignored :). Done

PR #2205, opened 2022-06-01
poll: return EINTR when interrupted

sporksmith:
Progress on #1889 .

* Adds Interruptor utility struct to test_utils, which arranges for the current thread to be sent a signal after a specified timeout.
* Adds tests to poll for being interrupted by a signal, and fixes poll to correctly return EINTR.
* The tests ran into the hard-coded limit on number of threads in shim_tls; I reworked to remove the limit.

### DIFF
--- a/src/test/poll/test_poll.rs
+++ b/src/test/poll/test_poll.rs
@@ -3,6 +3,8 @@
  * See LICENSE for licensing information
  */
 
+use nix::sys::signal;
+use nix::sys::signal::Signal;
 use std::time::Duration;
 use test_utils::set;
 use test_utils::TestEnvironment as TestEnv;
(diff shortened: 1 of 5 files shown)


src/test/poll/test_poll.rs
```diff
@@ -197,6 +200,22 @@ fn test_poll_args_common(
...
+                    signal::SaFlags::empty(),
+                    signal::SigSet::empty(),
+                ),
+            )
+        }
+        .unwrap();
+        let mut interruptor = test_utils::Interruptor::new(signal_time, signal::Signal::SIGALRM);
+        if signal_time.is_zero() {
+            interruptor.cancel();
+        }
```
stevenengler, line 217: Why cancel if it's zero? Couldn't the thread have already started and the signal already sent before this point?

stevenengler (approved):

src/test/poll/test_poll.rs
sporksmith: Possibly! Or it might not have and we'll interrupt the poll call without intending to. Cancelling it here would ensure the poll call won't be interrupted when signal_time is 0.

Changed this to use an `Option` and never create the Interruptor in the first place if it's unwanted, to be a bit clearer.

PR #2226, opened 2022-06-13
Replace 'Controller' with a rust version

stevenengler:
This PR doesn't change the simulation results in the [benchmark](https://github.com/shadow/benchmark-results/tree/master/tor/2022-06-17-T01-19-56):

run_time

The "Change random/seed/ordering behaviour to match older Shadow" commit is needed to obtain similar results/performance, otherwise the simulation performance [is noticeably different](https://github.com/shadow/benchmark-results/tree/master/tor/2022-06-13-T21-20-02) without this commit:

run_time

client_goodput exit


src/main/core/controller.rs
```diff
@@ -0,0 +1,492 @@
...
+}
+
+impl std::ops::Drop for Controller<'_> {
+    fn drop(&mut self) {
+        unsafe { c::dns_free(self.dns) };
+        self.dns = std::ptr::null_mut();
+    }
+}
+
+trait SimController {
```
sporksmith, line 228: What's the motivation for this trait, vs putting these methods directly in Controller? A comment and/or more-descriptive name might help.

src/main/core/sim_config.rs
```diff
@@ -121,6 +123,14 @@ impl SimConfig {
             }
         }
 
+        // arrange so that hosts with an IP address are at the beginning of the list, to match the
+        // old Shadow behaviour
+        hosts.sort_by(|a, b| match (a.ip_addr.is_some(), b.ip_addr.is_some()) {
```
sporksmith, line 128: Re this commit overall - is it worth this extra complexity to preserve exactly the previous behavior? It's useful to have done it to verify that this was the only source of deviation, but now that we have I wonder whether it's worth maintaining, or whether we ought to revert this commit? (Maybe something to discuss in the next team meeting; we can always revert this commit later)

PR #2230, opened 2022-06-16
Prepare manager for rust conversion

stevenengler:
This contains some of the changes needed before we can move the manager to rust. It simplifies some of the manager code, and removes the `globalmanager` (the global manager's syscall counter was removed, and the alloc/dealloc counters were moved to the worker module).


src/main/core/worker.rs
```diff
@@ -285,16 +288,51 @@ impl Worker {
...
+        .unwrap_or_else(|| {
+            // no live worker; fall back to the shared manager counter
+            unsafe { cshadow::manager_add_syscall_counts_global(syscall_counts) };
+        });
     }
 }
 
+/// Enable object counters. Should be called near the beginning of the program.
+pub fn enable_object_counters() {
+    USE_OBJECT_COUNTERS.store(true, std::sync::atomic::Ordering::SeqCst);
```
sporksmith, line 343: Optional nit - I think `Relaxed` would be fine here...? This atomic isn't protecting any other variables.

src/main/core/manager.c
```diff
@@ -872,12 +872,6 @@ void manager_add_syscall_counts(Manager* manager, const Counter* syscall_counts)
     _manager_unlock(manager);
 }
 
-void manager_add_syscall_counts_global(const Counter* syscall_counts) {
```
sporksmith: (no action needed) - I think this code path was because in the parallelism=1 case we used to not create any worker thread, and instead run directly from the manager thread. We tore that out though since creating a single worker thread has negligible cost and saves us from having to deal with corner cases like this one.

sporksmith (approved):

src/main/core/worker.rs
stevenengler: Changed.

src/main/core/manager.c
stevenengler: Ah okay that makes sense.

Issue #1648, opened 2021-09-17
Support fcntl `F_DUPFD_CLOEXEC`


kbleeke:
Hello!

**Describe the issue**
fcntl `F_DUPFD` and `F_DUPFD_CLOEXEC` are not supported (for Epoll FDs) and cause a failure with `EINVAL` when creating a [tokio](https://github.com/tokio-rs/tokio) runtime.

**To Reproduce**
Create a Rust Project using Tokio and attempt to build a Runtime like:
```rust
let runtime = tokio::runtime::Builder::new_multi_thread().enable_all().build().expect("build runtime");
```
This code panics with an `std::io::Error` containing `EINVAL`.

**Operating System (please complete the following information):**
Fedora 34
 - Kernel version: post the output of `uname -a`  
```
Linux wrex 5.13.15-200.fc34.x86_64 #1 SMP Wed Sep 8 15:51:46 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

**Shadow (please complete the following information):**
```
** Starting Shadow v2.0.0-pre.4-198-gae248676-dirty 2021-09-14--14:21:38 with GLib v2.68.4
```

**Additional context**
In similar fashion to #1637 I'm trying to run a consensus protocol in shadow. This (currently released) project uses the [tokio](https://github.com/tokio-rs/tokio) library for network IO. Unfortunately Tokio tries to immediately duplicate it's Epoll FD when the Runtime is created, causing creation to fail with `EINVAL`.

I was able to hack my was past this by simply returning the same FD for the `F_DUPFD_CLOEXEC` fcntl. This makes the project seem run fine (with the patch from #1637). 

I am wondering what it would take to properly implement FD duplication (Epoll would be enough for me). From looking at the Code it seems like this would be a larger undertaking because the Epoll "Instance" and the FD live in the same object. At least I can't see an obvious way to have two FDs refer to the same Epoll.

stevenengler:
> I am wondering what it would take to properly implement FD duplication (Epoll would be enough for me). From looking at the Code it seems like this would be a larger undertaking because the Epoll "Instance" and the FD live in the same object. At least I can't see an obvious way to have two FDs refer to the same Epoll.

[...]

PR #2240, opened 2022-06-24
Remove handle field from `LegacyDescriptor`

stevenengler:
The `LegacyDescriptor` struct combines the file descriptor and file description into a single object by combining the fd handle and CLOEXEC flag with the underlying file object. This forces them to have a 1:1 relationship, where they should have a n:1 relationship for n>=0 (any number of file descriptors for a file description).

This PR is a step towards removing this 1:1 relationship and towards supporting #1648. A follow-up PR will allow multiple fds for a `LegacyDescriptor`.

### DIFF
--- a/src/main/host/descriptor/socket.c
+++ b/src/main/host/descriptor/socket.c
@@ -269,7 +269,7 @@ void legacysocket_setSocketName(LegacySocket* socket, in_addr_t ip, in_port_t po
     gchar* ipString = address_ipToNewString(ip);
     GString* stringBuffer = g_string_new(ipString);
     g_free(ipString);
-    g_string_append_printf(stringBuffer, ":%u (descriptor %i)", ntohs(port), socket->super.super.handle);
+    g_string_append_printf(stringBuffer, ":%u (descriptor %p)", ntohs(port), &socket->super.super);
     socket->boundString = g_string_free(stringBuffer, FALSE);
 
     /* the socket is now bound */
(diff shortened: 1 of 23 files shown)


src/main/host/descriptor/socket.c
```diff
@@ -84,7 +84,7 @@ static void _legacysocket_close(LegacyDescriptor* descriptor, Host* host) {
 
     Tracker* tracker = host_getTracker(host);
     if (tracker != NULL) {
-        tracker_removeSocket(tracker, descriptor_getHandle(descriptor));
+        tracker_removeSocket(tracker, (guintptr)descriptor);
```
sporksmith, line 87: This commit introduces a lot of casting to `guintptr`, which seems error-prone. It'd be nice to encapsulate at least the cast itself. e.g. could these various `tracker_*` APIs take a `Descriptor*` or `Socket*`, and maybe in the tracker have a private helper for converting to a `guintptr` "handle"?

e.g.:

```
static guintptr _tracker_socketHandle(Descriptor* d) { 
  // XXX:  utility_assert to check that the Descriptor* is a Socket?
  return (guintptr) d;
}

void tracker_removeSocket(Tracker* tracker, Descriptor* d) {
  guintptr handle = _tracker_socketHandle(d);
  ...
}
```

Maybe even better to use a specialized value type instead of a guintptr, but maybe that's overkill if it can be purely internal to the Tracker.

sporksmith (approved):

src/main/host/descriptor/socket.c
stevenengler: My reason for using a `guintptr` was so that the `Tracker` stays independent of the socket type (for example the rust `Socket` or the C `LegacySocket` could both be used), and so that the `Tracker` doesn't try to call socket methods. But I don't think these are a big deal, so I'll change it to take pointers and cast them to `guintptr` internally. We'd probably want to rewrite this in rust anyways before calling it from rust sockets.

PR #2242, opened 2022-06-28
Restructure descriptors to support dup() for all descriptor/file types

### LINKED ISSUE #1648: Support fcntl `F_DUPFD_CLOEXEC`

stevenengler:
Closes #1648.

The `Descriptor` type now replaces the `CompatDescriptor`. The compatibility enum to support C `LegacyDescriptors` is a new `CompatFile` type contained within the `Descriptor`. This restructuring now means that we support `dup()` and related syscalls for all file descriptors, including for `LegacyDescriptors`.


sporksmith (approved):
Woohoo!

Issue #1987, opened 2022-03-24
Support subprocess creation and management (was "fork and exec")


sporksmith:
- [x] https://github.com/shadow/shadow/issues/2987
- [x] https://github.com/shadow/shadow/issues/2988
- [x] https://github.com/shadow/shadow/issues/2989
- [ ] ? https://github.com/shadow/shadow/issues/3123
- [ ] ? https://github.com/shadow/shadow/issues/3166

Supporting fork and exec would make it easier to delegate complexity to wrapper shell or python scripts instead of adding more features to Shadow itself.

e.g. rather than Shadow natively supporting compression (#1554), a user could use a config like:

```
-path /bin/bash
-args -c "tor | gzip -"
```

This could also be used to address clean shutdown of processes (#1491) with something like:

```
hostname:
  processes:
  - path: /bin/bash
    args: -c "tor -f torrc & PID=$! ; sleep 100 && kill $PID"
```

Using killall #1986 might be a better alternative for this particular case, but a shell script would allow for greater flexibility

The biggest potential blocker right now is that not all file descriptors support duplication yet. However, unix pipes and regular files are probably sufficient to cover a lot of use cases. In the meantime we can log a warning for any file descriptors we can't duplicate into the child.

stevenengler:
> The biggest potential blocker right now is that not all file descriptors support duplication yet. However, unix pipes and regular files are probably sufficient to cover a lot of use cases. In the meantime we can log a warning for any file descriptors we can't duplicate into the child.

We [now support](https://github.com/shadow/shadow/pull/2242) duplicating all descriptor types. There is still an [existing issue](https://github.com/shadow/shadow/issues/1780) with TCP sockets that would prevent `accept()` calls on duplicated listening sockets from working correctly in a forked process.

stevenengler:
We may also need to consider futexes shared between processes:

https://github.com/shadow/shadow/blob/ee157d00e39c2c7b1c6a7a3467d73f3bb1b2a49a/src/main/host/syscall/futex.c#L143-L147

sporksmith:
Another motivating example: arti's way of using the obs4 pluggable transport currently involves forking and execing an obs4 process.

Investigating whether it makes sense to add an alternative mechanism to arti that would allow it to use an independently started process...

trinity-1686a:
[...]

Issue #1780, opened 2021-11-19
Incoming TCP connections should not be added to the descriptor table until they're accept()ed


stevenengler:
When a listening socket receives a SYN packet, it immediately registers a new socket in the descriptor table.

https://github.com/shadow/shadow/blob/138b84a8294434aeb185cc2d84b5408177fc97f0/src/main/host/descriptor/tcp.c#L1931-L1945

Shadow should wait and only register the socket in the descriptor table once the user space process calls `accept()`.

stevenengler:
An example that fails due to this bug (the second bind fails):

```rust
fn main() {
    let fd_server = unsafe { libc::socket(libc::AF_INET, sock_type | libc::SOCK_NONBLOCK, 0) };
    let fd_client = unsafe { libc::socket(libc::AF_INET, sock_type | libc::SOCK_NONBLOCK, 0) };
    assert!(fd_server >= 0);
    assert!(fd_client >= 0);

    // the server address
    let server_addr = libc::sockaddr_in {
        sin_family: libc::AF_INET as u16,
        sin_port: 11111u16.to_be(),
        sin_addr: libc::in_addr {
            s_addr: libc::INADDR_ANY.to_be(),
        },
        sin_zero: [0; 8],
    };

    // bind on the server address
    {
        let rv = unsafe {
            libc::bind(
                fd_server,
                &server_addr as *const libc::sockaddr_in as *const libc::sockaddr,
                std::mem::size_of_val(&server_addr) as u32,
            )
        };
        assert_eq!(rv, 0);
    }

    // listen for connections
    let rv = unsafe { libc::listen(fd_server, 10) };
    assert_eq!(rv, 0);

    // connect a socket to the server, but the server doesn't accept() the connection
    let rv = unsafe {
        libc::connect(
            fd_client,
            &server_addr as *const libc::sockaddr_in as *const libc::sockaddr,
            std::mem::size_of_val(&server_addr) as u32,
        )
    };
    let errno = std::io::Error::last_os_error().raw_os_error().unwrap();
    assert!(rv == 0 || (rv == -1 && errno == libc::EINPROGRESS));

    // shadow needs to run events
    let rv = unsafe { libc::usleep(2000) };
    assert_eq!(rv, 0);

    // close the server socket
    let rv = unsafe { libc::close(fd_server) };
    assert_eq!(rv, 0);

    let rv = unsafe { libc::sleep(5) };
    assert_eq!(rv, 0);

    // create a new server socket
[...]

PR #2245, opened 2022-06-28
Remove `ownerProcess` field from `LegacyDescriptor`

stevenengler:
A file description is not owned by any one process, so the `LegacyDescriptor` should not have this `ownerProcess` field. Removing this field should make it easier to support `fork()` (#1987) and to support a non-ref-counted rust `Process` object in the future.

The one place where this field is used is for listening TCP sockets, so the `TCPServer` struct now holds this process reference instead. The way that it's used is incorrect (#1780), but fixing that issue is much more difficult.


src/main/host/descriptor/tcp.c
```diff
@@ -1996,13 +2004,18 @@ static void _tcp_processPacket(LegacySocket* socket, Host* host, Packet* packet)
                 guint64 recvBufSize = host_getConfiguredRecvBufSize(host);
                 guint64 sendBufSize = host_getConfiguredSendBufSize(host);
 
+                /* We will register the child socket with whichever process called listen() on the
+                 * parent socket. This is incorrect and we should register the child sockt with
```
sporksmith, line 2008: *socket

sporksmith (approved):

src/main/host/descriptor/tcp.c
stevenengler: Fixed!

PR #2248, opened 2022-06-29
Fix unsafe accesses of host objects from the manager

stevenengler:
The manager accesses `Host` objects without acquiring the hosts' locks, which is not safe.

Edit:

The simulation results have changed slightly in this PR, so I will try to investigate where this difference is from.

1656599789_grim

Edit 2:

With the workaround commit, the results are the same. We can revert that commit in a new PR.

1656636436_grim

### DIFF
--- a/src/main/core/manager.c
+++ b/src/main/core/manager.c
@@ -593,17 +593,11 @@ DNS* manager_getDNS(Manager* manager) {
 }
 
 guint32 manager_getNodeBandwidthUp(Manager* manager, in_addr_t ip) {
-    // TODO: The `/ 1024 * 1024 / 1000 * 1000` is a workaround to keep the same networking
-    // performance/results. It should be removed in an isolated commit so that the performance
-    // change is caused by this small integer rounding change.
-    return controller_getBandwidthUpBytes(manager->controller, ip) / 1024 * 1024 / 1000 * 1000 / 1024;
+    return controller_getBandwidthUpBytes(manager->controller, ip) / 1024;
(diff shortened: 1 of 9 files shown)


src/main/core/manager.c
```diff
@@ -593,11 +593,11 @@ DNS* manager_getDNS(Manager* manager) {
 }
 
 guint32 manager_getNodeBandwidthUp(Manager* manager, in_addr_t ip) {
-    return controller_getBandwidthUpBytes(manager->controller, ip) / 1024;
+    return controller_getBandwidthUpBytes(manager->controller, ip) / 1024 * 1024 / 1000 * 1000 / 1024;
```
sporksmith, line 596: A comment here and below that this is to preserve legacy behavior, with a TODO to remove, would be useful.

Assuming you want to keep this "identical behavior" commit in the history, I'd lean towards adding another commit reverting it in this same PR, and then merging without squashing to preserve that history. This avoids having this weird state in the "main history" (git log --first-parent main), and the possibility of forgetting to revert it later.

sporksmith (approved):

src/main/core/manager.c
stevenengler: Added some comments and a commit to undo this behaviour.

stevenengler:
Some examples of the expected network performance change (the difference between PR #2248 and nightly):

transfer_time_51200 exit

transfer_time_5242880 exit

client_goodput_5MiB exit

PR #2249, opened 2022-06-30
Fixed misc memory leaks

stevenengler:
Thanks to valgrind!

### DIFF
--- a/src/main/host/futex.c
+++ b/src/main/host/futex.c
@@ -42,6 +42,9 @@ Futex* futex_new(PluginPhysicalPtr word) {
 
 static void _futex_free(Futex* futex) {
     MAGIC_ASSERT(futex);
+
+    g_hash_table_destroy(futex->listeners);
+
     MAGIC_CLEAR(futex);
     free(futex);
     worker_count_deallocation(Futex);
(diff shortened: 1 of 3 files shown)


src/main/host/futex.c
```diff
@@ -42,6 +42,9 @@ Futex* futex_new(PluginPhysicalPtr word) {
 
 static void _futex_free(Futex* futex) {
     MAGIC_ASSERT(futex);
+
+    g_hash_table_destroy(futex->listeners);
```
sporksmith, line 46: This is an exciting one, since it could keep leaking memory over the course of the simulation, though hypothetically I *should* have seen that in the profiling I did before if it were the case in practice for tor simulations

stevenengler: In the tor minimal test: "880,672 (330,912 direct, 549,760 indirect) bytes in 3,447 blocks are definitely lost in loss record 327 of 328" for the futex leak. It seems like a lot for a 30 second simulation, but tor might create most of its mutexes when it launches, and the rate over time might be low. But we'll see in the benchmark graph eventually.

sporksmith (approved):

stevenengler:
No noticeable change :(

https://github.com/shadow/benchmark-results/tree/master/tor/2022-07-01-T11-48-58

ram_simtime

PR #2262, opened 2022-07-07
More organization changs to the manager / worker pool

stevenengler:
These changes are to prepare for the rust migration of the `Manager`.

- Workers now add their local counters to the global counters --- Rather than having each worker add its counters to the manager through the global worker pool, each worker now adds its counters to the global counters and the manager then later reads from the global counters.
- Worker pool now accesses the controller, pid watcher, and config directly --- This is work towards making the controller/manager/scheduler/workerpool more composable and to limit parent references.

### DIFF
--- a/src/main/core/worker.c
+++ b/src/main/core/worker.c
@@ -41,9 +41,14 @@ static void _worker_shutdownHost(Host* host, void* _unused);
 static void _workerpool_setLogicalProcessorIdx(WorkerPool* workerpool, int workerID, int cpuId);
 
 struct _WorkerPool {
-    /* Unowned pointer to the object that communicates with the controller
-     * process */
-    Manager* manager;
+    /* Unowned pointer to the object that communicates with the controller process */
+    const Controller* controller;
+
+    /* Unowned pointer to the PID watcher for managed processes */
+    const ChildPidWatcher* pidWatcher;
+
+    /* Unowned pointer to the configuration options */
+    const ConfigOptions* config;
 
     /* Unowned pointer to the per-manager parallel scheduler object that feeds
      * events to all workers */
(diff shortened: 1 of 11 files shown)


src/main/core/worker.c
```diff
@@ -483,12 +483,11 @@ void worker_finish(GQueue* hosts, SimulationTime time) {
     /* cleanup is all done, send counters to manager */
```
sporksmith, line 488: Update comment?

src/main/core/worker.rs
```diff
@@ -524,4 +529,47 @@ mod export {
...
+
+    /// Add the global counter to the provided counter, and clear the global counter.
+    #[no_mangle]
+    pub extern "C" fn worker_addFromGlobalSyscallCounter(
+        syscall_counter: *mut Counter,
+    ) {
+        let syscall_counter = unsafe { syscall_counter.as_mut() }.unwrap();
+
+        let mut global_syscall_counter = SYSCALL_COUNTER.lock().unwrap();
+        syscall_counter.add_counter(&global_syscall_counter);
```
sporksmith, line 558: Nit - these would be a little clearer if we added something like `Counter::take(&mut self)` that returns the current current count and resets to 0.

src/main/core/worker.c
```diff
@@ -625,34 +630,41 @@ static void _worker_shutdownHost(Host* host, void* _unused) {
 }
 
 guint32 worker_getNodeBandwidthUp(in_addr_t ip) {
```
sporksmith, line 632: optional, since it's orthogonal to this change, but it'd be nice if these functions included the units. e.g. worker_getNodeBandwidthUpKiBps

sporksmith (approved):

src/main/core/worker.c
stevenengler: Added another commit with this change.

Issue #2266, opened 2022-07-08
Improve failure mode when /dev/shm is full


sporksmith:
Follow-up to: https://github.com/shadow/shadow/discussions/2264#discussioncomment-3109758

It appears that when /dev/shm gets filled, we can fail in a pretty baroque and disruptive way - managed programs go into an infinite SIGBUS loop.

Hopefully we can instead catch this in the shared memory allocator or where it's used and shut down with a helpful error message.

stevenengler:
These conditions can probably be replicated using the `--shm-size` option for `docker run`. For example, maybe running a tor sim in a container with `--shm-size="100m"` or something similar.

sporksmith:
I was able to reproduce a bus error using a small `--shm-size`. It didn't turn into the infinite loop we saw before; I think that'll end up depending whether shadow or the managed process is the first to try to access the memory.

I have a PR that seems to fix it; will send out shortly.

PR #2267, opened 2022-07-08
Pre-allocate space in shared memory files

sporksmith:
Replace usage of `ftruncate` with `posix_fallocate`. `ftruncate` sets
the size of a file without actually allocating the backing storage,
which is allocated on-demand. While this allows the true allocation to
be "lazy", and thereby allows us to do large allocations without wasting
resources if the whole allocation isn't used, it can cause unpredictable
and confusing failure modes.

With `posix_fallocate`, the space is eagerly allocated. If the call
succeeds, we are guaranteed that the storage is actually available.
When we run out of storage, the call fails and we can fail with a clear
error message.

We were previously allocating quite-large chunks of memory at a time;
about 132 MB. While this probably wouldn't actually cause problems for a
single shadow simulation (which typically only needs one such
allocation), it ends up using a lot of space when running the shadow
tests in parallel. This changes the default allocation to 1 MB, instead.

Fixes https://github.com/shadow/shadow/issues/2266

### DIFF
--- a/src/main/shmem/shmem_allocator.c
+++ b/src/main/shmem/shmem_allocator.c
@@ -13,7 +13,7 @@
 #include "main/shmem/shmem_file.h"
 #include "main/shmem/shmem_util.h"
 
-#define SHD_SHMEM_ALLOCATOR_POOL_NBYTES SHD_BUDDY_POOL_MAX_NBYTES
+#define SHD_SHMEM_ALLOCATOR_POOL_NBYTES (1<<20)
 
 // when to change allocation strategies
 #define SHD_SHMEM_ALLOCATOR_CUTOVER_NBYTES                                     \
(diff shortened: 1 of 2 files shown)


sporksmith:
Running a benchmark: https://github.com/shadow/benchmark/actions/runs/2638637530

RWails (approved):
Looks good. I'm glad the default mem block file size is being reduced; I remember setting it to a high value when we were prototyping, but I didn't realize that the 100 MB value stuck!

I spot-checked to make sure that most of the allocations being made are under 500 k, and I believe this is the case. (Otherwise, we'd lose small-object allocation optimizations and the shmem allocator would make a shmem syscall for every allocation, which is how it handles large objects).

Thanks Jim!

sporksmith:
> Looks good. I'm glad the default mem block file size is being reduced; I remember setting it to a high value when we were prototyping, but I didn't realize that the 100 MB value stuck!

It was completely reasonable when using `ftruncate`, since the actual allocation happened on-demand anyway, but yeah is a bit too large now :).

> I spot-checked to make sure that most of the allocations being made are under 500 k, and I believe this is the case. 

Thanks! Yeah the [benchmark results](https://github.com/shadow/benchmark/actions/runs/2638637530) looks ok too; run-time and memory usage didn't change compared to the previous nightly run.

Issue #2273, opened 2022-07-12
Size of `__vdso_gettimeofday` is too small for trampoline


stevenengler:
In #2271, the shim panics with the message:

```
00:00:00.000001 error [patch_vdso.c:178] [_inject_trampoline] Can't inject 13 byte trampoline into 5 byte symbol '__vdso_gettimeofday'
```

This is failing on Ubuntu 22.04 with kernel 5.15.0-39-generic.

sporksmith:
via afressancourt:

Hello

Here is the result of the command on my machine (Ubuntu 22.04 LTS, 5.15.0-39-generic kernel)

```
> readelf -Ws /lib/modules/5.15.0-39-generic/vdso/vdso64.so

Symbol table '.dynsym' contains 13 entries:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
     0: 0000000000000000     0 NOTYPE  LOCAL  DEFAULT  UND
     1: 0000000000000c10     5 FUNC    WEAK   DEFAULT   11 clock_gettime@@LINUX_2.6
     2: 0000000000000bd0     5 FUNC    GLOBAL DEFAULT   11 __vdso_gettimeofday@@LINUX_2.6
     3: 0000000000000c20    96 FUNC    WEAK   DEFAULT   11 clock_getres@@LINUX_2.6
     4: 0000000000000c20    96 FUNC    GLOBAL DEFAULT   11 __vdso_clock_getres@@LINUX_2.6
     5: 0000000000000bd0     5 FUNC    WEAK   DEFAULT   11 gettimeofday@@LINUX_2.6
     6: 0000000000000be0    41 FUNC    GLOBAL DEFAULT   11 __vdso_time@@LINUX_2.6
     7: 0000000000000cb0   156 FUNC    GLOBAL DEFAULT   11 __vdso_sgx_enter_enclave@@LINUX_2.6
     8: 0000000000000be0    41 FUNC    WEAK   DEFAULT   11 time@@LINUX_2.6
     9: 0000000000000c10     5 FUNC    GLOBAL DEFAULT   11 __vdso_clock_gettime@@LINUX_2.6
    10: 0000000000000000     0 OBJECT  GLOBAL DEFAULT  ABS LINUX_2.6
    11: 0000000000000c80    37 FUNC    GLOBAL DEFAULT   11 __vdso_getcpu@@LINUX_2.6
[...]

PR #2283, opened 2022-07-15
Tweak shadow tests CI matrix

sporksmith:
* Add some configurations that run with a newer kernel, to help catch kernel incompatibilities such as https://github.com/shadow/shadow/issues/2273

* Only test clang in a couple configurations, to cut down the CI matrix a bit.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -43,24 +43,29 @@ jobs:
           - 'fedora:36'
           # EOL May 2024 https://www.centos.org/centos-stream/
           - 'quay.io/centos/centos:stream8'
-        cc: ['gcc', 'clang']
+        cc: ['gcc']
         buildtype: ['debug', 'release']
-        exclude:
-          # https://github.com/shadow/shadow/issues/1741
-          - container: 'quay.io/centos/centos:stream8'
-            cc: clang
-          # https://github.com/shadow/shadow/issues/1741
-          - container: 'ubuntu:21.10'
-            cc: clang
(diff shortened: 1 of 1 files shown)


.github/workflows/run_tests.yml
```diff
@@ -50,18 +43,42 @@ jobs:
...
+            buildtype: 'release'
+          - vm: 'ubuntu-22.04'
+            container: 'ubuntu:22.04'
+            cc: 'gcc'
+            buildtype: 'debug'
+
+          # Run some tests on the newest-available clang.  Testing clang on
+          # *every* platform is a bit overkill, but testing on at least one
+          # gives decent "bang for the buck" of testing compatibility with
+          # clang's most recent diagnostics and optimizations.
```
stevenengler, line 63: I think we should also run a test on a non-debian distro (maybe fedora), and on a distro with an older version of clang. Some distros have different clang warnings enabled by default, and since we use `-Wall` in the CI, we'd catch those.

stevenengler (approved):

.github/workflows/run_tests.yml
sporksmith: > Some distros have different clang warnings enabled by default, and since we use -Wall

Doesn't `-Wall` override the defaults?

In general I think newer versions of clang will catch the most problems; testing on older versions have pretty diminishing returns.

Going to go ahead and merge this now while I'm in a meeting with @robgjansen so we can coordinate changing the required actions, but happy to revisit

PR #2288, opened 2022-07-18
Fixed a warning in the unix socket code

stevenengler:
Noticed a warning in the unix socket code that was incorrectly being shown.

### DIFF
--- a/src/main/host/descriptor/socket/unix.rs
+++ b/src/main/host/descriptor/socket/unix.rs
@@ -1861,16 +1861,24 @@ struct UnixSocketCommon {
 
 impl UnixSocketCommon {
     pub fn close(&mut self, event_queue: &mut EventQueue) -> Result<(), SyscallError> {
-        if self.state.contains(FileState::CLOSED) {
-            log::warn!("Attempting to close an already-closed unix socket");
-        }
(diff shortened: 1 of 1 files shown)


src/main/host/descriptor/socket/unix.rs
```diff
@@ -1861,16 +1861,24 @@ struct UnixSocketCommon {
...
+            // do this before the conditional panic, otherwise rust gives us warnings
+            self.copy_state(
+                /* mask= */ FileState::all(),
+                FileState::CLOSED,
+                event_queue,
+            );
+
+            // panic in debug builds since the backtrace will be helpful for debugging
+            #[cfg(debug_assertions)]
+            panic!("{}", MSG);
```
sporksmith, line 1880: I think I've seen this idiom a few times now. Maybe worth adding a `debug_panic` macro, which delegates to `warn` in release builds, and `panic` in debug builds?

sporksmith (approved):

src/main/host/descriptor/socket/unix.rs
stevenengler: Yeah, that's a good idea. I'll leave it for now, but will add the macro tomorrow

PR #2293, opened 2022-07-19
Add shadow test workflow with incremental builds

sporksmith:
Adds a new workflow `run_tests_incremental.yml`, based on `run_tests.yml`.

When run manually, it gives the option of building and pushing Docker images with all dependencies installed, and a build of Shadow.

In other cases, it tries to *pull* such an image, and do an incremental build on top of the build baked into the image. In case the image can't be pulled, it falls back to building the image from scratch (but doesn't try to push to Dockerhub).

Note that no images exist yet in the `shadowsim:shadow-ci` repository, so in the runs on this PR it'll fall back to building images from scratch. We won't see the benefit until after merging this PR and doing a manual run that builds and pushes images.

The tentative plan is to *not* change the required tests when merging this PR. i.e. this adds a new set of tests "Shadow Tests Incremental", but it will run *in addition to* the current "Shadow Tests" workflow, and only the latter will be required.

After this PR is merged, we can kick off a workflow to build and push images.

After seeing how the new tests work for a bit, we can consider switching over to them - making the new tests required instead of the old ones, and disabling or deleting the old workflow.

### DIFF
--- a/.github/workflows/run_tests_incremental.yml
+++ b/.github/workflows/run_tests_incremental.yml
@@ -1,0 +1,152 @@
+# Syntax reference:
+# https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions
+
+name: Shadow Tests
+permissions: read-all
+
+defaults:
+  run:
+    shell: bash
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    types: [opened, synchronize]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  shadow:
+    strategy:
+      matrix:
+        vm:
+          # Oldest-available kernel from
+          # https://github.com/actions/virtual-environments
+          - ubuntu-18.04
+        container:
+          # End of standard support: April 2023 https://wiki.ubuntu.com/Releases
+          - 'ubuntu:18.04'
+          # End of standard support: April 2025 https://wiki.ubuntu.com/Releases
+          - 'ubuntu:20.04'
+          # End of standard support: April 2027 https://wiki.ubuntu.com/Releases
+          - 'ubuntu:22.04'
+          # EOL ~August 2022 https://wiki.debian.org/DebianReleases
+          - 'debian:10-slim'
+          - 'debian:11-slim'
+          # EOL June 7 2022 https://endoflife.date/fedora
+          - 'fedora:34'
+          - 'fedora:35'
+          - 'fedora:36'
+          # EOL May 2024 https://www.centos.org/centos-stream/
+          - 'quay.io/centos/centos:stream8'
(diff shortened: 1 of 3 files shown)


ci/run.sh
```diff
@@ -108,7 +116,7 @@ run_one () {
     # Replace all forward slashes with '-'
     CONTAINER_FOR_TAG="${CONTAINER_FOR_TAG//\//-}"
 
-    TAG="shadow:$CONTAINER_FOR_TAG-$CC-$BUILDTYPE"
+    TAG="$REPO:$CONTAINER_FOR_TAG-$CC-$BUILDTYPE"
```
stevenengler, line 71: There's a reference to the "shadow" tag name in https://shadow.github.io/docs/guide/ci.html which should maybe be updated to say "shadowsim/shadow-ci".

ci/run.sh
```diff
@@ -4,99 +4,50 @@
...
-  -b BUILDTYPES  Set build-types used in test matrix
-  -e EXTRAS      Set extra configurations to run
-  -o             Set only configurations to run
+  -i             Build image
+  -c CONTAINER   Set container
+  -C CC          Set C compiler
+  -b BUILDTYPE   Set build-types
   -n             nocache when building Docker images
-  -p             pushes images to dockerhub if set
+  -p             push image to dockerhub
```
stevenengler, line 27: These changes probably also require https://shadow.github.io/docs/guide/ci.html to be updated.

.github/workflows/run_tests_incremental.yml
```diff
@@ -96,34 +91,43 @@ jobs:
...
-          key: cargo-registry-${{ steps.get-month.outputs.month }}
-          restore-keys: |
-            cargo-registry-
-
-      - name: Build
-        run: . ci/container_scripts/build_and_install.sh
+          if [ "${{ github.event.inputs.build}}" == "true" ]
+          then
+            # Build the image
+            FLAGS+=("-i")
```
stevenengler, line 107: In the workflow dispatch API, booleans are actually strings and you could call the API with a value containing a `"` to break out of this bash string and run arbitrary code. Since we're the only ones who can run it, this is probably fine, but I think it's generally better to use the environment variable format for inputs to prevent that (ex: `env: INPUT_BUILD: ${{ github.event.inputs.build}}`).

stevenengler (approved):

.github/workflows/run_tests_incremental.yml
sporksmith: oof, nice catch

ci/run.sh
sporksmith: Done

ci/run.sh
sporksmith: Done

PR #2308, opened 2022-07-26
Remove EOL distros from CI and docs

### LINKED ISSUE #2307: Update CI images since Ubuntu 21.10 is EOL
Ubuntu 21.10 is EOL, upgrade path is to 22.04 LTS
https://wiki.ubuntu.com/Releases

robgjansen:
Ubuntu 21.10 is EOL. The upgrade path for Ubuntu 21.10 is to 22.04, which is already in the CI.
Ubuntu 22.10 is not scheduled to be released until October 20, 2022.
https://wiki.ubuntu.com/Releases

Fedora 34 is EOL. Fedora 35 and 36 are still supported.
https://docs.fedoraproject.org/en-US/releases/eol/

Closes #2307

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -30,8 +30,6 @@ jobs:
           - 'ubuntu:18.04'
           # End of standard support: April 2025 https://wiki.ubuntu.com/Releases
           - 'ubuntu:20.04'
-          # End of standard support: July 2022 https://wiki.ubuntu.com/Releases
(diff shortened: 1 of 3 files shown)


RWails (approved):

robgjansen:
In this case, Ubuntu 21.10 was EOL and so we could not download packages in the docker image, which basically blew up our CI pipeline. More generally, we should document how we decide to change our list of supported platforms: I opened #2309 for that.

PR #2320, opened 2022-08-03
run_tests_incremental: minor cleanup

sporksmith:
* quote variables in bash
* enable bash command echo

### DIFF
--- a/.github/workflows/run_tests_incremental.yml
+++ b/.github/workflows/run_tests_incremental.yml
@@ -104,7 +104,7 @@ jobs:
           FLAGS+=("-b" "${{ matrix.buildtype }}")
 
           # Get the generated tag for the image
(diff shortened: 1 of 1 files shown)


.github/workflows/run_tests_incremental.yml
```diff
@@ -98,13 +98,14 @@ jobs:
 
       - name: Test
         run: |
+          set -x
```
stevenengler, line 101: This shouldn't leak the docker secret right, because github will strip it from the output?

stevenengler (approved):

.github/workflows/run_tests_incremental.yml
sporksmith: Right. Did a quick test on my fork to verify: https://github.com/sporksmith/shadow/runs/7653802389?check_suite_focus=true

PR #2322, opened 2022-08-03
Support yaml merge keys and extension fields

stevenengler:
- [merge keys](https://yaml.org/type/merge.html): not technically valid yaml 1.2, but is common enough that I don't think it hurts to support it
- [extension fields](https://docs.docker.com/compose/compose-file/#extension): fields beginning with "x-" that are ignored; this follows the convention of docker compose, but we only allow them at the top level of the document

This is technically a breaking change since "<<" was previously an allowed shadow hostname, but this isn't generally a valid hostname and I don't think is worth worrying about.

These are currently activated by a `--use-extended-yaml true` option.

### DIFF
--- a/docs/shadow_config_overview.md
+++ b/docs/shadow_config_overview.md
@@ -77,7 +77,10 @@ Acceptable unit *prefixes* are:
 
 Examples: `20 B`, `100 MB`, `100 megabyte`, `10 kibibytes`, `30 MiB`, `1024 Mbytes`
 
-## YAML Extensions
+## YAML Extensions (Experimental)
+
+Shadow has experimental support for extended YAML conventions. These can be
+enabled using the `--use-extended-yaml true` command line option.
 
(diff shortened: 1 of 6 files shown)


docs/shadow_config_overview.md
```diff
@@ -76,3 +76,72 @@ Acceptable unit *prefixes* are:
 - tebi / Ti
 
 Examples: `20 B`, `100 MB`, `100 megabyte`, `10 kibibytes`, `30 MiB`, `1024 Mbytes`
+
+## YAML Extensions
+
```
sporksmith, line 81: Mention that this support is experimental, and must be enabled with an experimental flag?

sporksmith (approved):
Cool!

docs/shadow_config_overview.md
stevenengler: Added.

PR #2323, opened 2022-08-03
run_tests_incremental: save most of the build directory on failure

sporksmith:
We currently have a heisenbug for which it'd be useful to inspect the
generated makefiles on failure.

More generally, we aren't charged for storage, and if we exclude the
Rust target directories, everything else is only ~50 MB.


sporksmith:
Can't seem to get this working properly; the uploader keeps hanging at 99%. Shelving for now.

PR #2324, opened 2022-08-04
Add links between 'supported_platforms.md' and 'semver.md'

stevenengler:


docs/semver.md
```diff
@@ -19,7 +19,7 @@ versions:
   ([`general.data_directory`](shadow_config_spec.html#generaldata_directory))
   will not change.
 - The PID numbering (the values and their order) will not change.
-- Any supported platforms will not be dropped.
+- Any [supported platforms](supported_platforms.html) will not be dropped.
```
sporksmith, line 22: Shouldn't this (and other links) be `.md`? It looks like using `.html` works in the docbook, but not in GitHub's renderer.... Or does `.md` not work in docbook?

sporksmith (approved):

docs/semver.md
stevenengler: Oops, good catch. I also updated a couple other '.html' links.

PR #2325, opened 2022-08-04
Remove additional scheduler policies

stevenengler:
Removes the `scheduler_policy` experimental configuration option, and removes all scheduler policies except for the "host single" policy.


robgjansen (approved):
> +8 −1,439
🚀

PR #2326, opened 2022-08-05
Remove src/dst host pointers from `Event`, and reference counts from `Scheduler` and `SchedulerPolicy`

stevenengler:
I think it's better to pass the `Host` into the `Event` when running the event. The reference counts for `Scheduler` and `SchedulerPolicy` were unused.

### DIFF
--- a/src/main/core/worker.c
+++ b/src/main/core/worker.c
@@ -455,14 +455,14 @@ void* _worker_run(void* voidWorkerThreadInfo) {
     return NULL;
 }
 
-void worker_runEvent(Event* event) {
+void worker_runEvent(Event* event, Host* host) {
 
     /* update cache, reset clocks */
     worker_setCurrentEmulatedTime(
         emutime_add_simtime(EMUTIME_SIMULATION_START, event_getTime(event)));
 
     /* process the local event */
-    event_execute(event);
+    event_execute(event, host);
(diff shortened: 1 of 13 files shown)


src/main/core/worker.c
```diff
@@ -501,7 +501,11 @@ gboolean worker_scheduleTaskAtEmulatedTime(TaskRef* task, Host* host, EmulatedTi
         return FALSE;
     }
 
-    Event* event = event_new_(task, emutime_sub_emutime(t, EMUTIME_SIMULATION_START), host, host);
+    GQuark hostID = host_getID(host);
+    guint64 eventID = host_getNewEventID(host);
```
sporksmith, line 509: It's a little odd for this to be the caller's responsibility. Maybe `event_new_` could still take the Host pointers, or at least the source host pointer, and just not store it.

sporksmith (approved):

src/main/core/worker.c
stevenengler: I agree it's odd, but it's mainly because the "event ID" itself is odd (it only exists for determinism/ordering purposes, and we could remove it by replacing it with an internal atomic global counter instead). But I'll change `event_new_` to take `..., Host* srcHost, GQuark dstHostID` instead.

PR #2340, opened 2022-08-08
Move the scheduler's priority queue to rust

stevenengler:
Will run a performance benchmark before merging.

**Edit:** No noticeable performance change, and the results appear to be the same:

ram_simtime

run_time

transfer_time_5242880 exit

### DIFF
--- a/src/main/core/work/event_queue.rs
+++ b/src/main/core/work/event_queue.rs
@@ -1,0 +1,167 @@
+use std::cmp::Reverse;
+use std::collections::binary_heap::BinaryHeap;
+use std::sync::Mutex;
+
+use crate::core::support::emulated_time::EmulatedTime;
+
+use super::event::Event;
+
+/// A queue of [`Event`]s ordered by their times.
+#[derive(Debug)]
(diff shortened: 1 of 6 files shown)


src/main/core/work/event_queue.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+}
+
+mod export {
+    use super::*;
+    use crate::core::support::simulation_time::SimulationTime;
+    use crate::cshadow as c;
+
+    /// A wrapper for [`EventQueue`] that uses interior mutability to make the ffi simpler.
+    pub struct LockedEventQueue {
```
sporksmith, line 102: `LockedEventQueue` makes it sound like this object itself is a handle to an already-locked object. Maybe `ThreadSafeQueue`?

sporksmith (approved):

src/main/core/work/event_queue.rs
stevenengler: Renamed it to `ThreadSafeEventQueue`.

PR #2345, opened 2022-08-10
Fix use-after-free in scheduler

stevenengler:
Haven't seen any crashes caused by this, but it's definitely a bug.

**Edit:** The nightly benchmark failed, possibly due to this.

```
Progress: 44% — simulated: 00:08:00/00:18:00, realtime: 01:30:00, processes failed: 0
Progress: 45% — simulated: 00:08:01/00:18:00, realtime: 01:31:00, processes failed: 0
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', main/core/support/emulated_time.rs:62:46
fatal runtime error: failed to initiate panic, error 5
2022-08-10 05:44:57 1660110297.418442 [tornettools] [INFO] Cleaning up
2022-08-10 05:44:58 1660110298.407282 [tornettools] [INFO] Done simulating; shadow returned code '-6'
Error: -10 05:44:58 1660110298.423163 [tornettools] [ERROR] Shadow simulation did not complete successfully
Error: Process completed with exit code 250.
```

### DIFF
--- a/src/main/core/scheduler/scheduler.c
+++ b/src/main/core/scheduler/scheduler.c
@@ -188,13 +188,13 @@ gboolean scheduler_push(Scheduler* scheduler, Event* event, Host* sender, Host* 
     utility_debugAssert(receiver);
 
     /* push to a queue based on the policy */
-    schedulerpolicy_push(
+    eventTime = schedulerpolicy_push(
         scheduler->policy, event, sender, receiver, scheduler->currentRound.endTime);
 
     // Store the minimum time of events that we are pushing between hosts. The
     // push operation may adjust the event time, so make sure we call this after
     // the push.
-    worker_setMinEventTimeNextRound(event_getTime(event));
+    worker_setMinEventTimeNextRound(eventTime);
 
     return TRUE;
 }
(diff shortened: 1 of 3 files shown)


src/main/core/scheduler/scheduler.c
```diff
@@ -188,13 +188,13 @@ gboolean scheduler_push(Scheduler* scheduler, Event* event, Host* sender, Host*
     utility_debugAssert(receiver);
 
     /* push to a queue based on the policy */
-    schedulerpolicy_push(
+    eventTime = schedulerpolicy_push(
```
sporksmith, line 191: Slight preference toward calling `event_getTime` instead of adding the return value to `schedulerpolicy_push`.

Adding it as a return value makes the API seem like the scheduler might schedule it for some other time than the specified event time.

sporksmith (approved):

src/main/core/scheduler/scheduler.c
stevenengler: I'm not sure what you mean here. The `schedulerpolicy_push()` may change the event time, but it also takes ownership of the event so that we can't access it with `event_getTime` later.

One alternative could be to move the code where `schedulerpolicy_push()` changes the event time into this parent function. Another could be to make the `Event` reference-counted so that we can access it after calling `schedulerpolicy_push()`.

stevenengler: > Oh! I didn't realize that. I guess that's since moving the "cpu delay" code into the scheduler?

In this case, the `schedulerpolicy_push` may delay the event if `srcHost != dstHost` and the time is before the barrier (the end of the current round).

Issue #1341, opened 2021-05-11
Packets are dropped when net buffer is full even if receiver is trying to read


robgjansen:
Suppose a UDP buffer fits 10 packets. The receiver called read and is waiting for incoming packets. The sender sent 100 packets all at the same time.

Those 100 packets arrive at the receiver all at t=0. After the first one gets added to the buffer, we notice that the receiver needs to be woken up to read. We add an event to the queue to wake up the receiver. But if it gets added _after_ those other 99 packet events that are already in the queue, then:
- the first 10 packets will make it to the buffer
- the buffer will be full
- the next 90 packets will be dropped
- we wake up the receiver so they read those 10 packets

Clearly, we need a way to wake up the receiver in between some number of packet receive events, so they get a chance to read out some of the packets, so we don't drop any of them.

---

We have verified that an infinite size buffer fixes the problem.

One possible way to deal with this is that each host keeps 2 separate queues, one for packets (which other hosts can insert into), and one for local host tasks (which a host can only insert into itself). Then at a specific nanosecond `t` and packet processing threshold `p`, we:
- process all tasks with time=`t`
- process `p` packets with time=`t`
- repeat unless no more packets or tasks exist with time=`t`

This gives us better control over executing simulator tasks vs processing new packets.

sporksmith:
Another solution we talked about is to smear on the sender side; the sender shouldn't actually be able to send all 100 packets within a single nanosecond should it?

sporksmith:
> I think the sender-side approach breaks in case you have a large number of senders sending packets to the same receiver socket all at the same instant. For UDP where you don't have a socket buffer per "connection pair", I think this breaks.

In some sense that seems realistic - if many senders each send at the receiver's full capacity simultaneously, some packets are going to be dropped, right?

OTOH in a real network there'd be some latency jitter, and maybe some "smoothing" happening where packets from multiple senders arrive at a single upstream router and are buffered before being retransmitted, making it less likely that they'd all arrive "simultaneously" in practice. Maybe your proposal is an easier solution than trying to model those things :)

robgjansen:
We do model the upstream routers and upstream router buffering in `src/main/routing/router*`, separately from modeling the sender and receiver network rates in `shadow/src/main/host/network_interface.*`. I think the issue here is independent of the upstream router buffer or the network interface buffer.

[...]

PR #2347, opened 2022-08-10
Move the event queue into the host object

stevenengler:
In this PR the event processing has been simplified and moved into the host object. Each host's event queue is reference counted and stored by both the host and the global scheduler. This allows the host to access the event queue when adding local events or when processing/iterating events, and allows other hosts to add packet events through the global worker without needing to acquire the destination host's lock.

The future plan is:
1. Rather than having the worker add packet events directly to a host's event queue, the worker will store pending packet events in separate per-host queues, and once the current scheduling round completes, those packet events will be popped from the queues and added to their respective host event queues. This will allow us to remove both reference counting and locks from the host's event queue.
2. Have two separate event queues in the host rather than just one; one for local events and one for packet events. This will allow shadow to alternate between packet events and local events and is intended to solve #1341.

There does not seem to be any significant change to the performance or simulation results. (The performance does seem slightly better for the two benchmark runs I've done, but I don't think it's really significant.)

ram_simtime

run_time

transfer_time_1048576 exit

### DIFF
--- a/src/main/host/host.c
+++ b/src/main/host/host.c
@@ -52,6 +52,9 @@ struct _Host {
 
     HostParameters params;
 
+    /* for event scheduling */
+    const ThreadSafeEventQueue* eventQueue;
+
     /* The router upstream from the host, from which we receive packets. */
     Router* router;
 
@@ -138,6 +141,8 @@ Host* host_new(const HostParameters* params) {
 
     /* thread-level event communication with other nodes */
     g_mutex_init(&(host->lock));
+
+    host->eventQueue = eventqueue_new();
 
     host->interfaces = g_hash_table_new_full(g_direct_hash, g_direct_equal,
             NULL, (GDestroyNotify) networkinterface_free);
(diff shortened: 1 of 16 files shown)


src/main/host/host.c
```diff
@@ -116,7 +120,7 @@ static SimulationTime _maxUnappliedCpuLatencyConfig;
 ADD_CONFIG_HANDLER(config_getMaxUnappliedCpuLatency, _maxUnappliedCpuLatencyConfig)
 
 /* this function is called by manager before the workers exist */
-Host* host_new(const HostParameters* params) {
+Host* host_new(const HostParameters* params, EmulatedTime simEndTime) {
```
sporksmith, line 123: Is there a reason to make this a separate parameter instead of adding to HostParameters?

sporksmith (approved):

src/main/host/host.c
stevenengler: Not really. I've moved this into `HostParameters`.

PR #2353, opened 2022-08-11
Remove O_NOFOLLOW flag when mmap file in plugin

Congyu-Liu:
When the plugin mmaps a file opened with `O_NOFOLLOW`, shadow cannot open the file for plugin and return error for mmap.

example.c:

```c
#include <sys/mman.h>
#include <stdio.h>
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <stdlib.h>
#include <unistd.h>
#define errExit(msg)    do { perror(msg); exit(EXIT_FAILURE); \
                                   } while (0)
int main() {
        int fd = open("file", O_CREAT|O_NOFOLLOW|O_RDWR, 0666);
        if (fd == -1) {
                errExit("open");
        }
        int size = 32;
        if (ftruncate(fd, size) == -1) {
                errExit("ftruncate");
        }
        void *addr = mmap(NULL, size, PROT_READ, MAP_SHARED, fd, 0);
        if (addr == MAP_FAILED) {
                errExit("mmap");
        }
        return 0;
}
```

Running the above code will produce shadow log:

```
[memory_mapper.rs:499] [shadow_rs::host::memory_manager::memory_mapper] Handling mmap result for 7ffff7ffb000..+20
[managed_thread.c:238] [_managedthread_waitForNextEvent] received shim_event 3
[managed_thread.c:238] [_managedthread_waitForNextEvent] received shim_event 3
[mman.c:175] [_syscallhandler_openPluginFile] Failed to open path '/proc/2488142/fd/10' in plugin, error 40: Too many levels of symbolic links.
```

The error 40 (`ELOOP`) is caused by the `O_NOFOLLOW` flag, which prevents the plugin opening the path where the base name is a symbolic link. So reset `O_NOFOLLOW` since plugin is opening a symbolic link file in procfs. This flag is also unnecessary at this stage since it was already enforced when shadow opening the file.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,4 +9,6 @@ A list of user-facing changes since the latest Shadow release.
   as in a [guix](https://guix.gnu.org/) environment.
 * The `setup` script now has a `--search` option, which can be used to add additional directories
   to search for pkg-config files, C headers, and libraries. It obsoletes the options `--library` and `--include`.
+* Fixed a bug causing `mmap` to fail when called on a file descriptor that was
+opened with `O_NOFOLLOW`. https://github.com/shadow/shadow/pull/2353
 * (add entry here)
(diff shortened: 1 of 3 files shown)


sporksmith (changes requested):
Nice! We should add a regression test [here](https://github.com/shadow/shadow/blob/main/src/test/memory/test_mmap.rs), though.

LMK if you need more guidance, or if you'd rather not take that part on

Congyu-Liu:
@sporksmith Please check.

sporksmith (approved):
Thanks, looks good! Will merge shortly.

PR #2354, opened 2022-08-11
Magic: use TypeOf instead of an arbitrary integer

sporksmith:
I stumbled across TypeId and realized this would be a little nicer

### DIFF
--- a/src/main/core/work/event.rs
+++ b/src/main/core/work/event.rs
@@ -6,7 +6,7 @@ use super::task::TaskRef;
 
 #[derive(Debug)]
 pub struct Event {
-    magic: Magic<0x3adf260f>,
+    magic: Magic<Self>,
     task: TaskRef,
     time: EmulatedTime,
     src_host_id: HostId,
--- a/src/main/utility/mod.rs
+++ b/src/main/utility/mod.rs
@@ -22,6 +22,7 @@ pub mod stream_len;
 pub mod syscall;
 pub mod time;
 
+use std::marker::PhantomData;
 use std::os::unix::fs::{DirBuilderExt, MetadataExt};
 use std::path::{Path, PathBuf};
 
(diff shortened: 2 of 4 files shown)


src/main/core/work/event.rs
```diff
@@ -6,7 +6,7 @@ use super::task::TaskRef;
 
 #[derive(Debug)]
 pub struct Event {
-    magic: Magic<0x3adf260f>,
+    magic: Magic<Event>,
```
stevenengler, line 9: Can you use `Self` instead of `Event`?

src/main/utility/mod.rs
```diff
@@ -107,26 +108,27 @@ pub trait IsSync: Sync {}
...
-/// `python3 -c 'import random; print(random.randint(0, 2**32))'`
+/// T should be the type of the struct that contains the Magic.
 #[derive(Debug)]
-pub struct Magic<const MAGIC: u32> {
+pub struct Magic<T: 'static + ?Sized> {
     #[cfg(debug_assertions)]
-    magic: u32,
+    magic: std::any::TypeId,
+    _phantom: PhantomData<T>,
 }
```
stevenengler, line 117: When you use `PhantomData<T>` rust assumes that the object owns (and will drop) a `T` and makes safety assumptions about that, but that's not the case for `Magic` which doesn't access an object of that type at all. But I don't really know what you're supposed to use here. Maybe `_phantom: PhantomData<*const T>` would be slightly better [to avoid a drop check](https://doc.rust-lang.org/nomicon/phantom-data.html#table-of-phantomdata-patterns)? But I don't think it really matters either way, I just thought it was mildly interesting :)

stevenengler (approved):
Cool, looks a lot nicer!

src/main/utility/mod.rs
sporksmith: Huh, interesting.

If we use `* const T`, then `Magic` becomes `!Send` and `!Sync`. We could add them back, but then we'd need to be careful to ensure doing so is valid in the future. I think implementing our `IsSend` and `IsSync` traits for `TypeId` would do it... as long as we also ensured no other fields were added without also adding such a check.

Leaving as `PhantomData<T>` for now, but added a note.

PR #2360, opened 2022-08-16
Change host shuffle method to rust

stevenengler:
We'll want to change the shuffle method to a rust method at some point, and by changing only the shuffle method now (the shuffle is changed from C to rust), we can isolate simulation performance changes due to the different host order and not conflate it with other changes.

I also think shuffling in the manager makes more sense anyways.

### DIFF
--- a/src/main/bindings/rust/wrapper.rs
+++ b/src/main/bindings/rust/wrapper.rs
@@ -3040,11 +3040,7 @@ extern "C" {
     ) -> gboolean;
 }
 extern "C" {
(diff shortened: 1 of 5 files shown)


stevenengler:
@sporksmith I plan on waiting until the benchmark runner is back up before merging this.

sporksmith (approved):

stevenengler:
Does not appear to change the simulation results or runtime:

run_time
transfer_time_5242880 exit

PR #2361, opened 2022-08-16
Move some of `Controller`'s global state to the worker module

stevenengler:
The `Controller` is currently globally accessible through the global worker pool, but since the controller exists on the stack, this is not allowed in rust (unless we use `Box::leak`). This breaks out some of the global state into a new global object. Accesses to this global state is made explicit through a `worker::WORKER_SHARED` static variable.

### DIFF
--- a/src/main/core/worker.rs
+++ b/src/main/core/worker.rs
@@ -26,6 +26,12 @@ static ALLOC_COUNTER: Lazy<Mutex<Counter>> = Lazy::new(|| Mutex::new(Counter::ne
 static DEALLOC_COUNTER: Lazy<Mutex<Counter>> = Lazy::new(|| Mutex::new(Counter::new()));
 static SYSCALL_COUNTER: Lazy<Mutex<Counter>> = Lazy::new(|| Mutex::new(Counter::new()));
 
+std::thread_local! {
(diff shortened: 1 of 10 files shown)


src/main/core/worker.rs
```diff
@@ -11,25 +11,87 @@ use crate::host::process::Process;
...
+#[derive(Debug)]
+pub struct WorkerShared {
+    pub ip_assignment: IpAssignment<u32>,
+    pub routing_info: RoutingInfo<u32>,
+    pub host_bandwidths: HashMap<std::net::IpAddr, Bandwidth>,
+    pub dns: SyncSendPointer<cshadow::DNS>,
+}
+
+impl WorkerShared {
+    pub unsafe fn dns(&self) -> *mut cshadow::DNS {
```
sporksmith, line 52: I think this isn't actually unsafe? If it is unsafe, would be good to document the preconditions for calling this function safely with a `SAFETY` comment. https://rust-lang.github.io/rust-clippy/master/#missing_safety_doc

sporksmith (approved):

src/main/core/worker.rs
stevenengler: You're right; removed the unsafe.

Issue #2368, opened 2022-08-22
process hardening causes obscure errors


sporksmith:
arti recently started enabling some process hardening via the [secmem-proc](https://crates.io/crates/secmem-proc) crate. One of the things that does is to call `prctl` with `PR_SET_DUMPABLE` to disable core dumps *and* memory inspection from other processes. In Shadow, we allow unhandled `prctl` operations to execute natively, including this one. This causes shadow's memory accesses via `process_vm_readv` to start failing with `EPERM`.

For arti in particular, this can be worked around by disabling the `harden` feature at compile time, or setting the configuration flag `application.permit_debugging = true`.

It  probably makes sense to emulate this particular operation - return success without making the native prctl call.

More generally it might be worth validating that hardening via the `secmem-proc` crate either doesn't break shadow, or fails in a more-debuggable way.

PR #2370, opened 2022-08-22
Emulate PR_SET_DUMPABLE and misc improvements

sporksmith:
This is primarily to fix https://github.com/shadow/shadow/issues/2368

* Emulate PR_SET_DUMPABLE instead of allowing it to execute natively, which interferes with Shadow's ability to access managed process memory.

Some other minor improvements:

* Return an error for some other prctl operations that would interfere with Shadow
* Add explicit cases for all known prctl operations, and return errors for any others.
* Enable the getgroups syscall to just execute natively.
* Add explicit cases for the unimplemented sched_getaffinity and sched_setaffinity syscalls, so that the logged warnings give the syscall names instead of just the numbers.
* Fix clock_gettime and futex syscalls to handle the cases where a bad user pointer is provided.

### DIFF
--- a/src/main/host/syscall/futex.c
+++ b/src/main/host/syscall/futex.c
@@ -44,15 +44,16 @@ static SysCallReturn _syscallhandler_futexWaitHelper(SysCallHandler* sys, Plugin
     // Normally, the load/compare is done atomically. Since Shadow does not run multiple
     // threads from the same plugin at the same time, we do not use atomic ops.
     // `man 2 futex`: blocking via a futex is an atomic compare-and-block operation
-    const uint32_t* futexVal = process_getReadablePtr(sys->process, futexVPtr, sizeof(uint32_t));
-    if (!futexVal) {
+    uint32_t futexVal;
+    int result = process_readPtr(sys->process, &futexVal, futexVPtr, sizeof(futexVal));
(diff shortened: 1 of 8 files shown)


src/main/host/syscall/futex.c
```diff
@@ -44,15 +44,16 @@ static SysCallReturn _syscallhandler_futexWaitHelper(SysCallHandler* sys, Plugin
...
     // threads from the same plugin at the same time, we do not use atomic ops.
     // `man 2 futex`: blocking via a futex is an atomic compare-and-block operation
-    const uint32_t* futexVal = process_getReadablePtr(sys->process, futexVPtr, sizeof(uint32_t));
-    if (!futexVal) {
+    uint32_t futexVal;
+    int result = process_readPtr(sys->process, &futexVal, futexVPtr, sizeof(futexVal));
+    if (result) {
         warning("Couldn't read futex address %p", (void*)futexVPtr.val);
-        return syscallreturn_makeDoneErrno(EFAULT);
+        return syscallreturn_makeDoneErrno(-result);
```
stevenengler, line 51: Just curious, what was the error in the old code? Or is this just to make the code nicer?

stevenengler (approved):

PR #2371, opened 2022-08-22
workflows: Replace usage of ubuntu 18.04 image with 20.04

sporksmith:
The 18.04 VM image is now deprecated.
https://github.com/actions/runner-images/issues/6002

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -22,7 +22,7 @@ jobs:
     # use the oldest kernel supported by github's CI (make sure to update the
     # minimum supported kernel version in documentation when changing)
     # https://github.com/actions/virtual-environments
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     container:
       image: 'ubuntu:22.04'
       # the default shm-size for ubuntu:18.04, but with the size increased from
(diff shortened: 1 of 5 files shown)


sporksmith:
I tried using a symbolic name for the VM in the test matrix, and then a GH [expression](https://docs.github.com/en/actions/learn-github-actions/expressions#about-expressions) to map that to the actual VM name, but couldn't get it to work. GH doesn't tell me much other than that the line with the GH expression is invalid yml. It's here if you want to take a look: https://github.com/shadow/shadow/pull/2371/commits/58d20805c5b9168394b61320a2da06b34cb209d8

robgjansen:
Let me know when I should wave the magic wand to update the required status checks. :)

stevenengler (approved):
> I tried using a symbolic name for the VM in the test matrix, and then a GH expression to map that to the actual VM name, but couldn't get it to work. GH doesn't tell me much other than that the line with the GH expression is invalid yml. It's here if you want to take a look: 58d2080

I don't see anything wrong there. Maybe something needs extra quotes or something, but otherwise I'm not sure what could be wrong.

sporksmith:
> I don't see anything wrong there. Maybe something needs extra quotes or something, but otherwise I'm not sure what could be wrong.

Thanks - punting on it for now; had Rob update the required checks during our 1:1 so I could merge this at the same time.

PR #2372, opened 2022-08-22
Move more global state to the worker module

stevenengler:
These changes allow us to remove the `WorkerPool`s pointers to the `Scheduler` and the `ConfigOptions`, which brings us a step closer to being able to write the scheduler and work pool in rust. The manager is becoming a bit overloaded with stuff, but it'll be easier to reorganize things later once it's all in rust.

### DIFF
--- a/src/main/core/worker.rs
+++ b/src/main/core/worker.rs
@@ -7,13 +7,12 @@ use crate::core::scheduler::runahead::Runahead;
 use crate::core::sim_config::Bandwidth;
 use crate::core::support::emulated_time::EmulatedTime;
 use crate::core::support::simulation_time::SimulationTime;
+use crate::core::work::event::Event;
+use crate::core::work::event_queue::ThreadSafeEventQueue;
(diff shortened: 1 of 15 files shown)


src/main/core/worker.rs
```diff
@@ -499,6 +506,13 @@ mod export {
         Worker::with_mut(|w| w.shared.increment_plugin_error_count()).unwrap()
     }
 
+    /// SAFETY: This is unsafe since we're returning a pointer for a reference that's not valid
```
sporksmith, line 509: Nit - I think these conventionally such comments are phrased as describing what the caller's responsibility is when using this function to avoid unsoundness. e.g. "SAFETY: The returned pointer must not be accesses after this worker thread has exited"

sporksmith (approved):

src/main/core/worker.rs
stevenengler: Updated.

PR #2376, opened 2022-08-24
Replace the `CountDownLatch` with a rust `Barrier`

stevenengler:


stevenengler:
Closing as the performance (sim runtime) was much worse, likely due to the extra waiting required by the worker threads (the worker threads would also need to wait when counting down).

Issue #2379, opened 2022-08-25
Shadow uio is broken


Congyu-Liu:
**Describe the issue**

1. The return value of `writev` and `returnv` is incorrect. It returns the length of the last iovec. 

A test case for `writev`:
```c
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <string.h>
#include <sys/uio.h>
#include <sys/socket.h>
#include <netinet/in.h>
#include <arpa/inet.h>
#define ARRAY_LENGTH(a) (sizeof((a))/sizeof((a)[0]))
#define errExit(msg)    do { perror(msg); exit(EXIT_FAILURE); \
                                   } while (0)
int main(int argc, char *argv[]) {
        if (argc != 4) {
                exit(1);
        }
        char *serv_ip = argv[2];
        int serv_port = atoi(argv[3]);
        struct sockaddr_in sa;
        bzero(&sa, sizeof(sa));
        sa.sin_family = AF_INET;
        sa.sin_addr.s_addr = inet_addr(serv_ip);
        sa.sin_port = htons(serv_port);
        char *str_arr[] = {
                "hi",
                "hello",
        };
        struct iovec v_send[10];
        memset(v_send, 0, sizeof(v_send));
        int iov_size = ARRAY_LENGTH(str_arr);
        int total_len = 0;
        for (int i = 0; i < iov_size; i++) {
                v_send[i].iov_base = str_arr[I];
                v_send[i].iov_len = strlen(str_arr[i])+1;
                total_len += v_send[i].iov_len;
        }
        if (!strcmp(argv[1], "client")) {
                int s, ret;
                if((s = socket(AF_INET, SOCK_STREAM, 0)) == -1) {
                        errExit("socket");
                }
                
                if (connect(s, (struct sockaddr*)&sa, sizeof(sa))) {
                        errExit("connect");
                }
                ret = writev(s, v_send, iov_size);
                if (ret != total_len) {
                        fprintf(stderr, "return %d, expect %d\n", ret, total_len);
                        exit(1);
                }
        } else if (!strcmp(argv[1], "server")) {
                struct sockaddr_in ca;
                int s, ret, ca_len, c;

                if((s = socket(AF_INET, SOCK_STREAM, 0)) == -1) {
                        errExit("socket");
                }
                if ((bind(s, (struct sockaddr*)&sa, sizeof(sa)))) {
                        errExit("bind");
                }
                if ((listen(s, 10))) {
                        errExit("listen");
[...]

PR #2380, opened 2022-08-25
Fix shadow uio length issue

Congyu-Liu:


stevenengler (approved):
Looks good to me!

(For future reference, this fixes one of the issues in #2379. I don't think it's worth adding a test in this PR since iov is still broken in a few ways, so it would be better to add/enable tests once we get the other issues fixed.)

PR #2381, opened 2022-08-25
Move small worker functions to rust

stevenengler:


src/main/core/worker.rs
```diff
@@ -680,6 +680,26 @@ mod export {
         Worker::is_alive()
     }
 
+    #[no_mangle]
+    pub extern "C" fn worker_resolveIPToAddress(ip: libc::in_addr_t) -> *const cshadow::Address {
+        Worker::with_mut(|w| {
```
sporksmith, line 685: Optional - maybe have these use `with_ref` instead of `with_mut`? We'd either need to cast the dns  pointer to a mut pointer here, with an explanation that we know the C function doesn't actually mutate, or move the cast inside the C function when calling the hash table API.

sporksmith (approved):

src/main/core/worker.rs
stevenengler: Good catch, fixed (changed to `with` instead of `with_mut`)!

PR #2383, opened 2022-08-25
Replace the C `CountDownLatch` with a rust version

stevenengler:


src/main/utility/count_down_latch.rs
```diff
@@ -0,0 +1,78 @@
+use std::sync::{Condvar, Mutex};
```
sporksmith, line 1: Code looks fine, but could use some documentation and unit tests

sporksmith (approved):
Optional - in general I think it might be worth starting to create/move thing that don't have dependencies on the C code in the shadow-shim-helper-rs crate instead of the main crate, to help start to detangle some of the circular dependencies. Even if the shim doesn't need it, I don't think it hurts anything for it to be there.

Issue #2389, opened 2022-08-30
Release pre-built binaries


sporksmith:
Pre-built binaries would save some complexity and time in *other* CI that uses shadow (e.g. arti/tor), since those workflows could just download the binaries instead of installing all of the build dependencies and building. There are some other workarounds, such as caching built binaries, but still this might be nice.

I *think* it wouldn't be too difficult. For shadow itself the only runtime dependencies are glibc and glib. A pre-built binary should be usable on any system where those are present with a version no *older* than the ones shadow was built against. i.e. we'd want to build on the supported platform with the oldest versions of those libraries.

It'd also be useful if the shadow binary didn't depend on an absolute path. It currently sets an rpath to the absolute installation directory, which means that it would need to be in the same directory on any system. We could use `$ORIGIN` to make this path relative to the shadow binary though, in which case the shadow installation directory can work from wherever (e.g. /opt/shadow, ~/opt/shadow, ~/Downloads/shadow, ...).

sporksmith:
I've verified that  with #2390, I can build shadow in an ubuntu:18 docker container, copy the installation directory out to my host machine, and run shadow from there on my ubuntu:20 host machine. The flow is roughly:

```
# Build inside the ubuntu:18 image
$ ci/run.sh -i -c ubuntu:18.04 -C gcc -b release
# Create a container using that image
$ docker run shadowsim/shadow-ci:ubuntu-18.04-gcc-release  ls /opt/shadow
$ docker container ls --all
# Copy it out
$ docker cp 77a3fab26fd0:/opt/shadow shadow-release
```

PR #2391, opened 2022-08-30
Make shadow installation directory relocatable

sporksmith:
We probably especially want this if we decide to release pre-built binaries ourselves (#2389), but even if we don't, it gives users some additional flexibility.

Make shadow installation directory relocatable

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,7 +58,10 @@ endforeach(library)
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 
 ## use the installed shim path only when shadow is installed
-set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
+##
+## We use "$ORIGIN" here to make the shadow installation directory relocatable.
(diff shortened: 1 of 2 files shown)


ci/container_scripts/build_and_install.sh
```diff
@@ -34,5 +34,5 @@ case "$BUILDTYPE" in
         ;;
 esac
 
-./setup build -j4 --test --extra --werror $OPTIONS
-./setup install
+./setup build -j4 --test --extra --werror --prefix=/opt/shadow $OPTIONS
+./setup install
```
stevenengler, line 38: Why change the installation directory? I think it would be better to run tests using the default options here.

stevenengler (approved):

ci/container_scripts/build_and_install.sh
sporksmith: Installing into a dedicated directory `/opt/shadow` instead of merging into e.g. ~/.local with other things allows us to easily copy out a "releasable" shadow as in https://github.com/shadow/shadow/issues/2389#issuecomment-1232264491. It is a bit orthogonal from the meat of this PR though. I'll leave it out for now.

PR #2401, opened 2022-09-02
Run tests under Miri in CI

sporksmith:
This seems like a good idea with our increasing adventures into `unsafe`-land

While putting this together I found that we're not actually running tests in our rust crates that don't have a cmake target. This runs them "by accident" since I'm invoking cargo directly, and I made a couple small corresponding fixes.  It'd probably be good to ensure those run via `setup test` as well, but I think that'll be a separate PR.

### DIFF
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -1,0 +1,43 @@
+# Syntax reference:
+# https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions
+
+name: Rust Sanitizers
+permissions: read-all
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    types: [opened, synchronize]
+
+
+jobs:
+  miri:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        persist-credentials: false
+        # Run on PR head instead of merge result. Running on the merge
+        # result can give confusing results, and we require PR to be up to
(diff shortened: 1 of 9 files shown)


src/main/core/support/configuration.rs
```diff
@@ -1084,7 +1084,8 @@ pub fn parse_string_as_args(args_str: &OsStr) -> Result<Vec<OsString>, String> {
     Ok(args)
 }
 
-#[cfg(test)]
+// XXX: Won't run under miri due to foreign function calls to process_parseArgStr
+#[cfg(all(test, not(miri)))]
```
stevenengler, line 1088: Should we disable individual tests rather than all?

.github/workflows/sanitizers.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    steps:
+    - name: Install a nightly with miri
+      uses: actions-rs/toolchain@v1
+      with:
+          # Use a nightly that included miri
+          # https://rust-lang.github.io/rustup-components-history/
+          toolchain: nightly-2022-08-27
+          override: true
+          components: miri
```
stevenengler: It seems like `actions-rs/toolchain` is unmaintained? Maybe use https://github.com/dtolnay/rust-toolchain instead?

**Edit:** nvm I see this is changed later to use rustup built into the image.

stevenengler (approved):

src/main/core/support/configuration.rs
sporksmith: Done

PR #2403, opened 2022-09-02
shadow-shim-helper-rs: omit C APIs in rust library

sporksmith:
We build this library both as a system static library, which can be
linked from C, and as a Rust static library, which can be linked from
Rust. We end up pulling in both in the final linking of Shadow.

For the non-#[no_mangle] (C) api's, I think Rust knows how to sort this
out.

For the #[no_mangle] api's, though, the linker can end up with multiple
versions of the same symbol and fail. For some reason this currently
only happens in the coverage build.

We can omit the #[no_mangle] APIs when compiling the Rust version of the
library, since Rust code shouldn't call those APIs anyway.


sporksmith:
Going to back-burner this for the moment. I'm experimenting with compiling all the rust crates in a single cargo invocation, which I think gets closer to the root of the problem that crates are ending up compiled multiple times with different flags

PR #2412, opened 2022-09-14
Add rust synchronization primitives needed for the scheduler

stevenengler:
Adds the synchronization primitives (semaphore and count down latch) needed for the rust scheduler.

This count down latch is a little more complex than the C version since it supports multiple waiters. The main difficulties of using multiple waiters in the C version are:
1. There isn't a nice way to synchronize which waiting thread will reset the latch, since it must always be the last waiting thread that resets the latch.
2. If the waiters are modified so that the waiting thread that last takes the lock is the one to reset the latch, it leads to a new issue where the waiting threads could be notified, and one thread resumes and loops all the way back to the wait again before the other threads have a chance to resume.

### DIFF
--- a/src/main/utility/synchronization/count_down_latch.rs
+++ b/src/main/utility/synchronization/count_down_latch.rs
@@ -1,0 +1,379 @@
+use std::sync::{Arc, Condvar, Mutex};
+
+/// A latch counter.
+///
+/// If a counter is cloned, it will inherit the counter's state for the current generation. For
+/// example if a counter is cloned after it has already counted down, then the new counter will also
+/// be treated as if it had already counted down in the current generation. If a counter is cloned
+/// before it has counted down, then the new counter will also need to count down in the current
+/// generation.
+#[derive(Debug)]
+pub struct LatchCounter {
+    inner: Arc<LatchInner>,
+    /// An ID for this counter's count-down round.
+    generation: usize,
+}
+
+/// A latch waiter.
+///
+/// If a waiter is cloned, it will inherit the waiter's state for the current generation. For
+/// example if a waiter is cloned after it has already waited, then the new waiter will also be
+/// treated as if it had already waited in the current generation. If a waiter is cloned before it
+/// has waited, then the new waiter will also need to wait in the current generation.
+#[derive(Debug)]
+pub struct LatchWaiter {
+    inner: Arc<LatchInner>,
+    /// An ID for this waiter's count-down round.
+    generation: usize,
+}
+
+#[derive(Debug)]
+struct LatchInner {
+    lock: Mutex<LatchState>,
+    cond: Condvar,
+}
+
+#[derive(Debug)]
+struct LatchState {
+    /// The current latch "round".
+    generation: usize,
+    /// Number of counters remaining.
+    counters: usize,
+    /// Number of waiters remaining.
+    waiters: usize,
+    /// Total number of counters.
+    total_counters: usize,
+    /// Total number of waiters.
+    total_waiters: usize,
+}
+
+/// Build a latch counter and waiter. The counter and waiter can be cloned to create new counters
+/// and waiters.
+pub fn build_count_down_latch() -> (LatchCounter, LatchWaiter) {
+    let inner = Arc::new(LatchInner {
+        lock: Mutex::new(LatchState {
+            generation: 0,
+            counters: 1,
+            waiters: 1,
+            total_counters: 1,
+            total_waiters: 1,
+        }),
+        cond: Condvar::new(),
+    });
+
+    let counter = LatchCounter {
+        inner: Arc::clone(&inner),
+        generation: 0,
+    };
+
+    let waiter = LatchWaiter {
+        inner,
+        generation: 0,
+    };
+
+    (counter, waiter)
+}
+
+impl LatchState {
+    pub fn advance_generation(&mut self) {
+        debug_assert_eq!(self.counters, 0);
+        debug_assert_eq!(self.waiters, 0);
+        self.counters = self.total_counters;
+        self.waiters = self.total_waiters;
+        self.generation = self.generation.wrapping_add(1);
+    }
+}
+
+impl LatchCounter {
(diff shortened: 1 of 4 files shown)


src/main/utility/synchronization/count_down_latch.rs
```diff
@@ -0,0 +1,376 @@
...
+struct LatchInner {
+    lock: Mutex<LatchState>,
+    cond: Condvar,
+}
+
+#[derive(Debug)]
+struct LatchState {
+    /// The current latch "round".
+    generation: usize,
+    /// Number if counters remaining.
```
robgjansen, line 40: "if" -> "of"

src/main/utility/synchronization/count_down_latch.rs
```diff
@@ -0,0 +1,376 @@
...
+                let latch_gen = lock.generation;
+                std::mem::drop(lock);
+                panic!(
+                    "Counter generation does not match latch generation ({} != {})",
+                    self.generation, latch_gen
+                );
+            }
+
+            lock.counters = lock.counters.checked_sub(1).unwrap();
+            self.generation = self.generation.wrapping_add(1);
```
robgjansen, line 106: Can we do this increment while we're not holding the lock?

src/main/utility/synchronization/count_down_latch.rs
```diff
@@ -0,0 +1,376 @@
...
+
+        let mut lock = self
+            .inner
+            .cond
+            // wait until we're in the active generation and all counters have counted down
+            .wait_while(lock, |x| self.generation != x.generation || x.counters > 0)
+            .unwrap();
+
+        lock.waiters = lock.waiters.checked_sub(1).unwrap();
+        self.generation = self.generation.wrapping_add(1);
```
robgjansen, line 132: Could increment while not holding the lock (but maybe not even worth it).

src/main/utility/synchronization/semaphore.rs
```diff
@@ -0,0 +1,205 @@
+use std::cell::UnsafeCell;
+use std::sync::Arc;
+
+/// A libc semaphore that provides synchronization between threads. Any memory barrier/ordering
```
robgjansen, line 4: I guess there is not a rust equivalent in the std lib? Maybe worth explaining why we need our own (optional).

robgjansen (approved):

src/main/utility/synchronization/count_down_latch.rs
```diff
@@ -0,0 +1,376 @@
...
+    waiters: usize,
+    /// Total number of counters.
+    total_counters: usize,
+    /// Total number of waiters.
+    total_waiters: usize,
+}
+
+/// Build a latch counter and waiter. The counter and waiter can be cloned to create new counters
+/// and waiters.
+pub fn count_down_latch() -> (LatchCounter, LatchWaiter) {
```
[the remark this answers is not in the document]
stevenengler: I'll add that. The reason why I used this name was to make it similar to [`std::sync::mpsc::channel`](https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html). For example `pub fn channel<T>() -> (Sender<T>, Receiver<T>)`. But either way is good with me.

src/main/utility/synchronization/count_down_latch.rs
```diff
@@ -0,0 +1,376 @@
...
+///
+/// If a counter is cloned, it will inherit the counter's state for the current generation. For
+/// example if a counter is cloned after it has already counted down, then the new counter will also
+/// be treated as if it had already counted down in the current generation. If a counter is cloned
+/// before it has counted down, then the new counter will also need to count down in the current
+/// generation.
+#[derive(Debug)]
+pub struct LatchCounter {
+    inner: Arc<LatchInner>,
+    /// This counter's generation.
```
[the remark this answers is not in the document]
stevenengler: Changed the wording. It's more like an ID than the number of rounds since it will eventually wrap (and for example using less-than or greater-than would lead to errors), so I used "an ID for this counter's count-down round" instead.

src/main/utility/synchronization/count_down_latch.rs
```diff
@@ -0,0 +1,376 @@
...
+/// A latch waiter.
+///
+/// If a waiter is cloned, it will inherit the waiter's state for the current generation. For
+/// example if a waiter is cloned after it has already waited, then the new waiter will also be
+/// treated as if it had already waited in the current generation. If a waiter is cloned before it
+/// has waited, then the new waiter will also need to wait in the current generation.
+#[derive(Debug)]
+pub struct LatchWaiter {
+    inner: Arc<LatchInner>,
+    /// This waiter's generation.
```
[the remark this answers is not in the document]
stevenengler: As above.

src/main/utility/synchronization/count_down_latch.rs
```diff
@@ -0,0 +1,376 @@
...
+    let waiter = LatchWaiter {
+        inner,
+        generation: 0,
+    };
+
+    (counter, waiter)
+}
+
+impl LatchState {
+    pub fn next_generation(&mut self) {
```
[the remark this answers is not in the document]
stevenengler: Changed to `advance_generation`.

src/main/utility/synchronization/count_down_latch.rs
stevenengler: Fixed

src/main/utility/synchronization/count_down_latch.rs
stevenengler: Moved this line to the end.

src/main/utility/synchronization/count_down_latch.rs
stevenengler: Moved this line to the end.

src/main/utility/synchronization/count_down_latch.rs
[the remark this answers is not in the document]
robgjansen: Oh, ok :) I don't feel strongly.

PR #2420, opened 2022-09-28
Use bear instead of cmake to generate compile_commands.json

sporksmith:
Now that most of our C files are built via Rust, cmake's generation of compile_commands.json doesn't include them. Use the external tool [bear](https://github.com/rizsotto/Bear) instead.


setup
```diff
@@ -199,12 +204,20 @@ def build(args, remaining):
...
+
     # create directories
     if not os.path.exists(builddir): os.makedirs(builddir)
     if not os.path.exists(installdir): os.makedirs(installdir)
 
     # build up args string for the cmake command
-    cmake_cmd = "cmake " + rootdir + " -DCMAKE_INSTALL_PREFIX=" + installdir
+    cmake_cmd = ""
+    if args.do_compile_commands:
+        cmake_cmd += "bear --"
```
robgjansen, line 218: Do we want to add an entry in the docs for how to run `bear` externally, which may be useful for devs? Maybe here: https://shadow.github.io/docs/guide/coding.html

robgjansen (approved):

setup
sporksmith: Done

PR #2421, opened 2022-09-28
Run rust unit tests (including doc tests)

sporksmith:
Accidentally dropped running any rust unit tests in https://github.com/shadow/shadow/pull/2404

We were also previously not running doc tests, and I think not running tests for any crates for which we don't build a `staticlib` for C code to link to. (I'm not sure whether there were any such tests before, but there increasingly will be)

For simplicity and maintainability I just invoke `cargo test` now, instead of trying to directly run the test executables. While the cargo behavior of always rebuilding (if needed) may be slightly surprising, this approach should be more robust and maintainable.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -22,6 +22,11 @@ foreach(DEFINITION IN LISTS COMPILE_DEFINITIONS)
   set(RUST_CXXFLAGS "${RUST_CXXFLAGS} -D${DEFINITION}")
 endforeach()
 
+set(RUST_TARGETS "--lib")
+if(SHADOW_TEST STREQUAL ON)
+  set(RUST_TARGETS "${RUST_TARGETS} --tests")
+endif()
+
 ## build the rust library
 set(CARGO_ENV_VARS "${CARGO_ENV_VARS} RUSTFLAGS=\"${RUSTFLAGS}\"")
 set(CARGO_ENV_VARS "${CARGO_ENV_VARS} CFLAGS=\"${RUST_CFLAGS}\"")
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -49,6 +54,23 @@ foreach(LIBRARY shadow-shim-helper-rs logger shadow-shmem shadow-tsc shadow-rs)
...
+## the crates from cmake. (This is a larger set than the library-list above, which are only
+## the crates that we need to link from C code)
+##
+## Just using `cargo test` for now. It potentially (re)builds code, but seems
+## better than trying to # fight the "rust way" here. It also includes doc tests,
+## which normally don't have persistent executables at all.
+add_test(NAME rust-unit-tests
+         COMMAND bash -c "${CARGO_ENV_VARS} cargo test ${RUST_BUILD_FLAG} --manifest-path \"${CMAKE_CURRENT_SOURCE_DIR}\"/Cargo.toml --target-dir \"${TARGETDIR}\" --features \"${RUST_FEATURES}\"")
+## Longer timeout here and run serially, since it's running a whole test suite,
+#and may end up rebuilding code # if anything has changed.
```
robgjansen, line 71: There are two spurious `#` characters in the comments here.

robgjansen (approved):

src/CMakeLists.txt
sporksmith: Oops, I think this was at least partly an IDE auto-re-wrapping error; it doesn't seem to like the double ## leaders. Changed them to a single #.

PR #2424, opened 2022-09-29
Lint rust documentation in CI

stevenengler:
Linting the code documentation might be a little over-the-top, but I don't think it will be that bad and it makes sure that our documentation links are correct.


sporksmith (approved):
Good idea

PR #2429, opened 2022-09-30
Run C compiler in parallel

stevenengler:
Seems to significantly reduce the time to build Shadow's C code. For example:

#### Before:

1664575288_grim

#### After:

1664575271_grim


sporksmith (approved):
Nice - how did you generate those graphs btw?

stevenengler:
> Nice - how did you generate those graphs btw?

I added `--timings` to `cargo build`, then it generates an html file in the target directory.

PR #2443, opened 2022-10-03
cmake: set C compiler debug/release flags globally

sporksmith:
These are redundant for the C code built via Cargo, but still needed for the few remaining C modules built via cmake, including the openssl preload library.


sporksmith:
Kicked off a manual benchmark run to see if this explains the (whole) performance  regression, but we probably want to merge it either way. https://github.com/shadow/benchmark/actions/runs/3175726646

PR #2446, opened 2022-10-03
Use formatted help output

stevenengler:
This uses coloring and underlining in clap's help output (including help error messages).

Example:

1664825117_grim

### DIFF
--- a/src/main/core/main.rs
+++ b/src/main/core/main.rs
@@ -32,11 +32,14 @@ pub fn run_shadow<'a>(args: Vec<&'a OsStr>) -> anyhow::Result<()> {
     let options = match CliOptions::try_parse_from(args.clone()) {
         Ok(x) => x,
         Err(e) => {
+            // will print to either stdout or stderr with formatting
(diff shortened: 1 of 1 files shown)


src/main/core/main.rs
```diff
@@ -32,11 +32,10 @@ pub fn run_shadow<'a>(args: Vec<&'a OsStr>) -> anyhow::Result<()> {
     let options = match CliOptions::try_parse_from(args.clone()) {
         Ok(x) => x,
         Err(e) => {
+            e.print().unwrap();
             if e.use_stderr() {
-                eprint!("{}", e);
                 std::process::exit(1);
```
sporksmith, line 37: Semi-orthogonal, but could use a comment here why we're choosing the exit-code based on `use_stderr`

sporksmith (approved):

src/main/core/main.rs
stevenengler: Added some comments.

PR #2447, opened 2022-10-04
Allow for sourcing executables from $PATH

### LINKED ISSUE #2410: Adding NixOS support
Adding nix support would be pretty easy, but (like guix #2304) doesn't support absolute paths like `/bin/bash` and the like in shell scripts.

Would this project be open to a PR request that changed all the absolute paths to paths sourced from PATH and added some combination of `flake.nix`, `flake.lock` and/or a `shell.nix` to the root directory?

zyansheep:
This is a pretty large PR (with some possibly controversial changes) so just tell me what I should remove / change / split up to get this merged.

This PR does the following:
Reimplements `tilde_expansion` as `resolve_path`. It now takes a `&Path`, returns an `anyhow::Result<PathBuf>`, and looks in `$PATH` if input is a standalone string (e.g. `true` will now resolve to `/bin/true`)
Various edits to fix yaml files with absolute paths or relative paths that conflicted with the $PATH lookup.

### DIFF
--- a/src/main/core/sim_config.rs
+++ b/src/main/core/sim_config.rs
@@ -342,20 +342,34 @@ fn build_process(proc: &ProcessOptions) -> anyhow::Result<Vec<ProcessInfo>> {
             .with_context(|| format!("Failed to parse arguments: {x}"))?,
     };
 
-    // perform shell expansion
-    let plugin = tilde_expansion(proc.path.to_str().unwrap());
+    let proc_path = proc.path.to_str().unwrap();
 
-    // set argv[0] as the user-provided expanded string, not the canonicalized version
-    args.insert(0, plugin.clone().into());
+    // Check if the path has no '/' where it is likely to be a executable that should be sourced from $PATH
+    // TODO: In next breaking release, if likely_not_path is true, source from $PATH by default and remove deprecation warning below. Also make sure to fix all the broken tests that will result.
+    let likely_not_path = proc_path.chars().find(|c|*c == '/').is_none();
+    
+    let expanded_path = tilde_expansion(proc_path);
 
-    // get the full canonical path
-    let plugin = plugin
-        .canonicalize()
-        .with_context(|| format!("Failed to canonicalize plugin path '{}'", plugin.display()))?;
+    let plugin = match expanded_path
+        .canonicalize().with_context(|| format!("Failed to canonicalize plugin path '{}'", expanded_path.display())) {
+        Ok(plugin) => { // Warn that relative path "thing" should be changed to "./thing" 
+            if likely_not_path { log::warn!("The relative path \"{0}\" will be deprecated in favor of \"./{0}\"", proc_path) }
+            plugin
+        }
+        Err(_) if likely_not_path => {
+            let plugin = which::which(proc_path).with_context(|| format!("Failed to resolve \"{proc_path}\" from $PATH"))?;
+            log::info!("Resolved process path \"{proc_path}\" to \"{}\"", plugin.display());
+            plugin
+        }
+        Err(err) => Err(err)?,
+    };
 
     // verify that the path is a file and is executable
     verify_plugin_path(&plugin)
         .with_context(|| format!("Failed to verify plugin path '{}'", plugin.display()))?;
+
+    // set argv[0] as the user-provided expanded string, not the canonicalized version
+    args.insert(0, expanded_path.into());
 
     Ok(vec![
         ProcessInfo {
@@ -438,7 +452,7 @@ fn generate_routing_info(
 /// Check that the plugin path is valid.
 fn verify_plugin_path(path: impl AsRef<std::path::Path>) -> anyhow::Result<()> {
     let path = path.as_ref();
-    let metadata = std::fs::metadata(&path).unwrap();
+    let metadata = std::fs::metadata(&path)?;
 
     if !metadata.is_file() {
         return Err(anyhow::anyhow!("The path is not a file"));
--- a/src/main/core/sim_config.rs
+++ b/src/main/core/sim_config.rs
@@ -6,6 +6,7 @@ use std::path::PathBuf;
 use std::time::Duration;
 
 use anyhow::Context;
+use log::warn;
 use rand::{Rng, SeedableRng};
 use rand_xoshiro::Xoshiro256PlusPlus;
 
@@ -342,38 +343,56 @@ fn build_process(proc: &ProcessOptions) -> anyhow::Result<Vec<ProcessInfo>> {
             .with_context(|| format!("Failed to parse arguments: {x}"))?,
     };
 
-    let proc_path = proc.path.to_str().unwrap();
+    let expanded_path = tilde_expansion(proc.path.to_str().unwrap());
 
-    // Check if the path has no '/' where it is likely to be a executable that should be sourced from $PATH
-    // TODO: In next breaking release, if likely_not_path is true, source from $PATH by default and remove deprecation warning below. Also make sure to fix all the broken tests that will result.
-    let likely_not_path = proc_path.chars().find(|c|*c == '/').is_none();
-    
-    let expanded_path = tilde_expansion(proc_path);
+    // We currently use `which::which`, which searches the `PATH` similarly to a
+    // shell.
+    let new_canonical_path: Result<PathBuf, anyhow::Error> = which::which(&expanded_path)
+        .map_err(anyhow::Error::from)
+        // `which` returns an absolute path, but it may still contain
+        // symbolic links, .., etc.
+        .and_then(|p| p.canonicalize().map_err(anyhow::Error::from));
+    // We previously used only `std::fs::canonicalize`, which doesn't search
+    // `PATH`, and *does* search the current directory.
+    let legacy_canonical_path: Result<PathBuf, anyhow::Error> = expanded_path
+        .canonicalize()
+        .map_err(|e| anyhow::Error::from(e));
 
-    let plugin = match expanded_path
-        .canonicalize().with_context(|| format!("Failed to canonicalize plugin path '{}'", expanded_path.display())) {
-        Ok(plugin) => { // Warn that relative path "thing" should be changed to "./thing" 
-            if likely_not_path { log::warn!("The relative path \"{0}\" will be deprecated in favor of \"./{0}\"", proc_path) }
-            plugin
-        }
-        Err(_) if likely_not_path => {
-            let plugin = which::which(proc_path).with_context(|| format!("Failed to resolve \"{proc_path}\" from $PATH"))?;
-            log::info!("Resolved process path \"{proc_path}\" to \"{}\"", plugin.display());
-            plugin
-        }
-        Err(err) => Err(err)?,
+    let canonical_path = if new_canonical_path.is_ok()
+        && legacy_canonical_path.is_ok()
+        && *new_canonical_path.as_ref().unwrap() != *legacy_canonical_path.as_ref().unwrap()
+    {
+        warn!("Ambiguous path: {:?} resolves to either {:?} or {:?}. Using {:?} for backards compatibility, but a future version of shadow will not implicitly search the current working directory. Consider making absolute or prefixing with './'",
+            expanded_path,
+            legacy_canonical_path.as_ref().unwrap(),
+            new_canonical_path.as_ref().unwrap(),
+            legacy_canonical_path.as_ref().unwrap());
+        legacy_canonical_path.as_ref().unwrap()
+    } else if let Ok(p) = new_canonical_path.as_ref() {
+        p
+    } else if let Ok(p) = legacy_canonical_path.as_ref() {
+        warn!("Expanded path {:?} only via legacy fallback search, which implicitly searched current working dir. Consider prefixing with './'", proc.path.to_str().unwrap());
+        p
+    } else {
+        return Err(new_canonical_path
+            .with_context(|| format!("Failed to resolve plugin path '{:?}'", expanded_path))
+            .unwrap_err());
     };
 
-    // verify that the path is a file and is executable
-    verify_plugin_path(&plugin)
-        .with_context(|| format!("Failed to verify plugin path '{}'", plugin.display()))?;
+    verify_plugin_path(&canonical_path)
+        .with_context(|| format!("Failed to verify plugin path '{:?}'", canonical_path))?;
+    log::info!(
+        "Resolved binary path {:?} to {:?}",
+        proc.path,
+        canonical_path
+    );
 
     // set argv[0] as the user-provided expanded string, not the canonicalized version
     args.insert(0, expanded_path.into());
 
     Ok(vec![
         ProcessInfo {
-            plugin: plugin,
+            plugin: canonical_path.clone(),
             start_time,
             stop_time,
             args: args,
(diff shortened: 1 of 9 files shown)


sporksmith:
>  Adds flake.nix, shell.nix & flake.lock for creating NixOS developer environments

I don't think we're inclined to take this - minimally it should be a separate PR. We don't want to imply nix support by including these files without actually committing to keeping it working, which we don't have the resources to do right now.

Could you drop these changes from the PR for now to keep the discussion focused on the other part?

> Reimplements tilde_expansion as resolve_path. It now takes a &Path, returns an anyhow::Result<PathBuf>, and looks in $PATH if input is a standalone string (e.g. true will now resolve to /bin/true)
> Various edits to fix yaml files with absolute paths or relative paths that conflicted with the $PATH lookup.

I think it'd be good to have this or something like this. e.g. it'd also be nice if tor simulation configs didn't have to hard-code paths to tor, tgen, etc.

Locating binaries using the `PATH` environment variable is convenient, though opens the door to a little bit of confusion about what binary actually ran. I think this could be mitigated by one or both of:

* Include the fully resolved path in shadow's log and `processed-config.yaml`. @stevenengler do you know how difficult it'd be to do the latter?
* Maybe use a command-line option like `--path` or `--binary-search-path` instead of the environment variable, to make this behavior a bit more explicit.

@stevenengler and @robgjansen - thoughts?

robgjansen:
Overall this seems like a reasonable improvement.

I agree with sporksmith's points:

- We don't want to commit to supporting and maintaining the NixOS files; thanks for removing those.
- We really should be logging the absolute path to the binaries we are executing. It's not fun spending hours debugging unexpected sim behavior only to realize it's running the wrong binary. (I swear that's never happened to me 😢)
- I lean toward making the behavior explicit and therefore user-predictable; searching PATH is a bit of automation that may not be immediately obvious to users and may lead to confusion. A `--binary-search-path` option would also allow setting a different path for the simulator than the one you might want for your terminal.

stevenengler:
> @stevenengler and @robgjansen - thoughts?

Back when we switched to YAML, we had a few discussions about whether to search PATH for the binary, or to require a hard-coded binary path. At the time we decided that we didn't want to search the PATH, use a flag like `--binary-search-path`, or use any kind of YAML templating to deduplicate paths. I'm okay with searching the PATH or using a flag like `--binary-search-path`, but I'm wondering why we want to change it now. Is it because we've decided that our previous decision was wrong?

I like the `--binary-search-path`/`general.binary_search_path` option, but does this solve what the PR is intended for? This change to search the PATH is meant to make it so that a single shadow config file will work in different environments (such as NixOS). Will our examples and tests just use `--binary-search-path $PATH` everywhere? Is there ever a reason to use something other than `--binary-search-path $PATH`? If not, why not just search PATH directly instead?

> A --binary-search-path option would also allow setting a different path for the simulator than the one you might want for your terminal.

Why `shadow --binary-search-path "/home/user/.local/bin" config.yaml` rather than just something like `PATH="/home/user/.local/bin" shadow config.yaml`?

> Include the fully resolved path in shadow's log and processed-config.yaml. @stevenengler do you know how difficult it'd be to do the latter?

It wouldn't be difficult, but the reason I added the processed-config.yaml was to output the configuration after only the very basic config processing steps (merge the defaults, the yaml, and the cli options). If we want to add additional runtime-generated things to this file, then we'll need to make the configuration object mutable, make sure to update the config object after resolving paths, and then make sure we do that before we've written the config to a file. So we can do this, but it makes it less clear about what the processed-config.yaml file represents. I'd personally be in favour of just logging the paths to the binaries, but we could do it either way.

zyansheep:
> We really should be logging the absolute path to the binaries we are executing.

The way it resolves the path and inserts it into args at [0] makes sure the program being used is logged as the full path :+1: 
Screenshot_20221005_215059

sporksmith:
> I'm wondering why we want to change it now. Is it because we've decided that our previous decision was wrong?

Basically. I think we decided to kick the can on it and see if it was a worthwhile feature after more experience. After more experience with shadow I'm leaning towards this feature being worth it.

After @stevenengler 's comments and our offline chat about this, I think I've circled around to just using the `PATH` environment variable.

* It's what shells do
* We already do shell-like interprolation of `~`
* If we were to add a command-line option like `--binary-search-path`, typical usage is likely to be to set it to the `PATH` environment variable anyway, whether explicitly or because we make that the default.
* Users can still override it like any other environment variable when running shadow. e.g. `PATH="/path/to/experiment/bins" shadow shadow.yaml`

@robgjansen @stevenengler any objections?

src/main/core/manager.rs
```diff
@@ -732,7 +732,7 @@ impl<'a> Manager<'a> {
 
             // both ':' and ' ' are valid LD_PRELOAD separators
             for path in value.split(&[':', ' '][..]) {
-                let path = utility::tilde_expansion(path);
+                let path = utility::resolve_path(Path::new(path)).unwrap();
```
sporksmith, line 735: I *don't* think we want `$PATH` interpolation for the `LD_PRELOAD` environment variable. We should keep the `tilde_expansion` helper and still use it here.

src/main/core/sim_config.rs
```diff
@@ -267,7 +267,8 @@ fn build_host(
                 .options
                 .pcap_directory
                 .flatten_ref()
-                .map(|x| tilde_expansion(x)),
+                .map(|x| resolve_path(std::path::Path::new(x)).ok())
```
sporksmith, line 270: Still just use `tilde_expansion` here too

src/main/core/sim_config.rs
```diff
@@ -342,8 +343,9 @@ fn build_process(proc: &ProcessOptions) -> anyhow::Result<Vec<ProcessInfo>> {
             .with_context(|| format!("Failed to parse arguments: {x}"))?,
     };
 
-    // perform shell expansion
-    let plugin = tilde_expansion(proc.path.to_str().unwrap());
+    // resolve path
+    let plugin = resolve_path(&proc.path)
+        .with_context(|| format!("Failed to resolve plugin path \"{:?}\"", proc.path))?;
```
sporksmith, line 348: Please add an info-level log statement here with the resolved path. e.g.
`info!("Resolved process path {proc.path} to {plugin}")`

src/main/network/network_graph.rs
```diff
@@ -491,12 +492,12 @@ pub fn load_network_graph(graph_options: &GraphOptions) -> Result<String, NetGra
         GraphOptions::Gml(GraphSource::File(FileSource {
             compression: None,
             path: f,
-        })) => std::fs::read_to_string(tilde_expansion(f))
+        })) => std::fs::read_to_string(resolve_path(Path::new(f))?)
             .with_context(|| format!("Failed to read file: {f}"))?,
         GraphOptions::Gml(GraphSource::File(FileSource {
             compression: Some(Compression::Xz),
             path: f,
-        })) => read_xz(tilde_expansion(f))?,
```
sporksmith, line 499: Still use `tilde_expansion` here

src/test/bindc/bindc.yaml
```diff
@@ -7,5 +7,5 @@ hosts:
   testnode:
     network_node_id: 0
     processes:
-    - path: test-bind
+    - path: ./test-bind
```
sporksmith, line 10: Ah drat; I didn't realize we currently interpret `x` as `./x`. That makes this change a breaking change, since current yaml files will now have different behavior - either failing to run at all if the binary isn't in the `PATH`, or worse potentially running a different binary than they ran before.

Technically we could make this a non-breaking change by still searching the current directory first, but I don't think that's the behavior we ultimately want.

I think the best way to handle this for now is to put this feature behind an [`experimental`](https://shadow.github.io/docs/guide/shadow_config_spec.html#experimental) flag like `--enable-searching-path`  that's disabled by default. In the next major release we can change the default and/or remove the flag.

@stevenengler @robgjansen thoughts?

sporksmith (changes requested):

src/test/bindc/bindc.yaml
zyansheep: We could also keep everything the way it is and use an additional field like `exec: ` in the yaml which will (if `path: ` is not specified) search for its value in `$PATH` and set the path field in the processed-config.yaml.

sporksmith: Interesting idea, but I think ultimately the distinction between `path` and `exec` wouldn't be obvious based on the names.

The [`exec` man-page](https://www.man7.org/linux/man-pages/man3/exec.3.html) uses `file` vs `path`... but calls the old behavior we want to preserve `file`. So I don't think adding a `file` field and having that do the `$PATH` search would make sense.

I still lean towards enabling `$PATH`-search via a feature-flag

sporksmith: Actually another option is:

* Always search the current directory first for now (effectively prepend the current working directory to the path)
* Log a warning if the executable was found there via path-search, and wouldn't have been otherwise. (i.e. didn't have the `./` prefix and `PATH` didn't include the current directory) 
* In 3.0 stop searching the current directory first

That's maybe a slightly nicer migration strategy, though the logic for when to log a warning is a little tricky

zyansheep: Hmm, I kinda like that idea to search in CWD before resolving something using PATH as a temporary measure. I should be able to do that.

sporksmith: OK; I haven't heard from the others yet, but think it's probably safe to go forward with that plan.

I think the warning-logic isn't that bad after thinking about it some more. we warn if all of:

* There's no `./` prefix
* We found it in the CWD
* PATH-expansion *without* the CWD prefix gave a different result - either didn't find the binary at all or found a different one.

stevenengler: Either with an experimental flag or prepending "." to the path search string (and removing in 3.0) sounds fine to me. But another case to handle is non-absolute paths that contain a separator. For example `- path: foo/test-bind` should also find the `test-bind` binary in directory `foo`.

src/main/utility/mod.rs
```diff
@@ -189,27 +191,34 @@ impl Clone for ObjectCounter {
...
+            Ok(path)
+        } else {
+            anyhow::bail!("Couldn't find $HOME in env")
+        }
+    } else {
+        // Source exec path from $PATH if path is not a path
+        let path_str = path.to_str().ok_or(anyhow::anyhow!("invalid unicode"))?;
+        if !path_str.is_empty() && path_str.chars().find(|c| *c == '/' || *c == '.').is_none() {
+            which::which(path)
+                .with_context(|| format!("Failed to find path in $PATH for: {:?}", path))
```
stevenengler, line 216: Does this work correctly if the binary name has a '.' character in it? For example `foo.a`?

src/main/utility/mod.rs
```diff
@@ -268,37 +277,35 @@ mod tests {
...
+                resolve_exec("~/").unwrap(),
                 [home].iter().collect::<std::path::PathBuf>()
             );
 
             assert_eq!(
-                tilde_expansion("~someuser/test"),
-                ["/home", "someuser", "test"]
-                    .iter()
-                    .collect::<std::path::PathBuf>()
-            );
```
stevenengler, line 293: If we don't want to support this syntax, I think we should raise an error rather than return an incorrect path.

src/test/bindc/bindc.yaml
robgjansen: Both the experimental flag and the CWD prepend are good options. FWIW, I also lean toward the prepend CWD plan, and it would be great if the code is written such that it is really simple to remove that behavior when we reach 3.0 (or with sufficient comments so we understand/remember what needs to be changed later).

robgjansen:
It would be great if the code is written such that it is really simple to remove the behavior that we merge here but want to remove in 3.0, or at the very least please make sure to add sufficient comments so we understand/remember what needs to be changed later when we release 3.0. Thanks!

zyansheep:
Aight. I pretty much rewrote the entire commit, so I force-pushed. Should be a lot cleaner now

src/main/utility/mod.rs
zyansheep: Hmm... forgot this in the newest push, will add this now.

zyansheep: wait... can shell scripts be sourced from path?

stevenengler: > wait... can shell scripts be sourced from `$PATH`?

Yep as long as they're executable and have a shebang header, they're not very useful in Shadow though since we don't support fork/exec. But you could have non-shell files with a '.' in PATH as well (for example on my system I have a `/bin/python3.8`).

zyansheep: alr i'll fix that :+1:

src/main/core/sim_config.rs
```diff
@@ -342,21 +342,35 @@ fn build_process(proc: &ProcessOptions) -> anyhow::Result<Vec<ProcessInfo>> {
...
+        }
+        Err(err) => Err(err)?,
+    };
 
     // verify that the path is a file and is executable
     verify_plugin_path(&plugin)
         .with_context(|| format!("Failed to verify plugin path '{}'", plugin.display()))?;
 
+    // set argv[0] as the user-provided expanded string, not the canonicalized version
+    args.insert(0, expanded_path.into());
```
sporksmith, line 372: Am I reading correctly that we ultimately always set `argv[0]` the same way we were before - to the result of `tilde_expansion(proc.path)`?

I can see  how that might end up mostly working, since we ultimately call `execvpe`, which does its own PATH-based search. I wouldn't have thought that mattered since we're not setting the `PATH` in the child process's environment, but a) apparently it searches some standard locations anyway, possibly including the current directory, and b) the docs aren't clear whether it looks at the PATH in the `env` being passed to the function, or in the caller's environment.

Since this works in nix, where the binaries aren't in the "usual" locations, I guess it must be using the caller's environment?

I think we'll ultimately want to change how this works a bit to do all of the search and canonicalization ourselves, log the result, and pass the fully canonicalized path to `execvpe`. Otherwise there's a danger that our canonicalization/search is subtly different from what `execvpe` does, and we end up executing a different binary than we've logged. Are you interested in doing that, or would you rather this PR just does the bare minimum to fix the tests under nix?

sporksmith: I was a bit confused about what was going on here.

We were *already* passing the canonicalized path separately from argv[0].

The reason the current directory was being "searched" before, is that we were using `std::fs::canonicalize`, which interprets all non-absolute paths as relative to the current directory.

I think what you had here may have been correct, but I think it'd be a bit clearer and more robust to be more explicit about comparing the results  of the "old" and "new" ways. I pushed a commit to do that. Wdyt?

zyansheep: The code looks *beautiful* and all the tests seem to pass just fine :rocket:
Looking at the log file though, 

> 00:00:00.109174 [467600:shadow] n/a [WARN] [n/a] [sim_config.rs:374] [shadow_rs::core::sim_config] Expanded path "test-determinism" only via legacy fallback search, which implicitly searched current working dir. Consider prefixing with './'
> 00:00:00.109216 [467600:shadow] n/a [INFO] [n/a] [sim_config.rs:384] [shadow_rs::core::sim_config] Resolved binary path "test-determinism" to "/home/zyansheep/Repos/shadow/build/src/test/determinism/test-determinism"

This part seems like it works just fine

> 00:00:00.008566 [470057:shadow] n/a [INFO] [n/a] [sim_config.rs:384] [shadow_rs::core::sim_config] Resolved binary path "true" to "/nix/store/1xqr1166ji7xy80dp8d2nn1a6x8m6m5w-coreutils-9.1/bin/coreutils"

But shouldn't this part have a `/bin/coreutils/true` at the end?

Edit: nvm just figure it out, coreutils represents all the coreutils programs in one binary...

zyansheep: Looks good to me!

sporksmith (approved):

PR #2448, opened 2022-10-05
Rust thread-per-core scheduler

stevenengler:
This scheduler can be enabled using the experimental option `--scheduler thread-per-core`.


src/main/core/scheduler/pools/bounded.rs
```diff
@@ -175,6 +175,25 @@ impl ParallelismBoundedThreadPool {
 
     /// Create a new scope for the pool. The scope will ensure that any task run on the pool within
     /// this scope has completed before leaving the scope.
+    //
+    // SAFETY: This works because:
+    //
+    // 1. WorkerScope<'scope> is covariant over 'scope.
```
sporksmith, line 181: This is probably an embarrassing gap in my knowledge, but could you tell me or point me to definitions of covariant, invariant, and contravariant in this context?

src/main/core/scheduler/pools/unbounded.rs
```diff
@@ -0,0 +1,474 @@
...
+use std::sync::Arc;
+
+use atomic_refcell::AtomicRefCell;
+
+use crate::utility::synchronization::count_down_latch::{
+    build_count_down_latch, LatchCounter, LatchWaiter,
+};
+use crate::utility::synchronization::semaphore::LibcSemaphore;
+
+// If making substantial changes to this scheduler, you should uncomment each test at the end of
```
sporksmith, line 13: Optional - we're now running doc tests, and doc tests support ensuring that a snippet *won't* compile. See `compile_fail` in https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html

src/main/core/scheduler/pools/unbounded.rs
```diff
@@ -0,0 +1,474 @@
...
+            has_thread_panicked: AtomicBool::new(false),
+        });
+
+        let (task_end_counter, task_end_waiter) = build_count_down_latch();
+
+        let mut thread_handles = Vec::new();
+        let mut task_start_semaphores = Vec::new();
+
+        for i in 0..num_threads {
+            let task_start_semaphore = LibcSemaphore::new(0);
```
sporksmith, line 57: optional bikeshed, orthogonal to this PR - it's a little confusing that cloning below gives a reference to the same object. Maybe this type should be called something like `LibcSemaphoreArc`?

I also noticed the `SAFETY` comments in that type's definition - maybe it'd be worth using `Pin` to express in the type system that it can't be moved?

src/main/core/scheduler/pools/unbounded.rs
```diff
@@ -0,0 +1,474 @@
...
+            let shared_state_clone = Arc::clone(&shared_state);
+            let task_start_semaphore_clone = task_start_semaphore.clone();
+            let task_end_counter_clone = task_end_counter.clone();
```
sporksmith, line 60: Same here that it's a little unclear from the naming that this is cloning a reference rather than the object itself

src/main/core/scheduler/pools/unbounded.rs
```diff
@@ -0,0 +1,474 @@
...
+    impl<'a> std::ops::Drop for PoisonWhenDropped<'a> {
+        fn drop(&mut self) {
+            // if we panicked, then inform other threads that we panicked and allow them to exit
+            // gracefully
+            self.0.has_thread_panicked.store(true, Ordering::Relaxed);
+        }
+    }
+
+    let shared_state = shared_state.as_ref();
+    let poison_when_dropped = PoisonWhenDropped(shared_state);
```
sporksmith, line 240: I think this is ok, but would it be simpler / more idiomatic to use [catch_unwind](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html)?

src/main/core/scheduler/pools/unbounded.rs
```diff
@@ -0,0 +1,474 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use std::sync::atomic::{AtomicBool, AtomicU32};
+
+    use super::*;
+
+    #[test]
+    #[cfg_attr(miri, ignore)]
```
sporksmith, line 273: It's too bad we can't run miri on these. I suppose it's because of the FFI calls for the semaphore etc? Maybe worth a comment

src/main/core/scheduler/thread_per_core.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    /// See [`crate::core::scheduler::Scheduler::scope`].
+    pub fn scope<'scope>(
+        &'scope mut self,
+        f: impl for<'a, 'b> FnOnce(SchedulerScope<'a, 'b, 'scope>) + 'scope,
+    ) {
+        // we can't swap after the below `pool.scope()` due to lifetime restrictions, so we need to
+        // do it before instead
+        if self.hosts_need_swap {
+            #[cfg(debug_assertions)]
```
sporksmith, line 81: Optional nit - alternatively something like `debug_assert!(self.thread_hosts.iter().all(|queue| queue.len() == 0)`

sporksmith (approved):

src/main/core/scheduler/pools/unbounded.rs
stevenengler: I'll take a look at that, but I'm also considering trying https://github.com/dtolnay/trybuild, which allows us to ensure that not only it fails to compile, but also that it fails to compile for the correct reason. For example, I want these tests to fail due to lifetime issues rather than a syntax error or because we renamed something.

sporksmith: ah cool - agreed that's a weakness of the `compile_fail` doc tests

src/main/core/scheduler/thread_per_core.rs
stevenengler: Added.

src/main/core/scheduler/pools/unbounded.rs
stevenengler: I'm going to address this in a follow-up PR since I'll also want to change the thread-per-host scheduler at the same time.

src/main/core/scheduler/pools/unbounded.rs
stevenengler: I'm going to address this in a follow-up PR since I'll also want to change the thread-per-host scheduler at the same time.

src/main/core/scheduler/pools/unbounded.rs
stevenengler: I'm going to address this in a follow-up PR since I'll also want to change the thread-per-host scheduler at the same time.

src/main/core/scheduler/pools/unbounded.rs
stevenengler: I'm going to address this in a follow-up PR since I'll also want to change the thread-per-host scheduler at the same time.

src/main/core/scheduler/pools/unbounded.rs
stevenengler: I'm going to address this in a follow-up PR since I'll also want to change the thread-per-host scheduler at the same time.

PR #2450, opened 2022-10-05
Move HostId to Rust

sporksmith:
HostId should be defined in shadow-shim-helper-rs, since we need it in the shim as well as in shadow.
    
Also move the canonical definition into Rust.


sporksmith:
Unsetting reviewer for the moment - going to get rid of the string interning

src/main/core/manager.rs
```diff
@@ -533,7 +540,7 @@ impl<'a> Manager<'a> {
         let c_host = {
             let params = c::HostParameters {
                 // the manager sets this ID
-                id: unsafe { c::g_quark_from_string(hostname.as_ptr()) },
+                id: host_id,
```
stevenengler, line 543: I think this also allows us to remove `g_quark_from_string` from main/build.rs.

stevenengler (approved):

src/main/core/manager.rs
sporksmith: Done

PR #2461, opened 2022-10-12
Build an rlib for shadow-rs

stevenengler:
This adds doc tests in shadow-rs.

### DIFF
--- a/src/main/Cargo.toml
+++ b/src/main/Cargo.toml
@@ -6,7 +6,8 @@ publish = false
 
 [lib]
 path = "lib.rs"
-crate-type = ["staticlib"]
+# the rlib crate is so that we build and run doctests
+crate-type = ["rlib", "staticlib"]
 
 [dependencies]
 anyhow = { version = "1.0.65", features = ["backtrace"] }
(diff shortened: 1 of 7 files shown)


src/main/Cargo.toml
```diff
@@ -6,7 +6,7 @@ publish = false
 
 [lib]
 path = "lib.rs"
-crate-type = ["staticlib"]
+crate-type = ["rlib", "staticlib"]
```
sporksmith, line 9: Probably worth a comment here that the `rlib` is there to enable doc-tests

src/main/core/support/units.rs
```diff
@@ -2,13 +2,17 @@
 Types for parsing/deserializing unit values.
 
 ```
+# use shadow_rs::core::support::units::*;
+# use std::str::FromStr;
+# fn main() {
```
sporksmith, line 7: I don't think you need the `fn main()` ?

sporksmith (approved):

src/main/core/support/units.rs
stevenengler: Oops, fixed!

src/main/Cargo.toml
stevenengler: Added a comment.

PR #2463, opened 2022-10-12
Cleanup scheduler-related code

stevenengler:
This should resolve the review comments from #2448.

### DIFF
--- a/src/main/core/scheduler/mod.rs
+++ b/src/main/core/scheduler/mod.rs
@@ -5,7 +5,7 @@ pub use thread_per_core::ThreadPerCoreSched;
 pub use thread_per_host::ThreadPerHostSched;
 
 mod logical_processor;
-mod pools;
+pub mod pools;
 mod thread_per_core;
 mod thread_per_host;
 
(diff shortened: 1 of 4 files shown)


src/main/core/scheduler/mod.rs
```diff
@@ -5,7 +5,7 @@ pub use thread_per_core::ThreadPerCoreSched;
 pub use thread_per_host::ThreadPerHostSched;
 
 mod logical_processor;
-mod pools;
+pub mod pools;
```
sporksmith, line 8: Did this end up being needed for the doc tests to run?

sporksmith (approved):

src/main/core/scheduler/mod.rs
stevenengler: It isn't needed for them to run, but it's needed so that they will compile, otherwise `use shadow_rs::core::scheduler::pools::bounded::*;` won't be able to access the types we need in the test.

sporksmith: Ahh, right makes sense

PR #2465, opened 2022-10-12
bindgen and cbindgen: put generated files in build dir

sporksmith:
This eliminates spurious rebuilds between consecutive calls to `cargo test` (e.g. via `./setup test rust-unit-tests`). Since the generated bindings were in the source directory, it always detected that something in the source directory had changed and so needed to rerun.

There's still a rebuild on the first call to `cargo test` after the initial build. I suspect that this is to rebuild everything with `cfg(test)` and/or `cfg(doctest)` set, but I'm not sure.


src/lib/shadow-shim-helper-rs/build.rs
```diff
@@ -18,7 +18,7 @@ fn run_cbindgen(build_common: &ShadowBuildCommon) {
         })
         .generate()
         .expect("Unable to generate bindings")
-        .write_to_file("./shim_helper.h");
+        .write_to_file("../../../build/src/lib/shadow-shim-helper-rs/shim_helper.h");
```
stevenengler, line 21: What about putting the c bindings in OUT_DIR as well for consistency, and adding OUT_DIR to the include path for the C code? But either way is fine.

stevenengler (approved):

src/lib/shadow-shim-helper-rs/build.rs
sporksmith: Yeah, it'd be nice to do this, but I don't know of a way to add *this* crate's OUT_DIR to the header search path of dependent crates.

PR #2467, opened 2022-10-13
Use thread park/unpark instead of semaphores in work pools

stevenengler:
The interface of this 'thread_parking' module is a little weird, but it was written this way to support both work pools, which take different approaches to how the flag is shared between threads.

There's no noticeable change in performance when switching from semaphores to park/unpark. On Linux, park/unpark is implemented [with a futex](https://twitter.com/m_ou_se/status/1312803599689355266) so this isn't surprising.

Thread-per-host:

run_time_tph

Thread-per-core:

run_time_tpc

### DIFF
--- a/src/main/core/scheduler/pools/bounded.rs
+++ b/src/main/core/scheduler/pools/bounded.rs
@@ -9,7 +9,7 @@ use crate::core::scheduler::logical_processor::LogicalProcessors;
 use crate::utility::synchronization::count_down_latch::{
     build_count_down_latch, LatchCounter, LatchWaiter,
 };
-use crate::utility::synchronization::semaphore::LibcSemaphoreArc;
+use crate::utility::synchronization::thread_parking::{ThreadUnparker, ThreadUnparkerUnassigned};
 
 // If making substantial changes to this scheduler, you should verify the compilation error message
 // for each test at the end of this file to make sure that they correctly cause the expected
@@ -56,8 +56,8 @@ pub struct SharedState {
 
 /// Scheduling state for a thread.
 pub struct ThreadScheduling {
-    /// Semaphore used to wait for a new task.
-    task_start_semaphore: LibcSemaphoreArc,
+    /// Used to unpark the thread when it has a new task.
+    unparker: ThreadUnparker,
     /// The OS pid for this thread.
     tid: nix::unistd::Pid,
     /// The logical processor index that this thread is assigned to.
@@ -102,8 +102,9 @@ impl ParallelismBoundedThreadPool {
             .iter()
             .cycle()
             .zip(&tids)
-            .map(|(processor_idx, tid)| ThreadScheduling {
-                task_start_semaphore: LibcSemaphoreArc::new(0),
+            .zip(&thread_handles)
+            .map(|((processor_idx, tid), handle)| ThreadScheduling {
+                unparker: ThreadUnparkerUnassigned::new().assign(handle.thread().clone()),
                 tid: *tid,
                 logical_processor_idx: AtomicUsize::new(processor_idx),
             })
@@ -162,7 +163,7 @@ impl ParallelismBoundedThreadPool {
 
         // send the sentinel task to all threads
         for thread in &self.shared_state.threads {
-            thread.task_start_semaphore.post();
+            thread.unparker.unpark();
         }
 
         for handle in self.thread_handles.drain(..) {
@@ -358,11 +359,11 @@ fn work_loop(
     let poison_when_dropped = PoisonWhenDropped(shared_state);
 
     let thread_data = &shared_state.threads[thread_idx];
-    let start_semaphore = &thread_data.task_start_semaphore;
+    let thread_parker = thread_data.unparker.parker();
 
     loop {
         // wait for a new task
-        start_semaphore.wait();
+        thread_parker.park();
 
         // scope used to make sure we drop everything (including the task) before counting down
         {
@@ -431,7 +432,7 @@ fn start_next_thread(
         }
 
         // start the thread
-        next_thread.task_start_semaphore.post();
+        next_thread.unparker.unpark();
     }
 }
 
@@ -461,7 +462,7 @@ mod tests {
 
     use super::*;
 
-    // these tests don't use miri since they use `LibcSemaphoreArc` and `nix::unistd::gettid()`
+    // these tests don't use miri since they use `nix::unistd::gettid()`
 
     #[test]
     #[cfg_attr(miri, ignore)]
--- a/src/main/utility/synchronization/thread_parking.rs
+++ b/src/main/utility/synchronization/thread_parking.rs
@@ -1,10 +1,16 @@
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
+#[cfg(debug_assertions)]
+use std::sync::Mutex;
+
 /// Used to unpark a thread, but which hasn't been assigned a specific thread yet.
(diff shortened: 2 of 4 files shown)


src/main/utility/synchronization/thread_parking.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    /// Park the current thread until [`ThreadUnparker::unpark()`] is called. You must only call
+    /// `park()` from the thread which the corresponding `ThreadUnparker` is assigned, otherwise a
+    /// deadlock may occur.
+    pub fn park(&self) {
+        while self
+            .ready_flag
+            .compare_exchange(true, false, Ordering::SeqCst, Ordering::Relaxed)
+            .is_err()
+        {
```
sporksmith, line 77: I think this can deadlock?

```
thread A: l72 park
thread A: l73 compare_exchange
thread B:     l53 unpark
thread B:     l54 ready_flag.store
thread B:     l55 thread.unpark
thread A: l78 thread::park
```

I'm not sure there's a way to fix it without a semaphore.

If the goal is primarily to allow the thread-pools etc to run under miri, maybe an alternative is for your custom semaphore type to delegate to Rust's std semaphore instead of the pthread semaphore when `cfg(miri)`?

src/main/utility/synchronization/thread_parking.rs
```diff
@@ -0,0 +1,104 @@
...
+        Self { ready_flag }
+    }
+
+    /// Park the current thread until [`ThreadUnparker::unpark()`] is called. You must only call
+    /// `park()` from the thread which the corresponding `ThreadUnparker` is assigned, otherwise a
+    /// deadlock may occur.
+    pub fn park(&self) {
+        while self
+            .ready_flag
+            .compare_exchange(true, false, Ordering::SeqCst, Ordering::Relaxed)
```
sporksmith, line 75: I'm not sure `SeqCst` makes sense here; I *think* it's only meaningful if other relevant operations are `SeqCst`, and that it might not be needed here (depending how the potential deadlock is resolved)

sporksmith (changes requested):

src/main/utility/synchronization/thread_parking.rs
stevenengler: > I think this can deadlock?

I wouldn't be surprised if it does :) But I don't see the deadlock yet.

In your example, I think the `ready_flag` starts as false. The `compare_exchange()` in thread A fails to change `ready_flag` from true to false, then thread B changes `ready_flag` to true and calls `unpark()`. Thread A then calls `park()` which will return immediately because of the previous `unpark()`, loops back to the `compare_exchange()` which should now succeed since `ready_flag` is true, and we exit the loop.

> If the goal is primarily to allow the thread-pools etc to run under miri, maybe an alternative is for your custom semaphore type to delegate to Rust's std semaphore instead of the pthread semaphore when `cfg(miri)`?

There is no semaphore in the standard library unfortunately.

sporksmith: > Thread A then calls park() which will return immediately because of the previous unpark()

Ahh, that's the part I was missing. I was thinking of `park` and `unpark` like a `FUTEX_WAIT` and `FUTEX_WAKE`, but it looks like it actually behaves more like a semaphore, which is what we want here.

src/main/utility/synchronization/thread_parking.rs
sporksmith: I *think* this `SeqCst` can be `Relaxed`. Up to you, but maybe worth a comment about why you don't think it can be, or a `TODO` to further investigate if you want to just lean towards being more conservative for the moment.

sporksmith (approved):

src/main/utility/synchronization/thread_parking.rs
stevenengler: Yeah, I've been thinking of park and unpark like a binary semaphore. I'll add a comment about this.

The [documentation](https://doc.rust-lang.org/nightly/std/thread/fn.park.html) says:

> The `unpark` method on a `Thread` atomically makes the token available if it wasn’t already. Because the token is initially absent, `unpark` followed by `park` will result in the second call returning immediately.

src/main/utility/synchronization/thread_parking.rs
stevenengler: I agree that `SeqCst` isn't the right option here. What I want to prevent is any reordering with respect to the park() and unpark(), but I guess the memory orderings only synchronize what happens before the atomic operation, and there's nothing before the atomic operation that we need to synchronize. And here I think we have to hope/assume that the park() and unpark() operations have full memory barriers so that the prior atomic reads and writes will happen before the park() and unpark()?

sporksmith: I think `Release` might make sense here. It's also free on x86-64, and what they use in the corresponding part of the example: https://doc.rust-lang.org/std/thread/fn.park.html#examples

I agree though that it feels like it's depending a little on the implementation details of park and unpark. It wouldn't hurt to throw in a compiler fence to ensure the atomic operation doesn't get moved to after the `park`. https://doc.rust-lang.org/std/sync/atomic/fn.compiler_fence.html

src/main/core/scheduler/pools/bounded.rs
```diff
@@ -56,8 +56,8 @@ pub struct SharedState {
 
 /// Scheduling state for a thread.
 pub struct ThreadScheduling {
-    /// Semaphore used to wait for a new task.
-    task_start_semaphore: LibcSemaphoreArc,
+    /// Used to unpark the thread when it's waiting for a new task.
```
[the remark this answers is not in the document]
stevenengler: Updated.

src/main/utility/synchronization/thread_parking.rs
stevenengler: Looking at the rust source, it has comments that say:

> So, since this parker is just a single atomic variable, we only need to look at the ordering guarantees we need to provide to the 'outside world'.
>
> The only memory ordering guarantee that parking and unparking provide, is that things which happened before unpark() are visible on the thread returning from park() afterwards. Otherwise, it was effectively unparked before unpark() was called while still consuming the 'token'.
>
> In other words, unpark() needs to synchronize with the part of park() that consumes the token and returns.

https://github.com/rust-lang/rust/blob/master/library/std/src/sys_common/thread_parker/futex.rs#L21-L27

So I think this solves the park/unpark reordering problem. Otherwise I don't think a compiler fence would completely fix it since the memory accesses could still have been reordered.

> I think Release might make sense here. It's also free on x86-64, and what they use in the corresponding part of the example: https://doc.rust-lang.org/std/thread/fn.park.html#examples

I've changed it to be `Release`.

src/main/utility/synchronization/thread_parking.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    /// Assign this to a thread that will be unparked.
+    #[must_use]
+    pub fn assign(self, thread: std::thread::Thread) -> ThreadUnparker {
+        ThreadUnparker::new(self.ready_flag, thread)
+    }
+
+    /// Get a new [`ThreadParker`]. The `ThreadParker` must only be used from the thread which we
+    /// will later assign ourselves to using `assign()`. This is useful if you want to pass a
+    /// `ThreadParker` to a new thread before you have a handle to that thread.
```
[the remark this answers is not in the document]
stevenengler: Added this is a second commit behind `cfg(debug_assertions)` flags.

Issue #2466, opened 2022-10-13
Determinism test is sometimes failing


sporksmith:
Saw this failure in a PR's CI run, but fairly confident the PR wouldn't have caused it: https://github.com/shadow/shadow/actions/runs/3238878172/jobs/5307594292

I downloaded the data-dirs and found the diff output:

```
$ diff determinism1*-shadow.data/hosts/testnode1/testnode1.test-determinism.1000.strace 
6c6
< 00:00:01.000 [tid 1000] open(<pointer>, O_LARGEFILE | O_RDONLY, 0x7fffffffe866 <invalid>) = 3
---
> 00:00:01.000 [tid 1000] open(<pointer>, O_LARGEFILE | O_RDONLY, 0x7fffffffe856 <invalid>) = 3
```

I think we only recently added the parameter logging for `open`, so this may have been around for a while before.

stevenengler:
I noticed before that `open()` syscalls sometimes have these invalid modes, but I'm not sure why. My guess is that we're again using broken shim libc wrappers. In this case, the shorter version of `open`, which is `open(const char *pathname, int flags)`, should be setting a mode of 0 before making the syscall, but since we override libc's `open` call, that never happens and our shadow syscall handler is reading the mode from uninitialized memory.

sporksmith:
Thanks - I'll take a look

PR #2468, opened 2022-10-13
open, openat: use manual wrappers to make mode deterministic

sporksmith:
libc uses varargs to make the last option, `mode`, optional. When the caller doesn't explicitly provide a `mode`, we were getting unitialized data from the stack, which was sometimes causing the determinism tests to fail.

Fixes https://github.com/shadow/shadow/issues/2466

### DIFF
--- a/src/lib/libc_preload/libc_impls.c
+++ b/src/lib/libc_preload/libc_impls.c
@@ -13,6 +13,7 @@
 
 #include <assert.h>
 #include <errno.h>
+#include <fcntl.h>
 #include <netdb.h>
 #include <stdarg.h>
 #include <stdlib.h>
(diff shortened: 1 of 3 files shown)


src/lib/libc_preload/libc_impls.c
```diff
@@ -195,3 +196,41 @@ int __fxstat64(int ver, int a, struct stat64* b) {
...
+
+    // "If neither O_CREAT nor O_TMPFILE is specified in flags, then mode is
+    // ignored"
+    //
+    // We explicitly set to 0 here so that strace logging doesn't log an
+    // arbitrary value for `mode` when it wasn't explicitly provided by the
+    // caller.
+    if (!(flags & (O_CREAT | O_TMPFILE))) {
+        mode = 0;
+    }
```
stevenengler, line 214: It looks like this is what glibc does as well:

https://github.com/bminor/glibc/blob/b92a49359f33a461db080a33940d73f47c756126/sysdeps/unix/sysv/linux/open.c#L30-L41

```c
int mode = 0;

if (__OPEN_NEEDS_MODE (oflag))
{
  va_list arg;
  va_start (arg, oflag);
  mode = va_arg (arg, int);
  va_end (arg);
}
```

stevenengler (approved):

PR #2471, opened 2022-10-13
Move build info

sporksmith:
Move shadow build info (which changes on every build) to a separate header, which we only include from the top-level main.c.

This brings an incremental build down to 1.5s on my machine (from ~30s).

It looks like cmake's `configure_file` command already avoids changing the timestamp of the file if the content didn't change, so `shd-config.h` no longer gets an updated timestamp on every build.


stevenengler (approved):
> This brings an incremental build down to 1.5s on my machine (from ~30s).

🎉

PR #2474, opened 2022-10-15
Add signal handler to flush logger on SIGINT/SIGTERM

### LINKED ISSUE #2188: Add signal handler to flush shadow log on SIGINT
It would be nice if using ctrl-c to stop shadow would flush the log before exiting. Probably would want to handle SIGTERM and SIGHUP as well.

valdaarhun:
Closes #2188 .

Hi. I am quite new to rust and am trying to learn on the fly. Please do let me know if this PR can be improved.

### DIFF
--- a/src/Cargo.lock
+++ b/src/Cargo.lock
@@ -1090,6 +1090,7 @@ dependencies = [
  "shadow-shim-helper-rs",
  "shadow_shmem",
  "shadow_tsc",
+ "signal-hook",
  "syscall-logger",
  "system-deps",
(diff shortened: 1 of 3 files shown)


stevenengler:
Thanks, looks great! Can you add an info-level log message before the flush with something like "Received signal {signal}. Exiting..." or something like that?

valdaarhun:
> Can you add an info-level log message before the flush

Hi. Sure, will do.

stevenengler (approved):

PR #2476, opened 2022-10-18
Support the 'bounded' thread pool tests under miri

stevenengler:
`gettid()` and `sched_setaffinity()` are not allowed under miri.


sporksmith (approved):
Nice!

PR #2477, opened 2022-10-18
Split the host's event queue into local event and packet event queues

stevenengler:


stevenengler:
Closing since we're considering making a separate queue for "immediate" events (a fifo queue for events we want to run immediately before resuming the plugin), and since the event queue would then only contain packet and timer events, there wouldn't be a reason for this PR. This PR was originally intended so that we could alternate between packet events and "0 ns"/"1 ns" events that process packets. Also this PR is very out-of-date, and we would need to rewrite it.

PR #2479, opened 2022-10-19
Fix order of operations in codel control law

### LINKED ISSUE #2478: Incorrect order of operations in CoDel control law implementation 🤦
In Shadow's implementation of the CoDel control law, the order of operations is incorrect.

The [CoDel RFC](https://www.rfc-editor.org/rfc/rfc8289#section-5.6) defines the control law as the following function:

```C
time_t codel_queue_t::control_law(time_t t, uint32_t count)
   {
       return t + INTERVAL / sqrt(count);
   }
```

The following expression is equivalent but makes the correct order of operations explicit:

```C
       return t + (INTERVAL / sqrt(count));
```

Shadow's implementation orders things the wrong way: it does the addition operation first, and then the division:

https://github.com/shadow/shadow/blob/8581e745afc1916bef0e17aa806b6381291f62d3/src/main/routing/router_queue_codel.c#L199-L206

The result is that the schedule for dropping packets will be incorrect and unexpected.

robgjansen:
Fixes #2478

### DIFF
--- a/src/main/routing/router_queue_codel.c
+++ b/src/main/routing/router_queue_codel.c
@@ -197,12 +197,9 @@ static Packet* _routerqueuecodel_dequeueHelper(QueueManagerCoDel* queueManager, 
 }
 
 static CSimulationTime _routerqueuecodel_controlLaw(guint count, CSimulationTime ts) {
-    CSimulationTime newTS = ts + CODEL_PARAM_INTERVAL_SIMTIME;
-
-    double result = ((double)newTS) / sqrt((double)count);
-    double rounded = round(result);
-
-    return (CSimulationTime)rounded;
+    // control_law: t + (INTERVAL / sqrt(count));
+    double result = ((double)CODEL_PARAM_INTERVAL_SIMTIME) / sqrt((double)count);
+    return ts + ((CSimulationTime)result);
 }
 
 static Packet* _routerqueuecodel_dequeue(QueueManagerCoDel* queueManager) {


src/main/routing/router_queue_codel.c
```diff
@@ -197,12 +197,10 @@ static Packet* _routerqueuecodel_dequeueHelper(QueueManagerCoDel* queueManager,
...
 static CSimulationTime _routerqueuecodel_controlLaw(guint count, CSimulationTime ts) {
-    CSimulationTime newTS = ts + CODEL_PARAM_INTERVAL_SIMTIME;
-
-    double result = ((double)newTS) / sqrt((double)count);
+    // control_law: t + (INTERVAL / sqrt(count));
+    double result = ((double)CODEL_PARAM_INTERVAL_SIMTIME) / sqrt((double)count);
     double rounded = round(result);
-
-    return (CSimulationTime)rounded;
+    return ts + ((CSimulationTime)rounded);
```
stevenengler, line 203: Is the goal to be exactly the same as the reference? If so, should it just cast the float to an integer instead of rounding?

stevenengler (approved):
Nice catch! I see you're running a benchmark, so it will be nice to know what the impact is on the tor network performance.

robgjansen:
Benchmark results show that performance is non-deterministic but not changed too much:
https://github.com/shadow/benchmark-results/tree/master/tor/2022-10-19-T15-26-44

robgjansen:
Manually ran a larger 20% Tor network using `tornettools`, here are the results: [tornet-codel-bugfix.plot.pages.pdf](https://github.com/shadow/shadow/files/9873512/tornet-codel-bugfix.plot.pages.pdf)

Based on the results, the CoDel bug didn't appear to affect Tor network performance enough to lead us to believe that previous Tor simulations are invalid.

PR #2487, opened 2022-10-21
Rename Host and Process

sporksmith:
Thread is similarly renamed in https://github.com/shadow/shadow/pull/2486


stevenengler (approved):

sporksmith:
The Host part is obsoleted by #2500, and I'm not as sure anymore about the Process naming. I'll leave that alone until actually migrating Process to Rust to avoid churn.

PR #2495, opened 2022-10-27
setup: Support --compile-commands in incremental builds

sporksmith:
Use `bear`'s `--append` flag to update an existing `compile_commands.json`, so that we can run it on incremental builds as well without erasing entries for files that weren't rebuilt.

### DIFF
--- a/docs/coding.md
+++ b/docs/coding.md
@@ -12,16 +12,10 @@ firefox build/guide/index.html
 
 Many tools benefit from a [compiler command
 database](https://clangd.llvm.org/design/compile-commands), conventionally in a
-file called `compile_commands.json`. To generate such a database, use the
-`setup` script's `--compile-commands` option when building. This must be done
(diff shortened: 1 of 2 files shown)


sporksmith:
Maybe we should get rid of the `setup` flag altogether, and just do it automatically if we can find the `bear` executable, and maybe log a warning if we can't?

robgjansen (changes requested):
> Maybe we should get rid of the setup flag altogether, and just do it automatically if we can find the bear executable, and maybe log a warning if we can't?

+1 I prefer this simpler approach without the setup flag.

It would be nice to log if bear isn't installed in any case, but I wonder if we ought to make it less severe than warn since most people using Shadow probably don't need or want bear?

sporksmith:
Done. I didn't add a new log line, since we already log the `make` command-line and can see whether or not it used `bear`

robgjansen (approved):

PR #2497, opened 2022-10-28
Build shadow-tests in the same target directory

stevenengler:
In Shadow we run `cargo build` twice, once on the workspace (in `src/`) and once on the "shadow-tests" package (in `src/test/`). Each cargo invocation is given a different target directory. This means that we end up building "shadow-tests" twice, once in each target directory. This hasn't really been an issue for us yet since for the workspace build we only build `--lib --tests` and not `--bins`, so none of the binary crates from "shadow-tests" were being built in the workspace build.

We need the two cargo builds since "shadow-tests" should always be built with the dev profile, and cargo [doesn't yet](https://doc.rust-lang.org/cargo/reference/unstable.html#per-package-target) let you build a workspace using different profiles.

I think it's better to build both "shadow-tests" and the remainder of the workspace into the same target directory, and include only "shadow-tests" when building "shadow-tests", and excluding only "shadow-tests" when building the workspace. This also allows us to create binary crates in shadow, which previously wouldn't have been built since we used to only specify `--lib --tests`.

This PR also makes it so that both `cargo build`s don't run at the same time anymore, and they are run in sequence instead.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -18,7 +18,7 @@ foreach(DEFINITION IN LISTS COMPILE_DEFINITIONS)
   set(RUST_CXXFLAGS "${RUST_CXXFLAGS} -D${DEFINITION}")
 endforeach()
 
-set(RUST_TARGETS "--lib")
+set(RUST_TARGETS "--lib --bins")
 if(SHADOW_TEST STREQUAL ON)
   set(RUST_TARGETS "${RUST_TARGETS} --tests")
 endif()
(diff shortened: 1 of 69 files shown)


src/CMakeLists.txt
```diff
@@ -87,7 +104,17 @@ endforeach()
 # better than trying to fight the "rust way" here. It also includes doc tests,
 # which normally don't have persistent executables at all.
 add_test(NAME rust-unit-tests
-         COMMAND bash -c "${CARGO_ENV_VARS} cargo test ${RUST_TARGET_FLAG} ${RUST_BUILD_FLAG} --manifest-path \"${CMAKE_CURRENT_SOURCE_DIR}\"/Cargo.toml --target-dir \"${TARGETDIR}\" --features \"${RUST_FEATURES}\"")
+         COMMAND bash -c " \
+            ${CARGO_ENV_VARS} \
+            cargo test \
+            --workspace --exclude shadow-tests \
```
sporksmith, line 110: Optional - Do we need to exclude shadow-tests here? If there aren't any unit tests there, it shouldn't do anything, but if we write unit tests e.g. for test utilities we'd probably want to run them here?

sporksmith (approved):

src/CMakeLists.txt
stevenengler: Removed this line.

PR #2500, opened 2022-10-29
Host rust

sporksmith:
* The C implementation of `Host` is renamed to `HostCInternal` and its
  methods `host_*` renamed to `hostc_*`. This type and these methods are
  now meant only to be used from the Rust `Host`.
* The Rust `Host` is now the canonical `Host` type. It is currently not much
  more than a wrapper around a `HostCInternal`, with the `host_*` methods
  implemented by calling their original `hostc_*` implementations, but we
  should now be able to start moving fields from `HostCInternal` to `Host`,
  and start replacing the thin proxy methods with Rust native implementations.
* The `Host` now has a `Root` object that we can use for cheap interior
  mutability and reference counting of objects associated with that `Host`.
* `Host` is no longer ref-counted. There is no longer a `host_ref` nor
  `hostc_ref`. This is because `Host` is `!Sync`, so could not be stored
  inside a ref-counted wrapper without re-adding a `Host` lock wrapper
  to make it `Sync`. e.g. it would need to be something like an
  `Arc<AtomicRefCell<Host>>`. While the extra locking overhead in the scheduler
  probably wouldn't be that noticeable, it'd make it difficult to make access to the
  `&Host` or its internal `&Root` cheaply globally accessible.
* A reference to the current `Host` can now be retrived from the `Worker` via
  `Worker::with_active_host`. This lets us get to the `Host` from places
  that previously fetched it from a stored pointer.
* `Host` methods now all take `&Host`; there are no `&mut Host` methods.
  With so much inside the `Host`, and since we want to be able to access it via
  a global (at least for now), anything holding a `&mut Host` would prevent
  anything else being able to retrieve any reference to the `Host`.
* We do the same with the `Worker`, for analagous reasons.
* The scheduler now moves the `Host` value instead of passing a `&Host`.
  This allows the `Host` to be stored in the `Worker`'s thread-local,
  and borrowed as-needed via `Worker::with_active_host`.

In a separate commit, the Host's `EventQueue` is moved from `HostCInternal` to `Host`, along with some corresponding method implementations. This doesn't strictly need to be part of this PR, but is a proof of concept of what the next stages of this migration will look like.


sporksmith:
I think this should be performance neutral, but kicked off a benchmark to be sure https://github.com/shadow/benchmark/actions/runs/3362534049

sporksmith:
> I think this should be performance neutral, but kicked off a benchmark to be sure https://github.com/shadow/benchmark/actions/runs/3362534049

It looks like there may be a slight performance regression: https://github.com/shadow/benchmark-results/blob/master/tor/2022-10-31-T15-28-48/plots/run_time.png

Looking into it; probably one of the first things I'll try is boxing the Host as @stevenengler suggested, so that moving it around is a bit cheaper

src/main/host/host.rs
```diff
@@ -197,11 +213,60 @@ impl Host {
...
+        let cpu = unsafe { cshadow::hostc_getCPU(self.chost()) };
+        loop {
+            let mut event_queue = self.event_queue.lock().unwrap();
+            match event_queue.next_event_time() {
+                Some(t) if t < until => {}
+                _ => break,
+            };
+
+            let mut event = event_queue.pop().unwrap();
+            drop(event_queue);
```
stevenengler, line 225: nit: I think a scope is more readable than using `drop()`:

```rust
let mut event = {
    let mut event_queue = self.event_queue.lock().unwrap();
    match event_queue.next_event_time() {
        Some(t) if t < until => {}
        _ => break,
    };

    event_queue.pop().unwrap()
};
```

stevenengler (approved):

sporksmith:
Locally on the tor-minimal test it looks like boxing the Host went from making this PR a slight perf loss to a slight perf gain, but it's all close to being in the noise.

New benchmark running with the boxed host: https://github.com/shadow/benchmark/actions/runs/3371218645

src/main/core/scheduler/thread_per_core.rs
```diff
@@ -195,14 +184,14 @@ pub struct HostIter<'a> {
...
 }
 
 impl<'a> HostIter<'a> {
     /// See [`crate::core::scheduler::HostIter::next`].
-    pub fn next(&mut self) -> Option<&mut Host> {
-        self.return_current_host();
+    pub fn next(&mut self, prev: Option<Host>) -> Option<Host> {
+        if let Some(prev) = prev {
+            self.thread_hosts_to.push(prev).unwrap();
+        }
```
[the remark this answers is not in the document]
sporksmith: Good idea; done

src/lib/shadow-shim-helper-rs/src/rootedcell/refcell.rs
```diff
@@ -1,5 +1,8 @@
 use super::{Root, Tag};
-use std::cell::{Cell, UnsafeCell};
+use std::{
+    cell::{Cell, UnsafeCell},
+    fmt::Debug,
+};
```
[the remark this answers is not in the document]
sporksmith: Done

sporksmith:
After boxing the Host it looks about the same as the last nightly: https://github.com/shadow/benchmark-results/blob/master/tor/2022-11-01-T16-50-29/plots/run_time.png

src/main/host/host.rs
```diff
@@ -80,107 +136,516 @@ impl Host {
...
         crate::core::logger::log_wrapper::c_to_rust_log_level(level).map(|l| l.to_level_filter())
     }
 
-    pub fn random(&mut self) -> &mut impl rand::Rng {
-        let ptr = unsafe { cshadow::host_getRandom(self.chost()) };
+    pub fn random(&self) -> &mut impl rand::Rng {
+        let ptr = unsafe { cshadow::hostc_getRandom(self.chost()) };
         let random = unsafe { ptr.as_mut() }.unwrap();
         &mut random.0
     }
```
stevenengler, line 168: Sorry for the last-minute comment. I don't think this function is safe, since it's returning a mutable reference without any kind of interior mutability. For example, the following code would cause two mutable references to the same data:

```rust
let rng_1 = host.random();
let rng_2 = host.random();
```

sporksmith: Oops, good catch. I'll probably move the Rng into the Rust Host and wrap it in a RootedRefCell

PR #2516, opened 2022-11-02
Move `worker_sendPacket` to the rust worker

stevenengler:


src/main/core/worker.rs
```diff
@@ -270,6 +274,106 @@ impl Worker {
         .unwrap();
     }
 
+    pub fn send_packet(src_host: &Host, packet: *mut cshadow::Packet) {
```
sporksmith, line 277: Nit - I'd consider this `unsafe`. Even though the Rust code doesn't itself dereference the `packet`, it's still the caller's responsibility that `packet` is valid and not being accessed  simultaneously from other threads to avoid UB in the C code

src/main/core/worker.rs
```diff
@@ -270,6 +274,106 @@ impl Worker {
...
+        let dst_ip = std::net::IpAddr::V4(dst_ip);
+
+        // check if network reliability forces us to 'drop' the packet
+        let reliability: f64 = Worker::with(|w| w.shared.reliability(src_ip, dst_ip).unwrap())
+            .unwrap()
+            .try_into()
+            .unwrap();
+        let chance: f64 = src_host.with_random_mut(|r| r.gen());
+
+        // don't drop control packets with length 0, otherwise congestion control has problems
```
sporksmith, line 316: I know this is just ported over from the C code, but I wonder if we should do something about this at some point. For someone using Shadow to model a lossy network, having 0-length-payload packets be 100% reliable doesn't seem right.

Maybe open an issue to revisit and leave a link here? Or update the comment to be more clear about whether this is/isn't actually a bug in shadow?

src/main/host/host.rs
```diff
@@ -171,6 +172,15 @@ impl Host {
         crate::core::logger::log_wrapper::c_to_rust_log_level(level).map(|l| l.to_level_filter())
     }
 
+    pub unsafe fn upstream_router(&self) -> *mut Router {
```
sporksmith, line 175: `unsafe` not needed? The `unsafe` operation of dereferencing the pointer is left to the caller.

src/main/host/host.rs
```diff
@@ -171,6 +172,15 @@ impl Host {
...
+        unsafe { cshadow::hostc_getUpstreamRouter(self.chost()) }
+    }
+
+    pub unsafe fn interface(&self, handle: Ipv4Addr) -> *mut cshadow::NetworkInterface {
```
sporksmith, line 179: `unsafe` not needed? The `unsafe` operation of dereferencing the pointer is left to the caller.

src/main/core/worker.rs
stevenengler: Added.

src/main/host/host.rs
stevenengler: Removed.

src/main/core/worker.rs
stevenengler: I converted this into an issue, and left a comment in the code.

https://github.com/shadow/shadow/issues/2517

PR #2518, opened 2022-11-02
Move remainder of worker code from C to Rust

stevenengler:
Also fixes a small bug in the reliability calculation where a packet could be sent successfully on an edge with 0% reliability.


sporksmith (approved):
:tada:

PR #2519, opened 2022-11-02
Add unit tests for the thread-per-core scheduler

stevenengler:
This PR adds unit tests for the thread-per-core scheduler. We can't add them for the thread-per-host scheduler since the unit tests require the scheduler to be generic over the host type, and the thread-per-host scheduler uses a thread-local that cannot be generic.

This also moves the worker functions out of the scheduler and into a helper function in the manager. I think this is good since it makes the manager not dependent on any external code, and it doesn't hide away the worker's global state inside the scheduler, which was something that made the old C scheduler hard to deal with.


sporksmith (approved):
Nice!

PR #2524, opened 2022-11-07
Add a (currently disabled) regression test for #1341

### LINKED ISSUE #1341: Packets are dropped when net buffer is full even if receiver is trying to read

stevenengler:
We should enable this test when we fix #1341. Adding this test now since it was already written and it will probably be a while before we fix the issue.

### DIFF
--- a/src/test/regression/1341/test_regression_1341.rs
+++ b/src/test/regression/1341/test_regression_1341.rs
@@ -1,0 +1,123 @@
+use std::io::Write;
+use std::net::ToSocketAddrs;
+
+use nix::errno::Errno;
+use nix::sys::socket::{sockopt::RcvBuf, AddressFamily, MsgFlags, SockFlag, SockType};
+
+const NUM_MSGS: usize = 100;
+const MSG_SIZE: usize = 100;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let arg = std::env::args().nth(1).ok_or("Missing arg[1]")?;
+    let addr_or_port = std::env::args().nth(2).ok_or("Missing arg[2]")?;
+
+    match arg.as_str() {
+        "sender" => run_sender(addr_or_port.to_socket_addrs()?.next().unwrap())?,
+        "receiver" => run_receiver(addr_or_port.parse()?)?,
+        _ => return Err("Invalid argument".into()),
+    };
+
+    Ok(())
+}
+
+fn run_sender(addr: std::net::SocketAddr) -> Result<(), Box<dyn std::error::Error>> {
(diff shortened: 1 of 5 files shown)


src/test/regression/1341/test_regression_1341.rs
```diff
@@ -0,0 +1,123 @@
...
+    nix::sys::socket::bind(fd, &bind_addr)?;
+
+    // wait for the first packet
+    let mut poll_fds = [nix::poll::PollFd::new(fd, nix::poll::PollFlags::POLLIN)];
+    assert_eq!(1, nix::poll::poll(&mut poll_fds, -1)?);
+
+    let mut num_bytes_received = 0;
+    let mut num_bytes_received_since_eagain = 0;
+
+    // The socket's receive buffer should be full and we should be able to call recv() repeatedly to
```
sporksmith, line 77: Maybe worth having a second test here where the receiver takes too long to process packets, s.t. packets *do* get dropped? Maybe a nanosleep after processing each packet...? I'd expect such a test to pass now, but we'd want to be careful not to make it start failing when fixing #1341.

sporksmith (approved):

stevenengler:
I'm closing this since I've tried a few times to understand what's going on in #1341 and this PR, and I'm not really following it anymore. I'm also hesitant to add a test that's completely commented out and never runs. If someone decides to tackle #1341 (which I think requires significant changes to shadow's networking code), then feel free to reopen this PR or copy any of the code to a new PR.

PR #2528, opened 2022-11-09
Add uses of 'let-else' statements

stevenengler:
I don't really think the code needed to change, I just keep forgetting that this exists and wanted to see some places where it would be useful. Might as well merge them I think since it is slightly more readable.

### DIFF
--- a/src/lib/syscall-logger/src/lib.rs
+++ b/src/lib/syscall-logger/src/lib.rs
@@ -88,10 +88,9 @@ pub fn log_syscall(args: TokenStream, input: TokenStream) -> TokenStream {
             ctx: &mut crate::host::context::ThreadContext,
             args: &crate::host::syscall_types::SysCallArgs,
         ) -> crate::host::syscall_types::SyscallResult {{
-            let strace_fmt_options = match ctx.process.strace_logging_options() {{
-                Some(x) => x,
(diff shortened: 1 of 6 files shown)


src/lib/syscall-logger/src/lib.rs
```diff
@@ -88,10 +88,9 @@ pub fn log_syscall(args: TokenStream, input: TokenStream) -> TokenStream {
             ctx: &mut crate::host::context::ThreadContext,
             args: &crate::host::syscall_types::SysCallArgs,
         ) -> crate::host::syscall_types::SyscallResult {{
-            let strace_fmt_options = match ctx.process.strace_logging_options() {{
-                Some(x) => x,
+            let Some(strace_fmt_options) = ctx.process.strace_logging_options() else {{
```
sporksmith, line 91: Extra pair of brackets here?

sporksmith (approved):

src/lib/syscall-logger/src/lib.rs
stevenengler: Ah, GitHub doesn't show that it's actually part of a `format!()` string, so the first bracket is needed to escape the second one. (We should be using the syn crate here, but I didn't originally add it to avoid depending on it, but we use a lot of crates that depend on it now so I don't think that would be an issue.)

PR #2529, opened 2022-11-09
Fix potential mutable aliasing of DNS object

stevenengler:


robgjansen (approved):
Looking forward to not having to interact with C code from Rust anymore :D

PR #2536, opened 2022-11-10
Save counters to `sim-stats.json` file

stevenengler:
Saves the total syscall counters and alloc/dealloc counters to a json file.

Example:

```json
{
  "objects": {
    "alloc_counts": {
      "TaskRef": 18349,
      "Event": 8928,
      "Packet": 7616,
      "Payload": 7316,
      "SysCallCondition": 517,
      "Timer": 492,
      "LegacyDescriptor": 161,
      "TCP": 82,
      "UDP": 74,
      "StatusListener": 26,
      "RegularFile": 4,
      "NetworkInterface": 2,
      "Epoll": 1,
      "FutexTable": 1,
      "HostCInternal": 1,
      "ManagedThread": 1,
      "Process": 1,
      "Router": 1,
      "SysCallHandler": 1
    },
    "dealloc_counts": {
      "TaskRef": 18349,
      "Event": 8928,
      "Packet": 7616,
      "Payload": 7316,
      "SysCallCondition": 517,
      "Timer": 492,
      "LegacyDescriptor": 161,
      "TCP": 82,
      "UDP": 74,
      "StatusListener": 26,
      "RegularFile": 4,
      "NetworkInterface": 2,
      "Epoll": 1,
      "FutexTable": 1,
      "HostCInternal": 1,
      "ManagedThread": 1,
      "Process": 1,
      "Router": 1,
      "SysCallHandler": 1
    }
  },
  "syscalls": {
    "poll": 14466,
    "sendto": 12649,
    "write": 2658,
    "read": 2133,
    "close": 615,
    "nanosleep": 491,
    "recvfrom": 353,
    "socket": 348,
    "getsockname": 229,
    "bind": 183,
    "connect": 160,
    "socketpair": 91,
    "accept4": 84,
    "listen": 84,
    "brk": 8,
    "rt_sigaction": 5,
    "mmap": 3,
    "munmap": 3,
    "sigaltstack": 3,
    "exit_group": 1,
    "fstat": 1,
    "getrandom": 1,
    "mprotect": 1,
    "openat": 1,
    "shadow_init_memory_manager": 1
  }
}
```

### DIFF
--- a/src/main/core/manager.rs
+++ b/src/main/core/manager.rs
@@ -16,7 +16,7 @@ use crate::core::controller::{Controller, ShadowStatusBarState, SimController};
 use crate::core::scheduler::runahead::Runahead;
 use crate::core::scheduler::{HostIter, Scheduler, ThreadPerCoreSched, ThreadPerHostSched};
 use crate::core::sim_config::{Bandwidth, HostInfo};
-use crate::core::sim_stats::SimStats;
+use crate::core::sim_stats;
 use crate::core::support::configuration::{self, ConfigOptions, Flatten, LogLevel};
 use crate::core::worker;
 use crate::cshadow as c;
@@ -497,13 +497,13 @@ impl<'a> Manager<'a> {
         // since the scheduler was dropped, all workers should have completed and the global object
         // and syscall counters should have been updated
 
-        let mut sim_stats = SimStats::new();
+        let mut stats = sim_stats::SimStats::new();
 
         // log syscall counters
         if self.config.experimental.use_syscall_counters.unwrap() {
             worker::with_global_syscall_counter(|counter| {
                 log::info!("Global syscall counts: {}", counter);
-                sim_stats.syscalls = counter.clone();
+                stats.syscalls = counter.clone();
             });
         }
 
@@ -512,8 +512,8 @@ impl<'a> Manager<'a> {
             worker::with_global_object_counters(|alloc_counter, dealloc_counter| {
                 log::info!("Global allocated object counts: {}", alloc_counter);
                 log::info!("Global deallocated object counts: {}", dealloc_counter);
-                sim_stats.objects.alloc_counts = alloc_counter.clone();
-                sim_stats.objects.dealloc_counts = dealloc_counter.clone();
+                stats.objects.alloc_counts = alloc_counter.clone();
+                stats.objects.dealloc_counts = dealloc_counter.clone();
 
                 if alloc_counter == dealloc_counter {
                     log::info!("We allocated and deallocated the same number of objects :)");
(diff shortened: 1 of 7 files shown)


sporksmith (approved):

src/main/core/sim_stats.rs
```diff
@@ -0,0 +1,27 @@
...
+    pub fn new() -> Self {
+        Self {
+            objects: ObjectStats {
+                alloc_counts: Counter::new(),
+                dealloc_counts: Counter::new(),
+            },
+            syscalls: Counter::new(),
+        }
+    }
+}
```
robgjansen, line 27: Should this end in a newline?

src/main/core/manager.rs
```diff
@@ -516,6 +524,17 @@ impl<'a> Manager<'a> {
...
+        let stats_file = std::fs::File::create(&stats_filename)
+            .with_context(|| format!("Failed to create file '{}'", stats_filename.display()))?;
+
+        serde_json::to_writer_pretty(stats_file, &sim_stats).with_context(|| {
+            format!(
+                "Failed to write stats json to file '{}'",
+                stats_filename.display()
+            )
+        })?;
+
```
robgjansen, line 537: I would prefer to push this code inside of `sim_stats.rs` to try to reduce complexity as much as possible in the already-complex manager. Is there a good argument against doing that?

src/main/core/manager.rs
```diff
@@ -494,10 +497,13 @@ impl<'a> Manager<'a> {
         // since the scheduler was dropped, all workers should have completed and the global object
         // and syscall counters should have been updated
 
+        let mut sim_stats = SimStats::new();
```
robgjansen, line 500: It would be nicer to create a `SimStats` object at the start of the experiment that internally stores the object and syscall counters. Then when those counters need to be updated, we call through to the counters stored inside of the `SimStats` instance.

That way we can remove the current counters that exist there, and we can avoid cloning at the end. So, `worker::with_global_syscall_counter` and `worker::with_global_object_counters` would become `worker::with_global_sim_stats`.

src/main/core/manager.rs
```diff
@@ -506,6 +512,8 @@ impl<'a> Manager<'a> {
             worker::with_global_object_counters(|alloc_counter, dealloc_counter| {
                 log::info!("Global allocated object counts: {}", alloc_counter);
                 log::info!("Global deallocated object counts: {}", dealloc_counter);
+                sim_stats.objects.alloc_counts = alloc_counter.clone();
+                sim_stats.objects.dealloc_counts = dealloc_counter.clone();
```
robgjansen, line 516: Super super optional nit: I personally prefer to keep internal members private and use `set()` methods, but I think that's just a style thing. It does require more boilerplate for the set methods.

robgjansen (changes requested):
I think we should move all of the global and worker counters and mutexes into the SimStats module to help simplify the manager and centralize the stats-related code. This will move some of the counter-related complexity out of the manager, and help keep the conceptually related code in a simpler module.

src/main/core/sim_stats.rs
stevenengler: There should be a newline there, github would put a red icon there if there was no newline at the end of the file.

src/main/core/manager.rs
stevenengler: I moved this to `sim_stats`.

src/main/core/manager.rs
stevenengler: This has been changed. Now we have the two public structs `LocalSimStats`, `SharedSimStats`, and a private struct `SimStatsForOutput`.

src/main/core/manager.rs
stevenengler: I left this as public struct fields for now, which I think help keeps the boilerplate code down for these "collection" objects, but we can always change it later.

robgjansen (approved):
Looks great, thanks for the improvements!

PR #2537, opened 2022-11-10
Rewrite Cpu in Rust

sporksmith:


src/main/host/cpu.rs
```diff
@@ -0,0 +1,201 @@
+use std::time::Duration;
+
+use shadow_shim_helper_rs::{emulated_time::EmulatedTime, simulation_time::SimulationTime};
+
+pub struct Cpu {
```
robgjansen, line 5: It would be nice to have a comment about the intended function of the module.

src/main/host/cpu.rs
```diff
@@ -0,0 +1,201 @@
...
+    #[test]
+    fn round_up() {
+        let precision = SimulationTime::from_millis(100);
+        let mut cpu = Cpu::new(
+            1_000_000,
+            1_000_000,
+            Some(SimulationTime::NANOSECOND),
+            Some(precision),
+        );
+        cpu.add_delay(Duration::from_millis(151));
```
robgjansen, line 198: I think 150 should round up too, right?

src/main/core/manager.rs
```diff
@@ -537,7 +537,7 @@ impl<'a> Manager<'a> {
                 // the manager sets this ID
                 id: host_id,
                 // the manager sets this CPU frequency
-                cpu_frequency: self.raw_frequency_khz,
+                cpu_frequency_khz: self.raw_frequency_khz,
```
robgjansen, line 540: Should we standardize on Hz instead of Khz?

I regret using Khz (and KiBps in other parts of the code) and it has caused enough confusion over the years that I'd be happy to use base units instead.

robgjansen (approved):

src/main/host/cpu.rs
```diff
@@ -0,0 +1,201 @@
...
+        // round the adjusted delay to the nearest precision if needed
+        if let Some(precision) = self.precision {
+            let remainder = adjusted_delay % precision;
+
+            // first round down (this is also the first step to rounding up)
+            adjusted_delay -= remainder;
+
+            // now check if we should round up
+            let half_precision = precision / 2;
+            if remainder > half_precision {
```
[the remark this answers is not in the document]
sporksmith: :shrug: whether to round up or down at midpoint is completely arbitrary AFAICT, which is also why I left it out of the unit tests :). I'm fine with rounding up and specifying that through docs and tests though.

src/main/core/manager.rs
sporksmith: Yeah, I was debating whether or not to change it. I'll take a look but might push it to a follow-up.

src/main/host/cpu.rs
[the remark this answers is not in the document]
robgjansen: So we should round up half the time and down half the time to avoid introducing a rounding bias? 😝  I'm fine with whatever you decide :)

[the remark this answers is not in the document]
sporksmith: Haha - I guess it makes sense to document and test which way it goes so that we don't change it without noticing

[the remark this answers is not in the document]
sporksmith: Done

src/main/core/manager.rs
sporksmith: Done

PR #2538, opened 2022-11-10
Move defaultAddress from HostCInternal to Host

sporksmith:


src/main/host/host.rs
```diff
@@ -569,6 +576,10 @@ impl Host {
 
 impl Drop for Host {
     fn drop(&mut self) {
+        let default_addr = self.default_address.borrow(&self.root).ptr();
+        if !default_addr.is_null() {
+            unsafe { cshadow::address_unref(default_addr) };
+        }
         unsafe { cshadow::hostc_unref(self.chost()) };
     }
```
stevenengler, line 584: nit: maybe also set `default_address` to `NULL`, just in case for example `hostc_unref` were to call something that calls `host_getDefaultAddress` which then dereferences the dangling pointer.

src/main/host/host.rs
```diff
@@ -92,6 +92,9 @@ pub struct Host {
     random: RootedRefCell<Xoshiro256PlusPlus>,
     params: HostParameters,
 
+    // TODO: Use a Rust address type.
+    default_address: RootedRefCell<SyncSendPointer<cshadow::Address>>,
```
stevenengler, line 96: I think we could use just a `RefCell` here since it only needs to be `Send`? But either is fine I think.

stevenengler (approved):

src/main/host/host.rs
stevenengler: nvm, i forgot the C host doesn't have access to the rust host

src/main/host/host.rs
sporksmith: Oops, right. Done.

PR #2540, opened 2022-11-11
Host shared memory: Move from HostCInternal to Host

sporksmith:


src/lib/shmem/src/allocator.rs
```diff
@@ -92,6 +80,26 @@ where
     extern "C" fn _validate_stable_layout(_: T) {}
 }
 
+impl<'origin, T> Drop for ShMemBlock<'origin, T>
+where
+    T: Sync + VirtualAddressSpaceIndependent,
+{
+    fn drop(&mut self) {
+        unsafe { std::ptr::drop_in_place(self.internal.p as *mut T) };
```
stevenengler, line 86: Also add a SAFETY comment to `ShMemBlock::new()` stating that you must not create two `ShMemBlock` that point to the same shared memory?

src/lib/shmem/src/allocator.rs
```diff
@@ -227,8 +212,9 @@ impl Serializer {
     /// SAFETY:
     /// * `block` must have been created from a `ShMemBlock<T>`
     /// * The backing memory must be live for the lifetime of the returned
-    ///   ShMemBlock. e.g. `ShMemBlock::deallocate` must not have been called
-    ///   on any block referencing that memory.
+    ///   ShMemBlock. i.e. the original allocated ShMemBlock must outlive the
+    ///   returned one. We can't guarantee this with normal lifetime analysis, since
+    ///   the original block may be in another process.
     pub unsafe fn deserialize<'a, T>(&'a self, block: &ShMemBlockSerialized) -> ShMemBlock<'a, T>
```
stevenengler, line 219: How should the calling code enforce this, and in what situations can the backing memory be destroyed (for example, if the process crashes? the block is dropped in another process?)?

src/main/host/host.rs
```diff
@@ -573,12 +594,28 @@ impl Host {
...
+        // We're extending this lifetime to extend the lifetime of `lock`, below, without
+        // having to `transmute` the type itself.
+        //
+        // SAFETY:
+        // * We ensure that `self.shim_shmem_lock` doesn't outlive `self.shim_shmem`.
+        //   We never drop `self.shim_shmem` before `self` itself is dropped, and in `Self`'s
+        //   `Drop` implementation we validate that the lock isn't held.
+        // * `ShMemBlock` guarantees that its data doesn't move even if the block does.
+        //    So moving `shim_shmem` (e.g. by moving `self`) doesn't invalidate the lock.
+        let shim_shmem: &'static ShMemBlock<HostShmem> = unsafe { &*shim_shmem };
```
stevenengler, line 609: Here you're making a `'static` reference to the `self.shim_shmem`, so you're essentially making a `'static` reference to the `Host` (`Host` is now a self-referential structure, where the self-reference lifetime isn't tracked). If the host were to move, this reference would be invalid.

This method should still be `unsafe`, with the condition that the `Host` is not moved, swapped (`std::mem::replace`), mutated (no functions called that take `&mut Host` otherwise we'd have aliasing borrows, but I don't think we currently have any of these), etc until the next call to `unlock_shmem()`.

stevenengler (approved):

src/lib/shmem/src/allocator.rs
sporksmith: * Shadow allocates all the blocks, and won't deallocate them before any managed code that might have deserialized the block has stopped executing.
* In managed code, before deserializing any shmem blocks, we use a prctl to tell the OS to kill the current process if the parent dies. So if Shadow exits abnormally while the managed code is still running, the OS should kill the managed code before it can access the freed shared memory. (There might be a race condition here, but I think in the worst case the managed could would segfault trying to access the freed memory)

src/main/core/support/configuration.rs
```diff
@@ -130,6 +130,37 @@ impl ConfigOptions {
...
+
+    pub fn model_unblocked_syscall_latency(&self) -> bool {
+        self.general.model_unblocked_syscall_latency.unwrap()
+    }
+
+    pub fn max_unapplied_cpu_latency(&self) -> SimulationTime {
+        let nanos = match self.experimental.max_unapplied_cpu_latency {
+            Some(x) => x.convert(units::TimePrefix::Nano).unwrap().value(),
+            // shadow uses a value of 0 as "not set" instead of SIMTIME_INVALID
+            None => 0,
```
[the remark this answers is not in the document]
sporksmith: The C wrapper was removed in a later commit :).

Ultimately I want to implement a default of `SimulationTime::ZERO` here, and return a `SimulationTime` instead of an `Option<SimulationTime>`. I think we originally made the field in the config struct an `Option<SimulationTime>` so that we could tell whether the user provided it or not, but I can't think of why we need that. Should I just change the config struct to be a `SimulationTime` with a default value of `ZERO` instead of an `Option` with a default value of `None`?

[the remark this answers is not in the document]
sporksmith: Correction, the default is already `Some(...)`, which is indeed what we want. I'm still not sure whether I ought to remove the Option wrapper though. I think maybe it had something to do with the processed config we write back out? This way if the user didn't provide an override, the processed config won't have an override either, and a newer version of shadow run with that processed config, with an updated default, will pick up the new value...?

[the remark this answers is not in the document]
sporksmith: Oops, that doesn't make sense either, since we'd only get here if the user explicitly set this flag to None, not if they didn't explicitly set it at all.

So I guess go ahead and change the config field to not be an `Option`?

src/main/host/host.rs
sporksmith: I reviewed the rustnomicon ownership rules; I still think this is ok.

With respect to the [reference aliasing rules](https://doc.rust-lang.org/nomicon/references.html), I don't think a reference to the `ShMemBlock`'s data is considered to alias a reference to the `Host` itself, since they're different allocations. (i.e. because of the pointer indirection in `ShMemBlock`; similar to if it were just a `Box`)

We're certainly circumventing the [lifetime borrow checker](https://doc.rust-lang.org/nomicon/lifetimes.html), but my understanding is that doing so isn't automatically unsound; it's just the guard rails to prevent you from breaking the aliasing rules in safe Rust.

stevenengler: > Internally, ShMemBlock has a pointer to the actual data; moving the ShMemBlock (e.g. by moving the Host that contains the ShMemBlock) doesn't move the data. I added this as part of ShMemBlock's API contract and added a unit test to verify it.

Okay I agree that moving the `Host` should be fine.

> I agree this seems like a little bit of a gray area, but I don't think holding a reference to something that borrowed from the Host is equivalent to holding a reference to the Host in Rust's aliasing rules.

This doesn't seem right to me. Otherwise for `struct Foo { x: u32 }`, you could have both a `&mut foo` and a `&foo.bar` existing at the same time.

> I mocked up an example in the Rust playground and verified it passes miri. Wdyt?

In this example if you add the method:

```rust
fn cause_ub(&mut self) {
    let _ = std::mem::replace(self.shmem.as_mut(), RefCell::new(10));
}
```

to the `Host`, and then call `host.cause_ub()` after `host.lock()`, miri will raise an error. It's very likely we'd have this specific code in shadow, but it shows a way to invoke UB just by adding safe code.

> With respect to the reference aliasing rules, I don't think a reference to the `ShMemBlock`'s data is considered to alias a reference to the `Host` itself, since they're different allocations. (i.e. because of the pointer indirection in `ShMemBlock`; similar to if it were just a `Box`)

I don't think this would be allowed with a `Box` either? For example with `struct Foo { x: Box<u32> }`, you can't have both a `&foo.x.as_ref()` (a `&u32`) and a `&mut foo` (a `&mut Foo`) existing at the same time.

src/main/core/support/configuration.rs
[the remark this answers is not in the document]
stevenengler: Yeah, the user may provide it in either the cli options or the config options, but probably not both, so there needs to be a possibility for it to be `None`.

> We set it to Some(x) in the defaults though, and I don't think there's any reason to support the user explicitly setting it to an empty string, null, etc. So I just unwrapped the option here. LMK if that seems like the right way to go.

Yeah that's how I've been thinking about it. The documentation states what values/types are allowed for each field, and if they provide a `null` where a `null` isn't allowed, we may as well just unwrap it.

src/lib/shmem/src/allocator.rs
stevenengler: I think it would be clearer if it was split, since then you have some ideas about what things you should uphold based on the type. For example if you have a `ReferencedShMemBlock`, you know you should drop it before any `OwnedShMemBlock`. But definitely doesn't need to be changed for this PR.

src/main/host/host.rs
stevenengler: Maybe the difference between this code and the example rust playground is that `Box` gives you both `Deref` and `DerefMut`, whereas `ShMemBlock` gives you only `Deref`. I think you could still call `std::mem::replace(&mut self.shim_shmem, _)` to drop the current `ShMemBlock` while `shim_shmem_lock` holds a reference to it, which would let you cause UB with only safe code. This is maybe a bit nitpicky since we'd almost certainly never write `std::mem::replace(&mut self.shim_shmem, ..)` in practice, but I think it would still be good to mark a function like `lock()` as unsafe as long as there is the possibility to invoke UB with only safe code?

sporksmith: > This doesn't seem right to me. Otherwise for struct Foo { x: u32 }, you could have both a &mut foo and a &foo.x existing at the same time.
> ...
> I don't think this would be allowed with a Box either? For example with struct Foo { x: Box<u32> }, you can't have both a &foo.x.as_ref() (a &u32) and a &mut foo (a &mut Foo) existing at the same time.

Yeah I think you're right; the `Box` indirection is needed for it to be safe to move the `Host`, but isn't relevant for the reference aliasing safety. 

Added another caveat that this depends on code in this module (which can directly access `self.shim_shmem`) not mutating `self.shim_shmem` or borrowing `self.shim_shmem.protected` while the lock is held.

> I think you could still call std::mem::replace(&mut self.shim_shmem, _) to drop the current ShMemBlock while shim_shmem_lock holds a reference to it, which would let you cause UB with only safe code. This is maybe a bit nitpicky since we'd almost certainly never write std::mem::replace(&mut self.shim_shmem, ..) in practice, but I think it would still be good to mark a function like lock() as unsafe as long as there is the possibility to invoke UB with only safe code?

In general I think `SAFETY` arguments necessarily depend on invariants being upheld in `unsafe` code. Ideally those invariants can be verified very locally, but I've seen other cases where it relies on the rest of the safe code in the module (which has access to the non-pub internals) to *not* do something. e.g. it'd be weird for code outside this module to have to put calls to this method in an `unsafe` block, since the caller isn't responsible for upholding the safety invariants.

For this case I'll explicitly add the additional invariant that other code in this module must not mutate `self.shim_shmem` while the lock is held (or expose non-unsafe interfaces to code outside this module that would allow it to do so)

I know this is all a bit fragile; added a TODO to revisit once the lock doesn't need to be accessed from C code. At that point we can probably just take the lock on-demand, though will need to check the performance.

sporksmith: Wdyt of wrapping `Host::shim_shmem` and `Host::shim_shmem_lock` in `UnsafeCell`? Not because we need the extra layer of interior mutability, but because it would for us to internally use `unsafe` when accessing those fields; and we could move the SAFETY requirements to the field definitions?

sporksmith: Done; I think that's an improvement.

I think I'll let this simmer for tonight. LMK if you have more concerns or would like more time to look at it again.

src/lib/shmem/src/allocator.rs
sporksmith: SG; I'll do a follow-up

src/main/host/host.rs
stevenengler: Looks good to me.

PR #2542, opened 2022-11-11
Implement Host::get_random_free_port

sporksmith:
This replaces the C implementation `hostc_getRandomFreePort`

### DIFF
--- a/src/main/host/host.rs
+++ b/src/main/host/host.rs
@@ -9,10 +9,10 @@ use crate::host::network_interface::NetworkInterface;
 use crate::network::router::Router;
 use crate::utility::{self, SyncSendPointer};
 use atomic_refcell::AtomicRefCell;
-use log::{info, trace};
+use log::{info, trace, warn};
 use logger::LogLevel;
 use once_cell::unsync::OnceCell;
-use rand::SeedableRng;
+use rand::{Rng, SeedableRng};
 use rand_xoshiro::Xoshiro256PlusPlus;
 use shadow_shim_helper_rs::emulated_time::EmulatedTime;
 use shadow_shim_helper_rs::rootedcell::refcell::RootedRefCell;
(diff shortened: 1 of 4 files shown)


src/main/host/host.rs
```diff
@@ -581,6 +585,101 @@ impl Host {
...
+        let mut port = start;
+        loop {
+            port = if port == u16::MAX {
+                MIN_RANDOM_PORT
+            } else {
+                port + 1
+            };
+            if port == start {
+                break;
+            }
```
stevenengler, line 668: What about:

```rust
let ports = MIN_RANDOM_PORT..=u16::MAX;
let start_offset = self
    .random
    .borrow_mut(&self.root)
    .gen_range(0..ports.len());
for port in ports.cycle().skip(start_offset).take(ports.len()) {
    ...
}
```

stevenengler (approved):

src/main/host/host.rs
sporksmith: Wouldn't the `skip` be `O(n)` in `start_offset`? Or is there a clever optimization I don't know about?

(I also thought cycle required storing all previous values of the iterator, but afaict it just clones the original iterator, so doesn't have to keep the individual values around)

sporksmith: Ah, it uses  [nth](https://stackoverflow.com/questions/66287676/how-does-skip-works-in-rust), which for a range iterator would presumably be constant time

sporksmith: Done. Thanks, much nicer!

stevenengler: That's what I was thinking as well, but looking at it again, I don't know if the `Cycle` in the middle messes it up. `Cycle::nth()` seems to call `Iterator::advance_by(n)` on the range iterator. A normal range iterator `Range` has a specialized [`advance_by()`](https://doc.rust-lang.org/src/core/iter/range.rs.html#750) implementation, but a `RangeInclusive` seems to default to the default `Iterator` [`advance_by()`](https://doc.rust-lang.org/src/core/iter/traits/iterator.rs.html#328) implementation.

So I don't know anymore :) Maybe it's better to stick with something we know is optimal like the original implementation.

sporksmith: I tried to approximate on godbolt - I don't claim to fully understand the disassembly, but I don't see anything that looks like e.g. dynamically allocating somewhere to store past 

I made a contrived example with ridiculously large values. It hangs when run, even in release builds, so it definitely seems like non-constant time after all. I'll revert that last change.

```
use std::convert::TryInto;

pub fn main() -> std::process::ExitCode {
    let ports = 0..=u64::MAX;
    let random_offset = u64::MAX/2;
    let start_offset = unsafe { std::ptr::read_volatile(&random_offset) };

    let result = ports.clone().cycle().skip(start_offset.try_into().unwrap()).take(u64::MAX.try_into().unwrap()).next().unwrap();
 
    (result as u8).into()
}
```

sporksmith: Btw I just tried changing my example to use `0..u64::MAX` instead of `0..=u64::MAX`, so I think you're right that it's a problem specific to `RangeInclusive`.

I wonder if there's a fundamental reason it doesn't have an specialized `advance_by`, or whether it's just an oversight.

PR #2548, opened 2022-11-16
Move more fields from HostCInternal to Host

sporksmith:


src/main/host/host.rs
```diff
@@ -412,8 +412,9 @@ impl Host {
         crate::core::logger::log_wrapper::c_to_rust_log_level(level).map(|l| l.to_level_filter())
     }
 
-    pub fn upstream_router(&self) -> *mut Router {
-        unsafe { cshadow::hostc_getUpstreamRouter(self.chost()) }
+    pub fn with_upstream_router_mut<Res>(&self, f: impl FnOnce(&mut Router) -> Res) -> Res {
+        let router = unsafe { &mut *cshadow::hostc_getUpstreamRouter(self.chost()) };
+        f(router)
     }
```
stevenengler, line 470: This doesn't seem safe? It creates a `&mut` from ` &` without any runtime checks?

src/main/host/host.rs
```diff
@@ -413,8 +421,8 @@ impl Host {
...
 
     pub fn with_upstream_router_mut<Res>(&self, f: impl FnOnce(&mut Router) -> Res) -> Res {
-        let router = unsafe { &mut *cshadow::hostc_getUpstreamRouter(self.chost()) };
-        f(router)
+        let mut router = self.router.borrow_mut();
+        f(router.deref_mut())
     }
```
stevenengler, line 449: I'm personally a fan of giving the calling code a `&RefCell<Router>` and letting the calling code decide if it want to `borrow()` or `borrow_mut()`, and so it has more control over when these borrows get dropped so that other code can borrow. And also to reduce the number of closures, for example it might get to be a lot if we have code like:

```rust
host.with_upstream_router_mut(|router| {
    host.with_interface_mut(|interface| {
        host.with_random(|rand| {
             todo!();
        })
    })
})
```

But either way is fine with me.

src/main/host/host.rs
stevenengler: I see that a `RefCell` was added in a later commit.

src/main/host/host.rs
```diff
@@ -425,6 +429,21 @@ impl Host {
         f(router.deref_mut())
     }
 
+    /// Calls `f` with the Host's tracker, if it has one.
+    pub fn with_tracker_mut<Res>(
```
stevenengler, line 452: `#[must_use]`?

src/main/host/host.rs
```diff
@@ -814,6 +824,10 @@ impl Drop for Host {
             unsafe { cshadow::tracker_free(tracker.ptr()) };
         };
 
+        let futex_table = self.futex_table.borrow().ptr();
```
stevenengler, line 854: I think should be `borrow_mut()` since it mutates the ref count and frees the table.

stevenengler (approved):

src/main/host/host.rs
sporksmith: Yeah, I've been on the fence about doing it that way or not. The down-side is that making the `RefCell` part of the interface makes it harder to change later, e.g. if we decide we want to put something in a RootedRefCell or AtomicRefCell instead.

Another option is to wrap these up in new types, making the specific type of *Cell opaque. It's a fair bit of boilerplate, though. It could probably be reduced with macros, but I'm not sure it's worth that complexity.

Maybe just going ahead and exposing the RefCell type is the way to go; at most call sites the type won't be named explicitly anyway, so refactoring if we change it shouldn't be that big of a deal.

sporksmith: I looked into this. For the router it's not too bad, though at the current 1 usage site I ended up making a new scope to limit the lifetime of the borrow, avoiding a panic, anyway, actually making the code slightly longer:

```
+++ b/src/main/core/worker.rs
@@ -354,9 +354,10 @@ impl Worker {
         let packet_task = TaskRef::new(move |host| {
             let packet = packet.take().expect("Packet task ran twice");
 
-            let became_nonempty = host.with_upstream_router_mut(|router| unsafe {
-                crate::network::router::router_enqueue(router, packet.into_inner())
-            });
+            let became_nonempty = {
+                let mut router = host.upstream_router().borrow_mut();
+                unsafe { crate::network::router::router_enqueue(&mut *router, packet.into_inner()) }
+            };
 
             if became_nonempty {
                 host.packets_are_available_to_receive();
```

For other fields it leaks more implementation details such as that it's currently stored as a pointer, and exposes the option. e.g. exposing a `&RefCell<Option<SyncSendPointer<cshadow::Tracker>>>` pushes dealing with the pointer to the caller, and would actually let them mutate the enclosing `Option` to add/remove/substitute the `Tracker`.

I'm open to revisiting, but I think I'll leave these taking closures for now. My thought is to add non-mut variants if/when they're useful (e.g. `with_upstream_router`), but right now there wouldn't be any users.

src/main/host/host.rs
sporksmith: done

src/main/host/host.rs
sporksmith: done

PR #2549, opened 2022-11-16
Use tgen to test shadow network performance

### LINKED ISSUE #810: Add tests of Shadow's TCP simulation accuracy
Some of these tests might initially not pass, and hence should be disabled in automatic CI. Such tests will become part of the completion criteria of https://github.com/shadow/shadow/milestone/18.

Ideally these tests will be written in Rust; in which case this issue is blocked on #807.

An example test proposed by @robgjansen is whether two simulated machines connected by a gigabit link actually achieve close to a gigabit of bandwidth.

robgjansen:
Closes #810


src/test/tgen/fixed_duration/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+
+foreach(client 1stream 10streams 100streams 1000streams)
+    add_custom_target(tgen-duration-${client}-setup ALL
+                    COMMAND ${CMAKE_COMMAND}
+                      -E copy
+                      ${CMAKE_CURRENT_SOURCE_DIR}/client.${client}.graphml
+                      ${CMAKE_CURRENT_BINARY_DIR}/client.${client}.graphml)
+
+  foreach(network 1mbit_300ms 10mbit_200ms 100mbit_100ms 1gbit_10ms)
+    add_custom_target(tgen-duration-${network}-${client}-setup ALL
```
stevenengler, line 16: nit: maybe unindent the first `add_custom_target` by two spaces? it took me a little bit to understand what was going on with the loops here.

src/test/tgen/fixed_size/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+
+foreach(client 1stream_1b_1000x 1stream_1kib_100x 1stream_1mib_10x 10streams_1mib_10x 100streams_1kib_100x 1000streams_1b_1000x)
+    add_custom_target(tgen-size-${client}-setup ALL
+                    COMMAND ${CMAKE_COMMAND}
+                      -E copy
+                      ${CMAKE_CURRENT_SOURCE_DIR}/client.${client}.graphml
+                      ${CMAKE_CURRENT_BINARY_DIR}/client.${client}.graphml)
+
+  foreach(network 1mbit_300ms 10mbit_200ms 100mbit_100ms 1gbit_10ms)
+    add_custom_target(tgen-size-${network}-${client}-setup ALL
```
stevenengler, line 16: nit: indentation here as well?

docs/extra_tests.md
```diff
@@ -18,9 +18,9 @@ not complete before the timeout.
...
-# To exclude the Tor tests (for example if you built Shadow in debug mode)
-./setup test --extra -- --label-exclude tor
+# To exclude the TGen and Tor tests (for example if you built Shadow in debug mode)
+./setup test --extra -- --label-exclude tgen|tor
 ```
 
-If you change the version of tor located at `~/.local/bin/tor`, make sure to
-re-run `./setup build --test`.
+If you change the version of tgen located at `~/.local/bin/tgen` or tor located
+at `~/.local/bin/tor`, make sure to re-run `./setup build --test`.
```
stevenengler, line 26: The reason for this comment about tor is that we run `tor --list-torrc-options` to check what options are available in `src/test/tor/minimal/CMakeLists.txt`, and if you change the tor version, you should re-run this check. I don't think we need to re-run anything if the tgen version changes?

src/test/tgen/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# Make sure tests in subdirs append to previously collected tests.
+set(ALL_SHADOW_TESTS "${ALL_SHADOW_TESTS}")
+
+add_subdirectory(fixed_duration)
+add_subdirectory(fixed_size)
```
stevenengler, line 5: Maybe add a README to this tgen directory with a short explanation of what it's doing?

src/test/tgen/fixed_duration/client.10streams.graphml
```diff
@@ -0,0 +1,49 @@
...
+<graph edgedefault="directed"><node id="start">
+  <data key="d0">1 second</data>
+  <data key="d1">server:80</data>
+  <data key="d2">5 seconds</data>
+  <data key="d3">10 gib</data>
+  <data key="d4">10 gib</data>
+</node>
+<node id="end">
+  <data key="d5">1</data>
+</node>
```
stevenengler, line 17: My understanding is that there are X simultaneous streams that both send and receive 1 GiB of data, and when each stream completes (both download and upload are complete), it restarts?

stevenengler, line 17: I only see d0 to d4 defined above?

src/test/tgen/fixed_size/gen_conf.py
```diff
@@ -0,0 +1,44 @@
...
+        # Bidirectional transfer.
+        G.add_node(stream_name)
+        # Streams are launched in parallel.
+        G.add_edge("start", stream_name)
+        # Waits until all parallel streams are finished before checking end conditions.
+        G.add_edge(stream_name, "pause")
+
+    # If end conditions are not met, start another set of streams.
+    G.add_edge("pause", "end")
+    G.add_edge("end", "start")
```
stevenengler, line 39: What does this mean? So there are X total streams, and only Y run in parallel? And they are grouped, so it waits for Y parallel streams to complete, then starts another Y streams, until X total streams are complete?

src/test/tgen/fixed_duration/verify.sh
```diff
@@ -0,0 +1,55 @@
...
+# read speed of client is the effective write speed of server
+# tail -1 gets the fastest speed reached
+server_bps_read="$(grep -r --include="server.tgen.*.stdout" "driver-heartbeat" \
+    | cut -d' ' -f8 \
+    | cut -d'[' -f2 \
+    | cut -d',' -f1 \
+    | cut -d'=' -f2 \
+    | sort -n \
+    | tail -1)"
+client_bps_read="$(grep -r --include="client.tgen.*.stdout" "driver-heartbeat" \
```
stevenengler, line 21: Maybe rename to `Bps`? We use `bps` elsewhere in shadow to refer to "bits-per-second".

src/test/tgen/fixed_duration/verify.sh
```diff
@@ -0,0 +1,55 @@
...
+    | tail -1)"
+
+printf "Server bytes per second=${server_bps_read}\n"
+printf "Client bytes per second=${client_bps_read}\n"
+
+# Test passes if tgen application achieves >= 85% of configured speed.
+# We use a 85% threshold to allow for TCP overhead.
+expected=0
+if [[ "${network}" == "1mbit"* ]]; then
+    expected=106250
```
stevenengler, line 36: nit: maybe leave a comment on this first value with how it's computed? (1Mbit/s = 125000 bytes/s, 125000 bytes/s * 0.85 = 106250 bytes/s)

stevenengler (approved):

src/test/tgen/fixed_duration/client.10streams.graphml
stevenengler: Nvm, d5 is hidden behind the very long line above.

src/test/tgen/fixed_size/gen_conf.py
robgjansen: Yes, that's correct. The `pause` node without a specified pause time attribute acts as a "barrier" that causes tgen to wait until all incoming edges visits that node before it continues its walk through the graph.

In this case I am setting X to the same as Y so there will only be 1 round of "parallel" streams: the pause means tgen will not move forward until those Y streams complete, and then it moves to the end node which checks that X streams completed and ends because X==Y. But we could set up multiple rounds if we wanted by setting X greater than Y.

I'll add some comments to the script. :)

PR #2552, opened 2022-11-18
SelfContainedMutex: Strengthen compiler fence

sporksmith:
Using just `Acquire` here still lets the compiler move writes before the fence to after the fence. We need both `Acquire` and `Release` semantics.

Pointed out here:
https://masto.ai/@rfc1149/109359308547583544

And handy visual aid here:
https://masto.ai/@rfc1149/109359347916681723


sporksmith:
Closing in favor of a pending more-complete fix. See https://github.com/shadow/shadow/issues/2557

PR #2554, opened 2022-11-18
Increase timeout for tgen tests

robgjansen:


robgjansen:
Changing the timeout because a tgen test has already failed due to exceeding it.
https://github.com/shadow/shadow/actions/runs/3499489754/jobs/5861108159

stevenengler (approved):

Issue #2555, opened 2022-11-18
Use `MaybeUninit` in memory manager and `Pod`-related code


stevenengler:
The memory manager allows you to copy data to/from the managed process, and uses the `Pod` trait to restrict what types can be copied through the memory manager. While any initialized byte representation is valid for a "pod" type, that doesn't mean we can cast a pod type to a `&[u8]` since the type may have padding bytes that will invoke UB when read. For example `MemoryCopier::copy_to_ptr()` takes a `&[T: Pod]` as an argument and converts it to a `&[u8]`, which if read may invoke UB if `T` has padding bytes. There are other cases such as socket addresses where we have a pointer and length to copy to the plugin, but shouldn't cast them to a `&[u8]` since there may be padding bytes in the socket address. So it would be nice if the memory manager accepted a `&[MaybeUninit<u8>]` or more generally a `&[MaybeUninit<T: Pod>]` to copy to the managed process.

sporksmith:
Most of this is cleaned up now, but we still end up having to create slices of possibly-uninitialized-bytes inside the MemoryManager, to pass to nix's wrappers around process_vm_readv and process_vm_writev.

We should probably either:
1. Fix nix's wrappers, which currently take `&[u8]`, to take either:
  * `&[MaybeUninit<u8>]`
  * `*const u8` + length
  * `<T>`, so that callers with a sound `&T` with uninitialized padding bytes aren't forced to construct an unsound `&[u8]` out of them
2. Write our own wrappers that do one of the above.

It probably makes sense to write our own version to solve the problem for ourselves, and then try to upstream it.

PR #2559, opened 2022-11-21
Fix unsound methods in Pod and cleanup MemoryManager

sporksmith:
Progress on https://github.com/shadow/shadow/issues/2555, though maybe we should keep it open until we can fix the unsafe transformations from `&[MaybeUninit<u8>]` to `&[u8]`, which are currently needed to work with nix's wrappers around `process_vm_readv` and `process_vm_writev`.

### DIFF
--- a/src/main/utility/pod.rs
+++ b/src/main/utility/pod.rs
@@ -1,3 +1,5 @@
+use std::mem::MaybeUninit;
+
 /// Marker trait that the given type is Plain Old Data; i.e. that it is safe to
 /// interpret any pattern of bits as a value of this type.
 ///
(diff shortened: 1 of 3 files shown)


src/main/utility/pod.rs
```diff
@@ -10,29 +12,33 @@
...
+            slice.len() * std::mem::size_of::<MaybeUninit<T>>(),
         )
     }
 }
 
 /// Convert to a mut slice of raw bytes.
-pub fn to_u8_slice_mut<T>(slice: &mut [T]) -> &mut [u8]
+///
+/// Some bytes may be uninialized if T has padding.
+pub fn to_u8_slice_mut<T>(slice: &mut [T]) -> &mut [MaybeUninit<u8>]
```
stevenengler, line 33: I think the "mut" version of this function should be "unsafe". For example:

```rust
let mut x = [0u8; 10];
{
    let mut y = to_u8_slice_mut(&mut x);
    y[0] = MaybeUninit::uninit();
}
println!("{}", x[0]);
```

stevenengler (approved):

src/main/utility/pod.rs
sporksmith: Nice catch; done.

PR #2560, opened 2022-11-21
Host accessors: use impl trait instead of taking closures

sporksmith:
I got this idea from what look like a style guide for `std` development, but now can't find it again :/

Anyway, the idea is to use [impl Trait](https://doc.rust-lang.org/stable/rust-by-example/trait/impl_trait.html) to return borrowed objects, while still isolating the caller from most of the implementation details about how they're stored.

The previous approach of `with_*` hides the implementation details, but forces callers to write a closure, which is sometimes inconvenient or messy.

Using impl Trait involves a little more boilerplate in the accessors, particularly when we want to provide both const and mut options, but not as much as e.g. wrapping references to the cells themselves in newtypes.

This approach also still lets us use `Ref::map` and `RefMut::map` to transform the internals of the borrowed object (e.g. dereference pointers, unwrap options, etc), which we wouldn't be able to easily do by exposing a reference to the cell itself.

### DIFF
--- a/src/main/host/host.rs
+++ b/src/main/host/host.rs
@@ -22,7 +22,7 @@ use shadow_shim_helper_rs::simulation_time::SimulationTime;
 use shadow_shim_helper_rs::HostId;
 use shadow_shmem::allocator::ShMemBlock;
 use shadow_tsc::Tsc;
-use std::cell::{Cell, RefCell, UnsafeCell};
+use std::cell::{Cell, Ref, RefCell, RefMut, UnsafeCell};
 use std::ffi::{CString, OsString};
 use std::net::{Ipv4Addr, SocketAddrV4};
 use std::ops::{Deref, DerefMut};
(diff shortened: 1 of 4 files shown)


src/main/host/host.rs
```diff
@@ -490,27 +483,37 @@ impl Host {
...
+        addr: Ipv4Addr,
+    ) -> Option<impl Deref<Target = NetworkInterface> + DerefMut + '_> {
+        match self.interface_cell(addr) {
+            Some(rc) => {
+                let interface = rc.borrow_mut();
+                Some(RefMut::map(interface, |i| i.as_mut().unwrap()))
+            }
+            None => None,
+        }
+    }
```
stevenengler, line 500: Just an idea, but maybe worth adding a `#[track_caller]` to these? If anything in these functions panic, it would probably be more useful to know where the function was called from, rather than that for example the specific `borrow_mut()` function panicked.

stevenengler (approved):

src/main/host/host.rs
sporksmith: makes sense; done

Issue #2566, opened 2022-11-28
Add option to setup script for running loom tests and run them in continuous integration


sporksmith:
https://github.com/shadow/shadow/pull/2558 adds some tests designed to run with [loom](https://docs.rs/loom/latest/loom/). Doing so requires enabling the  rust configuration string "loom", which means rebuilding the relevant pieces of rust code with that config enabled. Right now there are comments in the code describing how to do this manually.

It'd be nice to automate this; e.g. add support to our `setup` script for running these tests and maybe run them in our continuous integration testing.

Before doing that it might make sense to move code with loom test (currently SelfContainedMutex, but later maybe our custom semaphore and spinlock) into their own crate with minimal dependencies, to minimize how much code needs to be rebuilt with `RUSTFLAGS="--cfg loom"`

PR #2567, opened 2022-11-28
SelfContainedMutex: move from shadow-shim-helper-rs to shadow_shmem

sporksmith:
shadow_shmem is a better home for "utility" things to work with shared memory that don't depend on the rest of Shadow.

This also lets us avoid having to compile shadow-shim-helper-rs when testing SelfContainedMutex in the loom configuration, which saves some work and lets us avoid some loom-specific workarounds in unrelated code.

Progress on https://github.com/shadow/shadow/issues/2566


src/lib/shmem/src/lib.rs
```diff
@@ -18,3 +18,5 @@ mod c_bindings {
     #![allow(improper_ctypes)]
     include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 }
+
+pub mod scmutex;
```
stevenengler, line 22: nit: put this above next to `pub mod allocator`?

stevenengler (approved):

src/lib/shmem/src/lib.rs
sporksmith: done

PR #2568, opened 2022-11-28
Reverse host ID order

stevenengler:
This is to see how many tests fail if we reverse the host ID order


robgjansen:
We lowered the TGen speed threshold in #2633, so the tgen tests should now be more resilient to minor changes in packet scheduling such as those that would be caused by reversing the host order.

PR #2574, opened 2022-11-29
Consolidate more

sporksmith:


sporksmith:
Here's what the full workflow looks like. It'd be nice to get the shadow build time down too, but at least this deduplicates it a bit.

https://github.com/shadow/shadow/actions/runs/3574968073
image

PR #2575, opened 2022-11-29
Extra Tests CI: Improve Rust caching

sporksmith:
* Cache a bit more of ~/.cargo. Unclear whether this ended up being needed in the end, but matches other examples, and didn't seem to increase compressed cache size significantly.
* Key the ~/.cargo cache on the hash of the lock file instead of the current YYYY-MM. This invalidates the cache when our dependencies change, instead of having an increasingly stale cache over the course of the month.
* Use sccache to cache *compiled* dependencies. This ends up only needing ~200 MB of cache, and speeds up the build quite a bit. (~10m-15m -> ~5m)

### DIFF
--- a/.github/workflows/extra_tests.yml
+++ b/.github/workflows/extra_tests.yml
@@ -33,11 +33,18 @@ env:
 
 jobs:
   build_shadow:
+    env:
+      # used by cargo
+      RUSTC_WRAPPER: sccache
+      SCCACHE_CACHE_SIZE: 2G
+      SCCACHE_DIR: /home/runner/.cache/sccache
+      # SCCACHE_RECACHE: 1 # Uncomment this to clear cache, then comment it back out
     runs-on: ubuntu-20.04
     container:
       # Should match env.CONTAINER.
       image: ubuntu:22.04
     steps:
+      - run: apt-get update
       - name: Checkout shadow
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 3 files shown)


sporksmith:
The sccache cache is small enough that we could think about using it for other jobs as well. It's about 200 MB, and GH allows 10 GB per repository. Maybe we could even go back to "full" builds for the full platform matrix.

I've had enough of shaving this yak for the moment though, and it wouldn't hurt to wait and see how our experience is with it in this workflow.

.github/workflows/extra_tests.yml
```diff
@@ -63,10 +70,36 @@ jobs:
           key: cargo-registry-${{ hashFiles('shadow/src/Cargo.lock') }}
           restore-keys: |
             cargo-registry-
+      - name: Restore sccache cache
+        uses: actions/cache@v3
+        continue-on-error: false
+        with:
+          path: ${{ env.SCCACHE_DIR }}
+          key: sccache-${{ env.CONTAINER }}-${{ env.CC }}-${{ env.BUILDTYPE }}-${{ hashFiles('shadow/src/Cargo.lock') }}
```
stevenengler, line 78: I think this should have the rust compiler version, since if the rust version changes everything needs to be rebuilt and we should invalidate the cache

.github/workflows/extra_tests.yml
```diff
@@ -63,10 +70,36 @@ jobs:
...
+          restore-keys: |
+                        sccache-${{ env.CONTAINER }}-${{ env.CC }}-${{ env.BUILDTYPE }}-
```
stevenengler, line 80: Will this cause the cache size to always increase? As we update dependency versions, will it continue to keep build artifacts from old dependencies since we never "reset" the cache?

.github/workflows/extra_tests.yml
```diff
@@ -33,11 +33,18 @@ env:
 
 jobs:
   build_shadow:
+    env:
+      RUSTC_WRAPPER: sccache
+      RUSTV: stable
```
stevenengler, line 38: Are these used anywhere?

stevenengler (approved):

.github/workflows/extra_tests.yml
sporksmith: Dug into this a bit; I think you're right. I'll add the month back into the cache key so it gets reset monthly.

.github/workflows/extra_tests.yml
sporksmith: done

.github/workflows/extra_tests.yml
sporksmith: RUSTC_WRAPPER is used by cargo; added a comment.

RUSTV isn't, afaict. Removed.

PR #2583, opened 2022-11-30
Host::add_application: take Rust objects and remove unsafe

sporksmith:


sporksmith:
This change is also included in https://github.com/shadow/shadow/pull/2584 ; up to you whether to review this separately or not

PR #2589, opened 2022-12-05
Addd `InetSocket` enum and placeholder `TcpSocket` struct

stevenengler:


src/main/host/descriptor/socket/inet/mod.rs
```diff
@@ -0,0 +1,251 @@
...
+use crate::utility::sockaddr::SockaddrStorage;
+use crate::utility::HostTreePointer;
+
+use tcp::TcpSocket;
+
+pub mod tcp;
+
+#[derive(Clone)]
+pub enum InetSocket {
+    Tcp(Arc<AtomicRefCell<TcpSocket>>),
```
robgjansen, line 19: I think we should name the version that wraps the C TCP stack inside of it as something other than `Tcp`, and reserve the name `Tcp` for the new Rust TCP stack. That way when we ultimately remove the C stack we'll have the Rust name already set.

robgjansen (approved):

PR #2599, opened 2022-12-09
Support `getsockname`/`getpeername()` for tcp wrapper

stevenengler:


src/main/host/descriptor/socket/inet/tcp.rs
```diff
@@ -99,11 +101,39 @@ impl TcpSocket {
     }
 
     pub fn getsockname(&self) -> Result<Option<SockaddrIn>, SyscallError> {
-        todo!()
+        let mut ip: libc::in_addr_t = unsafe { std::mem::zeroed() };
```
sporksmith, line 104: Alternatively, `Pod::zeroed`. (Not `unsafe`, since `in_addr_t` is `Pod`)

sporksmith (approved):

src/main/host/descriptor/socket/inet/tcp.rs
stevenengler: I forgot that a `in_addr_t` is just an integer, so I set it equal to 0 instead.

PR #2603, opened 2022-12-12
Add support for rust sockets in the network interface

stevenengler:
This adds support for the `InetSocket` type in `CompatSocket`. We don't actually use this variant of `CompatSocket` yet since the C bind/connect/listen syscall handlers work only with `LegacySocket` objects. When we write these syscall handlers in rust, we can add our `InetSocket` directly to the network interface instead of downcasting to a `TcpSocket` and then to a `LegacySocket` like the C syscalls do.

This does break the tracker's per-socket statistics, but we don't currently use them, we're considering removing them, and I will try to update the tracker in another PR.


src/main/utility/legacy_callback_queue.rs
```diff
@@ -33,9 +33,13 @@ impl LegacyCallbackQueue {
         // loop until there are no more events
         let mut count = 0;
 
-        // the mutable borrow is short-lived so that new callbacks can be added to the queue while
-        // running `f`
-        while let Some(f) = queue.borrow_mut().as_mut().unwrap().0.pop_front() {
```
sporksmith, line 38: Oh interesting; took me a bit to figure out why this is needed.

https://rustwiki.org/en/reference/destructors.html#drop-scopes

> Drop scopes are determined after replacing [for](https://rustwiki.org/en/reference/expressions/loop-expr.html#iterator-loops), [if let](https://rustwiki.org/en/reference/expressions/if-expr.html#if-let-expressions), and [while let](https://rustwiki.org/en/reference/expressions/loop-expr.html#predicate-pattern-loops) expressions with the equivalent expressions using [match](https://rustwiki.org/en/reference/expressions/match-expr.html).

I suppose the desugaring is something like

```
loop {
  match queue.borrow... {
    Some(f) => /* loop body */
    None => break
  };
}
```

and afaict from the other rules, the drop scope of the borrowed object becomes the whole match statement-expression. Definitely a bit surprising

TIL

sporksmith (approved):

PR #2605, opened 2022-12-13
Add examples for dynamic shadow config generation

sporksmith:


sporksmith:
I could be convinced to drop the nix example. I think it's neat, and the language has a lot more useful features for this kind of thing, but it's admittedly weird to point to given how coupled it is with the nix package manager.

sporksmith:
Oh! Just noticed that https://shadow.github.io/docs/guide/shadow_config_overview.html#yaml-extensions-experimental already talks about the yaml extensions and has examples.

I do kind of like having a separate page with single running example of the yaml extension + generating with other tools, though. Maybe it's fine to have both, or maybe it'd make sense to drop the examples from the overview page and have a forward reference to this page instead?

docs/shadow_config_spec.md
```diff
@@ -1,8 +1,9 @@
 # Shadow Configuration Specification
 
-Shadow uses the standard YAML format to accept configuration options from users.
-The following describes Shadow's YAML format and all of the options that Shadow
-supports that can be used to customize a simulation.
+Shadow uses the standard [YAML 1.2](https://yaml.org/spec/1.2.2/) format to
+accept configuration options from users.  The following describes Shadow's YAML
+format and all of the options that Shadow supports that can be used to customize
+a simulation.
```
stevenengler, line 6: nit: I've always wanted to remove the "from users" text. Since the person reading this is probably a user, this text seems weird.

docs/shadow_config_dynamic.md
```diff
@@ -0,0 +1,307 @@
...
+general:
+  stop_time: 10s
+network:
+  graph:
+    type: 1_gbit_switch
+hosts:
+  fast_foo:
+    <<: *FastHost
+    processes:
+    - path: *FooPath
```
stevenengler, line 128: One suggestion is that I find examples with a lot of indirection (for example with stuff like anchors and merging) to be easier to understand when they use concrete things instead of abstract things. So instead of "foo", to be a server or something. When using abstract names like "foo", it can be hard to relate it to a situation where these features would be useful in the real world, and it takes some mental energy to remember what a "foo" is (for example in this case "foo" is both a host name and a process path).

docs/SUMMARY.md
```diff
@@ -20,6 +20,7 @@
     - [Simulation Configuration]()
         - [Shadow Config Overview](shadow_config_overview.md)
         - [Shadow Config Specification](shadow_config_spec.md)
+        - [Programmatic Config Generation](shadow_config_dynamic.md)
```
stevenengler, line 23: Since this file contains some stuff that isn't "programmatic" such as yaml anchors, maybe something like "Advanced Config Generation" would be better?

stevenengler (approved):
> I do kind of like having a separate page with single running example of the yaml extension + generating with other tools, though. Maybe it's fine to have both, or maybe it'd make sense to drop the examples from the overview page and have a forward reference to this page instead?

Either sounds good to me.

docs/shadow_config_dynamic.md
```diff
@@ -0,0 +1,307 @@
+# Programmatically Generating Shadow Configurations
+
+It is sometimes useful to generate shadow configuration files programmatically.
+Since Shadow accepts configuration files in [YAML
+1.2](https://yaml.org/spec/1.2.2/) format, there are many options
+available; even more so since [JSON](https://en.wikipedia.org/wiki/JSON) is
+also valid YAML 1.2.
```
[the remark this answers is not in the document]
sporksmith: Oh interesting. The spec also has:

> The YAML 1.2[3](https://yaml.org/spec/1.2.2/#fn:1-2-spec) specification was published in 2009. Its primary focus was making YAML a strict superset of JSON

but I suppose this is a bit handwavy. They *focused* on making it a strict superset... but didn't actually do it?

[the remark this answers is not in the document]
sporksmith: Ah, so I think YAML is still a superset here on a technicality. From the [JSON spec](https://www.rfc-editor.org/rfc/rfc7159#section-4)

> The names within an object SHOULD be unique.

> An object whose names are all unique is interoperable in the sense
   that all software implementations receiving that object will agree on
   the name-value mappings.  When the names within an object are not
   unique, the behavior of software that receives such an object is
   unpredictable.  Many implementations report the last name/value pair
   only.  Other implementations report an error or fail to parse the
   object, and some implementations report all of the name/value pairs,
   including duplicates.

So, failing to parse duplicate keys is valid behavior for a JSON parser

docs/shadow_config_dynamic.md
sporksmith: I'd originally started with something more concrete using curl clients and python web servers, but that brought in too much incidental complexity.

I search-replaced foo -> client and bar -> server to make it a little more concrete, without making it too weird that I'm glossing over e.g. host names and ports.

docs/SUMMARY.md
sporksmith: I think "advanced" makes it sound like advanced configuration options in Shadow, rather than advanced usage and generation of YAML.

Maybe "Managing Complex Configurations"?

stevenengler: Sure, any of the above sound good to me.

PR #2606, opened 2022-12-13
Remove `Host::setup()`

stevenengler:
I wanted to move around some of the host's network objects, but it was a little complicated with the `setup()` function, so I moved the code from the `Host::setup()` into `Host::new()`. Hopefully this doesn't overlap with any other WIP changes.


sporksmith (approved):
:tada:

PR #2607, opened 2022-12-13
Move host networking to new `NetworkNamespace` object

stevenengler:
This serves two main purposes:

1. Consolidate the host's networking-related code into a new structure.
2. When we have a function like `Socket::bind()` that needs to be able to associate the socket with a network interface, we can pass it just this `NetworkNamespace` object instead of needing to pass the entire `Host`. This provides better encapsulation.

### DIFF
--- a/src/main/network/net_namespace.rs
+++ b/src/main/network/net_namespace.rs
@@ -183,16 +183,7 @@ impl NetworkNamespace {
         // to a linear search to make sure we get a free port if we have one.
         // but start from a random port instead of the min.
         let start = rng.gen_range(MIN_RANDOM_PORT..=u16::MAX);
-        let mut port = start;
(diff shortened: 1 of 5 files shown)


src/main/network/net_namespace.rs
```diff
@@ -183,16 +183,7 @@ impl NetworkNamespace {
...
-        loop {
-            port = if port == u16::MAX {
-                MIN_RANDOM_PORT
-            } else {
-                port + 1
-            };
-            if port == start {
-                break;
-            }
+        for port in (start..=u16::MAX).chain(MIN_RANDOM_PORT..start) {
```
sporksmith, line 186: Nice :)

sporksmith (approved):

PR #2614, opened 2022-12-15
Enforce clippy

sporksmith:
Fixes remaining clippy warnings and adds a clippy job to our lint github workflow. Also pins the default Rust version with a rust-toolchain.yaml file.


stevenengler:
Just wanted to comment while I was thinking about it: I think we should either make this test not-required, or set a fixed rust version in the CI. We don't want our CI to start failing because a new clippy version was released which adds new warnings, similar to how we don't fail on code warnings, since sometimes newer rust versions add new warnings and we don't want our tests to fail because of them.

sporksmith:
Ok, I think I've got this sorted out now.

I ultimately added a rust-toolchain.toml file to set the default toolchain in general, since it is an issue we've run into with other lints etc. It can be overridden locally various ways if desired, but it probably makes sense to mostly just have everything use that one version.

LMK whether you'd like me to clean up the history a bit before you review/continue-to-review

src/main/core/work/task.rs
```diff
@@ -82,7 +82,9 @@ pub mod export {
     }
 
     impl CTaskHostTreePtrs {
-        /// SAFETY: Given that the host lock is held when execution of a
+        /// # Safety
+        ///
+        /// Given that the host lock is held when execution of a
```
robgjansen, line 87: The column width is inconsistent with the following lines.

We're generally inconsistent with where we place the newlines in multi-row comments though, so maybe it's not worth fixing here. (Perhaps we need a lint for column width too? :)

src/main/host/host.rs
```diff
@@ -188,6 +188,10 @@ impl std::fmt::Debug for Host {
 }
 
 impl Host {
+    /// # Safety
+    ///
+    /// `dns` must be a valid pointer.
+    /// TODO: does this take ownership of `dns`?
```
robgjansen, line 194: I think there is one global DNS object that is shared to all of the hosts. And since the DNS here is a C pointer, I assume that's still true.

robgjansen (approved):

docs/coding_style.md
```diff
@@ -20,15 +20,19 @@ tool on the command-line to modify the lines touched by your commits.
...
+To format your Rust code, run `cargo fmt` in the `src` directory.
 
-Examples:
+```bash
+(cd src && cargo fmt)
+```
+
+### Clippy
+
+We use [Clippy](https://doc.rust-lang.org/stable/clippy/index.html) to helps
```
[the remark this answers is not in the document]
sporksmith: Done

src/main/core/work/task.rs
sporksmith: Fixed this one. I try to remember to reflow paragraphs, but yeah it's easy to miss, and rustfmt won't do it. :shrug:

It looks like rustfmt has [a feature for wrapping overly-long comment lines](https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#wrap_comments), but I'm unclear whether it'd reflow "too short" lines like in this case. Also appears to be in [stabilization purgatory](https://github.com/rust-lang/rustfmt/issues/3347)

src/main/host/host.rs
sporksmith: Fixed

stevenengler:
Just a note: Each time we change the rust version in `rust-toolchain.toml` it will download a new toolchain without deleting the old version, so you'll have to remember to manually uninstall old toolchains (for example `rustup toolchain uninstall 1.66.0`) if you want to save disk space.

Issue #2604, opened 2022-12-12
verify calls to _regularfile_getOSBackedFD are checked for -1, not 0


trinity-1686a:
**Describe the issue**
Some calls to `_regularfile_getOSBackedFD` checked for zero result, but the actual bad value is -1/OSFILE_INVALID, and 0 is a valid file descriptor.
Someone should verify if such calls still exists and fix them.

PR #2616, opened 2022-12-16
regular_file.c: don't treat 0 as a bad file descriptor

sporksmith:
I wanted to avoid comparing with equality to OSFILE_INVALID everywhere, in case we ever want to add other negative error values. Added an inline helper `_fd_isValid` instead.

Most of these checks could probably be omitted altogether in favor of letting the native file operation fail and return EBADF for us or us later failing to find the fd in our descriptor table, but probably not worth changing it right now.

Fixes https://github.com/shadow/shadow/issues/2604

### DIFF
--- a/src/main/host/descriptor/regular_file.c
+++ b/src/main/host/descriptor/regular_file.c
@@ -393,10 +393,6 @@ static size_t _regularfile_readvRandomBytes(RegularFile* file, const Host* host,
 ssize_t regularfile_read(RegularFile* file, const Host* host, void* buf, size_t bufSize) {
     MAGIC_ASSERT(file);
 
-    if (!_fd_isValid(_regularfile_getOSBackedFD(file))) {
-        return -EBADF;
-    }
-
     if (file->type == FILE_TYPE_RANDOM) {
         _regularfile_readRandomBytes(file, host, buf, bufSize);
         return (ssize_t)bufSize;
@@ -406,6 +402,10 @@ ssize_t regularfile_read(RegularFile* file, const Host* host, void* buf, size_t 
         ssize_t read = regularfile_pread(file, host, buf, bufSize, file->inMemoryFile.cursor);
         file->inMemoryFile.cursor += read;
         return read;
+    }
+
+    if (!_fd_isValid(_regularfile_getOSBackedFD(file))) {
+        return -EBADF;
     }
 
     trace("RegularFile %p will read %zu bytes from os-backed file %i at path '%s'", file, bufSize,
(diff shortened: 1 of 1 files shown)


src/main/host/descriptor/regular_file.c
```diff
@@ -531,11 +531,11 @@ ssize_t regularfile_preadv2(RegularFile* file, const Host* host, const struct io
 ssize_t regularfile_write(RegularFile* file, const void* buf, size_t bufSize) {
     MAGIC_ASSERT(file);
 
-    if (!_fd_isValid(_regularfile_getOSBackedFD(file))) {
+    if (file->type == FILE_TYPE_IN_MEMORY) {
         return -EBADF;
     }
 
-    if (file->type == FILE_TYPE_IN_MEMORY) {
+    if (!_fd_isValid(_regularfile_getOSBackedFD(file))) {
```
robgjansen, line 538: I don't understand why this change is necessary, or the other similar changes below, because I think the code is equivalent with or without the change?

Since the two `if` statements are executed in sequence, isn't it equivalent to

```C
if (file->type == FILE_TYPE_IN_MEMORY || !_fd_isValid(_regularfile_getOSBackedFD(file))) {
    return -EBADF;
}    
```

in which case the order should not matter?

src/main/host/descriptor/regular_file.c
```diff
@@ -888,6 +884,9 @@ int regularfile_fcntl(RegularFile* file, unsigned long command, void* arg) {
         arg = (void*)arg_int;
     }
 
+    if (!_fd_isValid(_regularfile_getOSBackedFD(file))) {
+        return -EBADF;
+    }
```
robgjansen, line 889: Above we are potentially making a change to some file state in Shadow, and returning some values in the arg, but then returning an `EBADF` error. But if we return an error, doesn't that mean that we don't want to make any changes to shadow state (and not write to arg memory) based on this erroneous call?

If this _is_ still a valid path, then a comment to explain why would be useful.

robgjansen (approved):

src/main/host/descriptor/regular_file.c
sporksmith: Yup! Basically I was moving the `_isValid` checks to just-before the native calls, to better convey/preserve the intent that it's there primarily to avoid a native syscall that we know would fail.

src/main/host/descriptor/regular_file.c
sporksmith: Yeah, this one is a bit of a grey area. Reverted this one.

PR #2620, opened 2022-12-19
Issue 2619; fixes TCP header sizes in pcap files

### LINKED ISSUE #2619: Shadow writes the wrong TCP header length in pcap files
**Describe the issue**

Shadow writes the wrong TCP header length in pcap files

**Additional context**

Shadow currently reports that the TCP segment length is 32 bytes. That's wrong: TCP header sizes are currently 20 bytes.

Here's the offending line:

https://github.com/shadow/shadow/blob/894a63ca370db1bff903143d4f25a61b47f30422/src/main/network/packet.rs#L202

This should instead match the CONFIG_HEADER_SIZE_TCP macro constant defined in `definitions.h`.

RWails:
Fixes #2619

### DIFF
--- a/src/main/network/packet.rs
+++ b/src/main/network/packet.rs
@@ -201,7 +201,7 @@ fn display_tcp_bytes(packet: *const c::Packet, mut writer: impl Write) -> std::i
     };
 
     // c::CONFIG_HEADER_SIZE is in bytes. Ultimately, TCP header len is represented in 32-bit
-    // words, so we divide by 4. The right-shift of 4 is because the header len is represented
+    // words, so we divide by 4. The left-shift of 4 is because the header len is represented
     // in the top 4 bits.
     let mut header_len: u8 = c::CONFIG_HEADER_SIZE_TCP.try_into().unwrap();
     header_len /= 4;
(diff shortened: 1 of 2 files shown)


src/main/network/packet.rs
```diff
@@ -199,7 +199,14 @@ fn display_tcp_bytes(packet: *const c::Packet, mut writer: impl Write) -> std::i
     } else {
         0u32.to_be_bytes()
     };
-    let header_len: u8 = 0x80;
+
+    // c::CONFIG_HEADER_SIZE is in bytes. Ultimately, TCP header len is represented in 32-bit
+    // words, so we divide by 4. The right-shift of 4 is because the header len is represented
```
stevenengler, line 204: "right-shift" -> "left-shift"?

stevenengler (approved):
Nice catch!

stevenengler:
(Rebasing should fix the CI failure.)

src/main/network/packet.rs
RWails: yep, fixed.

RWails:
I also corrected the pcap writer to not produce TCP options fields as they are not being used right now

PR #2623, opened 2022-12-20
Rename borrowing accessors

sporksmith:
Follow-up to @stevenengler's feedback on https://github.com/shadow/shadow/pull/2612.

* Rename APIs that return borrow-counted objects
* add `#[track_caller]` to borrowing APIs


robgjansen (approved):
Nit: I think you want the extended commit message of bd643c18939a5a4efee3c3f803ec46af98e49962 to start with "When" instead of "This".

PR #2626, opened 2022-12-21
Add rust `TcpSocket` to network interface as `LegacySocket`

stevenengler:
When #2611 was merged, it added the first use of rust `InetSocket` objects in the network interface. This seems to have affected the graphs in the tgen benchmark.

1671657833_grim
1671657857_grim

I don't think the simulation itself changed, just the tracker ~was~ became broken since only metrics that came from [Shadow](https://github.com/shadow/benchmark-results/blob/master/tgen/2022-12-20-T03-10-21/plots/shadow.results.pdf) changed, but not metrics that came from [tgen](https://github.com/shadow/benchmark-results/blob/master/tgen/2022-12-20-T03-10-21/plots/tgen.viz.pdf).

This PR changes it back to adding the C `LegacySocket` to the network interface instead, which causes the shadow metrics to go back to how they were before.

1671658236_grim
1671658261_grim

### DIFF
--- a/src/main/host/descriptor/socket/inet/mod.rs
+++ b/src/main/host/descriptor/socket/inet/mod.rs
@@ -333,7 +333,9 @@ fn associate_socket(
         return Err(Errno::EADDRINUSE.into());
     }
 
-    let socket = unsafe { c::compatsocket_fromInetSocket(&socket) };
+    // TODO: create the CompatSocket from the InetSocket
+    let InetSocket::Tcp(socket) = socket;
+    let socket = unsafe { c::compatsocket_fromLegacySocket(socket.borrow().as_legacy_socket()) };
 
     // associate the interfaces corresponding to addr with socket
     unsafe { net_ns.associate_interface(&socket, protocol, local_addr, peer_addr) };


sporksmith:
> I don't think the simulation itself changed, just the tracker was broken since only metrics that came from [Shadow](https://github.com/shadow/benchmark-results/blob/master/tgen/2022-12-20-T03-10-21/plots/shadow.results.pdf) changed, but not metrics that came from [tgen](https://github.com/shadow/benchmark-results/blob/master/tgen/2022-12-20-T03-10-21/plots/tgen.viz.pdf).

If we think the differences in the PR were due to fixing the tracker, why revert? Or is the plan to confirm that more carefully but revert in the meantime...?

sporksmith:
Oh I think I mis-parsed. You're saying the previous change *broke* the tracker, so this should fix it?

sporksmith (approved):

stevenengler:
> > I don't think the simulation itself changed, just the tracker was broken since only metrics that came from [Shadow](https://github.com/shadow/benchmark-results/blob/master/tgen/2022-12-20-T03-10-21/plots/shadow.results.pdf) changed, but not metrics that came from [tgen](https://github.com/shadow/benchmark-results/blob/master/tgen/2022-12-20-T03-10-21/plots/tgen.viz.pdf).
> 
> If we think the differences in the PR were due to fixing the tracker, why revert? Or is the plan to confirm that more carefully but revert in the meantime...?

There are two parts to this. In #2603 support for rust `InetSocket`s was added to the network interface, and in #2611 we added the first use of these `InetSocket`s in the network interface. So the issue was probably introduced in #2603, but that code path was only "activated" in #2611. I'm not entirely sure what the underlying issue is since I had looked into this beforehand and didn't think it was an issue, but for now it's simple enough to revert to the old behaviour and I'll look into it more later.

PR #2631, opened 2022-12-22
Migrate most of Process from C to Rust

sporksmith:
Best reviewed as individual commits.

This moves each member of the C Process struct into the Rust struct, and adds corresponding Rust methods. Many `process_` methods are now thin wrappers around a `_process_` method that operates on the Rust process.

I ran out of time to do the thread list, but don't foresee any major complications there. The next steps will be roughly:
* Move the last member (thread list) of Process to Rust
* Implement the remaining `process_` methods in Rust, such that they're all thin wrappers around `_process_` methods taking the Rust Process.
* The big changeover:
  * Delete process.c and process.h.
  * Rename or alias RustProcess to Process in the C code.
  * Rename the `_process_*` accessors to `process_*` - i.e. they'll have almost the same signature as the methods they're replacing, except the first parameter will be a const pointer to the RustProcess.


sporksmith:
Started a tor benchmark: https://github.com/shadow/benchmark/actions/runs/3762181499

stevenengler:
@sporksmith Is it straightforward for you to move a6a0045 to a new PR? I need more time to look through this, but the rest looks good to me and don't want to hold it up.

src/main/host/process.rs
```diff
@@ -124,8 +126,20 @@ pub struct Process {
...
+    // TODO: Remove these when we've migrated Shadow off of the APIs that need
+    // them (probably by migrating all the calling code to Rust).
+    //
+    // Must be before memory_manager for drop order.
+    unsafe_borrow_mut: RefCell<Option<UnsafeBorrowMut>>,
+    unsafe_borrows: RefCell<Vec<UnsafeBorrow>>,
+
+    // Must come after `unsafe_borrows` and `unsafe_borrow_mut`.
+    // Boxed to avoid invalidating those if Self is moved.
+    memory_manager: Box<RefCell<Option<MemoryManager>>>,
```
stevenengler, line 142: Maybe mark these comments with "SAFETY" so that they're more noticable (in vim it's a highlighted word, I'm not sure about other editors)?

src/main/host/process.rs
```diff
@@ -484,20 +500,236 @@ impl Process {
...
+
+    /// Free this reference, writing back to process memory.
+    fn flush(&mut self) -> Result<(), Errno> {
+        self.memory.take().unwrap().flush()
+    }
+
+    /// Free this reference without writing back to process memory.
+    fn noflush(&mut self) {
+        self.memory.take().unwrap().noflush()
     }
```
stevenengler, line 714: Is there a reason these don't take `self`? I think this would allow you to get rid of the `Option`? But if there's a reason for the `Option`, this could use a comment explaining why.

stevenengler (approved):

src/main/host/process.rs
```diff
@@ -185,6 +186,33 @@ impl Process {
...
+            let task = TaskRef::new(move |host| {
+                let process = host.process_borrow(id).unwrap();
+                let cprocess = unsafe { process.borrow(host.root()).cprocess() };
+                unsafe { cshadow::process_start(cprocess) };
+            });
+            host.schedule_task_at_emulated_time(task, self.start_time);
+        }
+
+        if let Some(stop_time) = self.stop_time {
+            if stop_time > self.start_time {
```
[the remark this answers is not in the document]
sporksmith: Done; rearranged a bit to add an early exit if stop_time is <= start_time.

src/main/host/process.rs
```diff
@@ -189,6 +189,17 @@ impl Process {
...
+            CString::new(format!(
+                "SHADOW_LOG_FILE={}",
+                process
+                    .borrow(host.root())
+                    .output_file_name(host, "shimlog")
+                    .to_str()
+                    .unwrap()
+            ))
+            .unwrap(),
+        );
```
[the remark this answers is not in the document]
sporksmith: Added a TODO

src/main/host/process.c
```diff
@@ -63,12 +63,6 @@
 static bool _use_shim_syscall_handler = true;
 ADD_CONFIG_HANDLER(config_getUseShimSyscallHandler, _use_shim_syscall_handler)
 
-// Shadow 1.x did not adjust the plugins working directories, but Shadow now runs each plugin with
-// the working directory of the host data path. Using the legacy working directory is useful when
-// running the same experiment in multiple versions of Shadow for performacne comparison purposes.
```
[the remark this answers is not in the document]
sporksmith: Done

src/main/host/process.rs
sporksmith: Done

src/main/host/process.rs
sporksmith: Yup, taking `self` is better; done. Still need the `Option`, though.

PR #2645, opened 2023-01-06
Allow higher-level host log levels

### LINKED ISSUE #2643: Host-specific log level appears to be broken
If I add the following to the tor minimal test case (with shadow built in debug mode), I do not see trace-level logs for that host in shadow's log:

```yaml
torhiddenclient:
  network_node_id: 0
  options:
    log_level: trace
  processes:
    [...]
```

stevenengler:
Previously if a host had a log level higher than the simulation log level, the host log level wouldn't take effect. Now you can specify higher host log levels (for example "trace") even if you specify a lower simulation log level (for example "info").

Closes #2643.

This doesn't appear to have any effect on runtime.

https://github.com/shadow/benchmark-results/tree/master/tor/2023-01-06-T17-24-59

run_time

### DIFF
--- a/src/main/core/logger/log_wrapper.c
+++ b/src/main/core/logger/log_wrapper.c
@@ -12,7 +12,7 @@ static void _flush(Logger* logger) { rustlogger_flush(); }
 
 static bool _isEnabled(Logger* logger, LogLevel level) { return rustlogger_isEnabled(level); }
 
-static void _setLevel(Logger* logger, LogLevel level) { rustlogger_setLevel(level); }
+static void _setLevel(Logger* logger, LogLevel level) { warning("Setting the log level is not supported"); }
 
 Logger* rustlogger_new() {
     Logger* logger = malloc(sizeof(*logger));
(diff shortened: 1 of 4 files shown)


src/main/core/logger/log_wrapper.c
```diff
@@ -12,7 +12,7 @@ static void _flush(Logger* logger) { rustlogger_flush(); }
 
 static bool _isEnabled(Logger* logger, LogLevel level) { return rustlogger_isEnabled(level); }
 
-static void _setLevel(Logger* logger, LogLevel level) { rustlogger_setLevel(level); }
+static void _setLevel(Logger* logger, LogLevel level) { warning("Setting the log level is not supported"); }
```
sporksmith, line 15: Remove?

sporksmith (approved):

src/main/core/logger/log_wrapper.c
stevenengler: The `logger_setLevel` is still needed for the shim, so we can't remove `setLevel` from `Logger`. Since the shim and shadow use the same `Logger` interface, we need to set this `setLevel` function to something in shadow, so it's set to this dummy `_setLevel()` function. I don't see a nice way of removing this function without decoupling the shim and shadow loggers.

PR #2646, opened 2023-01-06
Log syscalls that are handled in the shim

stevenengler:
Examples of shim-handled syscalls:

```
00:10:00.301000001 [tid 1000] epoll_wait(...) = 1
00:10:00.301000001 [tid 1000] read(7, 0x7ffffffee6d0, 65536) = 1460
000000600301000001 [tid 1000] clock_gettime(...) = 0
00:10:00.301000001 [tid 1000] epoll_wait(...) = 1
00:10:00.301000001 [tid 1000] epoll_wait(...) = 1
00:10:00.301000001 [tid 1000] read(7, 0x7ffffffee6d0, 65536) = 1460
000000600301000001 [tid 1000] clock_gettime(...) = 0
00:10:00.301000001 [tid 1000] epoll_wait(...) = 1
00:10:00.301000001 [tid 1000] epoll_wait(...) = 1
00:10:00.301000001 [tid 1000] read(7, 0x7ffffffee6d0, 65536) = 1460
000000600301000001 [tid 1000] clock_gettime(...) = 0
00:10:00.301000001 [tid 1000] epoll_wait(...) = 1
00:10:00.301000001 [tid 1000] epoll_wait(...) = 1
```

Examples of native syscalls:

```
00:10:00.000000000 [tid 1000] munmap(0x7ffff7ffb000, 118) = 0
000000600000000000 [tid 1000] gettimeofday(...) = 0
00:10:00.000000000 [tid 1000] access(...) = <native>
000000600000000000 [tid 1000] ^^^ = 0
00:10:00.000000000 [tid 1000] access(...) = <native>
000000600000000000 [tid 1000] ^^^ = 0
00:10:00.000000000 [tid 1000] openat(-100, "../../../conf/tgen.client.graphml.xml", O_LARGEFILE | O_RDONLY, (empty)) = 3
00:10:00.000000000 [tid 1000] readlink(...) = <native>
000000600000000000 [tid 1000] ^^^ = 27
```

The time isn't formatted properly in these log lines, but I'll leave that for a future PR.

### DIFF
--- a/src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
+++ b/src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
@@ -500,20 +500,6 @@ pub mod export {
     ///
     /// Pointer args must be safely dereferenceable.
     #[no_mangle]
-    pub unsafe extern "C" fn shimshmemprocess_init(
-        process_mem: *mut ShimShmemProcess,
-        lock: *const ShimShmemHostLock,
-    ) {
-        let lock = unsafe { lock.as_ref().unwrap() };
-        let m = ProcessShmem::new(&lock.root, lock.host_id);
-        assert_shmem_safe!(ProcessShmem, _test_process_shmem);
-        unsafe { process_mem.write(m) }
-    }
-
-    /// # Safety
-    ///
-    /// Pointer args must be safely dereferenceable.
-    #[no_mangle]
     pub unsafe extern "C" fn shimshmem_getEmulatedTime(
         host_mem: *const ShimShmemHost,
     ) -> CEmulatedTime {
--- a/src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
+++ b/src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
@@ -226,14 +226,16 @@ impl ProcessShmemProtected {
 #[repr(C)]
 pub struct ThreadShmem {
     host_id: HostId,
+    tid: libc::pid_t,
 
     protected: RootedRefCell<ThreadShmemProtected>,
 }
 
 impl ThreadShmem {
-    pub fn new(host: &HostShmemProtected) -> Self {
+    pub fn new(host: &HostShmemProtected, tid: libc::pid_t) -> Self {
         Self {
             host_id: host.host_id,
+            tid,
             protected: RootedRefCell::new(
                 &host.root,
                 ThreadShmemProtected {
@@ -680,11 +682,21 @@ pub mod export {
     pub unsafe extern "C" fn shimshmemthread_init(
         thread_mem: *mut ShimShmemThread,
         lock: *const ShimShmemHostLock,
+        tid: libc::pid_t,
     ) {
         let lock = unsafe { lock.as_ref().unwrap() };
-        let t = ThreadShmem::new(lock);
+        let t = ThreadShmem::new(lock, tid);
         assert_shmem_safe!(ThreadShmem, _test_thread_shmem);
         unsafe { thread_mem.write(t) }
+    }
+
+    /// # Safety
+    ///
+    /// Pointer args must be safely dereferenceable.
+    #[no_mangle]
+    pub unsafe extern "C" fn shimshmem_getThreadId(thread: *const ShimShmemThread) -> libc::pid_t {
+        let thread_mem = unsafe { thread.as_ref().unwrap() };
+        thread_mem.tid
     }
 
     /// # Safety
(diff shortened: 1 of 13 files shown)


stevenengler:
I think the failed tests are because I'm calling `dprintf()` from the shim, but we should only be calling async-signal safe functions (which `dprintf` isn't). I think the best solution might be a rust function which does the formatting and writing to the fd.

sporksmith:
> Edit 3: @sporksmith Since the shim's vdso just makes the syscall, this patch will log vdso time calls as syscalls. I think this doesn't hurt at all, but wanted to mention it here in case you think we shouldn't log these vdso calls in the strace log. The normal strace tool wouldn't log these, but we don't need to follow the strace tool.

Yeah, I think logging them is probably what we want.

src/lib/shim/shim_sys.c
```diff
@@ -142,6 +154,24 @@ bool shim_sys_handle_syscall_locally(long syscall_num, long* rv, va_list args) {
...
+        // TODO: format the time
+        uint64_t emulated_time_ms = shim_sys_get_simtime_nanos();
+        pid_t tid = shimshmem_getThreadId(shim_threadSharedMem());
+
+        bool oldNativeSyscallFlag = shim_swapAllowNativeSyscalls(true);
+        char buf[100] = {0};
+        int len = snprintf(buf, sizeof(buf), "%018ld [tid %d] %s(...) = %ld\n", emulated_time_ms,
+                           tid, syscallName, *rv);
+        len = MIN(len, sizeof(buf));
+        if (write(straceFd, buf, len) < 0) {
```
sporksmith, line 169: technically I think we ought to loop here and keep track of how much was actually written. A successful `write` call could write anything from 0 to `len` bytes. OTOH this probably won't happen in practice.

src/lib/shim/shim_syscall.c
```diff
@@ -181,6 +182,27 @@ static SysCallReg _shim_emulated_syscall_event(const ShimEvent* syscall_event) {
...
+                    // TODO: format the time
+                    uint64_t emulated_time_ms = shim_sys_get_simtime_nanos();
+                    pid_t tid = shimshmem_getThreadId(shim_threadSharedMem());
+
+                    bool oldNativeSyscallFlag = shim_swapAllowNativeSyscalls(true);
+                    char buf[100] = {0};
+                    int len = snprintf(buf, sizeof(buf), "%018ld [tid %d] ^^^ = %ld\n",
+                                       emulated_time_ms, tid, rv.as_i64);
+                    len = MIN(len, sizeof(buf));
+                    if (write(straceFd, buf, len) < 0) {
```
sporksmith, line 200: Same here - may be worth handling partial writes.

sporksmith (approved):

src/lib/shim/shim_sys.c
```diff
@@ -154,21 +154,36 @@ bool shim_sys_handle_syscall_locally(long syscall_num, long* rv, va_list args) {
...
         char buf[100] = {0};
         int len = snprintf(buf, sizeof(buf), "%018ld [tid %d] %s(...) = %ld\n", emulated_time_ms,
                            tid, syscallName, *rv);
         len = MIN(len, sizeof(buf));
-        if (write(straceFd, buf, len) < 0) {
-            warning("Unable to write to strace log");
+
+        int written = 0;
+        while (1) {
+            int write_rv = write(straceFd, buf + written, len - written);
```
sporksmith, line 173: `write(2)` returns -1 on error, and the specific error in `errno`

src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
```diff
@@ -167,6 +170,8 @@ pub struct ProcessShmemProtected {
     // actually delivering them. This is to handle the case where handlers are
     // defensively installed, but not used in practice.
     signal_actions: [shd_kernel_sigaction; SHD_SIGRT_MAX as usize],
+
+    pub strace_fd: FfiOption<libc::c_int>,
```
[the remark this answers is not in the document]
stevenengler: Yeah that's a good idea. I think when it was originally written for the C code it was more difficult to reorder the initialization, and I didn't check it again when I moved it over to the rust `Process`.

src/lib/shim/shim_sys.c
stevenengler: Yeah it seems to work fine in the shim log which just does one `write` call, but agreed that it's best to write it properly here. Fixed.

src/lib/shim/shim_syscall.c
stevenengler: Fixed.

src/lib/shim/shim_sys.c
stevenengler: Oops good catch. Fixed.

sporksmith (approved):

PR #2648, opened 2023-01-09
Support displaying more-complex syscall arguments

stevenengler:
In order to display syscall arguments in strace logging, we sometimes need to know the values of multiple syscall arguments. For example to display the byte buffer for `write()`, we need to know both `const void *buf` and `size_t count`. Shadow previously would just read bytes until a nul or a memory access error, but this wasn't correct since the buffer is often not nul-terminated.

To print a byte buffer like the one in `write()`, instead of using `*const libc::c_char` in the `log_syscall()` macro, we now have two types `SyscallStringArg` and `SyscallBufferArg<const LEN_INDEX: usize>` depending on whether the syscall takes a buffer (as in `write()`) or a nul-terminated string (as in `open()`). The constant generic in `SyscallBufferArg` allows us to specify the index of the `count` argument. This is a little weird, but allows us to continue using rust's type system for specifying how to display syscall arguments.

In the future we can use the same method to display other types of syscall arguments, like socket addresses or socket/ioctl options.

### DIFF
--- a/src/main/host/syscall/type_formatting.rs
+++ b/src/main/host/syscall/type_formatting.rs
@@ -205,8 +205,7 @@ impl TryFromSyscallReg for nix::sys::mman::MRemapFlags {
 simple_display_impl!(i8, i16, i32, i64, isize);
 simple_display_impl!(u8, u16, u32, u64, usize);
 
-// skip *const i8 since we have a custom string format impl below
-deref_pointer_impl!(i16, i32, i64, isize);
+deref_pointer_impl!(i8, i16, i32, i64, isize);
 deref_pointer_impl!(u8, u16, u32, u64, usize);
 
 deref_array_impl!(i8, i16, i32, i64, isize);
@@ -274,7 +273,10 @@ fn fmt_string(
     }
 }
 
-impl SyscallDisplay for SyscallVal<'_, *const i8> {
+/// Displays a nul-terminated string syscall argument.
+pub struct SyscallStringArg {}
+
+impl SyscallDisplay for SyscallVal<'_, SyscallStringArg> {
     fn fmt(
         &self,
         f: &mut std::fmt::Formatter<'_>,
--- a/src/main/host/syscall/type_formatting.rs
+++ b/src/main/host/syscall/type_formatting.rs
@@ -225,6 +225,55 @@ simple_debug_impl!(nix::sys::mman::ProtFlags);
 simple_debug_impl!(nix::sys::mman::MapFlags);
 simple_debug_impl!(nix::sys::mman::MRemapFlags);
 
+fn fmt_string(
+    f: &mut std::fmt::Formatter<'_>,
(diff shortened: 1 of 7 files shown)


src/main/host/syscall/type_formatting.rs
```diff
@@ -224,6 +224,51 @@ simple_debug_impl!(nix::sys::mman::ProtFlags);
...
+    options: FmtOptions,
+    mem: &MemoryManager,
+) -> std::fmt::Result {
+    const DISPLAY_LEN: usize = 40;
+
+    if options == FmtOptions::Deterministic {
+        return write!(f, "<pointer>");
+    }
+
+    // read up to one extra character to check if it's a null byte
```
sporksmith, line 240: Stale comment?

src/main/host/syscall/type_formatting.rs
```diff
@@ -224,6 +224,51 @@ simple_debug_impl!(nix::sys::mman::ProtFlags);
...
+    let mem_ref = match mem.memory_ref_prefix(TypedPluginPtr::new::<u8>(ptr, len)) {
+        Ok(x) => x,
+        // the pointer didn't reference any valid memory
+        Err(_) => return write!(f, "{ptr:p}"),
+    };
+
+    let mut s = Vec::with_capacity(DISPLAY_LEN);
```
sporksmith, line 247: Nit - I think you could use a `String` here instead of later converting from a `Vec`? If so, I think that'd be slightly clearer.

sporksmith (approved):

src/main/host/syscall/type_formatting.rs
stevenengler: Fixed.

src/main/host/syscall/type_formatting.rs
stevenengler: I tried this, but the problem is that `std::ascii::escape_default` returns a `u8` iterator, and `String::extend` doesn't exist (a similar replacement is `String::push_str`). So for each call to `escape_default` I think you need to collect the iterator into a new allocated string, and push that to the main string. There could be a better way that I'm missing, but I for now I'll just keep it as a `Vec` to avoid the extra allocations.

sporksmith: You can push a character at a time, but yeah it is a little awkward. Up to you.

e.g.
```
use std::convert::TryFrom;

pub fn main() {
    let mut s = String::new();

    let bytes = [0u8; 10];

    for b in bytes {
        for b in std::ascii::escape_default(b) {
            s.push(char::try_from(b as u32).unwrap())
        }
    }

    println!("{}", s);
}
```

stevenengler: Okay that seems good to me. Added (using `char::from(u8)` rather than `char::try_from(u32)`).

PR #2650, opened 2023-01-09
Process: move thread list to Rust

sporksmith:


src/main/host/thread.rs
```diff
@@ -259,6 +259,13 @@ impl TryFrom<libc::pid_t> for ThreadId {
     }
 }
 
+impl From<ProcessId> for ThreadId {
+    fn from(value: ProcessId) -> Self {
+        // A process ID is also a valid thread ID
+        ThreadId(value.into())
+    }
+}
```
stevenengler, line 267: If shadow's process ID and thread ID are to just be equivalent to Linux's pid, I think it would make sense to delete the `ThreadId` and just use `ProcessId`? I think being able to convert between `ThreadId` and `ProcessId` like this can be confusing.

src/main/host/process.c
```diff
@@ -496,10 +469,11 @@ static void _unref_thread_cb(gpointer data) {
...
 void process_addThread(Process* proc, Thread* thread) {
     MAGIC_ASSERT(proc);
-    g_hash_table_insert(proc->threads, GUINT_TO_POINTER(thread_getID(thread)), thread);
+    _process_insertThread(proc->rustProcess, thread);
 
-    // Schedule thread to start.
-    thread_ref(thread);
+    // Schedule thread to start. We're giving the caller's reference to thread
+    // to the TaskRef here, which is why we don't increment its ref count to
+    // create the TaskRef, but do decrement it on cleanup.
```
stevenengler, line 476: I don't see a change to the caller of `process_addThread`, so was this incorrect before? Or has something unrelated changed that affects the caller?

src/main/host/process.rs
```diff
@@ -1551,4 +1557,143 @@ mod export {
...
+        signo: i32,
+    ) {
+        let proc = unsafe { proc.as_ref().unwrap() };
+        let host_lock = unsafe { host_lock.as_mut().unwrap() };
+        Worker::with_active_host(|host| {
+            let proc = proc.borrow(host.root());
+            let signal = Signal::try_from(signo).unwrap();
+            let host_shmem = host.shim_shmem_lock_borrow().unwrap();
+            let mut threads = proc.threads.borrow_mut();
+            for (_tid, thread) in threads.iter_mut() {
```
stevenengler, line 1656: nit: Iterate over `threads.values_mut()` instead?

src/main/host/process.rs
```diff
@@ -1551,4 +1557,143 @@ mod export {
...
+    ) {
+        let proc = unsafe { proc.as_ref().unwrap() };
+        Worker::with_active_host(|host| {
+            let proc = proc.borrow(host.root());
+            let thread = unsafe { ThreadRef::new(thread) };
+            let tid = thread.id();
+            proc.threads.borrow_mut().insert(tid, thread);
+        })
+        .unwrap()
+    }
```
stevenengler, line 1620: Could use a comment explaining if it takes ownership of the thread.

src/main/host/process.rs
```diff
@@ -1551,4 +1557,143 @@ mod export {
...
+                        thread_shmem.protected.borrow_mut(&host_shmem.root);
+                    let blocked_signals = thread_shmem_protected.blocked_signals;
+                    if blocked_signals.has(signal) {
+                        continue;
+                    }
+                }
+                let Some(mut cond) = thread.syscall_condition_mut() else {
+                    continue;
+                };
+                cond.wakeup_for_signal(host_lock, signal);
```
stevenengler, line 1669: Maybe add a comment for what the condition is where there is no syscall condition? Shouldn't all running threads be blocked on some syscall?

stevenengler (approved):

src/main/host/process.c
sporksmith: `_process_insertThread` increments the reference count

src/main/host/process.rs
sporksmith: Done

src/main/host/process.rs
sporksmith: done

src/main/host/process.rs
sporksmith: Yeah, I *think* it shouldn't happen. I tried adding a `panic` here, and none of the tests triggered it. I'm not  quite confident enough to leave that in though, so added a comment and logging a warning instead.

src/main/host/thread.rs
sporksmith: I think it's useful to have two distinct types to avoid accidentally confusing them, but occasionally we need to convert between them.

PR #2655, opened 2023-01-11
Bump Rust to 1.66.1

sporksmith:
This is a point release fixing a security vulnerability. https://blog.rust-lang.org/2023/01/10/Rust-1.66.1.html


rust-toolchain.toml
```diff
@@ -5,4 +5,4 @@
 # This is the version that will be used in CI when building
 # shadow, running lints, etc.
 [toolchain]
-channel = "1.66.0"
+channel = "1.66.1"
```
stevenengler, line 8: While we're editing this file, maybe remove the above comment since this version is used for shadow in all contexts, not just the CI?

sporksmith: Done

PR #2656, opened 2023-01-11
Changed register-to-int conversions from `From` to `TryFrom`

stevenengler:
This PR changes the (for example) `impl From<SysCallReg> for u32 { ... }` implementations to `impl TryFrom<SysCallReg> for u32 { ... }`. Previously, the `From` implementations would just cast the register value to the corresponding type. This can be error-prone if we accidentally cast to a different type than what the Linux syscall actually accepts. Since when we write syscall handlers we tend to follow the libc API rather than the syscall API, we could accidentally have our syscall handler accept an `i32` argument if the libc wrapper takes an `i32`, but maybe the actual syscall takes a `i64` instead. This could also occur if we originally accepted a `i64`, but then we changed something else in the syscall handler which changed rust's type inference, which changed what type we cast the register to (like what `NoTypeInference` with `TypedPluginPointer` is supposed to protect against).

There are two cases to consider:

1. Shadow reads a larger-width value: If the application writes an `-1i32` to a syscall register, and Shadow's syscall handler does a `i64::from(reg)`, Shadow will get a value of `4_294_967_295i64`, not `-1i64` since it won't be sign extended.
2. Shadow reads a smaller-width value: If the application writes a `u32::MAX as u64 + 1` to a syscall register, and Shadow's syscall handler does a `u32::from(reg)`, Shadow will get a value of `0`, not `u32::MAX + 1`.

There's no way we can catch (1), but we can try to catch (2) by checking if there are non-zero bits outside of the range that we're casting to. For example in the `u32::MAX as u64 + 1` case, we could check if the bits outside of the `u32` range are non-zero.

@sporksmith What do you think about this? I have two concerns: (1) Maybe it's better to not check this in release mode for performance reasons, but this probably isn't a big deal. (2) Maybe we should leave this as only `From`, but show a warning instead of a `Result::Err`.

### DIFF
--- a/src/main/host/syscall_types.rs
+++ b/src/main/host/syscall_types.rs
@@ -89,57 +89,9 @@ impl PartialEq for SysCallReg {
 
 impl Eq for SysCallReg {}
 
-impl From<u64> for SysCallReg {
-    fn from(v: u64) -> Self {
-        Self { as_u64: v }
-    }
-}
-
 impl From<SysCallReg> for u64 {
     fn from(v: SysCallReg) -> u64 {
         unsafe { v.as_u64 }
-    }
-}
-
-impl From<u32> for SysCallReg {
-    fn from(v: u32) -> Self {
-        Self { as_u64: v as u64 }
-    }
-}
-
-impl From<SysCallReg> for u32 {
-    fn from(v: SysCallReg) -> u32 {
-        (unsafe { v.as_u64 }) as u32
-    }
-}
-
-impl From<usize> for SysCallReg {
-    fn from(v: usize) -> Self {
-        Self { as_u64: v as u64 }
-    }
-}
-
-impl From<SysCallReg> for usize {
-    fn from(v: SysCallReg) -> usize {
-        unsafe { v.as_u64 as usize }
-    }
-}
-
-impl From<isize> for SysCallReg {
-    fn from(v: isize) -> Self {
-        Self { as_i64: v as i64 }
-    }
-}
-
-impl From<SysCallReg> for isize {
-    fn from(v: SysCallReg) -> isize {
-        unsafe { v.as_i64 as isize }
-    }
-}
-
-impl From<i64> for SysCallReg {
-    fn from(v: i64) -> Self {
-        Self { as_i64: v }
     }
 }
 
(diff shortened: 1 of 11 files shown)


src/main/host/syscall_types.rs
```diff
@@ -89,51 +89,78 @@ impl PartialEq for SysCallReg {
...
-    }
+macro_rules! try_from_syscallreg {
+    ($to_type:ty, $using_field:ident) => {
+        impl TryFrom<SysCallReg> for $to_type {
+            //type Error = <Self as TryFrom<u64>>::Error;
+            type Error = nix::errno::Errno;
+
+            fn try_from(v: SysCallReg) -> Result<Self, Self::Error> {
+                let val = unsafe { v.$using_field };
+
```
sporksmith, line 112: I think this could just be something like `val.try_into().map_err(|_| Self::Error::EINVAL)`?

In particular I'm not sure what you have now will work properly with signed integers; I think e.g. a -1 passed through a 32-bit int at the library level might get sign extended into a 64-bit -1 with all 1-bits for the syscall.

Maybe worth some unit tests either way?

sporksmith (changes requested):

src/main/host/syscall_types.rs
stevenengler: glibc doesn't sign extend them before making the syscall, so this doesn't work for signed types. But I guess other libc implementations could sign extend them.

sporksmith: Hmm, yeah I can see how it ends up not getting sign extended, though it's not clear to me whether we can depend on it never being sign extended.

Maybe allow the high bits to be all 1's if the converted value is signed and negative?

sporksmith: In any case, worth documenting why we don't do the simpler thing here and what we do and don't know about the arguments being sign-extended :)

sporksmith (approved):

stevenengler:
We discussed this more, and it's not clear that this is worthwhile.

1. We merged #2664 which makes it less likely that type inference will cause us to cast the register value as the wrong type.
2. When there are non-zero bits outside of the range that we're casting to, there's no way to know if Shadow is reading the argument incorrectly, or if the application is sending the argument incorrectly. For example if the syscall is supposed to accept an `i64`, Shadow reads it as an `i32`, and the application sends an `i64`, it's a bug in Shadow and we should panic if there are non-zero bits outside of the `i32` range that we're casting to. But if the syscall is supposed to accept an `i32`, Shadow reads it as an `i32`, and the application sends an `i64`, then Shadow should truncate it to 32 bits and not return an errno nor panic (this is what Linux does).

I think it's better to leave the conversion code as-is.

@sporksmith Any thoughts about this?

PR #2657, opened 2023-01-11
Fix `sched_{get,set}affinity` bug

### LINKED ISSUE #2625: Recent regression in launching java under shadow
**Describe the issue**
[A previous change](https://github.com/shadow/shadow/issues/2094) allowed java to launch under shadow.  [This commit](https://github.com/shadow/shadow/commit/4ba2a0f8f7ee98feb001e62c3e77252a89310a4e) seems to have caused a regression where loading java now crashes with
> fatal error: pthread_getattr_np failed with error = 3 

Error 3 is `ESRCH` and it looks like that commit added it as an error code in some situations

**To Reproduce**
Same instructions as from the original issue which enabled java support
```
general:
  stop_time: 15s
network:
  graph:
    type: 1_gbit_switch
hosts:
  h1:
    network_node_id: 0
    ip_addr: "11.0.0.1"
    processes:
    - path: /usr/bin/java
      args: --version
```

Running in an ubuntu docker image on mac osx

> lsb_release -d
Description:	Ubuntu 20.04.5 LTS
uname -a
Linux 134296edcd2f 5.15.49-linuxkit #1 SMP Tue Sep 13 07:51:46 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
java --version
openjdk 16.0.1 2021-04-20
OpenJDK Runtime Environment (build 16.0.1+9-Ubuntu-120.04)
OpenJDK 64-Bit Server VM (build 16.0.1+9-Ubuntu-120.04, mixed mode, sharing)

>shadow --version
Shadow v2.3.0-83-g4ba2a0f8 2022-12-12--22:41:52

more specifically though it is broken in every commit after the one referenced above

I don't think I am using any

stevenengler:
Closes #2625.

### DIFF
--- a/src/main/host/host.rs
+++ b/src/main/host/host.rs
@@ -7,6 +7,7 @@ use crate::cshadow;
 use crate::host::descriptor::socket::abstract_unix_ns::AbstractUnixNamespace;
 use crate::host::network_interface::{NetworkInterface, PcapOptions};
 use crate::host::process::Process;
+use crate::host::thread::ThreadId;
 use crate::network::net_namespace::NetworkNamespace;
 use crate::network::router::Router;
 use crate::utility::{self, SyncSendPointer};
@@ -716,6 +717,18 @@ impl Host {
         };
     }
 
+    /// Returns `true` if the host has a process that contains the specified thread.
+    pub fn has_thread(&self, virtual_tid: ThreadId) -> bool {
+        for process in self.processes.borrow().values() {
+            let process = process.borrow(self.root());
+            if process.thread_borrow(virtual_tid).is_some() {
+                return true;
+            }
+        }
+
+        false
+    }
+
     /// Locks the Host's shared memory, caching the lock internally.
     ///
     /// Dropping the Host before calling [`Host::unlock_shmem`] will panic.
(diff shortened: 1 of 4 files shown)


src/main/host/host.rs
```diff
@@ -716,6 +717,29 @@ impl Host {
         };
     }
 
+    /// Returns the specified process.
+    pub fn get_process(
```
sporksmith, line 721: Nit - I  think it's more conventional in Rust not to have the `get_` prefix.

src/main/host/process.rs
```diff
@@ -517,6 +517,10 @@ impl Process {
         self.itimer_real.borrow_mut()
     }
 
+    pub fn get_thread(&self, virtual_tid: ThreadId) -> Option<impl Deref<Target = ThreadRef> + '_> {
```
sporksmith, line 520: Drop `get_` prefix?

sporksmith (approved):

src/main/host/host.rs
sporksmith: Actually it looks the current caller could use `process_borrow` instead?

src/main/host/process.rs
sporksmith: Actually, call this `thread_borrow`, since it borrows the thread list? (And can/will borrow a thread object in that list)

src/main/host/host.rs
stevenengler: I missed that one. Have removed `get_process`.

src/main/host/process.rs
stevenengler: Oops yeah that's better. Updated to `thread_borrow`.

PR #2662, opened 2023-01-13
Set 40s timeout for 'bind-shadow' test

stevenengler:
The 'bind-shadow' test has been failing in the CI recently for debug builds. In the last month we've moved more code from C to Rust, and this test has become faster in release builds and slower in debug builds.


stevenengler:
#2663

sporksmith:
Hah oops :)

PR #2667, opened 2023-01-13
Don't stall when already delivered packets are queued

robgjansen:
On slow connections, it is possible that retransmitted packets end up in the input queue multiple times. Before this change, Shadow would stall because it was looking for the `n`th packet to deliver to the plugin, but the next in-order packet in the queue was the `n-1`th packet.

### DIFF
--- a/src/main/host/descriptor/tcp.c
+++ b/src/main/host/descriptor/tcp.c
@@ -810,7 +810,11 @@ static void _tcp_bufferPacketOut(TCP* tcp, Packet* packet) {
 static void _tcp_bufferPacketIn(TCP* tcp, Packet* packet) {
     MAGIC_ASSERT(tcp);
 
-    if(!priorityqueue_find(tcp->unorderedInput, packet)) {
+    // Don't store old packets whose data we already gave to the plugin.
+    PacketTCPHeader* hdr = packet_getTCPHeader(packet);
+    bool already_received = hdr->sequence < tcp->receive.next;
+
+    if (!already_received && !priorityqueue_find(tcp->unorderedInput, packet)) {
         /* TCP wants in-order data */
         priorityqueue_push(tcp->unorderedInput, packet);
         packet_ref(packet);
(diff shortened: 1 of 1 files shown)


stevenengler (approved):
Looks good to me! This definitely looks like the bug could cause some issues. The only thing that looks a little worrying is the transfer time graph from the benchmark, but I'm not sure what could be causing that.

transfer_time_5242880 exit
time_to_last_byte_recv_1048576_timeseries

src/main/host/descriptor/tcp.c
```diff
@@ -1325,7 +1329,13 @@ static void _tcp_flush(TCP* tcp, const Host* host) {
...
-        if(header->sequence == tcp->receive.next) {
+        if (header->sequence < tcp->receive.next) {
+            // This is a (probably retransmitted) copy of a packet we already stored
+            // and delivered to the plugin.
+            trace("Removing packet %u with duplicate data", header->sequence);
+            priorityqueue_pop(tcp->unorderedInput);
+            packet_unref(packet);
+        } else if (header->sequence == tcp->receive.next) {
             /* move from the unordered buffer to user input buffer */
             gboolean fitInBuffer = legacysocket_addToInputBuffer(&(tcp->super), host, packet);
```
stevenengler, line 1340: aside: Below the code does a `packet_unref(packet)` before the `packet_getPayloadSize(packet)`. Maybe while you're editing this code, reorder these lines so that we don't use the packet after we unref it?

robgjansen:
@stevenengler note that the graphs you posted are for #2632, not for this PR. (The graphs you posted are from a branch that does not yet contain this fix.)

I highly doubt that this PR will negatively affect performance, but could run a quick tgen to double check.

src/main/host/descriptor/tcp.c
```diff
@@ -1325,7 +1329,13 @@ static void _tcp_flush(TCP* tcp, const Host* host) {
         PacketTCPHeader* header = packet_getTCPHeader(packet);
 
         _rswlog(tcp, "I just received packet %d\n", header->sequence);
-        if(header->sequence == tcp->receive.next) {
+        if (header->sequence < tcp->receive.next) {
+            // This is a (probably retransmitted) copy of a packet we already stored
+            // and delivered to the plugin.
+            trace("Removing packet %u with duplicate data", header->sequence);
+            priorityqueue_pop(tcp->unorderedInput);
+            packet_unref(packet);
```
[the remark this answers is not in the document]
robgjansen: Oh, yeah! Thanks for catching that!

src/main/host/descriptor/tcp.c
robgjansen: Will do :)

PR #2671, opened 2023-01-18
Support `InetSocket` in the tracker

stevenengler:
This adds support for `InetSocket` in the tracker, ~and reverts the change from #2626~.


stevenengler:
> Looks plausible. It might be nice to have some tracker tests, but no need to block this on it.

Yeah I agree, but I did just run a tgen test and nothing changed for the `LegacySocket` sockets, so at least I think we didn't break any of the existing code.

Issue #2673, opened 2023-01-19
Eventfd with epoll EPOLLET event doesn't match Linux


haxxpop:
**Describe the issue**
When using eventfd with epoll with EPOLLET flag set, the result of epoll_wait in the real Linux is different from the one in Shadow

**To Reproduce**
Run the following code in Linux and in Shadow and see the difference.
```c
#include <sys/epoll.h>
#include <sys/eventfd.h>
#include <stdio.h>
#include <stdlib.h>
#include <pthread.h>
#include <unistd.h>

#define handle_error(msg) \
           do { perror(msg); exit(EXIT_FAILURE); } while (0)

void *thread_fn(void *ptr);

int timeout1 = 2;
int timeout2 = 4;
int efd, epollfd;

void main() {
     pthread_t thread1, thread2;
     int iret1, iret2;
#define MAX_EVENTS 10
     struct epoll_event ev, events[MAX_EVENTS];
     int nfds;

     epollfd = epoll_create1(EPOLL_CLOEXEC);

     efd = eventfd(0, EFD_CLOEXEC | EFD_NONBLOCK);
     if (efd == -1)
         handle_error("eventfd");

     ev.events = EPOLLIN | EPOLLRDHUP | EPOLLET;
     ev.data.fd = efd;
     if (epoll_ctl(epollfd, EPOLL_CTL_ADD, efd, &ev) == -1)
         handle_error("epoll_ctl: efd");

     iret1 = pthread_create(&thread1, NULL, thread_fn, (void*) &timeout1);
     iret2 = pthread_create(&thread2, NULL, thread_fn, (void*) &timeout2);

     for (;;) {
         nfds = epoll_wait(epollfd, events, MAX_EVENTS, -1);
         if (nfds == -1)
             handle_error("epoll_wait");

         for (int n = 0; n < nfds; ++n) {
             if (events[n].data.fd == efd) {
                 uint64_t u;
                 printf("event found\n");
             }
         }
         fflush(stdout);
     }
     pthread_join(thread1, NULL);
     pthread_join(thread2, NULL); 

     printf("Thread 1 returns: %d\n",iret1);
     printf("Thread 2 returns: %d\n",iret2);
     exit(0);
}

void *thread_fn(void *ptr) {
    int timeout = *((int *)ptr);
    sleep(timeout);
    printf("%d\n", timeout);
    fflush(stdout);

    uint64_t inc = 10;
    write(efd, &inc, sizeof(uint64_t));
}
```
[...]

PR #2675, opened 2023-01-20
WIP Edge-triggered epoll

ppopth:
Since edge-triggered epoll events in Shadow don't work similarly to Linux as discussed in https://github.com/shadow/shadow/issues/2673, I fixed it in this PR.

You can see that the patch is quite short. However, I haven't implemented the tests for this yet since I don't have bandwidth to do so. If you are interested in applying this patch, you can continue the work. (I will come back and do the tests when I have bandwidth to do so)

This PR fixes the edge-triggered epoll only for legacy sockets and eventfd, so there will probably be other types of file descriptors that don't work similarly to Linux.


robgjansen:
We rewrote epoll in rust in #3125, so this PR is no longer useful, however we will work on the general issue as part of #2673. Thanks again for reporting!

PR #2687, opened 2023-01-23
Add support for Fedora 37 in the CI

stevenengler:


.github/workflows/run_tests_incremental.yml
```diff
@@ -46,6 +46,7 @@ jobs:
           - 'fedora:34'
           - 'fedora:35'
           - 'fedora:36'
+          - 'fedora:37'
```
sporksmith, line 49: `# EOL 15 Dec 2023 https://endoflife.date/fedora` ?

.github/workflows/run_tests.yml
```diff
@@ -39,6 +39,7 @@ jobs:
           - 'fedora:34'
           - 'fedora:35'
           - 'fedora:36'
+          - 'fedora:37'
```
sporksmith, line 42: `# EOL 15 Dec 2023 https://endoflife.date/fedora` ?

ci/container_scripts/install_extra_deps.sh
```diff
@@ -69,7 +69,7 @@ then
 fi
 
 # For debugging
-cargo --version
+rustup --version
```
sporksmith, line 72: In #2688 I fix it so that `cargo --version` does work, and updated the comment. I think we do want to force installing the toolchain here.

It probably *does* make sense to run `rustup --version` here for debugging/logging purposes too though. So I guess whenever we resolve the conflict between the two we should keep both.

sporksmith (commented):
LGTM, and seems like we ought to do it since the other current supported versions are EOL, but probably worth running by @robgjansen since it's a long term maintenance "promise"

.github/workflows/run_tests.yml
stevenengler: Added.

.github/workflows/run_tests_incremental.yml
stevenengler: Added.

robgjansen (approved):
Yeah, people do use Fedora so it's nice to test it.

PR #2688, opened 2023-01-23
ci scripts: use rust-toolchain.toml when installing Rust

sporksmith:
This was working when run in the github CI, since there's already a checkout of the shadow source in the current directory with rust-toolchain.toml. But when running locally with ci/run.sh, the `cargo --version` command failed since the version of rust to use was unspecified.


ci/container_scripts/install_extra_deps.sh
```diff
@@ -68,7 +68,12 @@ then
...
 fi
 
-# For debugging
+# This forces installation of the toolchain. When used with Docker,
+# this causes the rust installation get "baked in" to the Docker image layer
+# that runs this script, which is typically what we want.
+#
+# The version specified in the current version of rust-toolchain.toml will still
+# ultimately be respected, installing that toolchain on demand if necessary.
 cargo --version
```
stevenengler, line 77: Maybe add a comment in `rust-toolchain.toml` that the docker images need to be rebuilt each time we change that file?

stevenengler (approved):

ci/container_scripts/install_extra_deps.sh
sporksmith: Done

PR #2691, opened 2023-01-24
Add `DescriptorHandle` and return `Result` from descriptor table methods 

stevenengler:
The descriptor table shouldn't ever allow descriptors to be registered with a handle larger than `libc::c_int::MAX`. The old code also had a loop that could wrap around unexpectedly in release builds.

While the libc API uses `int` values for file handles, the Linux syscall API uses `unsigned int` for file handles. This means that when you pass a file handle of -1 to a `close()` call for example, Linux will see this as 4,294,967,295. If we read file handles in Shadow's syscall handlers as `libc::c_uint` values, we run into problems in these edge cases (for example with `dup2(fd, -1)`). We can continue reading them as signed integers and casting them to unsigned for now, but I think it's nicer to force us to handle these edge cases.

The public API to `DescriptorTable` now uses `DescriptorHandle` objects. The `register_descriptor` and `register_descriptor_with_min_fd` methods also now return results if the descriptor could not be added.

### DIFF
--- a/src/main/host/syscall/handler/eventfd.rs
+++ b/src/main/host/syscall/handler/eventfd.rs
@@ -78,10 +78,11 @@ impl SyscallHandler {
         let fd = ctx
             .process
             .descriptor_table_borrow_mut()
-            .register_descriptor(desc);
+            .register_descriptor(desc)
+            .or(Err(Errno::ENFILE))?;
 
         log::trace!("eventfd() returning fd {}", fd);
 
-        Ok(fd.into())
+        Ok(fd.val().into())
     }
 }
--- a/src/main/host/descriptor/descriptor_table.rs
+++ b/src/main/host/descriptor/descriptor_table.rs
@@ -5,9 +5,12 @@ use std::collections::{BTreeSet, HashMap};
 
 use log::*;
 
+/// POSIX requires fds to be assigned as `libc::c_int`, so we can't allow any fds larger than this.
+pub const FD_MAX: u32 = i32::MAX as u32;
+
 /// Map of file handles to file descriptors. Typically owned by a Process.
 pub struct DescriptorTable {
-    descriptors: HashMap<u32, Descriptor>,
+    descriptors: HashMap<DescriptorHandle, Descriptor>,
 
     // Indices less than `next_index` known to be available.
     available_indices: BTreeSet<u32>,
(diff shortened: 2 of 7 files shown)


src/main/host/syscall/handler/eventfd.rs
```diff
@@ -75,13 +76,20 @@ impl SyscallHandler {
...
-        let fd = ctx
+        let fd = match ctx
             .process
             .descriptor_table_borrow_mut()
-            .register_descriptor(desc);
+            .register_descriptor(desc)
+        {
+            Ok(x) => x,
+            Err(desc) => {
+                CallbackQueue::queue_and_run(|cb_queue| desc.close(ctx.host, cb_queue));
```
sporksmith, line 86: I forget, is there a reason we don't have Descriptors close themselves on Drop? Or maybe warn and/or debug_panic if they're dropped without closing?

sporksmith (approved):

src/main/host/syscall/handler/eventfd.rs
stevenengler: Good point. It used to be the case that you had to close them manually, and it would warn you if you didn't. But I forgot that I changed this. So a `Descriptor` holds an `OpenFile`, and the `OpenFile` will automatically close the ref-counted `File` if there are no more `OpenFile`s for that `File`. So it's actually fine to just drop the `Descriptor`, we'll just lose any errors returned by the file's close method, but we're ignoring those errors anyway.

https://github.com/shadow/shadow/blob/34fe9d83186fc084257cd3a50e681db754b28fd3/src/main/host/descriptor/mod.rs#L518-L521

I'll fix up the PR to remove the stuff about closing the descriptor.

stevenengler: Actually, this isn't entirely true. For `LegacyFile`s the `Descriptor` holds a `LegacyFileCounter`, which does not close the file when dropped ("This is similar to an `OpenFile` object, but does not close the file when dropped since we don't have a pointer to the host.").

https://github.com/shadow/shadow/blob/34fe9d83186fc084257cd3a50e681db754b28fd3/src/main/host/descriptor/mod.rs#L737-L743

We should be able to just get the `Host` from the worker and close it during the drop. I'll do this in a follow-up PR though.

sporksmith (approved):

src/main/host/syscall/handler/eventfd.rs
sporksmith: Sounds good. I guess that means we're potentially "leaking" the `LegacyFileCounter` if registering a `LegacyFile` fails, but we were already doing that before this PR as well

stevenengler: Yeah, we're not leaking them in the refcount sense, but they won't be cleaned up properly before the refcount reaches zero. I'll post a PR shortly for improving this.

PR #2703, opened 2023-01-27
Plan for interior mutability in Thread / ThreadRef

sporksmith:
At a high level, I think we're going to need interior mutability in Thread/ThreadRef for similar reasons as in Process and Host - it'd be difficult to avoid incompatible borrows of Thread itself without it.

More concretely, I ran into a problem when changing SysCallCondition to store a tid instead of a `Thread*`: some `SysCallCondition` methods need to act on the `Thread`. Since the condition itself is a member of the `Thread`, it'll be impossible to simultaneously hold a `&mut Thread` and a reference to the condition. Using a `&Thread` instead of a `&mut Thread` works around this.

I went a bit further here than strictly necessary and removed the `RootedRefCell` wrapper around `Process`'s `ThreadRef`s altogether. I think as with `Process` we'll end up making ~all methods take `&self` instead of `&mut self`, so will never need to get a `&mut` reference to the `ThreadRef`.

We *could* leave the `RootedRefCell` in place to make things easier in case we want to get `&mut` references later, but after working through the analagous problem in `Process` I think I'm leaning towards following the [yagni](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle here. Better to make the code simpler now and only support as much functionality as we know we need.

### DIFF
--- a/src/main/host/process.rs
+++ b/src/main/host/process.rs
@@ -694,7 +694,7 @@ impl Process {
         let mut host_shmem_protected = host.shim_shmem_lock_borrow_mut().unwrap();
         let threads = self.threads.borrow();
         for thread in threads.values() {
-            let mut thread = thread.borrow_mut(host.root());
+            let thread = thread.borrow_mut(host.root());
             {
                 let thread_shmem = thread.shmem();
                 let thread_shmem_protected = thread_shmem
--- a/src/main/host/process.rs
+++ b/src/main/host/process.rs
@@ -576,7 +576,6 @@ impl Process {
         // can drop the borrow of the thread list, and pass a borrowed mutable reference
         // to `thread` so that it doesn't need to be re-borrowed.
         let cthread = unsafe { thread.cthread() };
-        drop(thread);
         drop(threads);
 
         unsafe { cshadow::thread_resume(cthread) };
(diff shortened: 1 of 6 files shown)


sporksmith:
I think we could analogously remove the `RootedRefCell` around the `Process`s in the `Host`'s process list, but I think I'll work through migrating Thread with this approach before going back and messing with `Process`.

With the `RootedRefCell`s in both places gone, we *could* think about removing the `RootedRc` wrappers as well, but that'd make it so that we couldn't mutate the process and thread lists while holding a reference to a process or thread. There are only a couple places where we actually need to mutate those lists though, so we could probably work around that by scheduling an event to do the mutation, or maybe using a `CallbackQueue`.

sporksmith:
Oops, fixing clippy...

src/main/host/process.rs
```diff
@@ -694,7 +694,7 @@ impl Process {
         let mut host_shmem_protected = host.shim_shmem_lock_borrow_mut().unwrap();
         let threads = self.threads.borrow();
         for thread in threads.values() {
-            let mut thread = thread.borrow_mut(host.root());
+            let thread = thread.borrow_mut(host.root());
```
stevenengler: Should the be renamed to just `borrow()`?

stevenengler (approved):

src/main/host/process.rs
sporksmith: This line is removed in a later commit

PR #2706, opened 2023-01-29
Relay design improvements

robgjansen:


stevenengler (approved):

robgjansen:
TGen benchmark results:
https://github.com/shadow/benchmark-results/tree/master/tgen/2023-01-30-T21-14-49

stevenengler:
There does seem to have been a change to the retransmission overhead, but I don't see an obvious reason why this would be the case, so I think it's fine.

1675126861_grim

stevenengler:
Actually the previous benchmark run (the weekly) was run before #2690 was merged, so maybe that PR explains the difference shown in this graph.

PR #2707, opened 2023-01-29
Refill token buckets less often on slow connections

robgjansen:
The token bucket now returns the time that the next packet conforms, ie how long until we will have enough tokens to forward that packet. Returning this instead of how long until the next refill reduces the overall number of forwarding events we need to schedule, and is especially helpful on slow connections.

### DIFF
--- a/src/main/network/relay/token_bucket.rs
+++ b/src/main/network/relay/token_bucket.rs
@@ -40,6 +40,12 @@ impl TokenBucket {
         last_refill: EmulatedTime,
     ) -> Option<TokenBucket> {
         if capacity > 0 && refill_increment > 0 && !refill_interval.is_zero() {
+            log::trace!(
+                "Initializing token bucket with capacity {}, will refill {} tokens every {:?}",
+                capacity,
+                refill_increment,
+                refill_interval
+            );
             Some(TokenBucket {
                 capacity,
                 balance: capacity,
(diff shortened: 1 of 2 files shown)


robgjansen:
Tor benchmark results seem reasonable:
https://github.com/shadow/benchmark-results/tree/master/tor/2023-01-29-T15-04-36

src/main/network/relay/token_bucket.rs
```diff
@@ -72,10 +78,42 @@ impl TokenBucket {
...
+                num_refills
+            }
+        };
+
+        match num_required_refills {
+            0 => SimulationTime::ZERO,
+            1 => next_refill_span,
+            _ => next_refill_span.saturating_add(
+                self.refill_interval
+                    .saturating_mul(num_required_refills.saturating_sub(1)),
```
stevenengler, line 112: nit: maybe `num_required_refills.checked_sub(1).unwrap()`, since if this fails, then something else went wrong

stevenengler (approved):

stevenengler:
The simulation did seem to get a bit slower in the [tor](https://github.com/shadow/benchmark-results/tree/master/tor/2023-01-29-T15-04-36) benchmark, but I would have expected the opposite. Maybe worth running the simulation again to check if it was just within error since the weekly has large error bars. But I'm fine either way since I don't see anything wrong with this PR.

1675126861_grim

robgjansen:
Rebased and ran another Tor benchmark as suggested, and the performance is in line with the nightly from last night:

run_time

(We currently think the regression on the nightly was from #2632.)

PR #2708, opened 2023-01-31
Remove unused preload tests

stevenengler:
These tests aren't enabled and aren't relevant to Shadow's current design. @robgjansen Marking you as reviewer since these existed from pre-2020 and I don't know if there's anything here you want to keep.


robgjansen (approved):
These apply to Shadow v1 and I don't think are useful anymore.

PR #2711, opened 2023-02-01
Push local packets directly to netif, stop using events

robgjansen:
Local packets were previously "sent" to the netif by scheduling an event. This commit changes this behavior to instead just directly push local packets to the destination netif. This results in fewer events (it saves 1 event for every local packet) and should thus be more performant. It does change the order of operations a bit, so we expect non-deterministic results across this commit.


robgjansen:
TGen results are as expected. We don't see an performance improvement since the TGen benchmark is not sending packets over localhost, however the Tor benchmark does send localhost packets so we may see an improvement when that is next run.
https://github.com/shadow/benchmark-results/tree/master/tgen/2023-02-01-T19-13-56

stevenengler (approved):

PR #2715, opened 2023-02-02
Support Rust listeners on C `LegacyFile`s

stevenengler:
This should allow us to add rust state/status listener callback closures to `LegacyFile`s in the future. This means that we can write rust objects that add listeners to `LegacyFile`s. For example, we can write a rust epoll file object that listens for events on C `TCP` or `RegularFile` objects using rust closures.

### DIFF
--- a/src/main/host/descriptor/mod.rs
+++ b/src/main/host/descriptor/mod.rs
@@ -826,8 +826,11 @@ impl std::fmt::Debug for c::SysCallReturn {
 mod export {
     use super::*;
 
+    use shadow_shim_helper_rs::rootedcell::refcell::RootedRefCell;
+
     use crate::host::descriptor::socket::inet::tcp::TcpSocket;
     use crate::host::descriptor::socket::inet::InetSocket;
+    use crate::utility::legacy_callback_queue::RootedRefCell_StateEventSource;
 
     /// The new descriptor takes ownership of the reference to the legacy file and does not
     /// increment its ref count, but will decrement the ref count when this descriptor is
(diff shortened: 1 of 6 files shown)


stevenengler:
Runtime performance and memory usage seems to be slightly worse in a tor benchmark. I'm not sure why the memory usage appears worse. The simulation network results did not change.

https://github.com/shadow/benchmark-results/tree/master/tor/2023-02-07-T20-06-50

1675868563_grim
1675868575_grim
client_goodput exit

src/main/host/descriptor/mod.rs
```diff
@@ -1072,6 +1075,49 @@ mod export {
...
+        assert!(!event_source.is_null());
+        unsafe { Box::from_raw(event_source) };
+    }
+
+    #[no_mangle]
+    pub extern "C" fn eventsource_addLegacyListener(
+        event_source: *mut RootedRefCell_StateEventSource,
+        listener: *mut c::StatusListener,
+    ) {
+        let event_source = unsafe { event_source.as_mut() }.unwrap();
```
sporksmith, line 1106: IMO this should take a `*const RootedRefCell_StateEventSource`, and here we should use `as_ref` instead of `as_mut`. This helps ensure we don't accidentally create incompatible references.

src/main/host/descriptor/mod.rs
```diff
@@ -1072,6 +1075,49 @@ mod export {
...
+        })
+        .unwrap();
+    }
+
+    #[no_mangle]
+    pub extern "C" fn eventsource_removeLegacyListener(
+        event_source: *mut RootedRefCell_StateEventSource,
+        listener: *mut c::StatusListener,
+    ) {
+        let event_source = unsafe { event_source.as_mut() }.unwrap();
```
sporksmith, line 1120: Same here - consider avoiding an unchecked mutable reference here.

src/main/utility/legacy_callback_queue.rs
```diff
@@ -87,42 +35,55 @@ pub fn with_global_cb_queue<T>(f: impl FnOnce() -> T) -> T {
...
-    /// Returns `true` if successfully added to the queue, or `false` if there was no queue.
+    use crate::core::worker;
+
+    /// Notify listeners using the global callback queue. If the queue hasn't been set using
+    /// [`with_global_cb_queue`], the listeners will be notified here before returning.
     #[no_mangle]
-    pub unsafe extern "C" fn add_to_global_cb_queue(
-        listener: *mut c::StatusListener,
+    pub unsafe extern "C" fn notify_listeners_with_global_cb_queue(
+        event_source: *mut RootedRefCell_StateEventSource,
```
sporksmith, line 69: Same here - consider avoiding an unchecked mutable reference here.

src/main/host/descriptor/mod.rs
```diff
@@ -168,13 +168,13 @@ impl Drop for LegacyListener {
 
 /// Stores event listener handles so that `c::StatusListener` objects can subscribe to events.
 struct LegacyListenerHelper {
-    handle_map: std::collections::HashMap<usize, Handle<(FileState, FileState)>>,
+    handles: Vec<(usize, Handle<(FileState, FileState)>)>,
```
sporksmith, line 173: A comment here about what the `usize` represents would be helpful.

sporksmith (approved):

src/main/host/descriptor/mod.rs
stevenengler: Nice catch, fixed.

src/main/host/descriptor/mod.rs
stevenengler: Fixed.

src/main/utility/legacy_callback_queue.rs
stevenengler: Fixed.

src/main/host/descriptor/mod.rs
stevenengler: Added a comment.

PR #2722, opened 2023-02-06
Don't run workflows on merge to main

sporksmith:
We don't need to run our CI on commits to main since main is already a protected branch that only allows merging from a PR that is up to date with main and has already passed the CI.


stevenengler (approved):
I think we also want to remove the badge from:

https://github.com/shadow/shadow/blob/6064cc8a8eb4485d08a41587e9be1e2ba1a49397/README.md?plain=1#L3

PR #2724, opened 2023-02-06
CI: remove debug builds from the primary test matrix

sporksmith:
Testing both debug and release builds for every configuration is overkill. Testing a debug build for each compiler should be sufficient to catch most debug-only bugs; e.g. a debug assertion that fails.

Note that the same is not true of release builds - bugs arising from unsound/undefined behavior may only trigger on some platforms.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -42,7 +42,7 @@ jobs:
           # EOL May 2024 https://www.centos.org/centos-stream/
           - 'quay.io/centos/centos:stream8'
         cc: ['gcc']
-        buildtype: ['debug', 'release']
+        buildtype: ['release']
         include:
           # Run some tests on the newest-available base vm, and hence
           # newest-available kernel. https://github.com/actions/virtual-environments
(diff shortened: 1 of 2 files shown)


sporksmith:
@robgjansen This will need some of the debug builds removed from the required tests to be able to merge. I also modified the clang debug build.

So to merge this as-is, the set of "shadow" tests we'd want to require now is:

```
shadow (ubuntu-20.04, ubuntu:18.04, gcc, release)
shadow (ubuntu-20.04, ubuntu:20.04, gcc, release)
shadow (ubuntu-20.04, ubuntu:22.04, gcc, release)
shadow (ubuntu-20.04, debian:10-slim, gcc, release)
shadow (ubuntu-20.04, debian:11-slim, gcc, release)
shadow (ubuntu-20.04, fedora:34, gcc, release)
shadow (ubuntu-20.04, fedora:35, gcc, release)
shadow (ubuntu-20.04, fedora:36, gcc, release)
shadow (ubuntu-20.04, fedora:37, gcc, release)
shadow (ubuntu-20.04, quay.io/centos/centos:stream8, gcc, release)
shadow (ubuntu-22.04, ubuntu:22.04, gcc, release)
shadow (ubuntu-22.04, ubuntu:22.04, clang, release)
shadow (ubuntu-20.04, ubuntu:22.04, gcc, debug)
shadow (ubuntu-20.04, ubuntu:22.04, clang, debug)
```

robgjansen (approved):

PR #2725, opened 2023-02-06
Set shim's stdout/stderr to the shim log file

stevenengler:
Shadow now uses an fd passed through fork/exec for the shim log rather than passing a filename using `SHADOW_LOG_FILE`, and also uses this fd for stdout and stderr.

If the shim writes to stdout/stderr (or the plugin disables interposition and writes to stdout/stderr), it should be written to the shim log rather than to shadow's stdout/stderr (which may be the terminal or some other file).

@sporksmith If this overlaps with your `Thread` migration let me know and I'll hold off on merging.

### DIFF
--- a/src/lib/shim/shim.c
+++ b/src/lib/shim/shim.c
@@ -226,12 +226,13 @@ static void _shim_parent_init_logging() {
         logger_set_global_start_time_micros(logger_start_time);
     }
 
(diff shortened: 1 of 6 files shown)


sporksmith (approved):
It overlaps a bit, but that's ok

PR #2731, opened 2023-02-07
SysCallHandler: don't store pointers to Process and Thread

sporksmith:
These circular references otherwise somewhat complicate migrating Thread to Rust.

To facilitate getting rid of these, I added back `worker_getCurrentProcess` and `worker_getCurrentThread`. These are easier to implement safely now that they are wrapped in `RootedRc` objects, and we aren't trying to provide mutable access to the Rust objects. I tentatively still think it's better to prefer passing explicitly where possible, particularly in Rust code, but these are helpful in C code and for interop with C code during the migration.

### DIFF
--- a/src/main/core/worker.rs
+++ b/src/main/core/worker.rs
@@ -219,6 +219,10 @@ impl Worker {
         Worker::with(|w| w.active_process_info.borrow().as_ref().map(|p| p.id)).flatten()
     }
 
+    pub fn active_thread_id() -> Option<ThreadId> {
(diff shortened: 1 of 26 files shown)


sporksmith:
Queued a benchmark since this is somewhat hot-path https://github.com/shadow/benchmark/actions/runs/4118911692

stevenengler (approved):

sporksmith:
Benchmark results are a hair slower than the last nightly, but within the confidence interval of the weekly. https://github.com/shadow/benchmark-results/blob/master/tor/2023-02-08-T04-41-19/plots/run_time.png

PR #2739, opened 2023-02-09
Add RootedCell

sporksmith:
This is the Rooted equivalent of `std::cell::Cell`

### DIFF
--- a/src/lib/shadow-shim-helper-rs/src/rootedcell/cell.rs
+++ b/src/lib/shadow-shim-helper-rs/src/rootedcell/cell.rs
@@ -1,0 +1,151 @@
+use super::{Root, Tag};
+use std::cell::UnsafeCell;
+use vasi::VirtualAddressSpaceIndependent;
+
+/// Analagous to [std::cell::Cell]. In particular like [std::cell::Cell], it
+/// doesn't perform any atomic operations internally, making it relatively
+/// inexpensive.
+///
+/// Unlike [std::cell::Cell], this type is [Send] and [Sync] if `T` is
+/// [Send]. This is safe because the owner is required to prove access to the
+/// associated [Root], which is `![Sync]`, to access.
+#[derive(Debug)]
+#[repr(C)]
+pub struct RootedCell<T> {
+    tag: Tag,
+    val: UnsafeCell<T>,
+}
+
+// SAFETY: RootedCell is VirtualAddressSpaceIndependent as long as T is.
+unsafe impl<T> VirtualAddressSpaceIndependent for RootedCell<T> where
+    T: VirtualAddressSpaceIndependent
+{
+}
+
+impl<T> RootedCell<T> {
+    /// Create a RootedCell associated with `root`.
+    pub fn new(root: &Root, val: T) -> Self {
+        Self {
+            tag: root.tag(),
+            val: UnsafeCell::new(val),
+        }
+    }
+
+    pub fn get_mut(&mut self) -> &mut T {
+        // Since we have the only reference to `self`, we don't need to check the root.
+        unsafe { &mut *self.val.get() }
+    }
+
+    pub fn set(&self, root: &Root, val: T) {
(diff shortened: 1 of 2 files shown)


src/lib/shadow-shim-helper-rs/src/rootedcell/cell.rs
```diff
@@ -0,0 +1,163 @@
...
+        assert_eq!(
+            root.tag, self.tag,
+            "Expected {:?} Got {:?}",
+            self.tag, root.tag
+        );
+
+        let old = unsafe { *self.val.get() };
+        unsafe { *self.val.get() = val }
+        old
+    }
```
stevenengler, line 78: I didn't think about it too hard, but maybe double-check that it behaves correctly when the methods of `RootedCell` are called re-entrantly (I'll pretend that's a word). For example, do the `UnsafeCell` accesses behave correctly when `replace` is called while you're in the middle of executing `set`? For example, if `T` contains a reference to the outer `RootedCell` and `T` is dropped inside of `set`, the drop impl will be able to call methods on the `RootedCell` while it's in the middle of executing `set`. An example for `Cell` would be:

```rust
use std::cell::Cell;
use std::rc::{Rc, Weak};

struct Foo(i32, Option<Weak<Cell<Foo>>>);

impl std::ops::Drop for Foo {
    fn drop(&mut self) {
        println!("Droping id: {}", self.0);

        if let Some(weak) = &self.1 {
            if let Some(rc) = weak.upgrade() {
                // we're in the middle of set(), but we can call replace() here
                let _x = rc.replace(Foo(2, None));

                // here we drop Foo(1, _)
            }
        }
    }
}

fn main() {
    let x = Rc::new_cyclic(|weak| Cell::new(Foo(0, Some(Weak::clone(&weak)))));

    // will drop Foo(0, _) and Foo(1, _) during the set()
    x.set(Foo(1, None));

    // here we drop Foo(2, _)
}
```

src/lib/shadow-shim-helper-rs/src/rootedcell/cell.rs
```diff
@@ -0,0 +1,163 @@
...
+        // Prove that the root is held for this tag.
+        assert_eq!(
+            root.tag, self.tag,
+            "Expected {:?} Got {:?}",
+            self.tag, root.tag
+        );
+
+        let old = unsafe { *self.val.get() };
+        unsafe { *self.val.get() = val }
+        old
```
stevenengler, line 77: optional nit: Maybe consider using the pointer module functions instead, just because they're a little easier to reason about. For example sometimes dereferencing a pointer in rust actually causes it to create a reference. I wouldn't expect it to in this particular case, but I find dereferencing raw pointers to be confusing in rust because of it. For example `*x += 5;` where `x` is a pointer creates an implicit mutable reference to the integer (so it can be passed to the `AddAssign` trait), which is unexpected and may cause an aliasing issue if you already have another mutable reference to that integer.

The dereference-assignment here is fine I think, but it might be nicer to do a `std::ptr::replace` here anyways.

stevenengler (approved):

src/lib/shadow-shim-helper-rs/src/rootedcell/cell.rs
sporksmith: I don't see a problem with doing something analogous here. I considered adding this as a test just to be sure, but implementing `new_cyclic` looks a bit tricky. It also looks potentially useful and I think would simplify the Process and (soon Thread) constructors but I don't want to dive down that rabbit hole at the moment :). I could work around not having it here, but that gets to be a fairly complex unit test; I'm not sure it's worth it.

Anyway, I think it's clearer where things get dropped etc after the other changes. Wdyt?

stevenengler: Looks good to me!

stevenengler (approved):

PR #2742, opened 2023-02-10
Support `getsockopt()` and `setsockopt()` for tcp wrapper

stevenengler:
The logic for each socket option generally follows the logic from `src/main/host/syscall/socket.c`.


src/main/host/syscall/handler/mod.rs
```diff
@@ -214,6 +215,30 @@ pub fn read_sockaddr(
...
+    val: &T,
+    val_ptr: PluginPtr,
+    val_len: usize,
+) -> Result<usize, SyscallError> {
+    let val_len = std::cmp::min(val_len, std::mem::size_of_val(val));
+
+    let val = &pod::as_u8_slice(val)[..val_len];
+    let val_ptr = TypedPluginPtr::new::<MaybeUninit<u8>>(val_ptr, val_len);
+
+    mem.copy_to_ptr(val_ptr, val)?;
```
sporksmith, line 237: Hmm.... orthogonal to this PR, but I think we have a safety hazard in the `Pod` module. Since `MaybeUninit<T>` is declared as `Pod` where `T` is `Pod`, and we allow conversion from any `Pod` to `u8` slices, we have a path to go from uninitialized data to a regular `u8` slice without using `unsafe`.

I don't *think* we'll actually get unsound behavior in this specific instance since `copy_to_ptr` doesn't touch the bytes in userspace anyway.

src/main/host/descriptor/socket/inet/tcp.rs
```diff
@@ -406,6 +409,287 @@ impl LegacyTcpSocket {
...
+                Ok(bytes_written as libc::socklen_t)
+            }
+            (libc::SOL_SOCKET, libc::SO_ERROR) => {
+                // return error for failed connect() attempts
+                let conn_err = unsafe { c::tcp_getConnectionError(self.as_legacy_tcp()) };
+
+                let error = if conn_err == -libc::ECONNRESET || conn_err == -libc::ECONNREFUSED {
+                    // result is a positive errcode
+                    -conn_err
+                } else {
```
sporksmith, line 512: Why do we ignore other errors?

src/main/host/descriptor/socket/inet/tcp.rs
```diff
@@ -406,6 +409,287 @@ impl LegacyTcpSocket {
...
+                // Shadow's behaviour, but also prevents an application from setting this to
+                // something unnecessarily large like INT_MAX.
+                let val = std::cmp::min(val, 268435456); // 2^28 = 256 MiB
+
+                unsafe { c::legacysocket_setInputBufferSize(self.as_legacy_socket(), val) };
+                unsafe { c::tcp_disableReceiveBufferAutotuning(self.as_legacy_tcp()) };
+            }
+            (libc::SOL_SOCKET, libc::SO_REUSEADDR) => {
+                // TODO: implement this, tor and tgen use it
+                log::trace!("setsockopt SO_REUSEADDR not yet implemented");
```
sporksmith, line 675: Should these unsupported cases return errors? Or does that break our current simulations using tgen and tor?

sporksmith (approved):

src/main/host/descriptor/socket/inet/tcp.rs
stevenengler: No idea :) I'm just copying what the C code does:

https://github.com/shadow/shadow/blob/645a1788e09b1016e1f9bda4717d322bd4388d97/src/main/host/syscall/socket.c#L373-L386

src/main/host/descriptor/socket/inet/tcp.rs
stevenengler: It probably breaks tor or tgen, but I'm not sure. It has always been this way (from 2020):

https://github.com/shadow/shadow/blob/865c384824280c08cc8f8a76afef1285a72970c5/src/main/host/syscall/socket.c#L394-L405

src/main/host/syscall/handler/mod.rs
stevenengler: The `as_u8_slice` function actually returns a `&[MaybeUninit<u8>]`, which we don't cast to a `&[u8]` slice in the memory manager before passing to nix, so I think we're okay here?

PR #2745, opened 2023-02-13
Support `shutdown()` for tcp wrapper

stevenengler:
The logic for generally follows the logic from `src/main/host/syscall/socket.c`.

https://github.com/shadow/shadow/blob/8638dc4acd92b194215701c669259f23a29738b2/src/main/host/syscall/socket.c#L1179-L1211


stevenengler:
@sporksmith A ping just in case you missed this PR.

sporksmith (approved):

PR #2747, opened 2023-02-13
Rootscope

sporksmith:
Sketch of how we might implement the `Rooted*` methods with thread-local roots.

The basic idea for avoiding having to use closures everywhere is to store a `RefCell` of the root in the thread-local, and when replacing roots validate that there are no outstanding borrows.

We support multiple such roots by templating `RootScope` on a marker type. We currently would need this for our two "types" of roots - Host roots and Host shared memory roots - though I think we can eventually unify the two.

`ScopedRootedCell` is a thin wrapper around `RootedCell` that doesn't require a `root` in its APIs. It would be similarly simple to implement `ScopedRootedRc`. `ScopedRootedRefCell` is potentially a bit more boilerplate since the borrow objects would need to store the `Ref` to the thread-local-root, not just a `&Root`. (Though I think we might be able to templatize those types as well to avoid duplication)

I'm honestly not sure yet whether we want this, but wanted to see whether and how it would work out.

### DIFF
--- a/src/lib/shadow-shim-helper-rs/src/rootedcell/mod.rs
+++ b/src/lib/shadow-shim-helper-rs/src/rootedcell/mod.rs
@@ -2,6 +2,7 @@ use std::{
     cell::RefCell,
     marker::PhantomData,
     ops::Deref,
+    panic::UnwindSafe,
     sync::atomic::{AtomicU32, Ordering},
 };
 
@@ -134,10 +135,15 @@ impl<T> RootScope<T> {
             let outer = current.borrow();
             if let Some(root) = outer.as_ref() {
                 let ret = root.borrow();
+                type ReturnType = std::cell::Ref<'static, Root>;
+                static_assertions::assert_not_impl_all!(ReturnType: UnwindSafe);
                 // SAFETY:
                 // We only mutate the RefCell that Ref is borrowed from via `RefCell::take`,
                 // which validates at run-time that there are no borrows outstanding.
-                let static_ret: std::cell::Ref<'static, Root> = unsafe { std::mem::transmute(ret) };
+                //
+                // We validated above that the type we are using is `!UnwindSafe`, since
+                // otherwise `catch_unwind` could potentially circumvent this check.
+                let static_ret: ReturnType = unsafe { std::mem::transmute(ret) };
                 Some(static_ret)
             } else {
                 None
--- a/src/lib/shadow-shim-helper-rs/src/rootedcell/mod.rs
+++ b/src/lib/shadow-shim-helper-rs/src/rootedcell/mod.rs
@@ -1,5 +1,7 @@
 use std::{
+    cell::RefCell,
     marker::PhantomData,
+    ops::Deref,
     sync::atomic::{AtomicU32, Ordering},
 };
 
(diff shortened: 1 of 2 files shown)


src/lib/shadow-shim-helper-rs/src/rootedcell/mod.rs
```diff
@@ -104,3 +107,74 @@ impl Default for Root {
...
+                type ReturnType = std::cell::Ref<'static, Root>;
+                static_assertions::assert_not_impl_all!(ReturnType: UnwindSafe);
+                // SAFETY:
+                // We only mutate the RefCell that Ref is borrowed from via `RefCell::take`,
+                // which validates at run-time that there are no borrows outstanding.
+                //
+                // We validated above that the type we are using is `!UnwindSafe`, since
+                // otherwise `catch_unwind` could potentially circumvent this check.
+                let static_ret: ReturnType = unsafe { std::mem::transmute(ret) };
+                Some(static_ret)
```
stevenengler, line 147: I don't think setting `!UnwindSafe` really helps us here. For example, here's a very contrived example where we can move the `Ref` outside of the closure using `catch_unwind` and cause UB:

```rust
#[test]
fn foo() {
    thread_local! {
        static FOO: RefCell<Option<Box<dyn Deref<Target = Root> + 'static>>> = RefCell::new(None);
    }

    println!();

    let root = Root::new();
    let _ = std::panic::catch_unwind(|| {
        HostRootScope::with_current_set_to(root, || {
            FOO.with(|foo| {
                let root = HostRootScope::current().unwrap();
                foo.borrow_mut().replace(Box::new(root));
            });
            // FOO contains a Ref that points to the Root in CURRENT

            // will panic here
        });
    });

    // CURRENT still holds the Root because we panicked, so FOO still points to it
    let tag = FOO.with(|foo| {
        let foo = foo.borrow();
        let root = foo.as_ref().unwrap().deref().deref();
        root.tag()
    });

    let root = Root::new();
    HostRootScope::with_current_set_to(root, || {
        // CURRENT now holds a new Root
        FOO.with(|foo| {
            // when we read the Root from FOO, it's reading the new Root in CURRENT, which means
            // the memory was changed even though we hold an immutable reference to it
            let foo = foo.borrow();
            let root = foo.as_ref().unwrap().deref().deref();
            assert_eq!(
                tag,
                root.tag(),
                "We hold a `Ref<'_, Root>`, but the tag changed"
            );
        });
    });
}
```

stevenengler: One solution could be to "poison" the `CURRENT` thread local if `with_current_set_to` panics. So add a flag in `CURRENT`, and then have a struct with a drop handler that poisons `CURRENT` by setting the flag, and if everything completes successfully, then `mem::forget` the struct.

There's still a question about if you have the `Root` in one thread local and a `Ref<'static, Root>` in another thread local, the thread local containing the `Ref` might get dropped before the `Root`. So maybe you could surround `CURRENT` with a `ManuallyDrop` so that it's never dropped unless you do it explicitly.

sporksmith:
I'll go ahead and kill this; I think it's not worth the complexity.

PR #2750, opened 2023-02-14
Rooted*: make public APIs inline

sporksmith:
By default, Rust doesn't inline calls across crate boundaries. Since these functions are small, called frequently, and from across a crate boundary, we probably want them to be inline. We force this by marking them `#[inline]`, which is also done in the analogous functions in `std`.

A quick local benchmark using the tor minimal test shows there might be a measurable improvement here, though it's inconclusive since the confidence intervals overlap substantially:

Before:

```
Benchmark 1: ./setup test --extra tor-minimal
  Time (mean ± σ):     17.424 s ±  0.455 s    [User: 18.446 s, System: 6.024 s]
  Range (min … max):   16.818 s … 18.123 s    10 runs
```

After:

```
Benchmark 1: ./setup test --extra tor-minimal
  Time (mean ± σ):     17.119 s ±  0.482 s    [User: 18.211 s, System: 5.993 s]
  Range (min … max):   16.534 s … 18.020 s    10 runs
```

### DIFF
--- a/src/lib/shadow-shim-helper-rs/src/rootedcell/cell.rs
+++ b/src/lib/shadow-shim-helper-rs/src/rootedcell/cell.rs
@@ -24,6 +24,7 @@ unsafe impl<T> VirtualAddressSpaceIndependent for RootedCell<T> where
 
 impl<T> RootedCell<T> {
     /// Create a RootedCell associated with `root`.
+    #[inline]
     pub fn new(root: &Root, val: T) -> Self {
         Self {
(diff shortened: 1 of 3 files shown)


sporksmith:
Started a larger tor benchmark, mostly out of curiosity. I'm fairly confident this should be performance-neutral at worst. https://github.com/shadow/benchmark/actions/runs/4176543974

stevenengler (approved):

PR #2751, opened 2023-02-16
Add SendPointer

sporksmith:
This is analogous to `SyncSendPointer`, but only asserts that the pointee is `Send`, not that it is necessarily `Sync`.

This also changes  `SyncSendPointer<T>` and `HostTreePointer<T>` to unconditionally implement `Copy` and `Clone`, even when `T` does not.

### DIFF
--- a/src/main/utility/mod.rs
+++ b/src/main/utility/mod.rs
@@ -35,8 +35,18 @@ use shadow_shim_helper_rs::HostId;
 
 /// A type that allows us to make a pointer Send + Sync since there is no way
 /// to add these traits to the pointer itself.
-#[derive(Clone, Copy, Debug)]
+#[derive(Debug)]
 pub struct SyncSendPointer<T>(*mut T);
+
+// We can't automatically `derive` Copy and Clone without unnecessarily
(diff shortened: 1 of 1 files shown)


src/main/utility/mod.rs
```diff
@@ -35,17 +35,56 @@ use shadow_shim_helper_rs::HostId;
 
 /// A type that allows us to make a pointer Send + Sync since there is no way
 /// to add these traits to the pointer itself.
-#[derive(Clone, Copy, Debug)]
+#[derive(Debug)]
 pub struct SyncSendPointer<T>(*mut T);
 
+// T to be Copy and Clone. https://github.com/rust-lang/rust/issues/26925
+impl<T> Copy for SyncSendPointer<T> {}
+impl<T> Clone for SyncSendPointer<T> {
```
stevenengler, line 43: Is the comment cut-off here?

stevenengler (approved):

src/main/utility/mod.rs
sporksmith: Fixed

PR #2758, opened 2023-02-22
Add additional epoll tests

stevenengler:
Test the race-condition where Linux avoids spurious wakeups for `epoll_wait` if the epoll status changes before the threads are woken.

I wasn't sure if level-triggered `epoll_wait()` was guaranteed to be woken for all threads (it's not), so I tried it and figured it probably doesn't hurt to add a test for it.


src/test/epoll/test_epoll.rs
```diff
@@ -116,6 +124,112 @@ fn test_threads_level() -> anyhow::Result<()> {
...
+
+    let mut results = threads.map(|t| t.join().unwrap());
+
+    // One of the threads should have gotten an event, but we don't know which one.
+    // Sort results by number of events received.
+    results.sort_by(|lhs, rhs| lhs.events.len().cmp(&rhs.events.len()));
+
+    // One thread should have timed out with no events received.
+    ensure_ord!(results[0].epoll_res, ==, Ok(0));
+    ensure_ord!(results[0].duration, >=, timeout);
```
sporksmith, line 172: I'm confused how this passes without https://github.com/shadow/shadow/pull/2759 . It'd be nice if there's a small tweak or variation we could do to these tests to act as a regression test for that bugfix.

src/test/epoll/CMakeLists.txt
```diff
@@ -8,5 +8,5 @@ add_shadow_tests(BASENAME epoll)
 add_executable(test-epoll-writeable test_epoll_writeable.c)
 add_shadow_tests(BASENAME epoll-writeable LOGLEVEL debug)
 
-add_linux_tests(BASENAME epoll-rs COMMAND "../../target/debug/test_epoll")
+add_linux_tests(BASENAME epoll-rs COMMAND sh -c "../../target/debug/test_epoll --libc-passing")
```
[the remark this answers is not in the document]
stevenengler: I was sure I tried this and it didn't work, but I tried again and it seems to work, so I'll update it.

src/test/epoll/test_epoll.rs
stevenengler: I think it works because in this case the process is restarted early, the syscall condition is cancelled, and the `epoll_wait` syscall handler runs again, but sees that it should re-block. A new syscall condition is then created with a new timer. I'm not sure what happens to the old timer event though, maybe there's now two of them in the event queue?

I'm not sure of how we could write a regression test for #2759 though.

PR #2759, opened 2023-02-22
Fix backwards timeout condition

stevenengler:
The timeout condition in the `SysCallCondition` is backwards, causing the process to wake up before its timeout has expired. (And maybe not wake up at all when it does exceed the timeout?)

(This doesn't fix our tor onion service stream timeout issue.)


sporksmith (approved):

sporksmith:
I'm surprised this didn't cause any tests to fail. I suppose we need specifically a blocking syscall with a deadline with a condition that becomes true, but is falsified before the event runs...

stevenengler:
The tor benchmark networking results change slightly, but not significantly. The runtime and memory results do not change.

https://github.com/shadow/benchmark-results/tree/master/tor/2023-02-22-T15-06-28

Issue #2658, opened 2023-01-12
Improving conversions between `SysCallReg` and integers/pointers


stevenengler:
When making a system call, the plugin writes a value of type A to a "register", and shadow's syscall handler reads a value of type B from the register. If types A and B don't match (and if they don't have the same width), shadow's syscall handler may use the wrong value. We currently have no way of detecting these mistakes, and there may not be any good way of detecting them. We should try to detect these mistakes if we can, and we should also try to prevent them.

There are a few cases to consider:

1. Shadow reads a larger-width value: If the application writes an `-1i32` to a syscall register, and Shadow's syscall handler does a `i64::from(reg)`, Shadow may get a value of `4_294_967_295i64`, not `-1i64` since the value may not be sign extended to 64 bits. This is the case for glibc, but there is no guarentee that other libc implementations (or applications that don't use libc) haven't sign extended the value.
2. Shadow reads a smaller-width value: If the application writes a `u32::MAX as u64 + 1` to a syscall register, and Shadow's syscall handler does a `u32::from(reg)`, Shadow will get a value of `0`, not `u32::MAX + 1`.
3. Shadow reads a value with a different signedness: If the application writes a `u32` but Shadow's syscall handler reads it as an `i32`.
4. Shadow reads from a pointer of a different type: If the application writes a `*mut u32` pointer to a syscall register, and Shadow's syscall handler writes to it as a `*mut i64` pointer.

Some examples where these types of mistakes may happen:

1. When we write syscall handlers we tend to follow the libc API rather than the [syscall API](https://chromium.googlesource.com/chromiumos/docs/+/HEAD/constants/syscalls.md#x86_64-64_bit), so we could accidentally read a `SysCallReg` as the wrong type if the libc call doesn't match the syscall API.
2. If we originally cast the register value to some type, but then we changed something else in Shadow which changed rust's type inference, which changed what type we cast the register value to. For example, if we originally had:
    ```rust
    fn get_descriptor(i32) -> Option<Descriptor> { ... }
    fn syscall_write(ctx: &mut ThreadContext, args: &SysCallArgs) -> SyscallResult {
        let desc = get_descriptor(args.get(0).into());
        [...]
    }
    ```
[...]

PR #2762, opened 2023-02-23
Support explicit return types in the syscall handlers

stevenengler:
Similar to #2664 but for return types.

Syscalls can return various types (`int`, `size_t`, pointers, etc). We probably don't want to return the wrong type, which the plugin may interpret incorrectly. For example if a syscall is supposed to return an `int`, we probably don't want to return a `u64::MAX` from our syscall handler.

Currently our syscall handlers return a `SysCallReg`, which is a 64-bit union of various types, and there is no type checking to make sure our syscall handler is returning a valid type. This PR allows our syscall handlers to specify an explicit return type (ex: `libc::c_int` instead of `SysCallReg`).

Only the "sched.h" syscall handlers have been updated to use this. I didn't bother with the others. We can update them later if we want to, but for now this change allows us to use this for new syscall handlers that we write in the future.

(Related #2658.)

Before:

```rust
pub fn sched_yield(_ctx: &mut SyscallContext) -> SyscallResult {
    Ok(0.into())
}
```

After:

```rust
pub fn sched_yield(_ctx: &mut SyscallContext) -> Result<libc::c_int, SyscallError> {
    Ok(0)
}
```

### DIFF
--- a/src/lib/syscall-logger/src/lib.rs
+++ b/src/lib/syscall-logger/src/lib.rs
@@ -14,12 +14,12 @@ use quote::ToTokens;
 /// ```compile_fail
 /// # use syscall_logger::log_syscall;
 /// # use shadow_rs::host::syscall::handler::SyscallContext;
-/// # use shadow_rs::host::syscall_types::{SysCallArgs, SyscallResult};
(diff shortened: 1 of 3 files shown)


sporksmith (approved):
Cool!

PR #2765, opened 2023-02-24
Convert most syscall helper types to Rust

sporksmith:
In particular this is needed so that we can move the types that the shim also needs to access (e.g. SysCallArg) to the shadow-shim-helper-rs crate.

This reconfigures our cbindgen a bit so that most Rust items, and in particular struct definitions, are exported via `bindings-opaque.h`, which doesn't depend on any of the C shadow headers. This resolves a circular inclusion issue if we were to try exporting `SysCallArg` etc via `bindings.h` instead.

Migrating types used in unions that bindgen exports appears to cause bindgen to conservatively generate its own special union types instead of native Rust unions. This is resolved by migrating more types than I might have otherwise in this PR, including the union definitions themselves.


src/lib/shadow-shim-helper-rs/src/syscall_types.rs
```diff
@@ -0,0 +1,307 @@
...
+    pub fn number(&self) -> i64 {
+        self.number
+    }
+}
+
+impl PartialEq for SysCallReg {
+    fn eq(&self, other: &Self) -> bool {
+        unsafe { self.as_u64 == other.as_u64 }
+    }
+}
```
stevenengler, line 107: nit: move this down to after where `SysCallReg` is defined?

src/lib/shadow-shim-helper-rs/src/syscall_types.rs
```diff
@@ -0,0 +1,307 @@
...
+
+/// A register used for input/output in a syscall.
+#[derive(Copy, Clone, Eq)]
+#[repr(C)]
+pub union SysCallReg {
+    pub as_i64: i64,
+    pub as_u64: u64,
+    pub as_ptr: PluginPtr,
+}
```
stevenengler, line 116: do these fields need to be public if we have `From` implementations for them?

src/lib/shadow-shim-helper-rs/src/syscall_types.rs
```diff
@@ -0,0 +1,307 @@
+/// Represents a pointer to a virtual address in plugin memory.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[repr(C)]
+pub struct PluginPtr {
+    val: usize,
+}
```
stevenengler, line 6: Maybe add a comment here that says `SysCallArgs` requires this to be `repr(C)` since `transmute::<u64, PluginPtr>()` must be valid, so that we don't accidentally remove the `repr(C)` once we no longer access it from C.

src/lib/shadow-shim-helper-rs/src/syscall_types.rs
stevenengler, line 116: optional: maybe static_assert the size and alignment of `SysCallReg` as a sanity check to make sure we never accidentally break this

stevenengler (approved):

src/lib/shadow-shim-helper-rs/src/syscall_types.rs
sporksmith: Added static assertions and comments near the `SysCallArgs` definition - I think it makes more sense there.

src/lib/shadow-shim-helper-rs/src/syscall_types.rs
sporksmith: done

src/lib/shadow-shim-helper-rs/src/syscall_types.rs
sporksmith: It was for backwards compatibility with c::SysCallReg, but there were only a couple places that needed to be updated to make these non-pub. Done.

sporksmith: Done

src/lib/shadow-shim-helper-rs/src/syscall_types.rs
```diff
@@ -249,3 +249,61 @@ impl std::fmt::Debug for SysCallReg {
             .finish()
     }
 }
+
+// implement conversions from `SysCallReg`
+
+impl TryFrom<SysCallReg> for nix::fcntl::OFlag {
+    type Error = ();
```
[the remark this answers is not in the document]
sporksmith: > I think this change wouldn't be needed if TryFromSyscallReg were made non-pub or pub-crate?

No, I don't think that helps (and tried it to be sure). 

I think removing the blanket trait implemntation would fix it:

```
impl<T: TryFrom<SysCallReg>> TryFromSyscallReg for T {
    fn try_from_reg(reg: SysCallReg) -> Option<Self> {
        Self::try_from(reg).ok()
    }
}
```

but then we'd need to explicitly implement `TryFromSyscallReg` for the types that were covered by that before. That wouldn't be too bad, but OTOH I'm not sure it's worth the extra boilerplate vs getting rid of the indirection and using `TryFrom` directly. I can't think of cases where we'd want a `TryFromSyscallReg<T>` and specifically *not* want a `TryFrom<T>` that does the same conversion, but maybe I'm not thinking of something?

src/main/build.rs
```diff
@@ -88,11 +102,21 @@ fn run_cbindgen(build_common: &ShadowBuildCommon) {
     cbindgen::Builder::new()
         .with_crate(crate_dir)
         .with_config(cbindgen::Config {
+            // We want to avoid including any C headers from this crate here,
```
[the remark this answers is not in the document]
sporksmith: done

src/main/build.rs
```diff
@@ -88,11 +102,21 @@ fn run_cbindgen(build_common: &ShadowBuildCommon) {
...
+                // exported in the other header file, and need the C header files.
+                item_types: base_config
+                    .export
+                    .item_types
+                    .iter()
+                    .cloned()
+                    .filter(|t| *t != cbindgen::ItemType::Functions)
+                    .collect(),
                 ..base_config.export.clone()
             },
```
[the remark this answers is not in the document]
sporksmith: No, we want the struct definitions for e.g. `SysCallReg` to be exported here. It's ok to depend on C headers and their struct definitions, just not shadow headers, since the latter leads to circular include dependencies.

Maybe it's confusing that I kept the `-opaque.h` suffix, but it wasn't immediately obvious what else to call these.

src/lib/shadow-shim-helper-rs/src/syscall_types.rs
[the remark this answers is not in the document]
stevenengler: Okay my thinking was that the compiler error would be because we had a blanket implementation, but the trait was also public so another crate like `shadow-shim-helper` could also provide its own conflicting implementation. And if we made the trait non-pub, then other crates couldn't provide their own implementation of the trait and there couldn't be any conflict with the blanket implementation, so the compiler would be happy. But it sounds like this is wrong.

The original reason for this `TryFromSyscallReg` was to be able to convert from `SysCallReg` to other types without polluting the actual `TryFrom<SysCallReg>` trait with these conversions that are only needed by the strace logger. For example it isn't obvious if `impl TryFrom<SysCallReg> for nix::fcntl::OFlag` is good to have generally everywhere in Shadow since it implicitly converts `SysCallReg` to an `i32` (or maybe a `u32`, I'm not sure), and at the time I wanted to narrow the scope of this to just the strace logger.

But anyways I'm good with any change, including how it is now.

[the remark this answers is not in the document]
sporksmith: Ok - just added a comment in `TryFromSyscallReg` for now

PR #2768, opened 2023-02-28
Support `accept()` for tcp wrapper

stevenengler:
The logic generally follows the logic from `src/main/host/syscall/socket.c`.

https://github.com/shadow/shadow/blob/a8f49014f18d6da428f3548a914a7bdbbb88e9c2/src/main/host/syscall/socket.c#L156-L239

Since the TCP code has already (incorrectly) added the socket to the descriptor table, we need to remove it from the table so that we can re-add it later in the `accept()` syscall handler.

### DIFF
--- a/src/main/host/descriptor/socket/inet/tcp.rs
+++ b/src/main/host/descriptor/socket/inet/tcp.rs
@@ -11,7 +11,7 @@ use crate::cshadow as c;
 use crate::host::descriptor::socket::inet::{self, InetSocket};
 use crate::host::descriptor::socket::Socket;
 use crate::host::descriptor::{
-    File, FileMode, FileState, FileStatus, OpenFile, StateListenerFilter, SyscallResult,
+    CompatFile, File, FileMode, FileState, FileStatus, OpenFile, StateListenerFilter, SyscallResult,
 };
 use crate::host::host::Host;
 use crate::host::memory_manager::MemoryManager;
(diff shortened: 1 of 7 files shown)


src/main/host/descriptor/socket/inet/tcp.rs
```diff
@@ -614,11 +614,83 @@ impl LegacyTcpSocket {
...
+    pub fn accept(&mut self, _cb_queue: &mut CallbackQueue) -> Result<OpenFile, SyscallError> {
+        let is_valid_listener = unsafe { c::tcp_isValidListener(self.as_legacy_tcp()) } == 1;
+
+        // we must be listening in order to accept
+        if !is_valid_listener {
+            log::debug!("Socket is not listening");
+            return Err(Errno::EINVAL.into());
+        }
+
+        let mut peer_addr: libc::sockaddr_in = unsafe { std::mem::zeroed() };
```
sporksmith, line 626: optional - `pod::zeroed()` (without `unsafe`)

src/main/host/descriptor/socket/inet/tcp.rs
```diff
@@ -614,11 +614,83 @@ impl LegacyTcpSocket {
...
+        peer_addr.sin_family = libc::AF_INET as u16;
+        let mut accepted_fd = -1;
+
+        // now we can check if we have anything to accept
+        let errcode = Worker::with_active_host(|host| unsafe {
```
sporksmith, line 631: It's probably reasonable to pass the host and process in to this function explicitly? (and maybe just the whole syscall context object?)

sporksmith (approved):

src/main/host/descriptor/socket/inet/tcp.rs
stevenengler: Added.

src/main/host/descriptor/socket/inet/tcp.rs
stevenengler: I want to avoid this for now since ideally the socket objects shouldn't know anything about the host or process, and future rust versions shouldn't need to. The host and process are only needed for the C TCP socket, so for now I'd rather have them use the globals rather than pass the host/process in as arguments for all the socket types.

PR #2772, opened 2023-02-28
VirtualAddressSpaceIndpendent: use trait bounds in derive macro

sporksmith:
Extends the VirtualAddressSpaceIndependent Derive macro to handle generics, and replaces manual implementation of the trait by using the derive macro.

(Previously:)

> For discussion purposes - I'm a bit stumped how to get this to work. In particular I don't see a way to write the `static_assertions` for generics.
> 
> The current incarnation fails in a doc test that tries to use the macro with a generic type.
> 
> The test:
> ```
> /// #[derive(VirtualAddressSpaceIndependent)]
> /// struct MyWrapper<T> {
> ///   val: T,
> /// }
> ```
> 
> The failure:
> ```
>  --> src/lib.rs:120:8
>   |
> 6 | struct MyWrapper<T> {
>   |                  - type parameter from outer function
> 7 |   val: T,
>   |        ^ use of generic parameter from outer function
> ```
> 
> The problem is we end up generating something like this, and `T` isn't allowed in this position:
> 
> ```
> impl MyWrapper<T> where T: VirtualAddressSpaceIndependent {
>   fn _vasi_assertions() {
>     assert_impl_all!(T: vasi::VirtualAddressSpaceIndependent);
>   }
> }
> ```

### DIFF
--- a/src/lib/vasi-macro/src/lib.rs
+++ b/src/lib/vasi-macro/src/lib.rs
@@ -1,6 +1,6 @@
 use proc_macro2::TokenStream;
 use quote::quote;
-use syn::Field;
+use syn::{parse_quote, Attribute, GenericParam, Generics, Type};
 
 /// Implement `vasi::VirtualAddressSpaceIndependent` for the annotated type.
 /// Requires all fields to implement `vasi::VirtualAddressSpaceIndependent`.
@@ -110,8 +110,51 @@ use syn::Field;
 /// }
 /// ```
 ///
-/// TODO: Extend to support trait bounds. See e.g.
-/// <https://github.com/dtolnay/syn/blob/master/examples/heapsize/heapsize_derive/src/lib.rs#L16>
+/// A generic type *conditionally* implements VirtualAddressSpaceIndependent,
+/// if its type parameters do (as the derive macros in the std crate behave).
+/// ```
+/// use vasi::VirtualAddressSpaceIndependent;
+///
+/// #[derive(VirtualAddressSpaceIndependent)]
+/// struct MyWrapper<T> {
+///   val: T,
+/// }
+///
+/// static_assertions::assert_impl_all!(MyWrapper<i32>: vasi::VirtualAddressSpaceIndependent);
+/// static_assertions::assert_not_impl_all!(MyWrapper<* const i32>: vasi::VirtualAddressSpaceIndependent);
+/// ```
+///
+/// Generic type with existing bounds are also supported.
+/// ```
+/// use vasi::VirtualAddressSpaceIndependent;
+///
+/// #[derive(VirtualAddressSpaceIndependent)]
+/// struct MyWrapper<T: Copy> {
+///   val: T,
+/// }
+/// static_assertions::assert_impl_all!(MyWrapper<i32>: vasi::VirtualAddressSpaceIndependent);
+/// static_assertions::assert_not_impl_all!(MyWrapper<* const i32>: vasi::VirtualAddressSpaceIndependent);
+///
+/// #[derive(VirtualAddressSpaceIndependent)]
+/// struct MyWrapper2<T> where T: Copy {
+///   val: T,
+/// }
+/// static_assertions::assert_impl_all!(MyWrapper2<i32>: vasi::VirtualAddressSpaceIndependent);
+/// static_assertions::assert_not_impl_all!(MyWrapper2<* const i32>: vasi::VirtualAddressSpaceIndependent);
+///
+/// ```
+///
+/// As with e.g. Copy and Clone, a field that is dependent on a type parameter
+/// but still isn't VirtualAddressSpaceIndependent will cause the macro not to
+/// compile:
+/// ```compile_fail
+/// use vasi::VirtualAddressSpaceIndependent;
+///
+/// #[derive(VirtualAddressSpaceIndependent)]
+/// struct MyWrapper<T> {
+///   val: *const T,
+/// }
+/// ```
 #[proc_macro_derive(
     VirtualAddressSpaceIndependent,
     attributes(unsafe_assume_virtual_address_space_independent)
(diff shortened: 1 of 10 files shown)


sporksmith:
Oops, was supposed to be a Draft. It looks like there's no way to make a PR into a Draft once it's non-Draft?

sporksmith:
An alternative approach I'm tinkering with is to add a method to the Vasi trait, and access the fields by name in the body instead of trying to name the (possibly generic) types of the fields.

It looks like this could actually work but requires additional complexity to deal with unnamed fields, enum variants, union  fields (which need unsafe access), ...

Anyway the idea is to change the vasi trait declaration to:

```
pub unsafe trait VirtualAddressSpaceIndependent {
    /// Used by the derive macro to validate that fields are Vasi.
    fn is_vasi(&self) {}
}
```

and then in the macro, for `MyWrapper` above generate something like:

```
impl <T> VirtualAddressSpaceIndependent for MyWrapper<T> where T: VirtualAddressSpaceIndependent {
  fn is_vasi(&self) {
    VirtualAddressSpaceIndependent::is_vasi(&self.val);
  }
}
```

sporksmith:
@stevenengler if you get a chance to look at this, maybe you have a better idea how to do this?

stevenengler:
> Oops, was supposed to be a Draft. It looks like there's no way to make a PR into a Draft once it's non-Draft?

There is a button, but they don't make it obvious.

1677625743_grim

stevenengler:
I'm still not sure I understand completely, but could you do something more like the `Copy` derive where it's a compile-time failure if a non-generic field is not `Copy`, but then is derived conditionally for generic fields?

So something like:

```rust
#[derive(VirtualAddressSpaceIndependent)]
struct MyWrapper<T> {
    foo: Option<T>,
    bar: u32,
}
```

becomes:

```rust
struct MyWrapper<T> {
    foo: Option<T>,
    bar: u32,
}

assert_impl_all!(u32: vasi::VirtualAddressSpaceIndependent);

unsafe impl<T> VirtualAddressSpaceIndependent for MyWrapper<T> where
    Option<T>: VirtualAddressSpaceIndependent,
{}
```

sporksmith:
My understanding is that the derive macro for `Copy` is conditional on the generic type parameters being `Copy`, not the field types being `Copy`

According to https://stegosaurusdormant.com/understanding-derive-clone/#do-we-have-to-inspect-the-fields this is because making the constraints on the field types can leak private types into the public interface, which results in compilation errors. Maybe we could live with that limitation, though.

stevenengler:
Ah okay, I wondered then what happens if `T` were `Copy` but `Wrapper<T>` were `!Copy`. It seems like there's maybe some compiler magic that doesn't ever let this happen, so a better example is probably `Clone`. And your link shows why that would fail (the field access in the derived `clone()` method will try to call `clone()` on the field, but it not compile since `Wrapper<T>` doesn't always implement `Clone`).

stevenengler:
Here's another idea to avoid requiring a method in the trait. It still needs to access each field though:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=712fb56f0c303a52111503c4d0f34484

stevenengler:
Or maybe one of these so that you don't need to access each field:

https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=637fe36e09a2129005b96910a0767e02
https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=29f7895c524035eccf77ff7117e02b27

sporksmith:
I think we're miscommunicating. My understanding (and I think this matches what you're saying) is that e.g. for:

```
#[derive(Copy, Clone)]
struct MyStruct<T> {
    _x: T,
}
```

`MyStruct` is conditionally `Copy` and `Clone`. e.g. `MyStruct<Box<T>>` can be constructed, but won't be `Copy` since `Box` is never `Copy`.

OTOH for a struct definition is 
```
#[derive(Copy, Clone)]
struct MyStruct<T> {
    _x: Box<T>,
}
```

then there will be a compilation failure since `_x` isn't `Copy` even if `T` is.

i.e. this is different from your earlier example that would emit `where Option<T>: VirtualAddressSpaceIndependent,`; in particular that approach has the problem of exposing the field types in the public interface, and failing to compile if those types are private.

Anyways, I think I've gotten something working, and the ideas in your gist were definitely helpful - particularly using a trait const instead of a trait method seems a little nicer, and facilitates ensuring the code is never called at runtime (and probably dropped from the final binary).

It's currently checking the field values, though I may revisit your idea that uses the types instead. I'll take another look at that and clean up a bit before marking it for review.

sporksmith:
I think this is ready now! Thanks for the help and ideas!

src/lib/vasi-macro/src/lib.rs
```diff
@@ -126,50 +149,97 @@ pub fn derive_virtual_address_space_independent(
...
-        #assertions
-        /// SAFETY: All fields are vasi::VirtualAddressSpaceIndependent.
-        unsafe impl vasi::VirtualAddressSpaceIndependent for #name {}
+        unsafe impl #impl_generics vasi::VirtualAddressSpaceIndependent for #name #ty_generics #where_clause {
+            const IGNORE: () = {
+                let _ = || {
+                    fn check<T: vasi::VirtualAddressSpaceIndependent>() {}
+                    #calls_to_check
+                };
+            };
```
stevenengler, line 241: I don't think the closure is needed.

src/lib/vasi-macro/src/lib.rs
```diff
@@ -110,8 +110,31 @@ use syn::Field;
...
-/// <https://github.com/dtolnay/syn/blob/master/examples/heapsize/heapsize_derive/src/lib.rs#L16>
+/// A generic type *conditionally* implements VirtualAddressSpaceIndependent,
+/// if its type parameters do (as the derive macros in the std crate behave).
+/// ```
+/// use vasi::VirtualAddressSpaceIndependent;
+///
+/// #[derive(VirtualAddressSpaceIndependent)]
+/// struct MyWrapper<T> {
+///   val: T,
+/// }
```
stevenengler, line 121: Does it work if there are bounds on the generic types? Ex: `struct MyWrapper<T: Debug>` or `struct MyWrapper<T> where T: Debug`. (Not that we need to support that, but maybe leave as a TODO if it doesn't.)

src/lib/vasi-macro/src/lib.rs
```diff
@@ -126,50 +149,97 @@ pub fn derive_virtual_address_space_independent(
...
+    // This allows generic types to be conditionally VirtualAddressSpaceIndependent,
+    // iff their type parameters are.
+    let generics = add_trait_bounds(ast.generics);
+    let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
+
     let gen = quote! {
-        #assertions
-        /// SAFETY: All fields are vasi::VirtualAddressSpaceIndependent.
-        unsafe impl vasi::VirtualAddressSpaceIndependent for #name {}
+        unsafe impl #impl_generics vasi::VirtualAddressSpaceIndependent for #name #ty_generics #where_clause {
```
stevenengler, line 235: nit: maybe use `::vasi::VirtualAddressSpaceIndependent` instead so you're always using the `vasi` crate, and not something named `vasi` the user may have defined in their own code.

stevenengler (approved):
Cool!

src/lib/vasi-macro/src/lib.rs
sporksmith: Done

src/lib/vasi-macro/src/lib.rs
sporksmith: Done

src/lib/vasi-macro/src/lib.rs
sporksmith: Yup these are supported - added another doc test.

PR #2777, opened 2023-03-02
Unsafe scmutex

sporksmith:
This is an alternative to #2775 that removes the "reading" and "writing" states of `SelfContainedChannel`. Hopefully that will improve performance. The downside is that `send` and `receive` become `unsafe`, since we will have undefined behavior in the case of parallel calls to `send` or to `receive`.


sporksmith:
This gets back a little bit of performance, but still a little worse than nightly:

```
$ tornettools plot nightly/tornet-0.05-1/ 2023-03-01-T18-40-39/tornet-0.05-1/ 2023-03-03-T14-26-23/tornet-0.05-1 -l nightly rust-scmutex rust-scmutex-unsafe
```

image

sporksmith:
I think the improvement here is an illusion. It doesn't show up in [microbenchmarks](https://github.com/shadow/shadow/pull/2791), and the current version of IPC using SelfContainedMutex now appears to have the same performance as the C version. https://github.com/shadow/benchmark-results/blob/master/tor/2023-03-16-T13-26-54/plots/run_time.png

PR #2778, opened 2023-03-03
Separate sleeper atomic

sporksmith:
Another potential optimization to benchmark on top of #2777


sporksmith:
Started benchmark: https://github.com/shadow/benchmark/actions/runs/4327589703

sporksmith:
I think the improvement here is an illusion. It doesn't show up in [microbenchmarks](https://github.com/shadow/shadow/pull/2791), and the current version of IPC using SelfContainedMutex now appears to have the same performance as the C version. https://github.com/shadow/benchmark-results/blob/master/tor/2023-03-16-T13-26-54/plots/run_time.png

PR #2787, opened 2023-03-14
Scchannel align

sporksmith:
Try putting atomics on different cache lines.


sporksmith:
This doesn't appear to affect performance.

PR #2800, opened 2023-03-22
Use posix_spawn instead of vfork + exec

sporksmith:
Bare vfork + exec are tricky to use, since it's easy to accidentally corrupt the parent process's state from the child process. It's also unsupported in Rust (https://github.com/rust-lang/libc/pull/1574, https://github.com/rust-lang/libc/issues/1596).

I think we could work around this in Rust by using inline assembly or a C helper function to wrap the fork and exec, but `posix_spawn` basically does that for us already.

### DIFF
--- a/src/main/host/managed_thread.c
+++ b/src/main/host/managed_thread.c
@@ -6,6 +6,7 @@
 #include <inttypes.h>
 #include <sched.h>
 #include <search.h>
+#include <spawn.h>
 #include <string.h>
 #include <sys/prctl.h>
 #include <sys/syscall.h>
@@ -129,12 +130,13 @@ static gchar** _add_u64_to_env(gchar** envp, const char* var, uint64_t x) {
     return envp;
 }
 
-static pid_t _managedthread_fork_exec(ManagedThread* thread, const char* file,
-                                      const char* const* argv_in, const char* const* envp_in,
-                                      const char* workingDir, int straceFd, int shimlogFd) {
-    utility_debugAssert(file != NULL);
+static pid_t _managedthread_spawn(ManagedThread* thread, const char* file,
+                                  const char* const* argv_in, const char* const* envp_in,
+                                  const char* workingDir, int straceFd, int shimlogFd) {
+    // We use `posix_spawn` to spawn the child process. This should be functionally
+    // equivalent to vfork + execve, but wrapped in a safer and more portable API.
 
-    // execve technically takes arrays of pointers to *mutable* char.
+    // posix_spawn technically takes arrays of pointers to *mutable* char.
     // conservatively dup here.
     gchar** argv = g_strdupv((gchar**)argv_in);
     gchar** envp = g_strdupv((gchar**)envp_in);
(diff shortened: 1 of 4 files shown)


sporksmith:
A tor benchmark run shows no performance difference, as expected: https://github.com/shadow/benchmark-results/tree/master/tor/2023-03-22-T22-09-20

sporksmith:
Given that this is a relatively risky change, I'll hold off on merging until after the upcoming stable release.

src/main/host/managed_thread.c
```diff
@@ -129,72 +130,101 @@ static gchar** _add_u64_to_env(gchar** envp, const char* var, uint64_t x) {
...
+    // number of syscalls, so it might be better to continue using this slightly
+    // more awkward method anyway.
+    // https://github.com/bminor/glibc/commit/805334b26c7e6e83557234f2008497c72176a6cd
+    // https://austingroupbugs.net/view.php?id=411
+    if (posix_spawn_file_actions_adddup2(&file_actions, pipefd[1], STDOUT_FILENO) != 0) {
+        utility_panic("posix_spawn_file_actions_adddup2: %s", g_strerror(errno));
+    }
+    if (posix_spawn_file_actions_adddup2(&file_actions, STDOUT_FILENO, pipefd[1]) != 0) {
+        utility_panic("posix_spawn_file_actions_adddup2: %s", g_strerror(errno));
+    }
```
stevenengler, line 185: Maybe add a comment that it's fine to use `STDOUT_FILENO` as a temporary file handle since we're going to overwrite it later with the shim log handle. (When I first looked at this, I was worried that something writing to stdout in the shim would write to the pipe, potentially confusing Shadow's "process alive" checker thing, but that's not an issue since posix_spawn shouldn't write to stdout, and we set `STDOUT_FILENO` with the shimlog fd before the exec anyways.)

stevenengler (approved):
Cool! Hopefully this makes debugging easier as well.

src/main/host/managed_thread.c
```diff
@@ -146,55 +148,44 @@ static pid_t _managedthread_fork_exec(ManagedThread* thread, const char* file,
...
 
-            // clear the FD_CLOEXEC flag for the strace fd so that it's available in the shim
-            if (straceFd >= 0 && fcntl(straceFd, F_SETFD, 0)) {
-                die_after_vfork();
-            }
+    // Dup the write end of the pipe; the dup'd descriptor won't have O_CLOEXEC set.
+    if (posix_spawn_file_actions_adddup2(&file_actions, pipefd[1], pipefd[1]) != 0) {
+        utility_panic("posix_spawn_file_actions_adddup2: %s", g_strerror(errno));
+    }
```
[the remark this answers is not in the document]
stevenengler: Nvm I see this was fixed in a later commit.

src/main/host/managed_thread.c
sporksmith: It's in the block comment above; I broke it out into a separate paragraph for better visibility.

PR #2805, opened 2023-03-24
Support `sendmsg()` and `recvmsg()` for tcp wrapper

stevenengler:
We don't limit the send/recv buffer size using `SYSCALL_IO_BUFSIZE = 10 MiB` in this new version. I don't think this is needed with the current memory manager.

I duplicated some packet-related C functions to support passing the memory manager through, but this should hopefully be cleaned up once UDP sockets are migrated to rust.

The logic generally follows the logic from `src/main/host/syscall/socket.c`.

https://github.com/shadow/shadow/blob/ec2f888138ef9d466092d027441ad8fedc6ffbc5/src/main/host/syscall/socket.c#L407-L654

### DIFF
--- a/src/main/host/descriptor/tcp.c
+++ b/src/main/host/descriptor/tcp.c
@@ -904,15 +904,15 @@ static Packet* _tcp_createPacketWithoutPayload(TCP* tcp, const Host* host,
     return packet;
 }
 
-static Packet* _tcp_createDataPacket(TCP* tcp, const Thread* thread, enum ProtocolTCPFlags flags,
-                                     PluginVirtualPtr payload, gsize payloadLength) {
+static Packet* _tcp_createDataPacket(TCP* tcp, const Host* host, enum ProtocolTCPFlags flags,
+                                     PluginVirtualPtr payload, gsize payloadLength,
+                                     const MemoryManager* mem) {
     MAGIC_ASSERT(tcp);
 
-    const Host* host = thread_getHost(thread);
     bool isEmpty = payloadLength == 0;
     Packet* packet = _tcp_createPacketWithoutPayload(tcp, host, flags, isEmpty);
     if (!isEmpty) {
-        packet_setPayload(packet, thread, payload, payloadLength);
+        packet_setPayloadWithMemoryManager(packet, host, payload, payloadLength, mem);
     }
     return packet;
 }
@@ -2353,9 +2353,8 @@ static void _tcp_endOfFileSignalled(TCP* tcp, enum TCPFlags flags) {
 }
 
 /* Address and port must be in network byte order. */
-static gssize _tcp_sendUserData(LegacySocket* socket, const Thread* thread, PluginVirtualPtr buffer,
-                                gsize nBytes, in_addr_t ip, in_port_t port) {
-    TCP* tcp = _tcp_fromLegacyFile((LegacyFile*)socket);
+gssize tcp_sendUserData(TCP* tcp, const Host* host, PluginVirtualPtr buffer, gsize nBytes,
+                        in_addr_t ip, in_port_t port, const MemoryManager* mem) {
     MAGIC_ASSERT(tcp);
 
     /* return 0 to signal close, if necessary */
@@ -2379,13 +2378,26 @@ static gssize _tcp_sendUserData(LegacySocket* socket, const Thread* thread, Plug
     gsize maxPacketLength = CONFIG_TCP_MAX_SEGMENT_SIZE;
     gsize bytesCopied = 0;
 
+    /* Need non-NULL buffer. */
+    /* FIXME: should push this check to the point the data is actually read, to correctly handle
+     * non-NULL pointers that aren't accessible. This is currently in the Payload code; need to
+     * bubble up errors from there. If we do bubble up from the payload code, we also need to undo
+     * the TCP state changes made earlier, for example the sequence number increment in the
+     * _tcp_createPacketWithoutPayload code.
+     */
+    if (buffer.val == 0) {
+        return -EFAULT;
+    }
+
     /* create as many packets as needed */
     while(remaining > 0) {
         gsize copyLength = MIN(maxPacketLength, remaining);
 
         /* use helper to create the packet */
-        Packet* packet = _tcp_createDataPacket(
-            tcp, thread, PTCP_ACK, (PluginVirtualPtr){.val = buffer.val + bytesCopied}, copyLength);
+        Packet* packet = _tcp_createDataPacket(tcp, host, PTCP_ACK,
+                                               (PluginVirtualPtr){.val = buffer.val + bytesCopied},
+                                               copyLength, mem);
+
         if(copyLength > 0) {
             /* we are sending more user data */
             tcp->send.end++;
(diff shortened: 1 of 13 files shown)


src/main/host/descriptor/tcp.c
```diff
@@ -2379,13 +2378,19 @@ static gssize _tcp_sendUserData(LegacySocket* socket, const Thread* thread, Plug
     gsize maxPacketLength = CONFIG_TCP_MAX_SEGMENT_SIZE;
     gsize bytesCopied = 0;
 
+    if (buffer.val == 0) {
```
sporksmith, line 2381: Shouldn't need to explicitly check for NULL; `MemoryManager` will return an `EFAULT` in the usual case that 0x0 isn't accessible. I think it's technically possible to map a page at address 0, in which case an explicit NULL check would incorrectly return an error. NULL might also be ok if `nBytes` is 0.

As a general programming rule it's less duplicative and more reliable to try an operation and see if it succeeds than to try to figure out if an operation will succeed before trying it.

src/main/routing/packet.c
```diff
@@ -128,10 +130,27 @@ void packet_setPayload(Packet* packet, const Thread* thread, PluginVirtualPtr pa
...
+    utility_alwaysAssert(packet->payload != NULL);
     /* application data needs a priority ordering for FIFO onto the wire */
     packet->priority = host_getNextPacketPriority(thread_getHost(thread));
 }
 
+/* This is a copy of `packet_setPayload` but passes the memory manager through. Once we've moved UDP
+ * sockets to rust, we can remove `packet_setPayload` and rename this function to
+ * `packet_setPayload`. */
+void packet_setPayloadWithMemoryManager(Packet* packet, const Host* host, PluginVirtualPtr payload,
+                                        gsize payloadLength, MemoryManager* mem) {
```
sporksmith, line 142: I think `MemoryManager*` can be `const` here and the other functions that are only reading memory?

I'm also a little nervous about handing mutable pointers to Rust objects to C code in general, since it's a bit difficult to be confident that we won't break Rust aliasing rules. Instead of passing the `MemoryManager` around maybe we could pass the `const ProcessRefCell*` around and use the existing `process_readPtr` and `process_writePtr` APIs?

sporksmith (changes requested):

src/main/host/descriptor/tcp.c
stevenengler: I think I'd rather leave it the same as the original C code. I tried bubbling up the error from the memory manager, but then you also need to undo all of the TCP state changes that happen when creating the packet object. For example creating the packet object sometimes increases the TCP object's sequence number. So if you fail to set the packet's payload, then you need to decrease the sequence number back. But there are only certain conditions where the sequence number is increased, and so you need to copy that same logic to decide whether you need to decrease the sequence number. Since this code is fragile, I'd rather leave the control flow to be the same. This `buffer.val == 0` used to be in the C syscall handler, but I moved it here to the TCP code so that it wouldn't be used for all other socket types as well.

stevenengler: Right, previously Shadow would segfault if you give an invalid non-NULL plugin pointer. With this PR it will panic instead. I'll add a comment, but I'm not sure it's worth fixing for the C TCP code.

sporksmith: Ok, sounds good :)

src/main/routing/packet.c
sporksmith: > We call into the C code from LegacyTcpSocket::recvmsg(), which takes a &mut MemoryManager as an argument, so using process_writePtr would cause a double mutable borrow. 

Ah I see.

> Aside from signals and multi-threading, there wouldn't be any way to break aliasing rules.

I think another issue is Rust not enforcing borrowing rules when a pointer is propagated through the callstack. So e.g. we might have:

```
fn rust_fn(mem: &mut MemoryManager, ptr: TypedPluginPtr<u8>) {
  let first_borrow = mem.writer(ptr);
  unsafe { c_fn(mem as *mut _)};
```

and

```
void c_fn(MemoryManager* mem) {
  memorymanager_writePtr(mem, ...);
}
```

then `memorymanager_writePtr`  would construct a `&mut MemoryManager` when it's already borrowed further up the call stack.

I suppose the root cause of this bug is passing a mutable pointer to `MemoryManager` to `c_fn` inside an `unsafe` block while there is already a live reference to `MemoryManager`, knowing that the pointer may be used to create another, incompatible, reference. Maybe we can be careful to avoid such situations at those points.

Having worked through it, I think I'm leaning towards how you currently have it, but up to you.

sporksmith (approved):

src/main/routing/packet.c
sporksmith: If we stick with the current approach, maybe the new `memorymanager_` functions should be marked `unsafe`, with a `safety` comment that there must not exist (incompatible) borrows/references to the pointers being passed?

I know I haven't been super rigorous on this in general, and have been leaning on trying to only pass `const` pointers, though I suppose a similar condition exists even then that the caller must not hold a mutable borrow/reference.

sporksmith: > If we stick with the current approach, maybe the new memorymanager_ functions should be marked unsafe, with a safety comment that there must not exist (incompatible) borrows/references to the pointers being passed?
>
> I know I haven't been super rigorous on this in general, and have been leaning on trying to only pass const pointers, though I suppose a similar condition exists even then that the caller must not hold a mutable borrow/reference.

Err, I guess that doesn't directly address the example I gave either; the safety comment would belong on `c_fn`, which would be e.g. `tcp_sendUserData`. It's a little weird to have Rust safety comments on C code; I guess this ispart of the implicit safety conditions when calling a foreign C function with pointers to Rust objects...

sporksmith: Maybe it's not so bad - the conversion to a pointer itself requires borrowing the corresponding reference to the original object. It's possible to mess this up, but not trivially. https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=09dec6cb6b490c7ac04b7cdb38ed81b6

```
fn takes_ptr(p: *mut i32) {
    unsafe { *p += 1}
}

pub fn main() {
    let mut x = 42;
    
    /* borrow checker prevents this
    let borrow = &mut x;
    takes_ptr(&mut x as *mut _);
    println!("{borrow}");
    println!("{x}");
    */
    
    // Doesn't catch this unsoundness, but at this point we're really going out of our way
    // to fool the checker
    let ptr = &mut x as *mut _;
    let borrow = &mut x;
    takes_ptr(ptr);
    println!("{borrow}");
    println!("{x}");
}
```

stevenengler: > Maybe it's not so bad - the conversion to a pointer itself requires borrowing the corresponding reference to the original object. It's possible to mess this up, but not trivially. https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=09dec6cb6b490c7ac04b7cdb38ed81b6

Yeah, I think if you only convert a reference to a pointer at the C function callsite then you shouldn't run into issues (unless the C function itself does something that would break aliasing or lifetime rules). You don't need to even explicitly cast to a pointer in this case since rust will do it implicitly for you (for example `let borrow = &mut x; takes_ptr(borrow);`), maybe because it's always safe to do?

> Having worked through it, I think I'm leaning towards how you currently have it, but up to you.

Okay I think I'll leave it as-is, but I'll add some comments.

PR #2806, opened 2023-03-24
Unregister threads with ChildPidWatcher

sporksmith:
We were failing to save the ChildPidWatcher notification handles, which
in turn meant we weren't unregistering threads with the ChildPidWatcher
after they exited, thus creating a resource leak. The relevant data was
still getting freed when the whole Process exited, so it would have only
affected Processes that had many threads start and end over their
lifetime.

The callback registered for each thread is only about 16 bytes - a
function pointer and a data pointer.

Additionally this fixes a bug causing a panic when we do unregister a callback for a process that's already been unregistered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 A list of user-facing changes since the latest Shadow release.
 
-* item
+* Fixed a memory leak of about 16 bytes per thread due to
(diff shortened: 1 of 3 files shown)


sporksmith:
The ChildPidWatcher could use some more cleanup, but just doing the minimal fix for now to fix the resource leak.

stevenengler (approved):

PR #2808, opened 2023-03-27
Convert SysCallResult to Rust and rename to SyscallResult

sporksmith:


stevenengler (approved):
Do we construct a `SyscallReturn` from C at all? I don't see any, but just want to make sure we're not doing a `SyscallReturn ret = { 0 };` or something that would cause issues for rust.

sporksmith:
> Do we construct a SyscallReturn from C at all? I don't see any, but just want to make sure we're not doing a SyscallReturn ret = { 0 }; or something that would cause issues for rust.

Yeah, I don't think so. I created the helper functions for safely creating the union variants and switched all the constructions over to that a while ago.

I think technically `SyscallReturn ret = {0}` would currently be ok, since the enum tags start at `0`, but would certainly be better to avoid.

PR #2814, opened 2023-03-29
ChildPidWatcher: fix a race condition when unregistering a pid

sporksmith:
We were getting rare failures in CI unit tests, and consistent failures in large benchmarks. I was able to reproduce with:

```
for i in `seq 1000`; do
  echo trial $i
  if ! RUST_BACKTRACE=full cargo test --manifest-path=src/Cargo.toml -p shadow-rs childpid_watcher ; then
    break
  fi
done
```

I think what was happening is that we could get an epoll event for a dying process in the worker thread just as we were in the middle of unregistering it from another thread.

The simplest fix is to just let the worker thread handle unregistration so that it's serialized with other operations in that thread.

### DIFF
--- a/src/main/utility/childpid_watcher.rs
+++ b/src/main/utility/childpid_watcher.rs
@@ -27,6 +27,7 @@ pub type WatchHandle = u64;
 #[derive(Debug)]
 enum Command {
     RunCallbacks(Pid),
+    UnregisterPid(Pid),
     Finish,
 }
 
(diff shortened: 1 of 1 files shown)


sporksmith:
Actually I'm going to take a look at a more thorough revision moving all the "work" into the worker thread ...

sporksmith:
Though maybe it makes sense to get this in now since it's causing the regular CI to be flaky in addition to breaking the runner

stevenengler (approved):

PR #2815, opened 2023-03-29
Improve the strace formatting of `msghdr`

stevenengler:
Old:

```
00:00:04.857000000 [tid 1000] sendmsg(-1, 0x7fffffffd548, (empty)) = -9 (EBADF: Bad file number)
```

New:

```
00:00:04.857000000 [tid 1000] sendmsg(-1, msghdr { msg_name: 127.0.0.1:11111 (0x7fffffffdd24), msg_namelen: 16, msg_iov: 0x7fffffffd538, msg_iovlen: 1, msg_control: 0x0, msg_controllen: 0, msg_flags: (empty) } (0x7fffffffd548), (empty)) = -9 (EBADF: Bad file number)
```

### DIFF
--- a/src/main/host/syscall/type_formatting.rs
+++ b/src/main/host/syscall/type_formatting.rs
@@ -172,7 +172,6 @@ safe_pointer_impl!(libc::c_void);
 safe_pointer_impl!(libc::sockaddr);
 safe_pointer_impl!(libc::sysinfo);
 safe_pointer_impl!(libc::iovec);
-safe_pointer_impl!(libc::msghdr);
 
 simple_debug_impl!(nix::fcntl::OFlag);
 simple_debug_impl!(nix::sys::eventfd::EfdFlags);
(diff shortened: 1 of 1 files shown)


src/main/host/syscall/type_formatting.rs
```diff
@@ -341,3 +403,32 @@ impl<const LEN_INDEX: usize> SyscallDisplay for SyscallVal<'_, SyscallSockAddrAr
...
+
+        if options == FmtOptions::Deterministic {
+            return write!(f, "<pointer>");
+        }
+
+        // read the msghdr
+        let ptr = TypedArrayForeignPtr::new::<libc::msghdr>(ptr, 1);
+        let Ok(msg) = mem.memory_ref(ptr) else {
+            // if we couldn't read the memory, just show the pointer instead
+            return write!(f, "{:p}", ptr.ptr());
```
sporksmith, line 424: Maybe something like `"bad:{:p}"` to make it clearer that the pointer couldn't be read?

sporksmith (approved):

src/main/host/syscall/type_formatting.rs
stevenengler: I'll do this in a follow-up PR, since there's a bunch of places this would need to be updated to be consistent.

PR #2821, opened 2023-03-30
Add context to strace output when we can't read memory

stevenengler:
From https://github.com/shadow/shadow/pull/2815#discussion_r1152382178.

If the strace logger is unable to read some memory, a `<invalid>` tag will be added after the pointer. This matches when an invalid flag is passed to a syscall.

There are two downsides to this change.

1. Some syscall arguments allow NULL values, but the strace logger will show a NULL as "Invalid". For example this occurs with the address length pointer argument to `recvfrom`.
2. We don't always try to read pointers, so some pointers might be invalid but we never try to read them and they never get the "invalid" tag.

### DIFF
--- a/src/main/host/syscall/type_formatting.rs
+++ b/src/main/host/syscall/type_formatting.rs
@@ -105,7 +105,7 @@ macro_rules! deref_pointer_impl {
                 match (options, mem.memory_ref(ForeignArrayPtr::new(ptr, 1))) {
                     (FmtOptions::Standard, Ok(vals)) => write!(f, "{:?} ({:p})", &(*vals)[0], ptr),
                     // if we couldn't read the memory, just show the pointer instead
-                    (FmtOptions::Standard, Err(_)) => write!(f, "{ptr:p}"),
+                    (FmtOptions::Standard, Err(_)) => write!(f, "{ptr:p} <invalid>"),
                     (FmtOptions::Deterministic, _) => write!(f, "<pointer>"),
                 }
             }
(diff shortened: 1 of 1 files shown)


sporksmith (approved):
re the downsides you mentioned, we *could* special-case 0 to print "NULL" instead of "invalid", and/or use something like "unreadable" instead of "invalid"

I'm also fine with merging this as-is, though.

robgjansen:
After discussion, Rob will take over this PR. Here are the cases:

- we have a NULL ptr, we wont read it
- we have a real pointer, we tried to read but it is somehow invalid and can't be read
- we have a real pointer, but we decided not to read it because of its type
- we have a real pointer, and we read it because of its type, and we have some data

Rob will come up with tag names for these cases.

robgjansen:
@stevenengler I don't know if we still want this, or maybe I chose bad pointer suffix names, but here is my try at progressing this PR. I'm OK if we decide that it's not enough of an improvement and just want to close it instead.

src/main/host/syscall/type_formatting.rs
```diff
@@ -6,6 +6,30 @@ use crate::host::memory_manager::MemoryManager;
...
+    ptr: ForeignPtr<T>,
+    suffix: &str,
+) -> std::fmt::Result {
+    if ptr.is_null() {
+        write!(f, "{ptr:p} <null>")
+    } else {
+        write!(f, "{ptr:p} {}", suffix)
+    }
+}
+
```
robgjansen, line 32: Should these be macros since they are called from macros?

src/main/host/syscall/type_formatting.rs
```diff
@@ -24,7 +48,7 @@ macro_rules! simple_display_impl {
                 match <$type>::try_from(self.reg).ok() {
                     Some(x) => write!(f, "{x}"),
                     // if the conversion to type T was unsuccessful, just show an integer
-                    None => write!(f, "{:#x} <invalid>", u64::from(self.reg)),
+                    None => fmt_int_with_suffix(f, u64::from(self.reg), "<invalid-type>"),
```
stevenengler, line 51: What about "invalid-value" instead? The type should be correct, but the supplied value isn't valid for the type that the syscall takes.

stevenengler (commented):
Changes look good to me.

src/main/host/syscall/type_formatting.rs
stevenengler: A function is better here I think.

PR #2826, opened 2023-04-01
Fix some incorrect buffer sizes in the memory manager

stevenengler:
One fix is in a log statement, and the other is in `AllocdMem` which we only ever use as an `AllocdMem<u8>`, so neither of these currently cause incorrect behaviour.

### DIFF
--- a/src/main/host/memory_manager/mod.rs
+++ b/src/main/host/memory_manager/mod.rs
@@ -228,7 +228,7 @@ impl<'a, T: Debug + Pod> ProcessMemoryRefMut<'a, T> {
             CopiedOrMappedMut::Copied(copier, ptr, v) => {
                 trace!(
                     "Flushing {} bytes to {:x}",
-                    ptr.len(),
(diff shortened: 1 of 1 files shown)


sporksmith (approved):
Good catch!

Maybe we ought to add a method like `byte_count`?

Similarly I'm a little tempted to rename `len` to something like `item_count`. I think I originally used `len` to parallel the slice API, but in practice it's relatively common to want the length in bytes, and apparently easy to forget that `len` isn't that, at least for me :)

stevenengler:
> Maybe we ought to add a method like `byte_count`?
> 
> Similarly I'm a little tempted to rename `len` to something like `item_count`. I think I originally used `len` to parallel the slice API, but in practice it's relatively common to want the length in bytes, and apparently easy to forget that `len` isn't that, at least for me :)

I think these would sound good to me. It's nice to follow the slice API, but yeah if it makes the code clearer then I think renaming them makes sense.

PR #2813, opened 2023-03-29
Rename plugin pointer types

stevenengler:
- renamed `PluginPtr` to `ForeignPtr`
- removed `PluginVirtualPtr` typedef
- renamed `TypedPluginPtr` to `TypedArrayForeignPtr`
- renamed `PluginPhysicalPtr` to `ManagedPhysicalMemoryAddr`

This also renames some variables and function names to match.

@sporksmith If this ends up conflicting with your managed thread PR, let me know and I'll wait since it's probably easier for me to redo this than for you to deal with merge conflicts.

PR #2827, opened 2023-04-03
Add a generic type to `ForeignPtr`

stevenengler:
Follow-up to #2813.

This makes `ForeignPtr` typed. There aren't many places where we need untyped pointers in Shadow, so making pointers typed should be the default. This also means that for a lot of pointers, we can define their type directly in the syscall handler. Pointer types should now be self-documenting, and it should be harder to accidentally access the pointers as the wrong type. An `UntypedForeignPtr` was added for compatibility with the C code.

- added a generic type to `ForeignPtr`
- added `UntypedForeignPtr` for C code which maps to `ForeignPtr<()>` in rust code
- renamed `TypedArrayForeignPtr` to `ForeignArrayPtr`

The intermediate commits have a lot of casts between pointer types so that each commit would build and run, but the final code should have only a few pointer type conversions.

There are still some things that can be cleaned up, but I think this gets most of the code in.


src/lib/shadow-shim-helper-rs/src/util/mod.rs
```diff
@@ -0,0 +1,20 @@
+/// A trait to prevent type inference during function calls. Useful when you have a type that wraps
+/// a pointer (like `ForeignArrayPtr`) and you don't want Rust to infer the type of pointer
+/// during creation.  Instead, the caller must specify the generic type.
+///
+/// Example:
+///
+/// ```ignore
+/// let x: ForeignArrayPtr<u8>;
```
sporksmith, line 8: `let mut x` ?

src/main/host/managed_thread.rs
```diff
@@ -294,9 +294,9 @@ impl ManagedThread {
         &self,
         ctx: &ThreadContext,
         flags: libc::c_ulong,
-        child_stack: ForeignPtr,
-        ptid: ForeignPtr,
-        ctid: ForeignPtr,
+        child_stack: ForeignPtr<u8>,
+        ptid: ForeignPtr<()>,
```
sporksmith, line 298: Nit - `libc::pid_t`

sporksmith (approved):
Nice!

src/main/host/thread.rs
```diff
@@ -302,9 +310,9 @@ impl Thread {
         &self,
         ctx: &ProcessContext,
         flags: libc::c_ulong,
-        child_stack: ForeignPtr,
-        ptid: ForeignPtr,
-        ctid: ForeignPtr,
+        child_stack: ForeignPtr<u8>,
```
sporksmith, line 313: I think `<()>` might make more sense here? I think that makes it clearer that it's an opaque pointer of unknown size, not a pointer to a single `u8`.

src/lib/shadow-shim-helper-rs/src/syscall_types.rs
```diff
@@ -238,15 +322,17 @@ impl TryFrom<SysCallReg> for i16 {
...
+        Self {
+            as_ptr: v.cast::<()>(),
+        }
     }
 }
 
-impl From<SysCallReg> for ForeignPtr {
-    fn from(v: SysCallReg) -> ForeignPtr {
-        unsafe { v.as_ptr }
+impl<T> From<SysCallReg> for ForeignPtr<T> {
```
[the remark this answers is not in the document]
stevenengler: It does infer the type, which normally I wanted to avoid, but we need it so that `SyscallHandlerFn` can convert the `SysCallReg` to the correct pointer type. For example in the following code, if a syscall handler has an argument of type `ForeignPtr<u32>`, the `SyscallHandlerFn` trait needs to be able to convert from the syscall register to the `ForeignPtr<u32>`.

https://github.com/shadow/shadow/blob/5efe48f66d107d33f86fe95d50950faea504f524/src/main/host/syscall/handler/mod.rs#L171-L181

I'm not sure of a better way we can handle this.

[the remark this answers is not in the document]
sporksmith: Ah I see - yeah I don't see a better approach. A comment here for future reference would be helpful.

src/lib/shadow-shim-helper-rs/src/util/mod.rs
stevenengler: I don't think the `mut` is needed here since we assign it below but don't modify it. For example it's similar to

```rust
let x;
x = 0u32;
```

src/lib/shadow-shim-helper-rs/src/syscall_types.rs
```diff
@@ -1,60 +1,144 @@
...
+
+impl<T> Clone for ForeignPtr<T> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<T> std::fmt::Debug for ForeignPtr<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ForeignPtr")
```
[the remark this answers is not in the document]
stevenengler: I changed this to `f.debug_struct(std::any::type_name::<Self>())`. It's a little verbose (contains the full module path to `ForeignPtr`), but will contain the generic type.

src/lib/shadow-shim-helper-rs/src/syscall_types.rs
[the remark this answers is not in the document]
stevenengler: Added a comment.

src/main/host/thread.rs
stevenengler: Good idea, added.

src/main/host/managed_thread.rs
stevenengler: Added.

PR #2830, opened 2023-04-03
Add a `write` method to the memory manager

stevenengler:
This is similar to the existing `read` method, and I think it is a little nicer to use than `copy_to_ptr` in simpler cases.

(Also I'm not sure that the pid address write in `reap_thread()` is correct since I think it's supposed to write a pointer, not a `pid_t`, but for now I'm leaving this the same as it was before.)

### DIFF
--- a/src/main/host/memory_manager/mod.rs
+++ b/src/main/host/memory_manager/mod.rs
@@ -413,14 +413,28 @@ impl MemoryManager {
     ///
     /// Examples:
     ///
-    /// ```ignore
+    /// ```no_run
+    /// # use shadow_shim_helper_rs::syscall_types::ForeignPtr;
+    /// # use shadow_rs::host::memory_manager::MemoryManager;
+    /// # use nix::errno::Errno;
+    /// # fn foo() -> Result<(), Errno> {
+    /// # let memory_manager: MemoryManager = todo!();
     /// let ptr: ForeignPtr<u32> = todo!();
     /// let val: u32 = memory_manager.read(ptr)?;
+    /// # Ok(())
+    /// # }
     /// ```
     ///
-    /// ```ignore
+    /// ```no_run
+    /// # use shadow_shim_helper_rs::syscall_types::ForeignPtr;
+    /// # use shadow_rs::host::memory_manager::MemoryManager;
+    /// # use nix::errno::Errno;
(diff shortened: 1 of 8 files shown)


src/main/host/memory_manager/mod.rs
```diff
@@ -431,6 +431,19 @@ impl MemoryManager {
         Ok(unsafe { res.assume_init() })
     }
 
+    /// Writes a local value `val` into the memory at `ptr`.
+    ///
+    /// ```ignore
```
sporksmith, line 436: Maybe just `no_run` instead of `ignore`? https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes

sporksmith (approved):

sporksmith:
> (Also I'm not sure that the pid address write in reap_thread() is correct since I think it's supposed to write a pointer, not a pid_t, but for now I'm leaving this the same as it was before.)

I'm pretty sure it writes a `pid_t`. From `man set_tid_address`:

> When set_child_tid is set, the very first thing the new thread does is to write its thread ID at this address.
...
> When a thread whose clear_child_tid is not NULL terminates, then, if the thread is sharing memory with other threads, then 0 is written at the address specified in clear_child_tid and the kernel performs the following operation

src/main/host/memory_manager/mod.rs
stevenengler: Added (and updated example in `read`).

stevenengler:
> > (Also I'm not sure that the pid address write in reap_thread() is correct since I think it's supposed to write a pointer, not a pid_t, but for now I'm leaving this the same as it was before.)
> 
> I'm pretty sure it writes a `pid_t`. From `man set_tid_address`:
> 
> > When set_child_tid is set, the very first thing the new thread does is to write its thread ID at this address.
> > ...
> > When a thread whose clear_child_tid is not NULL terminates, then, if the thread is sharing memory with other threads, then 0 is written at the address specified in clear_child_tid and the kernel performs the following operation

Ah okay, I think I was getting confused by the prctl `PR_GET_TID_ADDRESS`, which writes the value of the `ForeignPtr` pointer (not the value pointed to by the `ForeignPtr`). I'll make a PR to try and add types here.

PR #2832, opened 2023-04-03
CI: don't run on multiple base images

sporksmith:
It turns out that the GitHub Action Ubuntu images both use the same
kernel version, so we probably don't get much value out of running on
multiple base images.

Both non-deprecated images currently use kernel 5.15.0-1034-azure.

https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md
https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md

Since our tests all run in containers, it's probably not worth pinning the base image to a specific image of Ubuntu at all, so this also changes the base images to "ubuntu-latest", which is currently 22.04.


sporksmith:
@robgjansen will need to change the required checks before it can be merged. No rush on this one; we could wait to discuss it in the group meeting.

stevenengler (approved):

robgjansen:
Done. Please double check that the correct tests are set as required and let me know if we need any more tweaks.

PR #2833, opened 2023-04-03
Remove hostname prefix from hosts' data files

### LINKED ISSUE #2218: Remove hostname prefix from data files
We currently have a lot of files with names of the form `<data-dir>/hosts/<hostname>/<hostname>.<processname>.<pid>.<type>`

Since each host is already in its own subdirectory, the `<hostname>` prefix is redundant.

Changing this would require a major version bump.

stevenengler:
Closes #2218.


sporksmith (approved):

sporksmith (changes requested):
Need to update semver.md as well

stevenengler:
Updated the semver doc, and fixed the `grep` commands in the tgen test.

PR #2839, opened 2023-04-04
Move and better-document waiting for thread exit

sporksmith:
This moves waiting for thread-exit from `Process` to `ManagedThread`, where it more logically belongs. It also adds a check in `ManagedThread`'s `Drop` implementation to validate that we never drop it while the native thread is still running, and adds documentation about why this is important - to prevent memory corruption due to shared memory regions being deallocated and reused.
    
This also fixes a bug - Before we *weren't* waiting for the thread to exit if the `clear_child_tid` attribute wasn't set. I belive this was a race condition, but in practice I think most higher level thread APIs (e.g. libc's `pthread` and Rust's `std::thread`) use this attribute, so wouldn't have wouldn't have been affected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ have been noticeable effect in typical simulations. In particular the per-thread
 data was already getting freed when the whole process exited, so it would only
 affect a process that created and terminated many threads over its lifetime.
 
+* Fixed a potential race condition when exiting managed threads that did not
+have the `clear_child_tid` attribute set. This is unlikely to have affected most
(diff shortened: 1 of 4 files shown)


stevenengler (approved):

sporksmith:
On one last look I realized that while iterating on this I'd somehow dropped the documentation I'd added about why it's important not to drop ManagedThread while the native thread is still running :facepalm: . Added this back in and on further reflection changed the drop implementation to always panic if the native thread is still running at drop time.

Issue #2210, opened 2022-06-03
Shadow doesn't detect when multi-threaded plugin is killed


stevenengler:
If a plugin is killed while it has more than one thread, Shadow doesn't correctly detect that the process was killed. Shadow will still report the simulation as successful.

Unexpected behaviour:

```yaml
general:
  stop_time: 10 seconds
network:
  graph:
    type: 1_gbit_switch
hosts:
  testnode:
    network_node_id: 0
    processes:
    - path: /usr/bin/python3
      args: '-c "import threading, time; threading.Thread(target=lambda: time.sleep(10)).start(); time.sleep(10)"'
      start_time: 5
    - path: /usr/bin/kill
      args: "1000 -KILL"
      start_time: 7
```

```
$ rm -rf shadow.data && build/src/main/shadow shadow.yaml > shadow.log
** Starting Shadow v2.0.0 with GLib v2.64.6
** Shadow completed successfully
```

---------

Expected behaviour:

```yaml
general:
  stop_time: 10 seconds
network:
  graph:
    type: 1_gbit_switch
hosts:
  testnode:
    network_node_id: 0
    processes:
    - path: /usr/bin/python3
      args: '-c "import time; time.sleep(10)"'
      start_time: 5
    - path: /usr/bin/kill
      args: "1000 -KILL"
      start_time: 7
```

```
$ rm -rf shadow.data && build/src/main/shadow shadow.yaml > shadow.log
** Starting Shadow v2.0.0 with GLib v2.64.6
** Shadow did not complete successfully: Controller exited with code -1
** See the log for details
```

sporksmith:
This appears to now be fixed:

```
$ rm -rf shadow.data && shadow 2210.yml > shadow.log
** Starting Shadow v2.5.0-115-g579814ee4 2023-04-05--12:26:09 with GLib v2.72.4
** Shadow did not complete successfully: Failed to run the simulation
**   1 managed processes exited with a non-zero error code
** See the log for details
```

and in the log:

```
00:00:00.100214 [4158237:shadow-worker] 00:00:07.000000000 [WARN] [testnode:11.0.0.1] [process.rs:1051] [shadow_rs::host::process] process 'testnode.python3.10.1000' exited with code 137 (error)
```

sporksmith:
Looks like it was fixed in v.2.4.0:

```
$ rm -rf shadow.data && shadow 2210.yml > shadow.log
[...]

PR #2842, opened 2023-04-05
Add regression test for #2210

stevenengler:


src/test/regression/CMakeLists.txt
```diff
@@ -34,3 +34,11 @@ add_shadow_tests(
       # the full timeout to fail otherwise.
       TIMEOUT 5
     )
+
+# Make sure tests in subdirs append to previously collected tests.
+set(ALL_SHADOW_TESTS "${ALL_SHADOW_TESTS}")
```
sporksmith, line 39: Maybe I'm misremembering how cmake scoping works, but I think this is a no-op?

sporksmith (approved):

src/test/regression/CMakeLists.txt
stevenengler: I have no idea, this is how it was done in #1131. @robgjansen Do you remember why you did it this way 2 years ago? :)

robgjansen: Without doing this the subdir tests would drop out of scope in the parent directory, meaning that the parent's list of tests would not include the subdir tests and they wouldn't run. Maybe that changed with a newer cmake version, but it would be pretty easy to test what happens if you don't include this blurb.

stevenengler: I removed it and the list of tests printed during the `setup build --test` don't change, so I'm guessing it's not doing anything on the version of cmake that I'm using.

I'll add another commit to remove this line everywhere we do it.

sporksmith: It looks like the only place we use this variable is to print a count of the number of tests?

```
$ git grep ALL_SHADOW_TESTS
src/test/CMakeLists.txt:   set(ALL_SHADOW_TESTS "${ALL_SHADOW_TESTS}" PARENT_SCOPE)
src/test/CMakeLists.txt:   list(APPEND ALL_SHADOW_TESTS ${SHADOW_TEST_NAME})
src/test/CMakeLists.txt:list(LENGTH ALL_SHADOW_TESTS ALL_SHADOW_TESTS_LENGTH)
src/test/CMakeLists.txt:message(STATUS "Configured to build ${ALL_SHADOW_TESTS_LENGTH} Shadow tests.")
src/test/CMakeLists.txt:message(STATUS "ALL_SHADOW_TESTS = ${ALL_SHADOW_TESTS}")
src/test/config/CMakeLists.txt:set(ALL_SHADOW_TESTS "${ALL_SHADOW_TESTS}")
src/test/config/CMakeLists.txt:set(ALL_SHADOW_TESTS "${ALL_SHADOW_TESTS}" PARENT_SCOPE)
src/test/examples/CMakeLists.txt:set(ALL_SHADOW_TESTS "${ALL_SHADOW_TESTS}")
src/test/examples/CMakeLists.txt:set(ALL_SHADOW_TESTS "${ALL_SHADOW_TESTS}" PARENT_SCOPE)
src/test/socket/CMakeLists.txt:set(ALL_SHADOW_TESTS "${ALL_SHADOW_TESTS}")
src/test/socket/CMakeLists.txt:set(ALL_SHADOW_TESTS "${ALL_SHADOW_TESTS}" PARENT_SCOPE)
src/test/tgen/CMakeLists.txt:set(ALL_SHADOW_TESTS "${ALL_SHADOW_TESTS}")
src/test/tgen/CMakeLists.txt:set(ALL_SHADOW_TESTS "${ALL_SHADOW_TESTS}" PARENT_SCOPE)
src/test/tor/CMakeLists.txt:set(ALL_SHADOW_TESTS "${ALL_SHADOW_TESTS}")
src/test/tor/CMakeLists.txt:set(ALL_SHADOW_TESTS "${ALL_SHADOW_TESTS}" PARENT_SCOPE)
```

stevenengler: Hmm, okay I'll leave things as-is then, and make a new PR to remove this list.

PR #2846, opened 2023-04-05
Switch the unbounded thread pool to use a latch instead of parkers

stevenengler:
I'm not sure that the `ThreadUnparker` works correctly when using rust's parking API, so I've updated a comment in it and replaced it in the thread-per-core scheduler (the unbounded thread pool) with a simple futex-based latch instead. The thread-per-host scheduler continues to use the (possibly broken) `ThreadUnparker`.

Another advantage is that it can futex-wake all threads with one syscall, rather than unparking each thread individually.

I ran a benchmark on an earlier version of this PR and there was no performance improvement.

### DIFF
--- a/src/main/utility/synchronization/simple_latch.rs
+++ b/src/main/utility/synchronization/simple_latch.rs
@@ -1,0 +1,238 @@
+use std::sync::atomic::{AtomicI32, Ordering};
+use std::sync::Arc;
+
+use nix::errno::Errno;
+
+/// A simple reusable latch. Multiple waiters can wait for the latch to open. After opening the
+/// latch with [`open()`](Self::open), you must not open the latch again until all waiters have
+/// waited with [`wait()`](LatchWaiter::wait) on the latch. In other words, you must not call
+/// `open()` multiple times without making sure that all waiters have successfully returned from
+/// `wait()` each time. This typically requires some other synchronization to make sure that the
+/// waiters have waited. If the latch and its waiters aren't kept in sync, the waiters will usually
+/// panic, but in some cases may behave incorrectly[^note].
+///
+/// [^note]: Since this latch uses a 32-bit wrapping integer to track the positions of the latch and
+/// its waiters, calling `open()` `u32::MAX + 1` times without allowing the waiters to wait will
+/// behave as if you did not call `open()` at all.
+///
+/// The latch uses release-acquire ordering, so any changes made before an `open()` should be
+/// visible in other threads after a `wait()` returns.
+#[derive(Debug)]
+pub struct Latch {
+    /// The generation of the latch.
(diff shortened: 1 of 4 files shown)


src/main/utility/synchronization/thread_parking.rs
```diff
@@ -107,10 +107,19 @@ impl ThreadUnparker {
...
+        // > Notice that being unblocked does not imply any synchronization with someone that
+        // > unparked this thread, it could also be spurious. For example, it would be a valid, but
+        // > inefficient, implementation to make both park and unpark return immediately without doing
+        // > anything.
+        //
+        // There is no guarentee that the change to `ready_flag` will be visible after the parked
+        // thread resumes, which may lead to a deadlock. In practice, rust currently does use an
+        // atomic with release-acquire ordering, but this might be relaxed in the future.
         // https://github.com/rust-lang/rust/blob/21b246587c2687935bd6004ffa5dcc4f4dd6600d/library/std/src/sys_common/thread_parker/futex.rs#L21-L27
         self.ready_flag.store(true, Ordering::Release);
```
sporksmith, line 124: Orthogonal to this PR, but I'm not sure I understand the `Release` orderings with no `Acquire`. Should `park` be using `Acquire` in the `success` ordering? Without that, I don't think the `ThreadParker` itself is providing synchronization?

I agree though that even so it wouldn't address the potential deadlock issue you're describing here.

src/main/utility/synchronization/simple_latch.rs
```diff
@@ -0,0 +1,238 @@
+use std::sync::atomic::{AtomicI32, Ordering};
+use std::sync::Arc;
+
+use nix::errno::Errno;
+
+/// A simple reusable latch. Multiple waiters can wait for the latch to open. After opening the
+/// latch with [`open()`](Self::open), you must not open the latch again until all watiers have
```
sporksmith, line 7: "waiters"

sporksmith (approved):

src/main/utility/synchronization/thread_parking.rs
stevenengler: Yeah I think that was a mistake. Ideally the thread `park`/`unpark` would be providing synchronization (and technically does as an implementation detail), and then I think the `ready_flag` operations could just be `Relaxed`. But since the documentation states that `park`/`unpark` does not provide synchronization, the `compare_exchange` should use `Acquire` in the success path to get the synchronization (but still doesn't fix the potential deadlock).

I'll change the `compare_exchange` to use `Acquire`.

src/main/utility/synchronization/simple_latch.rs
stevenengler: Fixed.

PR #2857, opened 2023-04-10
Fix broken example tests

stevenengler:
The tests would still pass even if shadow failed.

The etcd example still fails: #2858.

### DIFF
--- a/examples/curl/run.sh
+++ b/examples/curl/run.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -euo pipefail
 
 # first argument is the path to shadow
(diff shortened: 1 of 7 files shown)


examples/curl/run.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -e
```
sporksmith, line 2: My usual default is `set -euo pipefail`, which catches a few other footguns

sporksmith (approved):

examples/curl/run.sh
stevenengler: Thanks, changed it to this.

PR #2860, opened 2023-04-10
Replace process stop_time with shutdown_time and shutdown_signal

sporksmith:
Progress on https://github.com/shadow/shadow/discussions/2034, which is part of the planned breaking changes for 3.0: https://github.com/shadow/shadow/discussions/2496

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ prefixed with the hostname. For example a file that was previously named
 https://github.com/shadow/shadow/pull/2856
 
 * The per-process option `stop_time` has been replaced with `shutdown_time`.
+When set, the signal specified by `shutdown_signal` (a new option) will be sent
+to the process at the specified time. While shadow previously sent `SIGKILL` at
+a process's `stop_time`, the default `shutdown_signal` is `SIGTERM` to better
+support graceful shutdown.
 
 MINOR changes (backwards-compatible):
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ prefixed with the hostname. For example a file that was previously named
 [hostname(7)](https://man7.org/linux/man-pages/man7/hostname.7.html).
 https://github.com/shadow/shadow/pull/2856
 
+* The per-process option `stop_time` has been replaced with `shutdown_time`.
+
 MINOR changes (backwards-compatible):
 
 * Support the `MSG_TRUNC` flag for unix sockets.
--- a/docs/shadow_config_spec.md
+++ b/docs/shadow_config_spec.md
@@ -109,6 +109,7 @@ hosts:
 - [`hosts.<hostname>.processes[*].environment`](#hostshostnameprocessesenvironment)
 - [`hosts.<hostname>.processes[*].path`](#hostshostnameprocessespath)
 - [`hosts.<hostname>.processes[*].quantity`](#hostshostnameprocessesquantity)
+- [`hosts.<hostname>.processes[*].shutdown_signal`](#hostshostnameprocessesshutdown_signal)
 - [`hosts.<hostname>.processes[*].shutdown_time`](#hostshostnameprocessesshutdown_time)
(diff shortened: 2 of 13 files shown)


sporksmith:
Oops, nevermind for the moment; I need to make sure we still use SIGKILL at the simulation end time...

sporksmith:
> Oops, nevermind for the moment; I need to make sure we still use SIGKILL at the simulation end time...

Ah no, I did keep the SIGKILL at simulation shutdown

docs/shadow_config_spec.md
```diff
@@ -711,4 +710,14 @@ The simulated time at which to execute the process.
 Default: null  
 Type: String OR Integer OR null
 
-The simulated time at which to send a `SIGTERM` signal to the process.
+The simulated time at which to send 
+[`hosts.<hostname>.processes[*].shutdown_signal`](#hostshostnameprocessesshutdown_time)
```
stevenengler, line 714: "hostshostnameprocessesshutdown_time" -> "hostshostnameprocessesshutdown_signal"

docs/shadow_config_spec.md
```diff
@@ -711,4 +710,14 @@ The simulated time at which to execute the process.
...
+to the process.
+
+#### `hosts.<hostname>.processes[*].shutdown_signal`
+
+Default: "SIGTERM"
+Type: String OR Integer OR null
```
stevenengler, line 720: Needs two spaces after the `Default: "SIGTERM"` or else the next line will be on the same line.

stevenengler, line 720: What does a "null" signal mean?

CHANGELOG.md
```diff
@@ -29,6 +29,8 @@ prefixed with the hostname. For example a file that was previously named
 https://github.com/shadow/shadow/pull/2856
 
 * The per-process option `stop_time` has been replaced with `shutdown_time`.
+When set, the signal specified by `shutdown_signal` (a new option) will be sent
+to the process at the specified time.
```
stevenengler: Also document that the behaviour has changed (sends a SIGTERM by default rather than a SIGKILL; It's not enough to just rename `stop_time` with `shutdown_time` to keep the same behaviour)?

src/test/config/shutdown/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+    EXPECT_ERROR TRUE
+    CHECK_RETVAL FALSE)
+
+add_shadow_tests(
+    BASENAME shutdown_signal_sigterm
+    # `sleep` should die with signal SIGTERM (15), giving exitcode 128+15=143
+    POST_CMD "test `cat hosts/*/sleep.*.exitcode` -eq 143"
+    # Shadow should return failure
+    EXPECT_ERROR TRUE
+    CHECK_RETVAL FALSE)
```
stevenengler, line 15: Maybe also add a test for SIGKILL just because SIGKILL is handled kind of different from other signals?

stevenengler (approved):

docs/shadow_config_spec.md
```diff
@@ -113,6 +113,7 @@ hosts:
 - [`hosts.<hostname>.processes[*].quantity`](#hostshostnameprocessesquantity)
 - [`hosts.<hostname>.processes[*].start_time`](#hostshostnameprocessesstart_time)
 - [`hosts.<hostname>.processes[*].shutdown_time`](#hostshostnameprocessesshutdown_time)
+- [`hosts.<hostname>.processes[*].shutdown_signal`](#hostshostnameprocessesshutdown_signal)
```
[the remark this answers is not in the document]
sporksmith: Done

docs/shadow_config_spec.md
sporksmith: done

sporksmith: Hmm, yeah I think was confused here thinking that "null" meant that it didn't need to be specified by the user, but I think that's already covered by there being a `Default`.

Issue #2848, opened 2023-04-06
Reduce or eliminate usage of special environment variables in managed processes


sporksmith:
We currently pass some configuration options through environment variables to the shim inside the managed processes. Most such instances could use shared memory regions instead. It'd be good to do so for several reasons:

* This would consolidate shared config/state into the shared memory regions instead of having two different mechanisms.
* We currently access these environment variables via the libc function `getenv`, which is an undesired dependency on libc from the shim, and breaks if shadow's initialization happens before libc's early initialization. This is one problem in #2790.
* These environment variables are currently visible to, and maybe even mutable from, from managed code in the managed process.

Moving most such state to shared memory is trivial, but we currently bootstrap shared memory itself via environment variables. We could probably use some other mechanism instead, such as passing the serialized handles through the `stdin` file descriptor.

PR #2863, opened 2023-04-11
Move more env variables to shared memory

sporksmith:
Progress on https://github.com/shadow/shadow/issues/2848 and on refactoring process initialization and teardown.

### DIFF
--- a/src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
+++ b/src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
@@ -82,6 +82,9 @@ pub struct HostShmem {
     // Native pid of the Shadow simulator process.
     pub shadow_pid: libc::pid_t,
 
+    // Emulated CPU TSC clock rate, for rdtsc emulation.
+    pub tsc_hz: u64,
+
     // Current simulation time.
     pub sim_time: AtomicEmulatedTime,
 }
@@ -95,6 +98,7 @@ impl HostShmem {
         unblocked_syscall_latency: SimulationTime,
         unblocked_vdso_latency: SimulationTime,
         shadow_pid: libc::pid_t,
+        tsc_hz: u64,
     ) -> Self {
         Self {
             host_id,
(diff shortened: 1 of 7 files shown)


src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
```diff
@@ -79,6 +79,9 @@ pub struct HostShmem {
     // per-process option.
     pub unblocked_vdso_latency: SimulationTime,
 
+    // Native pid of the Shadow process.
+    pub shadow_pid: libc::pid_t,
```
stevenengler, line 83: How about "the Shadow simulator process"? When I first read "Shadow process", I thought it meant "Shadow `Process`" and was confused.

src/lib/shim/shim.c
```diff
@@ -228,41 +228,21 @@ static void _shim_parent_init_logging() {
...
-    // Ensure that the child process exits when Shadow does.  Shadow
-    // ought to have already tried to terminate the child via SIGTERM
-    // before shutting down (though see
-    // https://github.com/shadow/shadow/issues/903), so now we jump all
-    // the way to SIGKILL.
+    // Ensure that the child process exits when Shadow does. This is to avoid
+    // confusing behavior or a "stalled out" process in the case that Shadow
+    // exits abnormally. Shadow normally ensures all managed processes have
+    // exited before exiting itself.
     if (prctl(PR_SET_PDEATHSIG, SIGKILL) < 0) {
```
stevenengler, line 236: Is there a way we can set this with posix_spawn, for example if someone tries to run a static binary where this function isn't loaded/run?

stevenengler (approved):

src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
sporksmith: Done

src/lib/shim/shim.c
sporksmith: Unfortunately not. Added a comment.

PR #2865, opened 2023-04-11
Bump minimum glib version 2.32->2.58

sporksmith:


src/main/core/main.c
```diff
@@ -20,8 +20,8 @@
 
 int main_checkGlibVersion() {
     /* check the compiled GLib version */
-    if (!GLIB_CHECK_VERSION(2, 32, 0)) {
-        g_printerr("** GLib version 2.32.0 or above is required but Shadow was compiled against "
+    if (!GLIB_CHECK_VERSION(2, 58, 0)) {
+        g_printerr("** GLib version 2.58.0 or above is required but Shadow was compiled against "
                    "version %u.%u.%u\n",
                    (guint)GLIB_MAJOR_VERSION, (guint)GLIB_MINOR_VERSION, (guint)GLIB_MICRO_VERSION);
         return -1;
```
stevenengler, line 27: Do we need this check if this is checked by pkgconfig?

stevenengler (approved):

src/main/core/main.c
sporksmith: Dropped

PR #2867, opened 2023-04-12
Small latch improvements

stevenengler:
This changes the atomic from `AtomicI32` to `AtomicU32` since I realized that futex(2) in ubuntu 20.04 is out of date, and newer man pages show that its unsigned.

This also adds `FUTEX_PRIVATE_FLAG` which may improve performance (but I doubt it will in practice). This flag can be used since the latch will only ever be used in the same process.


src/main/utility/synchronization/simple_latch.rs
stevenengler, line 1: Temporary comment so that the PR isn't merged. I'll wait until after the next nightly benchmark before merging this.

sporksmith (approved):

PR #2868, opened 2023-04-12
Remove the host `quantity` config option

### LINKED ISSUE #2067: Remove configuration `hosts.<hostname>.quantity` field
Shadow supports a [`hosts.<hostname>.quantity`](https://shadow.github.io/docs/guide/shadow_config_spec.html#hostshostnamequantity) field to make it easier to start multiple hosts that run the same processes. For example, we use it in our curl example to start three clients:

```yaml
hosts:
  # a host with the hostname 'server'
  server:
    network_node_id: 0
    processes:
    - path: /usr/bin/python3
      args: -m http.server 80
      start_time: 3s
  # three hosts with hostnames 'client1', 'client2', and 'client3'
  client:
    network_node_id: 0
    quantity: 3
    processes:
    - path: /usr/bin/curl
      args: -s server
      start_time: 5s
```

I think we should consider removing this option, and instead recommend using yaml anchors instead. The previous example can be replaced by:

```yaml
hosts:
  server:
    network_node_id: 0
    processes:
    - path: /usr/bin/python3
      args: -m http.server 80
      start_time: 3s
  client1: &client_host
    network_node_id: 0
    processes:
    - path: /usr/bin/curl
      args: -s server
      start_time: 5s
  client2: *client_host
  client3: *client_host
```

There is also a `hosts.<hostname>.processes[*].quantity` field we may also want to consider removing.

This syntax is already supported in shadow today.

stevenengler:
Part of our Shadow 3.0 changes. Closes #2067.


src/test/determinism/determinism1.test.shadow.config.yaml
```diff
@@ -21,9 +21,36 @@ network:
         ]
       ]
 hosts:
-  testnode:
+  testnode1: &host
     network_node_id: 0
-    quantity: 10
+    ip_addr: 11.0.0.1
```
sporksmith, line 26: OOC why does this need explicit IPs now? Or maybe this is  an orthogonal improvement to not implicitly rely on shadow's auto IP assignment in this test?

sporksmith (approved):

src/test/determinism/determinism1.test.shadow.config.yaml
stevenengler: Previously with a `quantity`, show would create hosts `hostname{x}` for x in 0..10 and assigns each host an increasing IP address. So "hostname1" is assigned "11.0.0.1", "hostname2" is assigned "11.0.0.2", etc.

Without the `quantity` option, Shadow iterates over hosts in alphabetical order and assigns them an increasing IP address. But the hostname alphabetical order is "hostname1", "hostname10", "hostname2", "hostname3", etc. So "hostname1" is assigned "11.0.0.1", "hostname10" is assigned "11.0.0.2", "hostname2" is assigned "11.0.0.3", etc.

sporksmith: oh! I see. And I guess the test actually use the IP addresses instead of doing name resolution?

stevenengler: The simulation itself is fine I think, but the verification scripts after aren't. They do something along the lines of:

```
diff a/hosts/hostname{X}/11.0.0.{X}.pcap b/hosts/hostname{X}/11.0.0.{X}.pcap
```

sporksmith: Gotcha; an explanatory comment in the config files might be helpful in the future

stevenengler: Added some comments.

PR #2870, opened 2023-04-13
Refactor Process teardown

sporksmith:
Refactors Process to remove more state variables and make process teardown easier to follow.

### DIFF
--- a/src/main/host/managed_thread.rs
+++ b/src/main/host/managed_thread.rs
@@ -31,8 +31,8 @@ use crate::utility::{childpid_watcher, pod, syscall};
 pub enum ResumeResult {
     /// Blocked on a SysCallCondition.
     Blocked(SysCallCondition),
-    /// The native thread has exited.
-    ExitedThread,
+    /// The native thread has exited with the given code.
+    ExitedThread(i32),
     /// The thread's process has exited.
     ExitedProcess,
 }
@@ -202,8 +202,8 @@ impl ManagedThread {
                     // `set_tid_address`, to block here until the thread has
                     // actually exited.
                     if syscall.syscall_args.number == libc::SYS_exit {
-                        self.return_code
-                            .set(Some(syscall.syscall_args.args[0].into()));
+                        let return_code = syscall.syscall_args.args[0].into();
+                        self.return_code.set(Some(return_code));
                         // Tell mthread to go ahead and make the exit syscall itself.
                         // We *don't* call `_managedthread_continuePlugin` here,
                         // since that'd release the ShimSharedMemHostLock, and we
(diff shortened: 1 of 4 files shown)


src/main/host/thread.rs
```diff
@@ -13,12 +13,19 @@ use shadow_shmem::allocator::{Allocator, ShMemBlock};
...
 use crate::cshadow as c;
 use crate::host::syscall_condition::{SysCallConditionRef, SysCallConditionRefMut};
 use crate::utility::{syscall, IsSend, SendPointer};
 
+#[derive(Debug)]
+pub enum ResumeResult {
+    Blocked,
+    ExitedThread,
+    ExitedProcess,
+}
```
stevenengler: This could use some comments.

src/main/host/managed_thread.rs
```diff
@@ -25,6 +25,13 @@ use crate::cshadow;
 use crate::host::syscall_types::SyscallReturn;
 use crate::utility::{childpid_watcher, pod, syscall};
 
+#[derive(Debug)]
+pub enum ResumeResult {
+    Blocked(SysCallCondition),
+    ExitedThread,
+    ExitedProcess,
+}
```
stevenengler: This could use some comments.

src/main/host/managed_thread.rs
```diff
@@ -96,10 +100,11 @@ impl ManagedThread {
...
             Err(SelfContainedChannelError::WriterIsClosed) => {
-                trace!("Plugin exited while executing native syscall {n}");
-                ctx.process.mark_as_exiting();
-                self.cleanup_after_exit_initiated();
-                SysCallReg::from(-(Errno::ESRCH as i64))
+                // This should only happen if the syscall was SYS_exit or
+                // SYS_exit_group, or if the process was killed externally, e.g.
+                // by sending a signal from outside of shadow. These cases would be
+                // tricky to support, and we shouldn't need to.
+                panic!("Plugin exited while executing native syscall {n}");
```
stevenengler, line 107: Is it easy to get the exit status here and include it in the panic message? This might happen in practice if the oom manager kills some plugin process, which might be hard for the user to debug.

src/main/host/process.rs
```diff
@@ -545,6 +544,11 @@ impl Process {
     }
 
     pub fn stop(&self, host: &Host) {
+        if !self.is_running() {
+            info!("process {} has already stopped", self.name());
+            return;
+        }
```
stevenengler, line 555: nit: make this debug instead since it doesn't seem relevant to the user?

stevenengler (approved):

src/main/host/process.rs
sporksmith: done

PR #2871, opened 2023-04-13
Add experimental option to spin in the worker threads

stevenengler:
This adds the experimental option `use_worker_spinning`, which is disabled by default. If enabled, the worker threads will spin indefinitely in a sched_yield loop rather than wait on a futex while waiting for a new task to be ready. This has a large performance benefit on our benchmark runner (#2877). Using this option, we can test on other machines and see if/when we should enable this spinning behaviour by default.

https://github.com/shadow/benchmark-results/tree/master/tgen/2023-04-19-T21-22-45

1682457053_grim

https://github.com/shadow/benchmark-results/tree/master/tor/2023-04-25-T16-56-53

1682457065_grim

### DIFF
--- a/src/main/utility/synchronization/simple_latch.rs
+++ b/src/main/utility/synchronization/simple_latch.rs
@@ -32,6 +32,8 @@ pub struct LatchWaiter {
     gen: u32,
     /// The read-only generation of the latch.
     latch_gen: Arc<AtomicU32>,
+    /// Should we sched_yield in a spinloop indefinitely rather than futex-wait?
+    spin_yield: bool,
 }
 
 impl Latch {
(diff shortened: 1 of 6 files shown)


src/main/utility/synchronization/simple_latch.rs
```diff
@@ -106,27 +109,41 @@ impl LatchWaiter {
...
+                std::thread::yield_now();
+            }
         }
 
         self.gen = self.gen.wrapping_add(1);
     }
+
+    /// When waiting, should we `sched_yield` in a spinloop indefinitely rather than futex-wait?
+    /// This may improve performance in some workloads. This is disabled by default.
+    pub fn enable_spinning(&mut self, value: bool) {
```
sporksmith, line 144: Nit - is there a reason to have this as a method instead of a constructor parameter? I don't think we'll ever want to change it for a latch once constructed, will we?

sporksmith (approved):

src/main/utility/synchronization/simple_latch.rs
stevenengler: My reasoning was:

- The waiter should choose whether it spins or not, instead of the latch. A `LatchWaiter` can only be constructed through `Latch::waiter(&mut self)`, so having an `enable_spinning` method on the waiter allows you to make the decision closer to where you'll actually be doing the waiting.
- Enabling spinning is an optional feature, so I think it's nicer to have this feature be opt-in with a reasonable default (disabled), rather than requiring the constructor to specify explicitly if it's enabled or disabled (setting it in the constructor doesn't allow for a default behaviour, unless maybe we made it an `Option<bool>`).

sporksmith: > so having an enable_spinning method on the waiter allows you to make the decision closer to where you'll actually be doing the waiting.

Oops, I think what I meant was to add the field to the `waiter` method, not to the latch constructor.

> Enabling spinning is an optional feature, so I think it's nicer to have this feature be opt-in with a reasonable default (disabled),

Hmm, another way to achieve this would be for `waiter` to return a non-spinning waiter, and for a new method like `spinning_waiter` to return one with spinning enabled.

I'm also fine with the code as is, though.

sporksmith (approved):

PR #2872, opened 2023-04-13
Process: remove circular weak reference

sporksmith:
This unblocks some further cleanup of managing Process's internal state, and removes some complexity in itself.

### DIFF
--- a/src/main/host/thread.rs
+++ b/src/main/host/thread.rs
@@ -415,14 +415,7 @@ impl Thread {
                 let cond = cond.into_inner();
                 self.cond.set(unsafe { SendPointer::new(cond) });
                 if let Some(cond) = unsafe { cond.as_mut() } {
-                    unsafe {
-                        c::syscallcondition_waitNonblock(
-                            cond,
-                            ctx.host,
-                            ctx.process.cprocess(ctx.host),
-                            self,
-                        )
-                    }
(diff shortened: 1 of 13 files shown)


src/main/host/thread.rs
```diff
@@ -638,12 +631,12 @@ mod export {
...
     #[no_mangle]
-    pub unsafe extern "C" fn thread_getProcess(thread: *const Thread) -> *const ProcessRefCell {
+    pub unsafe extern "C" fn thread_getProcess(thread: *const Thread) -> *const Process {
         let thread = unsafe { thread.as_ref().unwrap() };
         Worker::with_active_host(|host| {
             let process = host.process_borrow(thread.process_id).unwrap();
-            let process: &ProcessRefCell = &process;
-            process as *const _
+            let p = &*process.borrow(host.root()) as *const _;
+            p
```
stevenengler, line 639: Maybe cast to the type explicitly here just to be safe (and then I think you can skip the pointer cast)?

```rust
&*process.borrow(host.root()) as &Process
```

stevenengler (approved):

src/main/host/thread.rs
sporksmith: Added an explicit type; we still need the pointer cast though

PR #2878, opened 2023-04-17
Move signal parsing into serde

sporksmith:
This takes a bit more code, but keeps with trying to do as much input validation as we can up front. We can also reuse it when adding process end state expectations (e.g. "killed with signal X") for https://github.com/shadow/shadow/discussions/2034

### DIFF
--- a/src/main/core/support/configuration.rs
+++ b/src/main/core/support/configuration.rs
@@ -570,7 +570,7 @@ pub struct ProcessOptions {
 
     /// The signal that will be sent to the process at `shutdown_time`
     #[serde(default = "default_shutdown_signal")]
-    pub shutdown_signal: String,
+    pub shutdown_signal: Signal,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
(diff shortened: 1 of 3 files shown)


src/main/core/support/configuration.rs
```diff
@@ -953,6 +953,89 @@ impl<'de> serde::Deserialize<'de> for ProcessArgs {
...
+                self.visit_str(&v)
+            }
+
+            fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+            where
+                E: serde::de::Error,
+            {
+                nix::sys::signal::Signal::from_str(v)
+                    .map(Signal)
+                    .map_err(|_e| E::custom(format!("Invalid signal string: {v}")))
```
stevenengler, line 992: Maybe add to the documentation (and maybe the doc comment / cli help) what format this needs to be in. For example, uppercase or lowercase, the shortened form (ex: `TERM`) or long form (`SIGTERM`), etc.

src/main/core/support/configuration.rs
```diff
@@ -953,6 +953,89 @@ impl<'de> serde::Deserialize<'de> for ProcessArgs {
...
+            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+                formatter.write_str("a signal string or integer")
+            }
+
+            fn visit_string<E>(self, v: String) -> Result<Self::Value, E>
+            where
+                E: serde::de::Error,
+            {
+                self.visit_str(&v)
+            }
```
stevenengler, line 984: nit: this is already the default behaviour for `visit_string`, so it's not needed

src/main/core/support/configuration.rs
```diff
@@ -953,6 +953,89 @@ impl<'de> serde::Deserialize<'de> for ProcessArgs {
...
+
+impl JsonSchema for Signal {
+    fn schema_name() -> String {
+        String::from("Signal")
+    }
+
+    fn json_schema(_gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+        // TODO: is there a way to generate a schema for "string or integer"?
+        schemars::schema::Schema::Bool(true)
+    }
```
stevenengler, line 1028: I'm not sure the proper way to construct the schema object, but in Shadow we only read the `description` and the `default` fields of the metadata in order to output cli help messages. Since the description and default should come come from the parent structure's `shutdown_signal` field (`Signal` doesn't implement `Default`, so the default has to be defined by `ProcessOptions`), I don't think anything used here will affect the cli help messages.

But we only uses this JSON schema information for options that can be set on the cli, but `shutdown_signal` is not configurable through the cli. So we don't really need schema information for this field anyways (or really the entire `ProcessOptions` or `HostOptions` I think). So alternatively you could add `#[schemars(skip)]` to `shutdown_signal` or remove the `JsonSchema` derive from all of `Signal`, `ProcessOptions`, and `HostOptions`.

stevenengler (approved):

src/main/core/support/configuration.rs
sporksmith: Done

src/main/core/support/configuration.rs
sporksmith: done

src/main/core/support/configuration.rs
sporksmith: Thanks - updated this comment.

PR #2881, opened 2023-04-18
Check if binary is dynamically linked

### LINKED ISSUE #2190: Detect statically linked binaries and show error
When running a Shadow simulation that uses statically linked binaries, Shadow will fail to interpose processes for those binaries. This can be confusing since the processes will run outside of Shadow's control, and Shadow will become blocked. We should detect if the user is trying to run statically linked binaries and show an appropriate error message.

stevenengler:
It might be better in the long term to read the ELF data and check if it's dynamically linked using something like [elf](https://docs.rs/elf/latest/elf/index.html), but using ld-linux could be a first step. (We need to read shadow's rpath anyways, so we'll need an ELF-parsing library at some point.)

On the other hand, we don't just want to test that it's dynamically linked. I think we specifically want to make sure it's using the GNU dynamic linker/loader which supports `LD_PRELOAD`, in which case I think testing it with ld-linux would be better. If we wanted to check based on ELF sections, I think we'd want to check that `.interp` is specifically the path to the GNU loader.

Closes #2190.


src/main/core/sim_config.rs
```diff
@@ -449,5 +451,33 @@ fn verify_plugin_path(path: impl AsRef<std::path::Path>) -> anyhow::Result<()> {
...
+
+        if ld_status.success() {
+            CHECKED_DYNAMIC_BINS
+                .write()
+                .unwrap()
+                .insert(path.to_path_buf());
+        } else {
+            // technically ld-linux could return errors for other reasons, but this is the most
+            // likely reason given that we already checked that the file exists
+            return Err(anyhow::anyhow!("The path is not dynamically linked"));
```
sporksmith, line 478: It might be helpful for debugging to include `ld.so`'s output. It looks like you can use `Command::output` instead of `Command::status` to also capture the output.

sporksmith (approved):
Nice!

I've also been toying with the idea to directly invoke the dynamic linker when we `exec`, but there might be some subtle differences vs the normal way of loading a program. e.g. I don't know how to get an `argv[0]` that's just the basename of the program.

```
$ bash -c 'echo $0'
bash
```

vs

```
$ /lib64/ld-linux-x86-64.so.2 /bin/bash -c 'echo $0'
/bin/bash
```

stevenengler:
> I've also been toying with the idea to directly invoke the dynamic linker when we exec, but there might be some subtle differences vs the normal way of loading a program. e.g. I don't know how to get an argv[0] that's just the basename of the program.

There is a way, but it's only supported in newer versions. For example see [ld.so(8) on ubuntu 22.04](https://manpages.ubuntu.com/manpages/jammy/man8/ld.so.8.html) vs [ld.so(8) on ubuntu 22.10](https://manpages.ubuntu.com/manpages/kinetic/man8/ld.so.8.html):

> ```
> --argv0 string (since glibc 2.33)
>              Set argv[0] to the value string before running the program.
> ```

I don't know if there's any workaround for older versions.

src/main/core/sim_config.rs
stevenengler: As far as I know the `--verify` option doesn't output anything. For example if I give it a wrong path or the path to a static binary, there's nothing printed on stdout or stderr. But I'll add this since it doesn't hurt if they do add output in the future.

PR #2882, opened 2023-04-19
Split Process internals by state

sporksmith:
Processes can be kept around as zombies after they exit, but it usually doesn't make sense to access most of their internal state at this point. Previously there were ad-hoc checks for whether the Process was still running, but by moving the internals into an enum we can't forget to make such checks.

This also helps ensure we clean up everything that ought to be cleaned up when the process exits (goes from the `Runnable` state to the `Zombie` state).

### DIFF
--- a/src/lib/std-util/src/nested_ref.rs
+++ b/src/lib/std-util/src/nested_ref.rs
@@ -1,0 +1,243 @@
+//! Tools for chaining borrows of [`std::cell::RefCell`].
+//!
+//! Useful for returning an abstract borrow, where multiple inner borrows are required. e.g.:
+//!
+//! ```
+//! use std::cell::RefCell;
+//!
+//! struct MyPrivateType {
+//!   x: RefCell<i32>
+//! }
+//!
+//! pub struct MyPublicType {
+//!   inner: RefCell<MyPrivateType>
+//! }
+//!
+//! impl MyPublicType {
+//!   pub fn borrow_x(&self) -> impl std::ops::Deref<Target=i32> + '_ {
+//!     std_util::nested_ref::NestedRef::map(self.inner.borrow(), |inner| inner.x.borrow())
+//!   }
+//! }
+//! ```
+//!
+//! Currently only supports nested RefCells; i.e. one level of nesting.
+//!
+//! TODO: It might be feasible to genericize the reference types, which would
+//! also add support for arbitrary levels of nesting.
+use std::cell::{Ref, RefMut};
+
+/// A nested [`std::cell::Ref`]. Useful for chaining borrows with [`std::cell::RefCell`].
+pub struct NestedRef<'a, Inner, Outer>
+where
+    Inner: 'static,
+{
+    // unsafely points to the Ref inside `_outer`.
+    // Must be declared first so that it's dropped first.
+    inner: Ref<'a, Inner>,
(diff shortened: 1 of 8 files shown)


src/lib/std-util/src/nested_ref.rs
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    #[inline]
+    pub fn filter_map(
+        outer: Ref<'a, Outer>,
+        borrow_fn: impl FnOnce(&Outer) -> Option<Ref<Inner>>,
+    ) -> Option<Self> {
+        let outer = Box::new(outer);
+        let inner: Ref<Inner> = borrow_fn(&outer)?;
+        let inner: Ref<'a, Inner> = unsafe { std::mem::transmute(inner) };
```
stevenengler, line 56: The transmute (and the one below) could use some documentation (that it's extending the lifetime and depends on the memory pointed to by the box not to change).

src/main/host/process.rs
```diff
@@ -332,31 +461,32 @@ impl Process {
...
+                        common,
+                        shim_shared_mem_block,
+                        memory_manager: Box::new(RefCell::new(memory_manager)),
+                        desc_table,
+                        itimer_real,
+                        strace_logging,
+                        dumpable: Cell::new(cshadow::SUID_DUMP_USER),
+                        native_pid,
+                        unsafe_borrow_mut: RefCell::new(None),
+                        unsafe_borrows: RefCell::new(Vec::new()),
```
stevenengler, line 707: Should these be left in `Common` so that we don't accidentally drop them while the C code still has borrows? Or are we sure that the process state will never become non-runnable while the C code still has borrows?

stevenengler (approved):

src/lib/std-util/src/nested_ref.rs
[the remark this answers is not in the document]
sporksmith: > but it could do a lot of other things like return a borrow() of some other RefCell, or return a Ref that was created outside of the closure.

The compiler won't let the closure return something with a shorter lifetime than its input reference. It could perhaps return a `Ref<'static, Inner>` that doesn't borrow from `outer`, but that would still be safe.

sporksmith: Oops; yeah done. I had some initially but it got lost in a refactor.

src/main/host/process.rs
sporksmith: > Should these be left in Common so that we don't accidentally drop them while the C code still has borrows? 

We'd then have to put the `memory_manager` in `Common` as well since these refer to it.

> Or are we sure that the process state will never become non-runnable while the C code still has borrows?

We only do the state change from in `Process`, where there shouldn't be any `Thread` or `SysCallHandler` code running, so there shouldn't be any outstanding borrows. I changed `RunnableProcess::into_common()` to assert that there are no remaining borrows instead of flushing them itself to validate this at runtime.

Ultimately it might make sense to pass a `&RunnableProcess` to thread methods and syscall handlers etc to make it clearer that it's guaranteed the process is running - that might let us also get rid of `NestedRef` etc. I started to sketch this out, but it's a big change to cram into this PR; e.g. we'd have to expose `RunnableProcess` to C code and change most of the C APIs to take it (or add ones that take it in case we still need Process in some places).

PR #2885, opened 2023-04-19
Small misc changes

stevenengler:
Removed some unused C macro constant definitions, fixed some mistakes in the documentation, and made some `prctl` changes.

@sporksmith I removed `PR_GET_SECCOMP` and `PR_SET_SECCOMP` support since I don't think we want the plugin changing seccomp settings, but wanted to check with you first.


sporksmith (approved):

sporksmith:
> I removed PR_GET_SECCOMP and PR_SET_SECCOMP support since I don't think we want the plugin changing seccomp settings, but wanted to check with you first.

I *had* been thinking that allowing seccomp might be ok, since filters can only be "layered on", not modified or removed. From `seccomp(2)`:

> If prctl(2) or seccomp() is allowed by the attached filter, further filters may be added.  This will increase  evaluation time, but allows for further reduction of the attack surface during execution of a thread.

OTOH if a filter is added that ends up blocking a syscall that the shim is trying to make natively, things could get... weird and broken. So yeah, blocking it for now makes sense until if and when we need it and actually think through and test it.

PR #2886, opened 2023-04-19
Add config option expected_end_state    

sporksmith:
* Allows the user to specify the expected state of a process at the end of a simulation.
* Changes the default behavior to treat a process still running at the end of the simulation as an error.
    
See https://github.com/shadow/shadow/discussions/2034

### DIFF
--- a/docs/shadow_config_spec.md
+++ b/docs/shadow_config_spec.md
@@ -201,7 +201,7 @@ Initialize randomness using seed N.
 *Required*  
 Type: String OR Integer
 
-The simulated time at which simulated processes are sent a SIGKILL signal.
+The simulated time at which the simulation ends.
 
 #### `general.template_directory`
 
--- a/docs/shadow_config_spec.md
+++ b/docs/shadow_config_spec.md
@@ -38,7 +38,7 @@ hosts:
     - path: /usr/sbin/nginx
       args: -c ../../../nginx.conf -p .
       start_time: 1
-      expected_final_state: !running
+      expected_final_state: running
   client1: &client_host
     network_node_id: 0
     host_options:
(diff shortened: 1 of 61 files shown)


docs/shadow_config_spec.md
```diff
@@ -661,6 +662,33 @@ environment:
 environment: { ENV_A: "1", ENV_B: foo }
 ```
 
+#### `hosts.<hostname>.processes[*].expected_final_state`
```
stevenengler, line 665: Maybe also double-check other documentation examples (for example "docs/getting_started_basic.md") and make sure those examples are updated to use this as well.

stevenengler, line 665: Also adding a comment here that tornettools and the benchmark runner's tor and tgen simulations will need to be updated.

stevenengler (approved):

src/main/host/process.rs
```diff
@@ -88,6 +88,28 @@ impl From<ProcessId> for libc::pid_t {
     }
 }
 
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum ExitStatus {
+    Normal(i32),
+    Signaled(nix::sys::signal::Signal),
```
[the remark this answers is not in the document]
stevenengler: Wait nvm, this isn't how things work :)

docs/shadow_config_spec.md
sporksmith: If I merge a benchmark change that sets `expected_final_state`, it'll break due to the config option being unknown until this PR is merged, won't it?

If I merge this PR first, the benchmarks will be broken until I merge changes to them.

I could be careful and avoid having a temporary broken state, but it's probably not worth the extra dev overhead vs just minimizing the time window that has the broken state

stevenengler: Okay I might have misunderstood. Yeah I think it would be best to not worry about a rigorous solution, just it's probably best for the benchmark to be updated the same day as this PR is merged to avoid breaking the nightly benchmarks. You shouldn't have to update tornettools in order to fix the benchmark to work (just updating the benchmark yaml is enough), but updating tornettools sooner rather than later would probably be good so that it doesn't block us from updating tornettools in the benchmarks if we need to in future PRs.

sporksmith:
Changed to use the untagged enum trick to avoid requiring the `!` explicit types in yaml. PTAL

docs/shadow_config_spec.md
```diff
@@ -675,30 +676,30 @@ environment: { ENV_A: "1", ENV_B: foo }
 
 #### `hosts.<hostname>.processes[*].expected_final_state`
 
-Default: !exited 0
-Type: !exited \<Integer\> OR !signaled \<Signal String or Integer\> OR !running
+Default: \{exited: 0\}  
+Type: \{exited: \<Integer\>\} OR \{signaled: [Unix Signal](./shadow_config_overview.md#unix-signals)\} OR "running"
```
stevenengler: nit: I think if we're using quotes around "running", we should also use quotes around "exited" and "signaled" which are also strings.

stevenengler (approved):

docs/shadow_config_spec.md
```diff
@@ -661,6 +662,33 @@ environment:
...
 
+#### `hosts.<hostname>.processes[*].expected_final_state`
+
+Default: !exited 0
+Type: !exited \<Integer\> OR !signaled \<Signal String or Integer\> OR !running
+
+The expected state of the process at the end of the simulation. If the process
+exits before the end of the simulation with an unexpected state, or is still running
+at the end of the simulation when this was not `!running`, shadow will log an error
+and return a non-zero status for the simulation.
```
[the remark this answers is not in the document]
sporksmith: Changed to not use `!`

docs/shadow_config_spec.md
sporksmith: done

docs/shadow_config_spec.md
sporksmith: https://github.com/shadow/benchmark/pull/7 is q'd up to address this

PR #2893, opened 2023-04-26
Replace hosts file with a memfd

stevenengler:
This means that Shadow's fake hosts file isn't written to the filesystem and won't clutter /tmp with hundreds of files.

### DIFF
--- a/src/main/routing/dns.c
+++ b/src/main/routing/dns.c
@@ -283,6 +283,9 @@ gchar* dns_getHostsFilePath(DNS* dns) {
 
     g_mutex_unlock(&dns->lock);
 
+    // TODO: there's a race condition here where another thread could clean up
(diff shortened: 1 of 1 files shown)


sporksmith (approved):
Cool! I wonder if we could do the same with the memory files created by the shmem allocator and the MemoryMapper.

I think at one time we thought it might be useful to have the files around later to inspect when debugging, but I don't think they've been useful in practice.

stevenengler:
> Cool! I wonder if we could do the same with the memory files created by the shmem allocator and the MemoryMapper.
> 
> I think at one time we thought it might be useful to have the files around later to inspect when debugging, but I don't think they've been useful in practice.

I think so, and one benefit would be that we don't run into issues due to a small /dev/shm mount as we do now in Docker containers.

PR #2895, opened 2023-04-27
Log system memory info during the heartbeat

stevenengler:
In the future, we should be able to use this data to compute memory usage in tornettools instead of running `/usr/bin/free` in a loop. While I think it would be best to not write json directly in the log file, I think this is fine until we have better structured logging. Even though it's json, we'll still consider it unsupported along with everything else in the log file.

Ex:

```
00:00:01.057619 [546333:shadow] n/a [INFO] [n/a] [manager.rs:765] [shadow_rs::core::manager] System memory usage in bytes at simtime 6120000000 ns reported by /proc/meminfo: {"mem_total":67169312768,"mem_free":48688345088,"swap_total":68719472640,"swap_free":55525654528,"buffers":0,"cached":7910858752,"s_reclaimable":0,"shmem":5169258496}
```

Double-checked that the performance didn't change:

https://github.com/shadow/benchmark-results/tree/master/tor/2023-04-27-T20-50-03

run_time

### DIFF
--- a/src/main/core/manager.rs
+++ b/src/main/core/manager.rs
@@ -16,6 +16,7 @@ use shadow_shim_helper_rs::simulation_time::SimulationTime;
 use shadow_shim_helper_rs::HostId;
 
 use crate::core::controller::{Controller, ShadowStatusBarState, SimController};
+use crate::core::resource_usage;
 use crate::core::scheduler::runahead::Runahead;
 use crate::core::scheduler::{HostIter, Scheduler, ThreadPerCoreSched, ThreadPerHostSched};
 use crate::core::sim_config::{Bandwidth, HostInfo};
@@ -52,6 +53,8 @@ pub struct Manager<'a> {
 
     check_fd_usage: bool,
     check_mem_usage: bool,
+
+    meminfo_file: std::fs::File,
 }
 
 impl<'a> Manager<'a> {
@@ -189,6 +192,9 @@ impl<'a> Manager<'a> {
             )
         })?;
 
+        let meminfo_file =
+            std::fs::File::open("/proc/meminfo").context("Failed to open '/proc/meminfo'")?;
+
         Ok(Self {
             manager_config: Some(manager_config),
             controller,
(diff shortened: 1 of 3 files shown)


sporksmith (approved):
I think as a general principle we've tried to keep this sort of "orthogonal" system monitoring outside of shadow itself for sake of modularity. I was going to ask @robgjansen 's opinion, but I see he added positive reactions to the PR description, so I'll take it he's in favor :)

I'm a little curious about the reasoning, but don't particularly object :)

robgjansen:
> I think as a general principle we've tried to keep this sort of "orthogonal" system monitoring outside of shadow itself for sake of modularity.

I don't immediately recall any such examples, can you? Maybe you're thinking about `tornettools` which currently runs `free` in a loop (and used to run `dstat` too, or maybe still does?...)?

I see what you mean about modularity, but I don't find it particularly odd if Shadow measures its own resource usage. This information is beneficial and very often desirable to understand an experiment, but users may not realize they should be wanting to run `free` during a sim. Running `free` in a loop is particularly brittle as some older versions don't support the `-w` (wide) option which can break out parsing scripts. Measuring and storing the information ourselves in a standardized format would alleviate some of these problems.

FWIW, Tor logs some auxiliary information about its own resource usage as well. In particular, it measures its upstream and downstream bandwidth per second, also some form of memory usage now that it needs to track it for DoS mitigation.

stevenengler:
My thoughts are that I also think in principle it's nice to keep things modular and not make Shadow bloated with things that could be done externally, but on the other hand tor simulations can be painful to run and a small change that simplifies things is worthwhile. For example people might need to run tor simulations without tornettools, and may need to run many sims with different parameters, which typically leads to a bunch of spaghetti bash/python scripts. And then if they need to also run background processes in their scripts (for example free) at the same time and later try to merge its data with the simulation data, it just makes running simulations a lot complicated and painful instead of just doing it easily in Shadow.

sporksmith:
Ok, sounds good

PR #2906, opened 2023-05-04
Drop per-process exitcode files

sporksmith:
The semantics of these were a bit awkward (e.g. 128+signum for death-by-signal; empty for still-running), and the per-process `expected_final_state` attribute replaces their primary use-case.

If we decide to bring these back, we should consider a richer format; maybe just the serialized actual process final state. e.g. "running", "{exited: 0}", "{signaled: SIGINT}", ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,10 @@ renamed the host's `options` field to `host_options`.
 an error by default. This can be overridden by the new per-process option
 `expected_final_state`. https://github.com/shadow/shadow/pull/2886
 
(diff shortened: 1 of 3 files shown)


stevenengler (approved):
I thought we had tests that checked that processes exited correctly by reading the exitcode file, but I guess not.

sporksmith:
> I thought we had tests that checked that processes exited correctly by reading the exitcode file, but I guess not.

I replaced those with using `expected_final_state` in https://github.com/shadow/shadow/pull/2886/commits/cb06833998101ac3648aed00e9128dc719f0aa9e

Issue #2877, opened 2023-04-15
Spinning in the scheduler


stevenengler:
In #2866 we found that if the worker threads spin while waiting for a new task rather than block, the runtime performance on the benchmark runner improves significantly. We should investigate this further, but should also test on systems other than the benchmark runner to make sure that the simulation performance doesn't become worse on other hardware platforms.

I've tried a few different spinning configurations:

## 1: Spin 100 times before blocking

Results: https://github.com/shadow/benchmark-results/tree/master/tgen/2023-04-13-T21-16-06
Shadow: dd651d82caaa3de43e5af5db16b18b1a6684d60b (diff: 0fe3be4ee57615bf91f7b30e223f923298f7ccc3...dd651d82caaa3de43e5af5db16b18b1a6684d60b)

A rust mutex spins 100 times when locking, so I wanted to try that here. But in a tgen simulation, there's no performance improvement:

![1681517713_grim](https://user-images.githubusercontent.com/3708797/232173141-0d3299f7-fe39-44e2-a853-46895f40a579.png)

## 2: Spin `u64::MAX` times before blocking

Results: https://github.com/shadow/benchmark-results/tree/master/tgen/2023-04-14-T02-21-54
Shadow: d0c2f9619e1967cbb36b5956d8ea87379ccb6190 (diff: 0fe3be4ee57615bf91f7b30e223f923298f7ccc3...d0c2f9619e1967cbb36b5956d8ea87379ccb6190)

Using `u64::MAX` to spin indefinitely has a large performance improvement (the same improvement we saw in #2866):

![1681517845_grim](https://user-images.githubusercontent.com/3708797/232173451-540d3423-fbdc-431e-be2c-b1939d9cd9a4.png)

## 3: Spin `u64::MAX` times before blocking and using `std::hint::spin_loop()`

Results: https://github.com/shadow/benchmark-results/tree/master/tgen/2023-04-14-T15-10-23
Shadow: 2017ac16be874dc12f2719a32fa15c2de57ba9bc (diff: 0fe3be4ee57615bf91f7b30e223f923298f7ccc3...2017ac16be874dc12f2719a32fa15c2de57ba9bc)

Like the previous version we spin `u64::MAX` times, but tell the CPU that it's a spin loop using `std::hint::spin_loop()`. We see the same performance improvement, but this is maybe better for energy efficiency:

[...]

PR #2909, opened 2023-05-04
Enable the experimental worker spinning by default

stevenengler:
This enables it in all cases unless opted out of. The performance impact is documented in #2877. We may change this before the 3.0 release.

### DIFF
--- a/src/main/core/support/configuration.rs
+++ b/src/main/core/support/configuration.rs
@@ -475,7 +475,7 @@ impl Default for ExperimentalOptions {
             unblocked_vdso_latency: Some(units::Time::new(10, units::TimePrefix::Nano)),
             use_memory_manager: Some(true),
             use_cpu_pinning: Some(true),
-            use_worker_spinning: Some(false),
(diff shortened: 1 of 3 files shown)


src/main/core/support/configuration.rs
```diff
@@ -475,7 +475,7 @@ impl Default for ExperimentalOptions {
             unblocked_vdso_latency: Some(units::Time::new(10, units::TimePrefix::Nano)),
             use_memory_manager: Some(true),
             use_cpu_pinning: Some(true),
-            use_worker_spinning: Some(false),
+            use_worker_spinning: Some(true),
```
sporksmith, line 478: Maybe worth mentioning in the CHANGELOG, since it'll potentially make shadow a bit more cpu-hungry?

sporksmith (approved):

src/main/core/support/configuration.rs
stevenengler: Oops good catch!

PR #2911, opened 2023-05-08
Fix `host_option_defaults` config option

stevenengler:
The option was named properly in the documentation but not the code.


src/main/core/support/configuration.rs
sporksmith, line 1: Probably worth a mention in the changelog :)

sporksmith (approved):

src/main/core/support/configuration.rs
stevenengler: Added.

PR #2913, opened 2023-05-08
MemoryCoper: use live thread ID for writes

sporksmith:
In particular this fixes a crash when reaping a thread and trying to
write to the `child_tid` address, since the current thread is no longer
live and can't be used with the `process_vm_writev` syscall.

This problem was mostly latent, since when the memory mapper is enabled,
we usually do the write through shared memory instead.


src/main/host/memory_manager/memory_copier.rs
```diff
@@ -215,16 +215,22 @@ impl MemoryCopier {
...
-
-        // Don't access another process's memory.
-        assert_eq!(active_pid, self.pid);
+        let tid = Worker::with_active_host(|host| {
+            Worker::with_active_process(|process| {
+                // Don't access another process's memory.
+                assert_eq!(process.native_pid(), self.pid);
+                let thread = process.first_live_thread_borrow(host.root()).unwrap();
+                let thread = thread.borrow(host.root());
+                thread.native_tid()
```
stevenengler, line 227: Does shadow require that threads share memory? My understanding is that Linux doesn't really have threads, just processes that may or may not have shared resources (depending on the arguments to `clone`) and can be grouped into "thread groups" if `CLONE_THREAD` is set. In shadow's `clone` handler, do we require that `CLONE_VM` is always set? And if so, maybe document in shadow's `clone` handler that we require it for the memory copier to work, so that we don't accidentally relax that requirement later and break the memory copier?

stevenengler (approved):

src/main/host/memory_manager/memory_copier.rs
sporksmith: Yes, shadow currently requires `CLONE_THREAD`; and in both Linux and shadow if `CLONE_THREAD` is set, then `CLONE_SIGHAND` and `CLONE_VM` must also be set.

Added a comment for the `memory_manager` field in `RunningProcess`

PR #2916, opened 2023-05-09
Update kernel support policy

sporksmith:


docs/semver.md
```diff
@@ -21,6 +21,9 @@ versions:
 - Support for any of Shadow's [supported platforms](supported_platforms.md)
   will not be dropped, unless those platforms no longer receive free updates
   and support from the distribution's developer.
+- We will not increase the minimum Linux kernel version beyond the version of the
+General Availability kernel for all supported platforms [supported
+platforms](supported_platforms.md).
```
stevenengler, line 26: "supported platforms" is stated twice. Also maybe add a disclaimer ("some exceptions apply")?

src/main/core/main.rs
```diff
@@ -178,6 +219,10 @@ pub fn run_shadow(build_info: &ShadowBuildInfo, args: Vec<&OsStr>) -> anyhow::Re
     unsafe { c::main_logBuildInfo(build_info) };
     log_environment(args.clone());
 
+    if let Err(e) = verify_supported_system() {
+        log::error!("Couldn't verify supported system: {e:?}")
+    }
```
stevenengler, line 224: nit: maybe `warn` instead of `error` since it's not something that will cause shadow to exit (not a hard error)?

stevenengler (approved):

docs/semver.md
stevenengler: Or maybe don't mention the "General Availability kernel" and just say that we won't increase it beyond what is stated in the supported platforms doc.

sporksmith: Tweaked the wording here - wdyt?

stevenengler: Looks good to me.

src/main/core/main.rs
stevenengler: > fwiw at some point I think we decided to use error for things that strongly indicate that the simulation results may be invalid.

I though it was the opposite, that we would never use the error log level since we could return an `anyhow::Result` in that case and bubble it up to the top. In the C code I think shadow used to always use the error log level when it was about to exit (for example if an assertion failed).

But yeah I agree we should have this documented. I don't think it was documented since we never really reached a consensus. There are some notes from that discussion in the riseup pad from 2022-03-24.

PR #2918, opened 2023-05-09
Move common C flags to root CMakeLists.txt

sporksmith:
In particular our remaining pure C targets, such as the shim, were missing the debug flags.


CMakeLists.txt
```diff
@@ -167,6 +167,18 @@ else()
     MESSAGE(FATAL_ERROR "Unknown build type '${CMAKE_BUILD_TYPE}'; valid types are 'Release' or 'Debug'")
 endif()
 
+# Sync with ShadowBuildCommon::cc_flags
```
stevenengler, line 170: this comment says to sync with `cc_flags`, but you removed the flags in `cc_build` so they're not in sync.

CMakeLists.txt
```diff
@@ -167,6 +167,18 @@ else()
...
+# TODO: move remaining C targets to Rust (even if just a Rust build script that
+# builds C files).
+add_compile_options(
+  # Always include debug symbols and frame pointers for ease of debugging
+  -ggdb
+  -fno-omit-frame-pointer
+  # Enable some particularly useful warnings
+  -Wreturn-type
+  -Wswitch
+)
```
stevenengler, line 180: do these get passed on to the rust compiler somehow? does cmake put these in `CFLAGS` for example?

stevenengler (approved):

CMakeLists.txt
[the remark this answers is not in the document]
sporksmith: Oops, I'd meant to delete this comment after verifying that these get propagated to the Rust build.

sporksmith: Yup, initially I thought we'd need to sync these, but they're already propagated from here to the cc_builds automatically.

CMakeLists.txt
sporksmith: Yup; added a comment to make this clearer.

PR #2925, opened 2023-05-11
Implement clone3

sporksmith:
Implements clone3, currently using clone internally.

This also requires restoring general purpose registers from the original context of the syscall, since glibc's clone3 implementation assumes they still have the original values from the parent in the newly created child thread.

### DIFF
--- a/src/lib/shim/shim_syscall.c
+++ b/src/lib/shim/shim_syscall.c
@@ -51,19 +51,13 @@ long __attribute__((noinline)) shim_clone(const ucontext_t* ctx, int32_t flags, 
     long rv = 0;
     register pid_t* r10 __asm__("r10") = ctid;
     register long r8 __asm__("r8") = newtls;
-    // Store address of shim_ensure_init.
-    // TODO: We ought to be able to put the literal "shim_ensure_init" in
-    // the inline assembly and have the assembler resolve the address for
-    // us, but on ubuntu 18.04's gcc this ends up generating a relocation it
-    // can't resolve at the link step.
-    register long r13 __asm__("r13") = (long)&shim_ensure_init;
     __asm__ __volatile__(
         "syscall\n"
         // If in the parent, done with asm.
         "cmp $0, %%rax\n"
         "jne shim_native_syscallv_out\n"
         // Initialize state for this thread
-        "callq *%%r13\n"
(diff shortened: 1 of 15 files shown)


src/lib/shim/shim_syscall.c
```diff
@@ -16,6 +16,60 @@
...
+    register pid_t* r10 __asm__("r10") = ctid;
+    register long r8 __asm__("r8") = newtls;
+    // Store clone_rip in a callee-save register, so that we can call shim_ensure_init
+    // without clobbering it.
+    register long r12 __asm__("r12") = (long)clone_rip;
+    // Store address of shim_ensure_init in a callee-save register.
+    // TODO: We ought to be able to put the literal "shim_ensure_init" in
+    // the inline assembly and have the assembler resolve the address for
+    // us, but on ubuntu 18.04's gcc this ends up generating a relocation it
+    // can't resolve at the link step.
```
stevenengler, line 58: Off topic: Is this fixable now?

src/main/host/syscall/handler/clone.rs
```diff
@@ -129,6 +120,80 @@ impl SyscallHandler {
...
+        if args_size != std::mem::size_of::<libc::clone_args>() {
+            // TODO: allow smaller size, and be careful to only read
+            // as much as the caller specified, and zero-fill the rest.
+            return Err(Errno::EINVAL.into());
+        }
+        let args = ctx.objs.process.memory_borrow().read(args)?;
+        trace!("clone3 args: {args:?}");
+        let Ok(flags) = i32::try_from(args.flags) else {
+            debug!("Couldn't safely truncate flags to 32 bits: {} ({:?})", args.flags,
+            CloneFlags::from_bits(args.flags));
```
stevenengler, line 161: super nit: indent the second line or something to make it a little clearer.

stevenengler (approved):

src/lib/shim/shim_syscall.c
sporksmith: Looks like it is!

src/main/host/syscall/handler/clone.rs
sporksmith: Done; apparently rustfmt doesn't handle these macros (maybe any macros?).

Issue #2766, opened 2023-02-26
shim unsafely uses the libc allocator via the shmem allocator


haxxpop:
I found that sometimes, when the native syscall is turned on (using `shim_swapAllowNativeSyscalls(true)`), the shim uses glibc's calloc to allocate memory. This can causes a problem because internally calloc can trigger a futex wait syscall.

You can see that calloc can trigger a futex wait syscall by looking at the following stack trace.
```
(gdb) bt
#0  0x0000155554df30b2 in shim_native_syscallv () from /root/.local/bin/../lib/../lib/libshadow-shim.so
#1  0x0000155554df3a5c in shim_syscallv () from /root/.local/bin/../lib/../lib/libshadow-shim.so
#2  0x0000155554df3b92 in shim_syscall () from /root/.local/bin/../lib/../lib/libshadow-shim.so
#3  0x0000155554df1fe8 in _shim_seccomp_handle_sigsys () from /root/.local/bin/../lib/../lib/libshadow-shim.so
#4  <signal handler called>
#5  futex_wait (private=0, expected=2, futex_word=0x15551c000030) at ../sysdeps/nptl/futex-internal.h:146
#6  __GI___lll_lock_wait_private (futex=futex@entry=0x15551c000030) at ./nptl/lowlevellock.c:34
#7  0x000015555501e8a0 in __libc_calloc (n=n@entry=1, elem_size=elem_size@entry=288) at ./malloc/malloc.c:3650
#8  0x0000155554e2bfd1 in shmemserializer_blockDeserialize (serializer=0x55555c906f00, serial=0x15554da08f88) at shmem_allocator.c:505
#9  0x0000155554df0803 in shim_newThreadStart () from /root/.local/bin/../lib/../lib/libshadow-shim.so
#10 0x0000155554df343c in shim_emulated_syscallv () from /root/.local/bin/../lib/../lib/libshadow-shim.so
#11 0x0000155554df3a3f in shim_syscallv () from /root/.local/bin/../lib/../lib/libshadow-shim.so
#12 0x0000155554df3b92 in shim_syscall () from /root/.local/bin/../lib/../lib/libshadow-shim.so
#13 0x0000155554df1fe8 in _shim_seccomp_handle_sigsys () from /root/.local/bin/../lib/../lib/libshadow-shim.so
#14 <signal handler called>
#15 clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:83
...
``` 
Or by looking at the source code of glibc directly at
[...]

PR #2926, opened 2023-05-12
Moving the shared memory alloactor to Rust

RWails:
Opening this as an early draft PR so that @sporksmith can roast me for my bad no-std system call code; see `src/lib/shmem/src/raw_syscall.rs`. :<)

More seriously, if the raw syscall code is more broadly useful, we should promote it above the shmem module. If there's a better way to spell it or a crate that already provides the functionality we need, we should ditch the `raw_syscall` code and use that instead.


src/lib/shmem/src/raw_syscall.rs
```diff
@@ -0,0 +1,222 @@
...
+
+unsafe fn interpret_str_to_char_ptr(s: &str) -> *const char {
+    unsafe { core::mem::transmute::<*const u8, *const char>(s.as_bytes().as_ptr()) }
+}
+
+fn interpret_i32_as_u64(x: i32) -> u64 {
+    u64::from_ne_bytes((x as i64).to_ne_bytes())
+}
+
+fn interpret_u64_as_i32(x: u64) -> i32 {
```
sporksmith, line 21: I think just `x as i32` works

src/lib/shmem/src/raw_syscall.rs
```diff
@@ -0,0 +1,222 @@
...
+    let x_bytes = x.to_ne_bytes();
+    if cfg!(target_endian = "big") {
+        i32::from_ne_bytes(x_bytes[4..8].try_into().unwrap())
+    } else {
+        i32::from_ne_bytes(x_bytes[0..4].try_into().unwrap())
+    }
+}
+
+fn convert_i32_rv_to_rv_errno(rv: i32) -> Result<i32, i32> {
+    if rv < -1 {
```
sporksmith, line 31: This isn't right for, e.g. `mmap`. It should be something like `if (-4095..=-1).contains(rv)`

https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/x86_64/sysdep.h;h=24d8b8ec20a55824a4806f8821ecba2622d0fe8e;hb=HEAD#l41

src/lib/shmem/src/raw_syscall.rs
```diff
@@ -0,0 +1,222 @@
...
+        Err(-1 * rv)
+    } else {
+        Ok(rv)
+    }
+}
+
+fn x86_64_syscall(
```
sporksmith, line 38: I've been experimenting a little bit with the [linux-syscall](https://crates.io/crates/linux-syscall) crate; I think it's probably worth using for its [`syscall!` macro](https://docs.rs/linux-syscall/1.0.0/linux_syscall/macro.syscall.html)

src/lib/shmem/src/raw_syscall.rs
```diff
@@ -0,0 +1,222 @@
...
+        filename,
+        interpret_i32_as_u64(flags),
+        mode as u64,
+        0,
+        0,
+        0,
+    ))
+}
+
+pub fn open(filename: &str, flags: i32, mode: u32) -> Result<i32, i32> {
```
sporksmith, line 83: I was going to suggest  taking a `&Path` here but whoops of course that's `std` :). In this case it's probably better to take something like a `&[u8]` and let the caller worry about any necessary conversion and ensuring that it's null-terminated. (But it'd be worth having an assertion or check here to ensure it really is NULL-terminated)

Ultimately it'd be nice use [`bitflags`](https://crates.io/crates/bitflags) to generate types to use here, and maybe to avoid pulling in the `libc` crate by generating our own bindings from the kernel headers. I started to do the latter in the [crate I'm sketching out](https://github.com/shadow/shadow/pull/2930)

src/lib/shmem/src/raw_syscall.rs
```diff
@@ -0,0 +1,222 @@
...
+        0,
+        0,
+        0,
+    )))
+}
+
+fn unlink_impl(filename: *const char) -> i32 {
+    const SYS_UNLINK: u64 = 87;
+
+    let filename = unsafe { core::mem::transmute::<*const char, u64>(filename) };
```
sporksmith, line 107: Should be possible to do this with regular casts, but as above need to be careful that the string is null terminated. I think just `filename as u64` should work.

src/lib/shmem/src/raw_syscall.rs
```diff
@@ -0,0 +1,222 @@
...
+pub fn munmap(bytes: &mut [u8]) -> Result<(), (&mut [u8], i32)> {
+    let p = unsafe { core::mem::transmute::<*mut u8, u64>(bytes.as_mut_ptr()) };
+    let nbytes = bytes.len();
+
+    let rv = munmap_impl(p, nbytes.try_into().unwrap());
+
+    if rv == 0 {
+        Ok(())
+    } else {
+        Err((bytes, -1 * rv))
```
sporksmith, line 162: No need to return `bytes`; the caller still has access to it.

sporksmith (commented):
Looks like a good start!

At a high level I think it's worth using the [linux-syscall](https://crates.io/crates/linux-syscall) crate for the raw `syscall` macro  (and syscall numbers)

I'll work on getting my ["linux api"](https://github.com/shadow/shadow/pull/2930) crate cleaned up and merged, and then I think the wrappers you're writing for individual syscalls (e.g. `open`) would make sense to add there

RWails:
Oops, I marked this ready for review before resolving two open comments. I'll fix and then all the changes should be ready.

src/lib/shmem/src/shmalloc_impl.rs
```diff
@@ -0,0 +1,739 @@
...
+pub(crate) type PathBuf = [u8; PATH_BUF_NBYTES];
+
+use vasi::VirtualAddressSpaceIndependent;
+
+fn get_null_path_buf() -> PathBuf {
+    [0; PATH_BUF_NBYTES]
+}
+
+#[cfg(debug_assertions)]
+type CanaryBuf = [u8; 4];
```
sporksmith, line 27: We don't need to do it now, but maybe we could unify this with [`Magic`](https://shadow.github.io/docs/rust/shadow_rs/utility/struct.Magic.html). I think we'd need to move it to a different crate  though, and make it `VirtualAddressSpaceIndependent`, which might mean not storing `TypeId` directly.

src/lib/shmem/src/shmalloc_impl.rs
```diff
@@ -0,0 +1,739 @@
...
+
+#[cfg(debug_assertions)]
+fn get_canary_buf() -> CanaryBuf {
```
sporksmith, line 30: optional - I think this could be a const? e.g.

`const CANARY: CanaryBuf =   [0xDE, 0xAD, 0xBE, 0xEF];`

src/lib/shmem/src/shmalloc_impl.rs
```diff
@@ -0,0 +1,739 @@
...
+    Open,
+    FTruncate,
+    MMap,
+    MUnmap,
+    Unlink,
+    WrongAllocator,
+    Leak,
+}
+
+pub(crate) fn log_error(error: AllocError, errno: Option<i32>) {
```
sporksmith, line 66: Probably better to use the `log` crate than write directly to `stderr` --- The `log` crate supports `no_std`, and we have a `no_std` logger in the shim.

Maybe even better to have the error type include the errno value, implement `Debug` for that type, and propagate errors up to the user of this crate to let them decide what to do with it.

src/lib/shmem/src/shmalloc_impl.rs
```diff
@@ -0,0 +1,739 @@
...
+        }
+        unsafe {
+            write(STDERR_FD, b"\n".as_ptr() as *const _, 1);
+        }
+    }
+}
+
+pub(crate) fn log_error_and_exit(error: AllocError, errno: Option<i32>) -> ! {
+    log_error(error, errno);
+    let _ = kill(0, linux_api::signal::Signal::SIGABRT.into());
```
sporksmith, line 104: `kill` with pid=0 doesn't do what you expect. This is a mistake I've made before, with "fun" results :)

kill(2):

> If pid equals 0, then sig is sent to every process in the process group of the calling process.

Also we're not guaranteed that the signal will be delivered to the current thread, in which case `kill` might return before the signal is actually delivered.

I think what we actually want is something like: `tgkill(getpid(), gettid(), SIGABRT)`. https://shadow.github.io/docs/rust/linux_api/signal/fn.tgkill.html

src/lib/shmem/src/shmalloc_impl.rs
```diff
@@ -0,0 +1,739 @@
...
+    unreachable!()
+}
+
+fn format_shmem_name(buf: &mut [u8]) {
```
sporksmith, line 108: Can probably simplify or replace this using [FormatBuffer](https://shadow.github.io/docs/rust/formatting_nostd/struct.FormatBuffer.html)

src/lib/shmem/src/shmalloc_impl.rs
```diff
@@ -0,0 +1,739 @@
...
+        self.canary_back = get_canary_buf();
+    }
+
+    fn canary_check(&self) -> bool {
+        self.canary_front == get_canary_buf() && self.canary_back == get_canary_buf()
+    }
+}
+
+impl Block {
+    /// Gets the data corresponding to the block
```
sporksmith, line 324: Stale comment block?

src/lib/shmem/src/shmalloc_impl.rs
sporksmith, line 1: I've only skimmed this file so far; I'll try to take a closer look next week.

Ultimately it'd be nice if we could decouple the core allocation logic from the shared memory allocator, using some parameterizable backing store/allocator. That'd let us validate the tricky pointer manipulation under miri (e.g. using `std::Box` as a backing allocator under test), and maybe use that as a global allocator in the shim using `mmap` or a statically allocated backing store.

Ok if we want to save that for a future potential improvement, though. Getting this code from C to Rust is a good first step before doing further refactoring...

RWails: Thanks for the amazing feedback so far @sporksmith !! I should be able to get thru a bunch of your comments before next week.

src/lib/shmem/src/shmalloc_impl.rs
```diff
@@ -0,0 +1,739 @@
...
+        self.canary_front == get_canary_buf() && self.canary_back == get_canary_buf()
+    }
+}
+
+fn allocate_shared_chunk(path_buf: &PathBuf, nbytes: usize) -> *mut Chunk {
+    let (p, fd) = create_map_shared_memory(path_buf, nbytes);
+
+    // Zero the memory so that we do not have to worry about junk between blocks.
+    unsafe {
+        core::ptr::write_bytes::<u8>(p.as_mut_ptr(), 0x00, nbytes);
```
sporksmith, line 250: Shouldn't be needed.

From `ftruncate(2)`
> If the file previously was shorter, it is extended, and the extended part reads as null bytes ('\0')

sporksmith (commented):
Took another look at the impl; just had another one or two small comments, though TBH I'm mostly leaning on our tests to find problems vs trying to prove to myself that it's correct :).

At a high level I think it's worth getting this in relatively soon and iterating on improvements, since e.g. it should immediately fix https://github.com/shadow/shadow/issues/2766. For most of the comments I've left I'd be happy with just adding `TODO` comments so we remember to follow-up later.

src/lib/shmem/src/raw_syscall.rs
RWails: fixed

src/lib/shmem/src/shmalloc_impl.rs
RWails: fixed

src/lib/shmem/src/shmalloc_impl.rs
RWails: fixed

src/lib/shmem/src/allocator.rs
```diff
@@ -1,335 +1,456 @@
...
-    ///   that `block` was serialized from. We can't guarantee this with normal
-    ///   lifetime analysis, since the original block may be in another process.
-    pub unsafe fn deserialize<'a, T>(
-        &'a self,
-        block: &ShMemBlockSerialized,
-    ) -> ShMemBlockAlias<'a, T>
+    /// This function can violate type safety if a template type is provided that does not match
+    /// original block that was serialized.
+    pub unsafe fn deserialize<T>(
+        &mut self,
```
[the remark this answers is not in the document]
RWails: Agreed, we should fix the design with a wrapper type. Going to leave this as a todo for now.

src/lib/shmem/src/allocator.rs
```diff
@@ -1,335 +1,456 @@
...
+        let t_nbytes: usize = core::mem::size_of::<T>();
+        let t_alignment: usize = core::mem::align_of::<T>();
+
+        let block = self.internal.alloc(t_nbytes, t_alignment);
+        unsafe {
+            (*block).get_mut_ref::<T>()[0] = val;
+        }
+
+        self.nallocs += 1;
+        ShMemBlock::<'alloc, T> {
```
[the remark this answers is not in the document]
RWails: Leaving this as a todo for now.

src/lib/shmem/src/raw_syscall.rs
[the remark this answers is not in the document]
RWails: Leaving this as a todo for now.

src/lib/shmem/src/shmalloc_impl.rs
RWails: Leaving this as a todo

src/lib/shmem/src/shmalloc_impl.rs
RWails: I like the idea of using a generic allocator frontend with a configurable backend; I do think we should eventually move to such a design. One piece we want to work out is if the allocator supports extensions, eg, serializable memory, the allocator itself will need some extra knowledge about the backing store. Should be doable with the right traits, I think. Let's plan for this kind of design, but go ahead and merge this for now.

RWails:
OK @sporksmith nagging you one more time. The latest round of changes should address your comments in (hopefully) unsurprising ways.

I did update the PR to tear out all the old C code. I think I updated the build system appropriately to remove any references, but you may want to peek just to make sure.

src/lib/shmem/src/allocator.rs
```diff
@@ -210,62 +228,54 @@ pub struct ShMemBlockSerialized {
...
+                .parse::<isize>()
+                .map_err(Err::<(), core::num::ParseIntError>)
+                .unwrap();
 
-        let mut path_buf: crate::shmalloc_impl::PathBuf =
-            [0; crate::shmalloc_impl::PATH_BUF_NBYTES];
+            let mut chunk_format = FormatBuffer::<{ crate::util::PATH_MAX_NBYTES }>::new();
 
-        path_buf.iter_mut().zip(rhs_itr).for_each(|(x, y)| *x = *y);
+            write!(&mut chunk_format, "{}", &path_str[1..]).unwrap();
```
sporksmith, line 259: This will panic if `path_str` is 0-length; e.g. if the original string started with `;`. It's also a bit unclear why the first character is being ignored. If the intent is to strip a leading '/', then maybe:

```
let Some(stripped_path_str) = path_str.strip_prefix('/') else {
  return Err(anyhow::anyhow!("missing leading /"))
};
```

src/lib/shmem/src/shmalloc_impl.rs
```diff
@@ -1,4 +1,4 @@
-//! This module implements a low-level, unsafe shared memory allocator that uses mmap'ed shared
+// //! This module implements a low-level, unsafe shared memory allocator that uses mmap'ed shared
```
sporksmith, line 1: Stray `//` ?

sporksmith (approved):

src/lib/shmem/src/shmalloc_impl.rs
RWails: Fixed

src/lib/shmem/src/allocator.rs
```diff
@@ -210,62 +228,54 @@ pub struct ShMemBlockSerialized {
...
-        let lhs_itr = string_buf.0.iter();
-        let mut rhs_itr = string_buf.0.iter();
-        rhs_itr.find(|x| **x == DELIM);
+    // Required method
+    fn from_str(s: &str) -> anyhow::Result<Self> {
+        use core::fmt::Write;
+        use formatting_nostd::FormatBuffer;
 
-        let offset_itr = lhs_itr.take_while(|x| **x != DELIM);
+        if let Some(split_point) = s.find(';') {
```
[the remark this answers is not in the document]
RWails: Good catch. The odd `&path_str[1..]` argument was due to the `;` still being included in the path_str. This adjustments fixes the issue.

PR #2928, opened 2023-05-12
Proof-of-concept no_std test shim

sporksmith:
Not ready for review, but demonstrates what a "no_std" test library, and eventually the real shim, might look like. The build will fail if this crate ends up pulling in `std`, e.g. through a dependency.

It looks like the vasi derive macro is ok despite depending on non-no_std macro crates, presumably since they're only actually used at at compile time and get dropped in the link step.


sporksmith:
I think this has more-or-less served its purpose. we've begun migrating the shim to be no_std.

PR #2935, opened 2023-05-15
clone handler: handle individual flags more explicitly

sporksmith:
This is mostly a refactor in preparation for supporting fork. https://github.com/shadow/shadow/issues/1987

There are also a couple minor behavioral changes, that shouldn't affect any of our supported thread libraries:

* Some flags that we don't actually care about are no longer required (e.g. CLONE_SYSVEM).
* CLONE_SETTLS *is* now required. Previously we would have tried to go ahead and execute the native clone without it, but the shim would misbehave since it depends on thread local storage.

### DIFF
--- a/src/main/host/syscall/handler/clone.rs
+++ b/src/main/host/syscall/handler/clone.rs
@@ -64,55 +64,161 @@ impl SyscallHandler {
             return Err(Errno::EINVAL.into());
         };
 
-        {
-            let required_flags = CloneFlags::CLONE_VM
-                | CloneFlags::CLONE_FS
-                | CloneFlags::CLONE_FILES
-                | CloneFlags::CLONE_SIGHAND
-                | CloneFlags::CLONE_THREAD
-                | CloneFlags::CLONE_SYSVSEM;
-            let missing_flags = required_flags.difference(flags);
-            if !missing_flags.is_empty() {
-                warn!("Missing required clone flags: {missing_flags:?}");
+        // We use this for a consistency check to validate that we've inspected
+        // and emulated all of the provided flags.
+        let mut handled_flags = CloneFlags::empty();
+
+        // The parameters that we'll pass to the native clone call.
+        let mut native_flags = CloneFlags::empty();
+
+        // We emulate the flags that would use these, so we always pass NULL to
+        // the native call.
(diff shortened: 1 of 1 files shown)


src/main/host/syscall/handler/clone.rs
```diff
@@ -64,55 +64,154 @@ impl SyscallHandler {
...
                 warn!("Exit signal is unimplemented");
                 return Err(Errno::ENOTSUP.into());
             }
+
+            assert!(!native_flags
+                .contains(CloneFlags::CLONE_CHILD_SETTID | CloneFlags::CLONE_CHILD_CLEARTID));
+            let ctid = ForeignPtr::<libc::pid_t>::null();
+
+            assert!(!native_flags.contains(CloneFlags::CLONE_PARENT_SETTID));
+            let ptid = ForeignPtr::<libc::pid_t>::null();
```
stevenengler, line 185: Add comment explaining what these assertions are for?

src/main/host/syscall/handler/clone.rs
```diff
@@ -64,55 +64,154 @@ impl SyscallHandler {
...
-                | CloneFlags::CLONE_CHILD_SETTID
-                | CloneFlags::CLONE_CHILD_CLEARTID;
-            let filtered_flags = flags.difference(flags_to_emulate);
             if raw_exit_signal != 0 {
                 warn!("Exit signal is unimplemented");
                 return Err(Errno::ENOTSUP.into());
             }
+
+            assert!(!native_flags
+                .contains(CloneFlags::CLONE_CHILD_SETTID | CloneFlags::CLONE_CHILD_CLEARTID));
```
stevenengler, line 181: This only asserts that both flags aren't in `native_flags`, but it's still fine if just one of the flags is in `native_flags`. Is this intended?

stevenengler (approved):

src/main/host/syscall/handler/clone.rs
sporksmith: Oops, you're right this was a bug.

src/main/host/syscall/handler/clone.rs
sporksmith: Rearranged and added comments

PR #2936, opened 2023-05-15
Fix two warnings on nightly

stevenengler:
```
warning: variable does not need to be mutable
  --> lib/syscall-logger/src/lib.rs:63:13
   |
63 |         let mut input_fn = match &mut input {
   |             ----^^^^^^^^
   |             |
   |             help: remove this `mut`
   |
   = note: `#[warn(unused_mut)]` on by default
```

```
warning: variable does not need to be mutable
   --> main/host/syscall/io.rs:392:9
    |
392 |     let mut plugin_msg = &mut mem_ref.deref_mut()[0];
    |         ----^^^^^^^^^^
    |         |
    |         help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default
```

### DIFF
--- a/src/lib/syscall-logger/src/lib.rs
+++ b/src/lib/syscall-logger/src/lib.rs
@@ -60,7 +60,7 @@ pub fn log_syscall(args: TokenStream, input: TokenStream) -> TokenStream {
     let context_arg_name;
 
     {
-        let input_fn = match &mut input {
+        let input_fn: &mut _ = match &mut input {
             syn::Item::Fn(input_fn) => input_fn,
             _ => panic!("Expected Item::Fn"),
         };
--- a/src/lib/syscall-logger/src/lib.rs
+++ b/src/lib/syscall-logger/src/lib.rs
@@ -60,7 +60,7 @@ pub fn log_syscall(args: TokenStream, input: TokenStream) -> TokenStream {
     let context_arg_name;
 
     {
-        let mut input_fn = match &mut input {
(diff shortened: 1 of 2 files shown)


src/lib/syscall-logger/src/lib.rs
```diff
@@ -60,7 +60,7 @@ pub fn log_syscall(args: TokenStream, input: TokenStream) -> TokenStream {
     let context_arg_name;
 
     {
-        let mut input_fn = match &mut input {
+        let input_fn = match &mut input {
```
sporksmith, line 63: At first glance it's not obvious how this compiles without the `mut`. Is it because  the inner captured type is already a `&mut`? Maybe it'd be helpful to have an explicit type in the `let`, if it's not horribly verbose.

stevenengler: Yeah, since just the reference has to be mutable, but the thing holding the reference (`input_fn`) doesn't need to be mutable. And if you match on a reference to an enum, the associated data for the variant will also be a reference.

I think it's similar to the following:

```rust
let mut x = Some(1);
let y = match &mut x {
    Some(a) => a,
    None => panic!(),
};
*y = 5;
```

stevenengler: I added a `&mut _` to show that it's a reference.

sporksmith: Thanks, LGTM!

PR #2937, opened 2023-05-15
ChildPidWatcher: use pidfd_open

sporksmith:
The previous design using a pipe would have been painful to use while implementing fork, since we would somewhow want the two ends of the pipe owned by shadow and its "grandchild" process being forked. https://github.com/shadow/shadow/issues/1987
    
Conversely, using pidfd_open is simpler and more flexible in general. We couldn't use it before since it requires Linux 5.3 or later, but we now require such a kernel.


src/main/host/managed_thread.rs
```diff
@@ -1,7 +1,7 @@
 use std::cell::{Cell, RefCell};
 use std::ffi::{CStr, CString};
-use std::fs::File;
-use std::os::fd::{FromRawFd, RawFd};
+
+use std::os::fd::RawFd;
 use std::sync::{atomic, Arc};
```
stevenengler, line 5: super nit: I think we want the block of `std` imports to be together.

src/main/utility/childpid_watcher.rs
```diff
@@ -266,35 +263,34 @@ impl ChildPidWatcher {
...
-    /// *solely* by `pid`, causing `read_fd` to become invalid when `pid` exits.
-    /// In a multi-threaded program care must be taken to prevent a concurrent
-    /// fork from leaking the write end of the pipe into other children. One way
-    /// to avoid this is to use O_CLOEXEC when creating the pipe, and then unset
-    /// O_CLOEXEC in the child before calling exec.
+    /// Register interest in `pid`.
     ///
-    /// Be sure to close the parent's write-end of the pipe.
+    /// Will succeed even if `pid` is already dead, in which case callbacks
+    /// registered for this `pid` will run immediately.
```
stevenengler, line 269: nit: maybe reword to not say immediately? To me, a function saying something will run immediately makes me think it will run in that function (before the function returns), but that's not true in this case.

stevenengler (approved):

src/main/utility/childpid_watcher.rs
```diff
@@ -224,28 +230,19 @@ impl ChildPidWatcher {
         fork_syscall: i64,
         child_fn: impl FnOnce(),
     ) -> Result<Pid, nix::Error> {
-        let (read_fd, write_fd) = nix::unistd::pipe2(OFlag::O_CLOEXEC)?;
-
         // TODO: Allow vfork when Rust supports it:
         assert!(fork_syscall == libc::SYS_fork);
         let raw_pid = unsafe { libc::syscall(fork_syscall) };
```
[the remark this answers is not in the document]
sporksmith: When I wrote it, initially there was also a `vfork_watchable` that also called `fork_watchable_internal`. Went ahead and inlined it.

I had tried using `impl FnOnce() -> !`, but it's not available in stable. Added a comment. https://github.com/rust-lang/rust/issues/35121

src/main/utility/childpid_watcher.rs
sporksmith: Changed to "will immediately be scheduled to run"

PR #2948, opened 2023-05-18
Rework process initialization for fork

sporksmith:
The previous initialization had a complex inter-relationship between ManagedThread, Thread, and Process; e.g. we needed the Process's shared memory handle to create the ManagedThread, but don't create the Process itself until the ManagedThread and Thread have been created. Additionally, the thread shared memory handle was being sent through an env variable, which won't work with fork.

The first thing this does is detangle that so that we create a ManagedThread first, then a Thread, then a Process. We do this by moving the Process's and Thread's shared memory handles into an initialization message that is sent the first time we want to run the thread, so that we no longer need them when *creating* thread.

That step creates a large initialization message, which increased the size of all shadow -> shim messages because of the  enum used for such messages. So we next move the initialization data out of band and change the "start handshake" so that the plugin first sends a pointer to where initialization data can be stored. Shadow writes the data directly, and finishes with another message to let the shim know the data has been initialized.

Progress on #1987


src/lib/shadow-shim-helper-rs/src/shim_event.rs
```diff
@@ -27,13 +27,14 @@ pub struct ShimEventAddThreadReq {
...
 #[derive(Copy, Clone, Debug, VirtualAddressSpaceIndependent)]
 #[repr(C)]
-pub struct ShimEventStart {
-    pub thread_shmem_block: ShMemBlockSerialized,
-    pub process_shmem_block: ShMemBlockSerialized,
+pub struct ShimEventStartReq {
+    /// Pointer to be initd by Shadow. Required.
+    pub thread_shmem_block_to_init: ForeignPtr<ShMemBlockSerialized>,
+    /// Pointer to be initd by Shadow. Optional.
+    pub process_shmem_block_to_init: ForeignPtr<ShMemBlockSerialized>,
```
stevenengler, line 37: Maybe clarify that these are pointers to plugin memory. Since these pointers are shared with both shadow and the shim, it's not immediately clear which is the "foreign" memory space (it's not really a "foreign pointer" when you're accessing these fields in the shim).

src/main/host/managed_thread.rs
```diff
@@ -183,16 +183,41 @@ impl ManagedThread {
...
-                    self.continue_plugin(
-                        ctx.host,
-                        &ShimEventToShim::Start(ShimEventStart {
-                            thread_shmem_block: ctx.thread.shmem().serialize(),
-                            process_shmem_block: ctx.process.shmem().serialize(),
-                        }),
+                    // Initialize the shim.
+
+                    // `wait_for_next_event` expects shmem to be unlocked.
+                    ctx.host.unlock_shmem();
```
stevenengler, line 189: It seems a little confusing to have an `unlock` without a corresponding `lock`. Can we unlock this closer to where it was previously locked and used?

src/main/host/managed_thread.rs
```diff
@@ -183,16 +183,41 @@ impl ManagedThread {
...
+                        Ok(ShimEventToShadow::StartReq(s)) => s,
+                        other => panic!("Unexpected result from shim: {other:?}"),
+                    };
+
+                    ctx.process
+                        .memory_borrow_mut()
+                        .write(
+                            start_req.thread_shmem_block_to_init,
+                            &ctx.thread.shmem().serialize(),
+                        )
```
stevenengler, line 205: Add a comment to explain what this is doing?

src/lib/shadow-shim-helper-rs/src/shim_event.rs
```diff
@@ -46,6 +47,9 @@ pub struct ShimEventStart {
 #[allow(clippy::large_enum_variant)]
 pub enum ShimEventToShadow {
     Null,
+    /// First message from the shim, indicating that it's ready to start
+    /// executing.
+    StartReq(ShimEventStartReq),
```
stevenengler, line 52: Maybe clarify what "Req" means? Is this "request"?

src/main/utility/pod.rs
```diff
@@ -278,3 +278,7 @@ unsafe impl Pod for libc::clone_args {}
 
 // shadow re-exports this definition from /usr/include/linux/tcp.h
 unsafe impl Pod for crate::cshadow::tcp_info {}
+
+// TODO: Move this module out of `main` so that `shadow_shmem` can use it,
+// and put this with the struct definition.
+unsafe impl Pod for shadow_shmem::allocator::ShMemBlockSerialized {}
```
stevenengler, line 284: Maybe add a comment to the original struct definition at `c::ShMemBlockSerialized` stating that it must be POD.

stevenengler (approved):

src/main/utility/pod.rs
sporksmith: done

PR #2953, opened 2023-05-19
We no longer need executable /dev/shm

sporksmith:
Now that the MemoryMapper uses `memfd_create` instead of creating files in `/dev/shm` (a422da50ea7e6ee8438486b534c251ae9c1e9a0c), `/dev/shm` no longer needs to be executable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ MINOR changes (backwards-compatible):
 
 PATCH changes (bugfixes):
 
-*
+* Updated documentation and tests to reflect that shadow no longer requires
+`/dev/shm` to be executable. (This requirement was actually removed in v3.0.0)
 
 Full changelog since v3.0.0:
 
--- a/docs/supported_platforms.md
+++ b/docs/supported_platforms.md
@@ -34,14 +34,13 @@ By these criteria, Shadow's oldest supported kernel version is currently 5.4
 ## Docker
 
 If you are installing Shadow within a Docker container, you must increase the
-size of the container's `/dev/shm` mount, add the "exec" mount option, and
-disable the seccomp security profile. You can do this by passing additional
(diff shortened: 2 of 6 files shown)


docs/supported_platforms.md
```diff
@@ -34,14 +34,13 @@ By these criteria, Shadow's oldest supported kernel version is currently 5.4
...
-disable the seccomp security profile. You can do this by passing additional
-flags to `docker run`.
+size of the container's `/dev/shm` mount and disable the seccomp security
+profile. You can do this by passing additional flags to `docker run`.
 
 Example:
 
 ```bash
-docker run -it --tmpfs /dev/shm:rw,nosuid,nodev,exec,size=1024g --security-opt seccomp=unconfined ubuntu:22.04
+docker run -it --tmpfs /dev/shm:rw,nosuid,nodev,size=1024g --security-opt seccomp=unconfined ubuntu:22.04
```
stevenengler, line 43: I think we can just use the simpler `--shm-size=1024g` instead of specifying all of the mount parameters.

CHANGELOG.md
```diff
@@ -14,7 +14,8 @@ MINOR changes (backwards-compatible):
 
 PATCH changes (bugfixes):
 
-*
+* Updated documentation and tests to remove the requirement that `/dev/shm` is executable.
+This requirement was actually removed in v3.0.0, but we didn't document it.
```
stevenengler, line 18: Maybe edit the 3.0 release notes to mention that?

sporksmith: done

sporksmith:
I think more recent versions of Docker may work without disabling seccomp as well; they updated their seccomp policy. Probably better to keep explicitly disabling it for now, though if we confirm, we could update the docs to say that disabling it only needed for versions < x.

PR #2959, opened 2023-05-22
Add the `LegacyTcp` wrapper to the interface qdisc 

stevenengler:
Previously the C `TCP` object would be added to the qdisc, but we want the rust `LegacyTcp` wrapper to be added instead. This requires a circular reference between the `TCP` and `LegacyTcp` objects since the `TCP` object adds itself to the qdisc, but the `TCP` object has been made to only hold a weak reference to the `LegacyTcp`.  Some additional changes needed to be made to prevent multiple mutable borrows of the `LegacyTcp`, and to extend the lifetime of the `LegacyTcp`, otherwise upgrading the weak reference won't work.

### DIFF
--- a/src/main/host/descriptor/socket/inet/legacy_tcp.rs
+++ b/src/main/host/descriptor/socket/inet/legacy_tcp.rs
@@ -64,7 +64,13 @@ impl LegacyTcpSocket {
             _counter: ObjectCounter::new("LegacyTcpSocket"),
         };
 
-        Arc::new(AtomicRefCell::new(socket))
+        let rv = Arc::new(AtomicRefCell::new(socket));
+
+        let inet_socket = InetSocket::LegacyTcp(rv.clone());
+        let inet_socket = Box::into_raw(Box::new(inet_socket.downgrade()));
+        unsafe { c::tcp_setRustSocket(legacy_tcp, inet_socket) };
+
+        rv
     }
 
     /// Get a canonical handle for this socket. We use the address of the `TCP` object so that the
--- a/src/main/host/descriptor/socket/inet/mod.rs
+++ b/src/main/host/descriptor/socket/inet/mod.rs
@@ -453,7 +453,13 @@ mod export {
     #[no_mangle]
     pub extern "C" fn inetsocket_drop(socket: *const InetSocket) {
         assert!(!socket.is_null());
-        unsafe { Box::from_raw(socket as *mut InetSocket) };
+        unsafe { Box::from_raw(socket.cast_mut()) };
+    }
+
+    /// Helper for GLib functions that take a `TaskObjectFreeFunc`. See [`inetsocket_drop`].
+    #[no_mangle]
+    pub extern "C" fn inetsocket_dropVoid(socket: *mut libc::c_void) {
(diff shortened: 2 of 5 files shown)


src/main/host/descriptor/socket/inet/legacy_tcp.rs
```diff
@@ -151,7 +151,9 @@ impl LegacyTcpSocket {
             return None;
         }
 
-        Some(Packet::from_raw(packet))
+        let packet = Packet::from_raw(packet);
+        unsafe { c::packet_ref(packet.borrow_inner()) }
```
sporksmith, line 161: Should `Packet::from_raw` make this call? Does it (and should it) `unref` on `Drop`?

src/main/host/descriptor/socket/inet/mod.rs
```diff
@@ -481,4 +500,25 @@ mod export {
         socket.borrow().update_packet_header(&mut packet);
         packet.into_inner();
     }
+
+    /// Decrement the ref count of the `InetSocketWeak` object. The pointer must not be used after
+    /// calling this function.
+    #[no_mangle]
+    pub extern "C" fn inetsocketweak_drop(socket: *const InetSocketWeak) {
```
sporksmith, line 531: I think it'd make sense for this to take a `mut` pointer, since it's effectively taking ownership (and destroying)

src/main/host/descriptor/socket/inet/mod.rs
```diff
@@ -481,4 +500,25 @@ mod export {
...
+        assert!(!socket.is_null());
+        unsafe { Box::from_raw(socket.cast_mut()) };
+    }
+
+    /// Upgrade the weak reference. May return `NULL` if the socket has no remaining strong
+    /// references and has been dropped.
+    #[no_mangle]
+    pub extern "C" fn inetsocketweak_upgrade(socket: *const InetSocketWeak) -> *const InetSocket {
```
sporksmith, line 539: I think it'd make sense for this to return a `mut` pointer, since the caller owns the returned reference

sporksmith (approved):

src/main/host/descriptor/socket/inet/legacy_tcp.rs
stevenengler: `Packet::from_raw` doesn't `ref`, but it does `unref` when dropped. The documentation for `from_raw` says that it takes ownership of the reference to the packet.

src/main/host/descriptor/socket/inet/mod.rs
stevenengler: Added.

src/main/host/descriptor/socket/inet/mod.rs
stevenengler: Added.

stevenengler:
A tor benchmark failed due to a "Assertion failed: inetSocket != NULL" in `_tcp_flush`, which was because I missed a scheduled TCP task. I'm adding a patch that should hopefully fix this.

sporksmith (approved):

Issue #2919, opened 2023-05-10
Convert shim and its dependencies to statically linked no_std (async signal safe)


sporksmith:
The shim currently has some dependencies on libc, which can lead to surprising and difficult to debug issues due to its usage via `LD_PRELOAD` and much of it being executed from a signal handler context.

To *really* be certain to avoid problems, we'd want to restrict anything that could be called from the shim to async-signal-safe functions (see [signal-safety(7)](https://man7.org/linux/man-pages/man7/signal-safety.7.html). However, this limits us to a subset of libc. Most libraries outside of libc, including Rust's `std`, have no concept of async signal safety.

In practice, we've mostly gotten by by avoiding memory allocation functions, but in most cases we can't be sure that some library call doesn't allocate internally.

In Rust, making the shim and its dependencies `no_std` would ensure that no allocation functions or other libc functions are called. We could still run into trouble if a function is non-reentrant, but it's not as likely outside of libc, and if we ensure we don't dynamically link any other libraries either. e.g. even if the shim and the managed program use the same library with non-reentrant functionality, if the shim has *statically* linked it, then they'll have separate copies of their internal global state. i.e. the shim will still be able to interfere with itself in the case of recursive signal handling, but it shouldn't have any surprising interactions with the managed program.

sporksmith:
I wasn't sure it's possible to make a statically linked `.so`, but it appears that you can. i.e. we should theoretically be able to make the shim completely statically linked.

Successful proof of concept:

lib.rs:
```
#![no_std]

use linux_syscall::ResultSize;

#[panic_handler]
fn panic(_info: &core::panic::PanicInfo) -> ! {
    let _ = unsafe { linux_syscall::syscall!(linux_syscall::SYS_kill, 0, 6) };
    loop {}
}

#[no_mangle]
pub extern "C" fn write(fd: i32, buf: *mut i8, count: usize) -> isize {
    if fd == 1 {
        unsafe { *buf = 'x' as i8 };
    }
    unsafe {
        linux_syscall::syscall!(linux_syscall::SYS_write, fd, buf, count)
            .try_isize()
            .unwrap()
    }
}
```

Cargo.toml:
```
[package]
name = "cnostd"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[lib]
crate-type = ["cdylib"]

[dependencies]
linux-syscall = "1.0.0"

[profile.dev]
panic = "abort"

[...]

PR #2965, opened 2023-05-23
Add Linux API crate

sporksmith:
Adds the crate linux-api, a no_std, no libc crate for interacting with the Linux kernel.

We use statically generated bindgen bindings from Linux kernel headers. This lets us avoid depending on the `libc` crate; while the `libc` crate *can* be used with `no_std` and won't result in a runtime libc dependency if no methods are called, this can be tricky. Worse, libc can use "adapter" types with same name as kernel types. Binding the linux headers directly avoids such problems.

Moves some existing signal and clone definitions into this new module and reworks the definitions to use the kernel bindings instead of libc types and constants.

Progress on #2919


src/lib/linux-api/src/signal.rs
```diff
@@ -0,0 +1,548 @@
...
+unsafe impl VirtualAddressSpaceIndependent for LinuxSigActionFlags {}
+
+#[derive(Copy, Clone, VirtualAddressSpaceIndependent)]
+#[allow(non_camel_case_types)]
+#[repr(C)]
+pub struct linux_siginfo_t {
+    // We don't wrap bindings::siginfo_t directly,
+    // since this would introduce a dependency on the original system
+    // header file in the cbindgen'd version of this type, and requiring our C code
+    // to include the kernel headers results in naming conflicts.
```
stevenengler, line 103: I don't understand this comment. Isn't the point of including the bindings for a specific kernel version that we don't need to worry about this? We just use the fields that exist in that specific kernel version?

Otherwise, I don't see why this wouldn't apply to all other structs in the Linux headers, and then we'd need to provide our own struct definitions for all Linux header structs.

stevenengler (commented):
Still looking through it, but just wanted to leave the comments I have so far.

src/lib/linux-api/src/signal.rs
```diff
@@ -0,0 +1,548 @@
...
+//
+// Copied from linux's include/uapi/asm-generic/signal-defs.h.
+pub const LINUX_SIG_DFL: usize = 0;
+pub const LINUX_SIG_IGN: usize = 1;
+pub const LINUX_SIG_ERR: usize = (-1_isize) as usize;
+
+// signal names
+#[derive(Debug, Copy, Clone, IntoPrimitive, TryFromPrimitive)]
+#[repr(i32)]
+pub enum LinuxSignal {
```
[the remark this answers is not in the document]
sporksmith: I tried the `prefix` export option. The main place I ran into trouble is with exported functions from *other* crates. e.g. if we use `sigset_t` here, but use the `prefix` export option to export as `linux_sigset_t`, then in the `shadow-shim-helper-rs` crate when exporting APIs that take a pointer to this type, we have to account for the renaming somehow.

e.g. we now have a function in `shadow-shim-helper-rs` declared as:

```
    pub unsafe extern "C" fn shimshmem_getProcessPendingSignals(
        lock: *const ShimShmemHostLock,
        process: *const ShimShmemProcess,
    ) -> linux_sigset_t {
```

If we use `linux_api::sigset_t` instead of `linux_sigset_t`, then the cbindgen'd header will just call it `sigset_t` in the function signature. (Actually I'm not sure how this is working right in the code @ head, where it is `sigset_t`, and the shim code looks like it'd be getting the libc definition, which is the wrong size).

Ideally we'd want to configure cbindgen in the `shadow-shim-helper-rs` crate to prefix anything from the `linux-api` crate, but I don't think there's a way to do that. We'd have to enumerate the individual symbols to be renamed. If we forget one, it'll name the libc type instead, and might seem to work (but potentially fail in confusing ways at runtime).

I think I'd be happy to drop the `Linux` prefix on everything in this crate once we don't need to `cbindgen` it anymore :)

src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
```diff
@@ -434,7 +438,9 @@ impl Default for SiginfoWrapper {
 
 impl From<libc::siginfo_t> for SiginfoWrapper {
     fn from(s: libc::siginfo_t) -> Self {
-        Self(s)
+        static_assertions::assert_eq_align!(SiginfoWrapper, libc::siginfo_t);
+        static_assertions::assert_eq_size!(SiginfoWrapper, libc::siginfo_t);
+        Self(unsafe { core::mem::transmute(s) })
     }
 }
```
stevenengler, line 445: Should make sure there's no padding bytes in `libc::siginfo_t` so that it doesn't transmute uninitialized bytes into initialized fields.

src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
```diff
@@ -402,27 +404,29 @@ impl SiginfoWrapper {
...
     /// # Safety
     ///
     /// Pointer fields must not be dereferenced from outside of their
     /// native virtual address space.
     pub unsafe fn as_siginfo(&self) -> &libc::siginfo_t {
-        &self.0
+        static_assertions::assert_eq_align!(SiginfoWrapper, libc::siginfo_t);
+        static_assertions::assert_eq_size!(SiginfoWrapper, libc::siginfo_t);
+        unsafe { core::mem::transmute(self) }
     }
```
stevenengler, line 430: Why is this function marked `unsafe`, but `linux_siginfo_t::as_bound_type` isn't?

src/lib/linux-api/src/signal.rs
```diff
@@ -0,0 +1,548 @@
...
+        if self.val == 0 {
+            return None;
+        }
+        for i in 1..=LINUX_SIGRT_MAX {
+            let s = LinuxSignal::try_from(i).unwrap();
+            if self.has(s) {
+                return Some(s);
+            }
+        }
+        unreachable!("");
```
stevenengler, line 219: I think you can just do `unreachable!()`.

stevenengler (changes requested):

src/lib/linux-api/src/signal.rs
stevenengler: So this sounds like it will be an issue for each struct we define? But yeah I don't see a better alternative.

sporksmith: One other option is to manually define the internals here, and validate that they match the bindings. e.g. that's the approach in `linux_sigaction`. It's a little more manual work up-front, but I suppose we shouldn't have to mess with it often.

```
pub struct linux_sigaction {
    // SAFETY: We do not dereference the pointers in this union, except from the
    // shim, where it is valid to do so.
    #[unsafe_assume_virtual_address_space_independent]
    u: LinuxSignalHandler,
    ksa_flags: LinuxSigActionFlags,
    // Rust guarantees that the outer Option doesn't change the size:
    // https://doc.rust-lang.org/std/option/index.html#representation
    //
    // SAFETY: We never dereference this field.
    #[unsafe_assume_virtual_address_space_independent]
    ksa_restorer: Option<extern "C" fn()>,
    ksa_mask: linux_sigset_t,
}

impl linux_sigaction {
    pub fn handler(&self) -> &LinuxSignalHandler {
        assert_eq!(
            memoffset::offset_of!(linux_sigaction, u),
            memoffset::offset_of!(bindings::sigaction, sa_handler)
        );
        &self.u
    }
}
```

Unfortunately I don't think there's a `const` equivalent of `memoffset::offset_of`, though maybe we could move those assertions to unit tests.

src/lib/linux-api/src/sched.rs
```diff
@@ -0,0 +1,37 @@
+use crate::bindings;
+
+bitflags::bitflags! {
+    // While `clone` is documented as taking an i32 parameter for flags,
+    // in `clone3` its a u64. Promote to u64 throughout.
+    #[derive(Copy, Clone, Debug)]
+    pub struct LinuxCloneFlags: u64 {
```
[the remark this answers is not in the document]
sporksmith: Done

src/lib/linux-api/src/sched.rs
```diff
@@ -0,0 +1,37 @@
...
+        const CLONE_CLEAR_SIGHAND = bindings::CLONE_CLEAR_SIGHAND;
+        const CLONE_INTO_CGROUP = bindings::CLONE_INTO_CGROUP;
+        const CLONE_NEWTIME = bindings::CLONE_NEWTIME as u64;
+
+        const CLONE_VM = bindings::CLONE_VM as u64;
+        const CLONE_FS = bindings::CLONE_FS as u64;
+        const CLONE_FILES = bindings::CLONE_FILES as u64;
```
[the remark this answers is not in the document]
sporksmith: Done

src/lib/linux-api/src/signal.rs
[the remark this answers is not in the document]
sporksmith: This is now a struct, so that it can cleanly represent realtime signals. While making other changes I noticed we were doing an unsound transmute to work around nix's Signal not explicitly representing them.

src/lib/linux-api/src/signal.rs
sporksmith: I ended up doing a bit of a hybrid here. I added/kept explicit field definitions where feasible and added tests to validate field ordering. I left the `siginfo_t` mess of nested unions as `MaybeUninit<[u8;...]>` for the moment.

src/lib/linux-api/src/signal.rs
```diff
@@ -0,0 +1,548 @@
...
+}
+static_assertions::assert_eq_align!(linux_siginfo_t, bindings::siginfo_t);
+static_assertions::assert_eq_size!(linux_siginfo_t, bindings::siginfo_t);
+
+impl linux_siginfo_t {
+    fn as_bound_type(&self) -> &bindings::siginfo_t {
+        static_assertions::assert_eq_align!(linux_siginfo_t, bindings::siginfo_t);
+        static_assertions::assert_eq_size!(linux_siginfo_t, bindings::siginfo_t);
+        unsafe { core::mem::transmute(self) }
+    }
```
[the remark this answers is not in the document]
sporksmith: Added a comment. The only potential padding is in the unions, which still require `unsafe` to read from.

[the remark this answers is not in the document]
sporksmith: done

src/lib/linux-api/src/signal.rs
```diff
@@ -0,0 +1,548 @@
...
+
+    fn as_bound_type_mut(&mut self) -> &mut bindings::siginfo_t {
+        static_assertions::assert_eq_align!(linux_siginfo_t, bindings::siginfo_t);
+        static_assertions::assert_eq_size!(linux_siginfo_t, bindings::siginfo_t);
+        unsafe { core::mem::transmute(self) }
+    }
```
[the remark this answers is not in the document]
sporksmith: Done

src/lib/linux-api/src/signal.rs
sporksmith: done

src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
sporksmith: The `unsafe` here is meant to reflect that pointers may not be valid in the current address space. I'm fairly certain there aren't actually any pointers in `siginfo_t` though.

I ended up removing this wrapper altogether.

src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
sporksmith: Removed. I got rid of most of the conversions to/from the libc type; I don't think there's any guarantee that it'll always be compatible with the kernel definition.

There's still one place in the shim where we unsafely assume this by casting the pointer; I added a TODO to avoid it by installing our native signal handlers without going through libc, so that our handler actually gets `linux_siginfo_t`.

sporksmith:
P.S. Sorry I realize there are quite a lot of commits here now; it's probably more work than it's worth to squash out the design iteration  without squashing everything. It's probably easiest to primarily review the final state. It's a bit large, but mostly boilerplate.

src/lib/linux-api/src/signal.rs
```diff
@@ -0,0 +1,1197 @@
...
+    Cld(SigInfoCodeCld),
+    Poll(SigInfoCodePoll),
+    Sys(SigInfoCodeSys),
+}
+
+#[derive(Debug, Copy, Clone)]
+pub struct SigInfoCodeFromRawError {
+    pub code: i32,
+    pub signo: i32,
+}
```
stevenengler, line 137: Implement `std::error::Error` for the error type?

src/lib/linux-api/src/signal.rs
```diff
@@ -0,0 +1,1197 @@
...
+
+impl SigInfoCode {
+    fn try_from_raw(raw_code: i32, raw_signo: i32) -> Result<Self, SigInfoCodeFromRawError> {
```
stevenengler, line 140: I think this could use a comment explaining what the function is used for.

src/lib/linux-api/src/signal.rs
```diff
@@ -0,0 +1,1197 @@
...
+        match signal {
+            Signal::SIGCHLD => {
+                if let Ok(cld_code) = SigInfoCodeCld::try_from(raw_code) {
+                    Ok(SigInfoCode::Cld(cld_code))
+                } else {
+                    Err(err)
+                }
+            }
+            Signal::SIGILL => {
+                if let Ok(code) = SigInfoCodeIll::try_from(raw_code) {
```
stevenengler, line 167: optional nit: maybe simpler as `SigInfoCodeCld::try_from(raw_code).map(SigInfoCode::Cld).or(Err(err))`

src/lib/linux-api/src/signal.rs
```diff
@@ -0,0 +1,1197 @@
...
+#[allow(non_camel_case_types)]
+pub type linux_siginfo_t = bindings::linux_siginfo_t;
+
+pub type SigInfoDetailsKill = bindings::linux__sifields__bindgen_ty_1;
+pub type SigInfoDetailsTimer = bindings::linux__sifields__bindgen_ty_2;
+pub type SigInfoDetailsRt = bindings::linux__sifields__bindgen_ty_3;
+pub type SigInfoDetailsSigChld = bindings::linux__sifields__bindgen_ty_4;
+pub type SigInfoDetailsSigFault = bindings::linux__sifields__bindgen_ty_5;
+pub type SigInfoDetailsSigPoll = bindings::linux__sifields__bindgen_ty_6;
+pub type SigInfoDetailsSigSys = bindings::linux__sifields__bindgen_ty_7;
```
stevenengler, line 363: This seems like it could be brittle if the kernel adds another field or adds a padding field somewhere in the middle. Maybe these types should be hard-coded here, and then if the bindings were to change and the types didn't match, it would cause a compile error?

stevenengler (commented):
I'll finish reviewing tomorrow morning, but figured I'd leave these comments now.

src/lib/linux-api/src/signal.rs
sporksmith: Done

src/lib/linux-api/src/signal.rs
sporksmith: done

src/lib/linux-api/src/signal.rs
sporksmith: A bit, but that'll only happen when we explicitly change what version of the Linux source we're building from. It should be fairly clear what caused the breakage and how to sort it out, e.g. looking at the diff of the bindings. 

When constructing a `linux_siginfo_t` we have to directly use the bindgen types. We could create our own same-shaped types with conversion functions to use in public output for this module, but IMO it's not worth the boilerplate since we'd still have to resolve the breakage inside this module if these type names change.

Added a block-comment here to help, and added an internal constructor in SigInfo to also avoid naming the generated union field names more than we have to.

src/lib/linux-api/src/signal.rs
```diff
@@ -0,0 +1,1197 @@
...
+    SigPoll(SigInfoDetailsSigPoll),
+    SigSys(SigInfoDetailsSigSys),
+}
+
+#[derive(Copy, Clone)]
+#[repr(transparent)]
+pub struct SigInfo(linux_siginfo_t);
+// Contains pointers, but they are understood to not necessarily be valid in the
+// current address space.
+unsafe impl VirtualAddressSpaceIndependent for SigInfo {}
```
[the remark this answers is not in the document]
sporksmith: My view is that these types defined in the kernel API are explicitly meant to be shared between address spaces (user and kernel). While there are sometimes pointers inside of them, we don't dereference those pointers here in the wrappers. At most we we make the pointers themselves available to the caller, who should understand whether it's a pointer in the current address space or not, and needs `unsafe` to deref.

One slightly gray area is the `transmute` from pointer to `function` in `SigAction::handler`. I don't think the `transmute` itself is unsound though, and it is transmuted to an `unsafe` fn, so the caller again needs `unsafe` to actually call it.

src/lib/linux-api/src/signal.rs
sporksmith: Forgot to update this thread - we now use a script to mangle the names in the bindgen output so that we can directly wrap the bindgen types without causing naming conflicts in the cbindgen output.

src/lib/linux-api/src/signal.rs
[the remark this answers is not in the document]
stevenengler: > My view is that these types defined in the kernel API are explicitly meant to be shared between address spaces (user and kernel).

We do share them between different user address spaces though, otherwise we wouldn't need them to be vasi. The kernel knows that any pointers given from userspace have to point to memory in the address space of the calling process, so there's no ambiguity about what address space the pointers are in. Whereas when we share pointers between user address spaces, it is ambiguous without knowing the larger context about who sets the values and when/if they should be dereferenced.

> At most we we make the pointers themselves available to the caller, who should understand whether it's a pointer in the current address space or not, and needs unsafe to deref.

Wouldn't this apply to any pointer type? What's special about this pointer, as opposed to `unsafe impl<T> VirtualAddressSpaceIndependent for *const T {}`?

src/lib/linux-api/src/signal.rs
stevenengler: > A bit, but that'll only happen when we explicitly change what version of the Linux source we're building from. It should be fairly clear what caused the breakage and how to sort it out, e.g. looking at the diff of the bindings.

I'm not really thinking of a compile breakage, but if for example an extra padding field was added between the fields of type `linux__sifields__bindgen_ty_6` and `linux__sifields__bindgen_ty_7`. If `linux__sifields__bindgen_ty_7` was originally a `u32` and the new `linux__sifields__bindgen_ty_7` padding field is a `u16`, `SigInfoDetailsSigSys` would change from a `u32` to a `u16` and it might break things in a way that still compiles but is incorrect, which might be difficult to detect.

But I didn't realize that all of these types are structs, so maybe this wouldn't happen then (a padding field wouldn't be a struct, so would probably be a compile error if the type changed from a struct to an integer).

src/lib/linux-api/src/signal.rs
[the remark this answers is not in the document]
sporksmith: > We do share them between different user address spaces though, otherwise we wouldn't need them to be vasi.

I don't think we do? I'm thinking of shadow's address space as "kernel space". Any given instance of one of these is only going to be used in Shadow (kernel) and a single managed process that it's associated with.

> Wouldn't this apply to any pointer type? What's special about this pointer, as opposed to `unsafe impl<T> VirtualAddressSpaceIndependent for *const T {}`?

We don't provide "safe" (non-`unsafe`) methods on these types that dereference internal pointers, (partly) *because* these wrappers are designed to be `VirtualAddressSpaceIndependent`.

`unsafe impl<T> VirtualAddressSpaceIndependent for *const T {}` would be dangerous because even if the outermost pointer itself is in the current address space, it ignores whether there are any pointers inside `T` and what address space they are in.

It would particularly be a problem because it would cause the `VirtualAddressSpaceIndependent` derive macro to work in places that it shouldn't.

e.g. suppose you had:

```
struct MyHashMapWrapper {
  internal: *const HashMap<u32, u32>
}

impl MyHashMapWrapper {
  pub fn get(&self, x: &u32) -> Option<&u32> {
      unsafe { internal.ref().unwrap() }.get(x)
  }
}
```

Someone could add the line

```
#[derive(VirtualAddressSpaceIndependent)]
```

and store a `MyHashMapWrapper` in shared memory, which would make the `get` method unsound, all without any additional use of `unsafe`.

src/lib/linux-api/src/signal.rs
sporksmith: > If linux__sifields__bindgen_ty_7 was originally a u32

I don't think bindgen would create the type `linux__sifields__bindgen_ty_7` in that case; it would just use `u32`. These type names are generated for anonymous union and anonymous struct types.

src/lib/linux-api/src/signal.rs
```diff
@@ -0,0 +1,1192 @@
...
+    pub const FULL: Self = Self(0);
+
+    #[allow(unused)]
+    fn deref_c(c: &bindings::linux_sigset_t) -> &Self {
+        unsafe { &*(c as *const _ as *const Self) }
+    }
+
+    fn deref_mut_c(c: &mut bindings::linux_sigset_t) -> &mut Self {
+        unsafe { &mut *(c as *mut _ as *mut Self) }
+    }
```
stevenengler, line 774: Why do these exist? Aren't they covered by the `TransparentWrapper<linux_sigset_t>` impl?

src/lib/linux-api/src/signal.rs
```diff
@@ -0,0 +1,1192 @@
...
+
+#[allow(non_camel_case_types)]
+pub type linux_sigset_t = bindings::linux_sigset_t;
+
+/// Compatible with the Linux kernel's definition of sigset_t on x86_64.
+///
+/// This is analagous to, but typically smaller than, libc's sigset_t.
+#[repr(transparent)]
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Default, VirtualAddressSpaceIndependent)]
+pub struct SigSet(linux_sigset_t);
```
stevenengler, line 760: nit: maybe make it `pub linux_sigset_t` so it can be accessed directly instead of through `peel_ref`?

src/lib/linux-api/src/signal.rs
```diff
@@ -0,0 +1,1192 @@
...
+    }
+
+    pub fn handler(&self) -> SignalHandler {
+        let as_usize = self.0.lsa_handler.map(|f| f as usize).unwrap_or(0);
+        if as_usize == Self::SIG_IGN {
+            SignalHandler::SigIgn
+        } else if as_usize == Self::SIG_DFL {
+            SignalHandler::SigDfl
+        } else if self.flags_retain().contains(SigActionFlags::SIGINFO) {
+            SignalHandler::Action(unsafe { core::mem::transmute(self.0.lsa_handler.unwrap()) })
```
stevenengler, line 955: I think this transmute could be more explicit, maybe with the types defined explicitly. For example `transmute::<extern "C" fn(arg1: ::core::ffi::c_int), extern "C" fn(i32, *mut SigInfo, *mut core::ffi::c_void)>(lsa_handler.unwrap())`.

I also don't understand how this is correct since the arguments are completely different. I'm guessing this has something to do with the libc version using a union for `sa_handler` and `sa_sigaction`, but I think this needs some comments explaining what's going on and why it's safe.

src/lib/linux-api/src/signal.rs
```diff
@@ -0,0 +1,1192 @@
...
+        sa: *const linux_sigaction,
+    ) -> LinuxSigActionFn {
+        let sa = SigAction::wrap_ref(unsafe { sa.as_ref().unwrap() });
+        match sa.handler() {
+            SignalHandler::Action(h) =>
+            // We need to transmute the function pointer here, to one that takes
+            // the exported C types.
+            unsafe { core::mem::transmute::<SignalActionFn, LinuxSigActionFn>(h) },
+            _ => unsafe { core::mem::transmute(core::ptr::null::<core::ffi::c_void>()) },
+        }
```
stevenengler, line 1178: Same as above transmute comments (about making the first transmute explicit, and the second transmute is UB).

stevenengler (changes requested):

src/lib/linux-api/src/signal.rs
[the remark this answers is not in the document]
stevenengler: > > We do share them between different user address spaces though, otherwise we wouldn't need them to be vasi.
> 
> I don't think we do? I'm thinking of shadow's address space as "kernel space". Any given instance of one of these is only going to be used in Shadow (kernel) and a single managed process that it's associated with.

I could see it this way, but I could also see a developer seeing it the other way too. Both shadow and the plugin are two separate processes with their own address spaces, and you're using shared memory to share the `SigInfo` between them.

> > Wouldn't this apply to any pointer type? What's special about this pointer, as opposed to `unsafe impl<T> VirtualAddressSpaceIndependent for *const T {}`?
> 
> We don't provide "safe" (non-`unsafe`) methods on these types that dereference internal pointers, (partly) _because_ these wrappers are designed to be `VirtualAddressSpaceIndependent`.

I see the `VirtualAddressSpaceIndependent` trait a little differently. For me, the `VirtualAddressSpaceIndependent` trait is a marker to say that a type is address space-independent in all cases, and anything given to you through the public API is also `VirtualAddressSpaceIndependent` (unless there's an unsafe method that has documentation that says otherwise).

With `SigInfo` (a type with the `VirtualAddressSpaceIndependent` marker), you can directly access an inner field that is not `VirtualAddressSpaceIndependent` through `sig_info.details().unwrap().l_call_addr`. In this case the `VirtualAddressSpaceIndependent` marker means "address space independent, but only as long as you don't deref any of the pointers that are given to you in the public API".

This would be like saying `UnsafeCell` should be `Sync`, because `UnsafeCell` only gives you pointers which require `unsafe` to deref. Or making a struct `Unpin` which contains a public pointer because the user code would need unsafe to dereference the pointer. In both cases we can't specify additional safety requirements on the deref since the deref is not part of the type's API.

> `unsafe impl<T> VirtualAddressSpaceIndependent for *const T {}` would be dangerous because even if the outermost pointer itself is in the current address space, it ignores whether there are any pointers inside `T` and what address space they are in.
>
> It would particularly be a problem because it would cause the VirtualAddressSpaceIndependent derive macro to work in places that it shouldn't.

But isn't that what `unsafe impl VirtualAddressSpaceIndependent for SigInfo {}` is doing? If you make a `struct Foo(SigInfo)`, it can automatically derive `VirtualAddressSpaceIndependent` even though the inner `SigInfo` has pointers which might be in another address space.

> Someone could add the line
>
> ```rust
> #[derive(VirtualAddressSpaceIndependent)]
> ```
>
> and store a MyHashMapWrapper in shared memory, which would make the get method unsound, all without any additional use of unsafe.

I'd say this is the fault of `MyHashMapWrapper`. It's doing an unsafe pointer dereference with a pointer that it doesn't know is valid. It would be like the `Foo(SigInfo)` wrapper above implementing a method to dereference a pointer in `SigInfo`.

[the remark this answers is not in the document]
sporksmith: Ok, I'm convinced. I dropped the `VirtualAddressSpaceIndependent` markers on `SigInfo` and `SigAction`, made the accessors that can return pointers `unsafe`, and documented that we don't internally dereference pointer.

src/lib/linux-api/src/signal.rs
```diff
@@ -115,8 +115,6 @@ bitflags::bitflags! {
         const SIGINFO = bindings::LINUX_SA_SIGINFO as u64;
     }
 }
-// We can't derive this since the bitflags field uses an internal type.
-unsafe impl VirtualAddressSpaceIndependent for SigActionFlags {}
```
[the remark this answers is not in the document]
sporksmith: Done

PR #2971, opened 2023-05-24
Make Process's strace file shareable across Processes

sporksmith:
When we fork a Process, the child will use the same strace file (if any). This is analogous to `strace -f`.

Progress on https://github.com/shadow/shadow/issues/1987.

### DIFF
--- a/src/main/host/process.rs
+++ b/src/main/host/process.rs
@@ -1195,10 +1195,9 @@ impl Process {
             unreachable!("Tried to handle process exit of non-running process");
         };
 
-        runnable
-            .strace_logging
-            .take()
-            .map(|s| RootedRc::safely_drop(s, host.root()));
+        if let Some(s) = runnable.strace_logging.take() {
+            RootedRc::safely_drop(s, host.root())
+        }
 
         #[cfg(feature = "perf_timers")]
         debug!(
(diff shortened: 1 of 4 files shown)


src/main/host/process.rs
```diff
@@ -191,7 +191,8 @@ pub struct RunnableProcess {
     // Shared memory allocation for shared state with shim.
     shim_shared_mem_block: ShMemBlock<'static, ProcessShmem>,
 
-    strace_logging: Option<StraceLogging>,
+    // strace logging file, if any. Shared with forked Processes.
+    strace_logging: Option<RootedRc<RootedRefCell<StraceLogging>>>,
```
stevenengler, line 195: Why is the `RootedRefCell` needed? I don't think we need to mutate the format options at runtime, and the file is already inside a `RefCell`. If it's needed for `Send`/`Sync` or something, maybe replace the `RefCell` in `StraceLogging` with `RootedRefCell` instead?

stevenengler (approved):

src/main/host/process.rs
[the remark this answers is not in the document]
sporksmith: Thanks, that sounds much simpler :)

sporksmith: Right, otherwise the inner `RefCell` can't go inside the `RootedRc`, since it's not `Sync`. I'll take a look at replacing the inner `RefCell` instead if `File::try_clone` doesn't work out.

sporksmith:
Obsoleted by adding `StraceLogging::try_clone` in https://github.com/shadow/shadow/commit/54f2e6868470da4383a445566e12e00a947cb5f6

PR #2981, opened 2023-05-30
Add a rust UDP implementation

stevenengler:
This adds a rust implementation of udp. A future PR will remove the C version, along with the C socket syscall handlers and other legacy code.

There are still some TODO comments in the udp code that should be looked at and have tests added for, but I'll leave those for future PRs. All of the existing udp tests pass with this new rust version, and it also supports `sendmsg`/`recvmsg` and `shutdown`. It should behave similar to the C version, but there are a few changes such as the socket buffer size acting as a soft limit rather than a hard limit. There are also some accesses of the host through the worker that shouldn't be there, but these are required until we rewrite the network interface in rust.

### DIFF
--- a/src/main/host/descriptor/socket/inet/udp.rs
+++ b/src/main/host/descriptor/socket/inet/udp.rs
@@ -1,0 +1,867 @@
+use std::io::Read;
+use std::net::{Ipv4Addr, SocketAddrV4};
+use std::sync::Arc;
+
+use atomic_refcell::AtomicRefCell;
+use nix::errno::Errno;
+use nix::sys::socket::{AddressFamily, MsgFlags, Shutdown, SockaddrIn};
+use shadow_shim_helper_rs::syscall_types::ForeignPtr;
+
+use crate::core::worker::Worker;
+use crate::cshadow as c;
+use crate::host::descriptor::socket::inet::{self, InetSocket};
+use crate::host::descriptor::socket::{RecvmsgArgs, RecvmsgReturn, SendmsgArgs};
+use crate::host::descriptor::{
+    File, FileMode, FileState, FileStatus, OpenFile, Socket, StateEventSource, StateListenerFilter,
+    SyscallResult,
+};
+use crate::host::memory_manager::MemoryManager;
+use crate::host::syscall::io::{write_partial, IoVec, IoVecReader};
+use crate::host::syscall_types::SyscallError;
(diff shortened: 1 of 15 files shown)


src/main/host/descriptor/socket/inet/udp.rs
```diff
@@ -0,0 +1,892 @@
...
+use crate::host::syscall::io::{write_partial, IoVec, IoVecReader};
+use crate::host::syscall_types::SyscallError;
+use crate::network::net_namespace::NetworkNamespace;
+use crate::network::packet::{PacketRc, PacketStatus};
+use crate::utility::callback_queue::{CallbackQueue, Handle};
+use crate::utility::sockaddr::SockaddrStorage;
+use crate::utility::{HostTreePointer, ObjectCounter};
+
+/// Maximum size of a datagram we are allowed to send out over the network.
+const CONFIG_DATAGRAM_MAX_SIZE: usize = 65507;
```
sporksmith, line 28: Where did this particular number come from? (Might be worth a comment)

src/main/host/descriptor/socket/inet/udp.rs
```diff
@@ -0,0 +1,892 @@
...
+        };
+
+        let mut peer_addr: std::net::SocketAddrV4 = (*peer_addr).into();
+
+        // https://stackoverflow.com/a/22425796
+        if peer_addr.ip().is_unspecified() {
+            peer_addr.set_ip(std::net::Ipv4Addr::LOCALHOST);
+        }
+
+        // TODO: handle AF_UNSPEC
```
sporksmith, line 529: stale comment? looks like it's handled above?

sporksmith (approved):

src/main/host/descriptor/socket/inet/udp.rs
stevenengler: Added a comment.

> 65,535 (2^16 - 1) - 20 (ip header) - 8 (udp header)

src/main/host/descriptor/socket/inet/udp.rs
stevenengler: This means to handle a `AF_UNSPEC` socket address (as opposed to a `AF_INET` socket address). The above code checks if the IP address is unspecified (is an `AF_INET` socket address with IP "0.0.0.0").

I moved this comment up to where we cast it to an `AF_INET` socket.

PR #2982, opened 2023-05-30
Remove unused C socket code

stevenengler:
There's probably more that can be removed, but this is a start.


src/main/host/syscall/handler/socket.rs
```diff
@@ -127,13 +126,9 @@ impl SyscallHandler {
...
 
-            let file = match desc.file() {
-                CompatFile::New(file) => file,
-                // if it's a legacy file, use the C syscall handler instead
-                CompatFile::Legacy(_) => {
-                    drop(desc_table);
-                    return Self::legacy_syscall(c::syscallhandler_bind, ctx);
-                }
+            let CompatFile::New(file) = desc.file() else {
+                // we don't have any C socket objects
```
sporksmith, line 130: Do we think these are now unreachable? Maybe `panic` or `debug_panic`?

sporksmith (approved):

src/main/host/syscall/handler/socket.rs
stevenengler: These can be reached if the syscall is called on a non-socket C file, for example `connect()` on a regular file. I originally did have a panic here, but luckily the socket tests caught it.

Issue #1559, opened 2021-08-04
Shim depends on native thread-local storage (via ShimThreadLocalStorage)


sporksmith:
We already had a known dependency on thread-local-storage for preload mode, for which we'd disabled  the "clone" test.

ptrace mode depends on it too, though, e.g. via the shim logger and the thread-specific shared memory area. Previously this was mostly working even without native thread local storage (in the "clone" test), but in a buggy, fragile way.

If we want to support threads without native thread local storage (such as in our clone test), we should extend the implementation of ShimThreadLocalStorage to not be implemented on top of native thread local storage. One potential solution is to have Shadow write a current-executing-thread-id to a process-global (not per-thread) shared memory location.

sporksmith:
This is partly fixed now. See e.g. #3055 and #3064. There are now alternate modes of operation that don't try to store anything in native TLS, but native TLS is still the default. I think we're about ready to switch the default and monitor the performance and stability impact.

PR #2984, opened 2023-05-31
Partly fix raw clone test

sporksmith:
This removes some dependencies from the shim on native thread local storage (https://github.com/shadow/shadow/issues/1559). Together with modifying the clone test to explicitly set an empty TLS descriptor, this allows the clone tests to work under shadow... sometimes.

I haven't been able to reproduce the remaining failures locally, but I suspect it's due to usage of libc from the shim, which assumes that native thread local storage is set up. In particular, most libc functions will try to set `errno` on error, which it expects to be in native thread local storage. These accesses result in a segfault in the test.

It would be nice if we could set up proper native thread local storage in these tests, but the process for doing so is not well documented, and I think would conflict with the ones that libc sets up (global state is involved). 

Still, getting this at least partly working will make it easier to test different flags to `clone` individually; e.g. to test `clone` without `CLONE_FILES` as a step towards implementing a full `fork`.

### DIFF
--- a/src/lib/vasi-sync/src/sync.rs
+++ b/src/lib/vasi-sync/src/sync.rs
@@ -31,85 +31,106 @@ loom::lazy_static! {
 }
 
 #[cfg(not(loom))]
-pub fn futex_wait(futex_word: &AtomicU32, val: u32) -> nix::Result<i64> {
-    nix::errno::Errno::result(unsafe {
-        libc::syscall(
-            libc::SYS_futex,
-            futex_word as *const _,
-            libc::FUTEX_WAIT,
-            val,
-            core::ptr::null() as *const libc::timespec,
-            core::ptr::null_mut() as *mut u32,
-            0u32,
-        )
-    })
-}
-#[cfg(loom)]
-pub fn futex_wait(futex_word: &AtomicU32, val: u32) -> nix::Result<i64> {
-    // From futex(2):
-    //   This load, the comparison with the expected value, and starting to
-    //   sleep are performed atomically and totally ordered with
-    //   respect to other futex operations on the same futex word.
-    //
-    // We hold a lock on our FUTEXES to represent this.
-    // TODO: If we want to run loom tests with multiple interacting locks,
-    // we should have per-futex mutexes here, and not hold a lock over the
-    // whole list the whole time.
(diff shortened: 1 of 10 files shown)


src/lib/vasi-sync/src/sync.rs
```diff
@@ -30,86 +30,158 @@ loom::lazy_static! {
...
+            )
+        };
+        match raw_res.check() {
+            Ok(()) => Ok(0),
+            Err(e) => Err(nix::errno::Errno::from_i32(e.into())),
+        }
+    }
+    // Miri understands futex syscalls made through the libc wrapper.
+    // It's ok to depend on libc for miri testing.
+    #[cfg(miri)]
```
stevenengler, line 60: Does miri only work with futex() libc calls but not direct futex syscalls?

src/lib/vasi-sync/src/sync.rs
```diff
@@ -30,86 +30,158 @@ loom::lazy_static! {
...
+        raw_res
+            .check()
+            .map_err(|e| nix::errno::Errno::from_i32(e.into()))
+    }
+    // Miri understands futex syscalls made through the libc wrapper.
+    // It's ok to depend on libc for miri testing.
+    #[cfg(miri)]
+    {
+        // The kernel just checks for equality of the value at `futex_word`,
+        // which it interprets as `* const u32`.
```
stevenengler, line 146: It would be nice to be able to dedup this logic for "miri" and "not-miri" so that we don't end up with bugs in only one implementation, but I don't really see a nice way of doing that. Maybe we could write wrapper syscall functions that have the same signature, and then we could assign that to a function pointer to call either using the same arguments. But I think this is fine for this PR, and maybe futex is the only syscall where we need to deal with this problem.

stevenengler (approved):
(Optional) Maybe worth running a benchmark to make sure the memory usage doesn't increase significantly? It seems like it will be ~1 MiB per process, so I don't think it would be much?

sporksmith:
> (Optional) Maybe worth running a benchmark to make sure the memory usage doesn't increase significantly? It seems like it will be ~1 MiB per process, so I don't think it would be much?

The extra memory is allocated lazily the first time we create a thread with no native TLS. In most normal simulations it should never be allocated.

PR #2990, opened 2023-06-01
Upgrade rustup toolchain to 1.70

stevenengler:
The main reason is so we can start using the sparse registry.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -7,4 +7,4 @@
 # that the newly specified version is baked in and doesn't have to be installed
 # (and rust deps rebuilt) during every incremental build.
 [toolchain]
-channel = "1.69.0"
+channel = "1.70.0"
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -28,7 +28,7 @@ jobs:
       run: |
           # Must be a version built with miri; check
           # https://rust-lang.github.io/rustup-components-history/
-          TOOLCHAIN=nightly-2023-03-13
+          TOOLCHAIN=nightly-2023-05-31
           rustup toolchain install ${TOOLCHAIN} --component miri
           rustup override set ${TOOLCHAIN}
           cargo miri setup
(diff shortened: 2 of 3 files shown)


rust-toolchain.toml
```diff
@@ -7,4 +7,4 @@
 # that the newly specified version is baked in and doesn't have to be installed
 # (and rust deps rebuilt) during every incremental build.
 [toolchain]
-channel = "1.69.0"
+channel = "1.70.0"
```
sporksmith, line 10: We'll need to bump the nightly version as well. (There probably ought  to be a comment and pointer here; and maybe we ought to make a rust-toolchain-nightly.toml that the corresponding workflows use somehow?)

```
.github/workflows/sanitizers.yml:          TOOLCHAIN=nightly-2023-03-13
ci/container_scripts/install_extra_deps.sh:  rustup override set nightly-2023-03-13
```

sporksmith (approved):

rust-toolchain.toml
stevenengler: Good catch! We do list these in our "CI Maintenance" workflow so that we don't forget them, but it's easy to forget them when we're doing it outside of the github issue that the workflow creates. I'm not sure how we could use a custom rust toolchain file though. Maybe we could just store it in a text file and then do something like `rustup override set "$(cat ../../rust-nightly-toolchain)"` or somethng?

sporksmith: Or maybe we could do something with symbolic links?

e.g.
* rename `rust-toolchain.toml` to `rust-toolchain-stable.toml`
* add `rust-toolchain-nightly.toml`
* add a symbolic link `rust-toolchain.toml` -> `rust-toolchain-stable.toml`

Then the workflows that need the nightly could just locally reset the symbolic link. That'd make it easier to switch back and forth locally between the "blessed" stable and nightly as well.

If you'd rather not sort it out in this PR, I'm also fine with just changing the version in the current 3 places for the moment :)

stevenengler: Okay that sounds like a good idea. For now I just updated the nightly version in this PR, but I'll make the symlink change in another PR.

PR #2992, opened 2023-06-01
logger: remove some libc usage (and flesh out more of linux-api)

sporksmith:
This is extracted out of an attempt at a more comprehensive migration of our C code from depending on libc headers that conflict with kernel headers (https://github.com/shadow/shadow/pull/2975)

The migration of `logger` is partial, but is some progress. The additions to the `linux-api` crate will be nice to get in though.


sporksmith:
Sorry; looks like there are some build issues to sort out

src/lib/linux-api/src/signal.rs
```diff
@@ -1175,6 +1184,17 @@ mod export {
         unsafe { SigInfo::peel(SigInfo::new_for_kill(signal, sender_pid, sender_uid)) }
     }
 
+    #[no_mangle]
+    pub unsafe extern "C" fn linux_kill(pid: i32, sig: i32) -> i32 {
+        let Ok(sig) = Signal::try_from(sig) else {
+            return Errno::EINVAL.to_negated_i32()
+        };
```
stevenengler, line 1191: Not really relevant for this PR, but it seems like it would be more useful to have this the other way around. To have the rust function take a `pid: i32, sig: i32` and this function would pass those directly. And maybe have a rust wrapper that takes a `Signal`. For example it might be nice in the future to use this library in the shadow integration tests, but we can't because they don't allow a signal that isn't known by the library. For example there's no way to do the equivalent of the C code: `kill(0, INT_MAX)`. This rust wrapper would return `EINVAL` and it would never make it to the kernel, even if that's not what the kernel would return.

src/lib/linux-api/src/time.rs
```diff
@@ -0,0 +1,68 @@
...
+pub use bindings::linux___kernel_clockid_t;
+
+// Clocks
+#[derive(Debug, Copy, Clone, Eq, PartialEq, IntoPrimitive, TryFromPrimitive)]
+#[repr(i32)]
+#[allow(non_camel_case_types)]
+pub enum ClockId {
+    CLOCK_REALTIME = bindings::LINUX_CLOCK_REALTIME as linux___kernel_clockid_t,
+    CLOCK_MONOTONIC = bindings::LINUX_CLOCK_MONOTONIC as linux___kernel_clockid_t,
+    CLOCK_PROCESS_CPUTIME_ID = bindings::LINUX_CLOCK_PROCESS_CPUTIME_ID as linux___kernel_clockid_t,
```
stevenengler, line 19: Since we're casting the Linux types in most of our structs/enums, maybe we should have static assertions for them to make sure we're not truncating the values?

src/lib/linux-api/src/time.rs
```diff
@@ -0,0 +1,68 @@
...
+    .check()
+    .map_err(Errno::from)?;
+    Ok(t)
+}
+
+mod export {
+    use super::*;
+
+    #[no_mangle]
+    pub unsafe extern "C" fn linux_clock_gettime(clockid: i32, res: *mut linux_timespec) -> i64 {
```
stevenengler, line 57: Maybe assert the pointer isn't null?

stevenengler (approved):

src/lib/linux-api/src/time.rs
sporksmith: done

src/lib/linux-api/src/errno.rs
```diff
@@ -0,0 +1,130 @@
...
+    EKEYEXPIRED = bindings::LINUX_EKEYEXPIRED as u16,
+    EKEYREVOKED = bindings::LINUX_EKEYREVOKED as u16,
+    EKEYREJECTED = bindings::LINUX_EKEYREJECTED as u16,
+    EOWNERDEAD = bindings::LINUX_EOWNERDEAD as u16,
+    ENOTRECOVERABLE = bindings::LINUX_ENOTRECOVERABLE as u16,
+    ERFKILL = bindings::LINUX_ERFKILL as u16,
+    EHWPOISON = bindings::LINUX_EHWPOISON as u16,
+
+    #[num_enum(catch_all)]
+    EUNKNOWN(u16),
```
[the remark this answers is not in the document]
sporksmith: Actually I ended up changing this from an enum to a struct, so this went away.

PR #2997, opened 2023-06-05
Migrate timerfd descriptor and syscall handlers to rust

robgjansen:


src/main/host/syscall/handler/timerfd.rs
```diff
@@ -0,0 +1,244 @@
...
+
+            // Error if the process disappeared.
+            let Some(runnable) = ctx.objs.process.borrow_runnable() else {
+                return Err(Errno::EINVAL);
+            };
+
+            // Set up the result values.
+            let result = libc::itimerspec {
+                it_value: remaining.try_into().unwrap(),
+                it_interval: interval.try_into().unwrap(),
```
robgjansen, line 114: To reviewer: please check if these unwraps are appropriate here.

src/main/host/syscall/handler/timerfd.rs
```diff
@@ -0,0 +1,244 @@
...
+        // Now we can adjust the timer with the new_value.
+        if new_value.it_value.tv_sec == 0 && new_value.it_value.tv_nsec == 0 {
+            // A value of 0 disarms the timer; it_interval is ignored.
+            TimerFd::disarm_timer(timerfd);
+            log::trace!("TimerFd {} disarmed", fd);
+        } else {
+            // Need to arm the timer, value may be absolute or relative.
+            let value: SimulationTime = new_value.it_value.try_into().unwrap();
+            let interval: SimulationTime = new_value.it_interval.try_into().unwrap();
+            let now = Worker::current_time().unwrap();
```
robgjansen, line 186: To reviewer: please check if these unwraps are appropriate here.

src/main/host/syscall/handler/timerfd.rs
```diff
@@ -0,0 +1,244 @@
...
+    syscall_types::{SyscallError, SyscallResult},
+};
+
+impl SyscallHandler {
+    #[log_syscall(/* rv */ libc::c_int, /* clockid */ libc::c_int, /* flags */ libc::c_int)]
+    pub fn timerfd_create(
+        ctx: &mut SyscallContext,
+        clockid: libc::c_int,
+        flags: libc::c_int,
+    ) -> SyscallResult {
```
stevenengler, line 26: We're moving towards using the actual linux syscall return type in the syscall handler return type. So this would be `-> Result<libc::c_int, SyscallError>`. For example the `timerfd_create` syscall returns an `int`, but you're returning `fd.val()` in this syscall handler which is a `uint`. Which isn't a big deal, but making the return type explicit will hopefully help detect bugs.

(And for the other syscall handlers below.)

src/main/host/syscall/handler/timerfd.rs
```diff
@@ -0,0 +1,244 @@
...
+        // Continue only if we support the clockid.
+        check_clockid(clockid)?;
+
+        let Some(flags) = TimerFlags::from_bits(flags) else {
+            log::warn!("Invalid timerfd_create flags: {}", flags);
```
stevenengler, line 31: nit: "Invalid timerfd_create flags: {flags}"

(I won't leave comments for any more of these string formatting things.)

src/main/host/syscall/handler/timerfd.rs
```diff
@@ -0,0 +1,244 @@
...
+            return Err(Errno::EINVAL.into());
+        };
```
stevenengler, line 33: I think this should be a `log::debug` or `trace` instead. If the application passes in a flag that isn't valid, I think returning `EINVAL` is the expected thing to do.

src/main/host/syscall/handler/timerfd.rs
```diff
@@ -0,0 +1,244 @@
...
+        log::trace!("timerfd_create() returning fd {}", fd);
+
+        Ok(fd.val().into())
+    }
+
+    #[log_syscall(/* rv */ libc::c_int, /* fd */ libc::c_int, /*curr_value*/ *const libc::c_void)]
+    pub fn timerfd_gettime(
+        ctx: &mut SyscallContext,
+        fd: libc::c_int,
+        curr_value_ptr: ForeignPtr<libc::itimerspec>,
```
stevenengler, line 66: Maybe add a comment that the syscall takes a `__kernel_itimerspec`, which we assume here is the same size/layout/etc as libc's `itimerspec`.

src/main/host/syscall/handler/timerfd.rs
```diff
@@ -0,0 +1,244 @@
...
+        Ok(())
+    } else if clockid == libc::CLOCK_BOOTTIME
+        || clockid == libc::CLOCK_BOOTTIME_ALARM
+        || clockid == libc::CLOCK_REALTIME_ALARM
+    {
+        log::debug!(
+            "Unsupported clockid {}, we support CLOCK_REALTIME and CLOCK_MONOTONIC.",
+            clockid
+        );
+        Err(Errno::ENOSYS.into())
```
stevenengler, line 226: Optional nit: I think it would be better to return `EINVAL` here too. It can be confusing when the logs show `ENOSYS`, since then it seems like we don't support this syscall at all, and that we're maybe not intercepting it properly.

src/main/host/syscall/handler/timerfd.rs
```diff
@@ -0,0 +1,244 @@
...
+
+/// Checks the clockid; returns `Ok(())` if the clockid is `CLOCK_REALTIME` or
+/// `CLOCK_MONOTONIC`, or the appropriate errno if the clockid is unknown or
+/// unsupported.
+fn check_clockid(clockid: libc::c_int) -> Result<(), SyscallError> {
+    if clockid == libc::CLOCK_MONOTONIC || clockid == libc::CLOCK_REALTIME {
+        Ok(())
+    } else if clockid == libc::CLOCK_BOOTTIME
+        || clockid == libc::CLOCK_BOOTTIME_ALARM
+        || clockid == libc::CLOCK_REALTIME_ALARM
```
stevenengler, line 220: nit: maybe simpler as `[libc::CLOCK_BOOTTIME, libc::CLOCK_BOOTTIME_ALARM, libc::CLOCK_REALTIME_ALARM].contains(&clock_id)`?

src/main/host/syscall/handler/timerfd.rs
```diff
@@ -0,0 +1,244 @@
...
+    ) -> Result<(), Errno> {
+        // Lookup the timer state.
+        let (remaining, interval) = {
+            let borrowed = timerfd.borrow();
+
+            // We return a zero duration if the timer is disarmed.
+            let remaining = match borrowed.get_timer_remaining() {
+                Some(simtime) => simtime,
+                None => SimulationTime::ZERO,
+            };
```
stevenengler, line 92: optional nit: maybe simpler as:

```rust
let remaining = borrowed.get_timer_remaining().unwrap_or(SimulationTime::ZERO);
```

src/main/host/syscall/handler/timerfd.rs
```diff
@@ -0,0 +1,244 @@
...
+
+            (remaining, interval)
+        };
+
+        // Write the result to the plugin.
+        {
+            // Need a location to write the result.
+            if value_ptr.is_null() {
+                return Err(Errno::EFAULT);
+            }
```
stevenengler, line 104: We shouldn't need to check for `NULL` here, it should be handled by the eventual memory manager `write()`.

src/main/host/syscall/handler/timerfd.rs
```diff
@@ -0,0 +1,244 @@
...
+
+            // Error if the process disappeared.
+            let Some(runnable) = ctx.objs.process.borrow_runnable() else {
+                return Err(Errno::EINVAL);
+            };
```
stevenengler, line 109: Should this be an unwrap/panic instead? I don't think there's a legitimate case where the process would disappear during a syscall handler?

Or is there a reason you can't do a `ctx.objs.process.memory_borrow_mut().write(value_ptr, &result)` instead without getting an intermediate process reference?

src/main/host/syscall/handler/timerfd.rs
stevenengler: Yep this is good.

src/main/host/syscall/handler/timerfd.rs
```diff
@@ -0,0 +1,244 @@
...
+        ctx: &mut SyscallContext,
+        fd: libc::c_int,
+        flags: libc::c_int,
+        new_value_ptr: ForeignPtr<libc::itimerspec>,
+        old_value_ptr: ForeignPtr<libc::itimerspec>,
+    ) -> SyscallResult {
+        let Some(flags) = TimerSetTimeFlags::from_bits(flags) else {
+            log::warn!("Invalid timerfd_settime flags: {}", flags);
+            return Err(Errno::EINVAL.into());
+        };
```
stevenengler, line 136: This can be a debug or trace log. (See earlier comment.)

src/main/host/syscall/handler/timerfd.rs
```diff
@@ -0,0 +1,244 @@
...
+        let File::TimerFd(ref timerfd) = file else {
+            return Err(Errno::EINVAL.into());
+        };
+
+        // Read in the new value from the plugin.
+        let new_value = {
+            // Need to be able to read the new timer config.
+            if new_value_ptr.is_null() {
+                return Err(Errno::EFAULT.into());
+            }
```
stevenengler, line 149: We don't need the NULL check.

src/main/host/syscall/handler/timerfd.rs
```diff
@@ -0,0 +1,244 @@
...
+
+            // Error if the process disappeared.
+            let Some(runnable) = ctx.objs.process.borrow_runnable() else {
+                return Err(Errno::EINVAL.into());
+            };
```
stevenengler, line 154: Same as comment above. Can we just do `ctx.objs.process.memory_borrow().read(new_value_ptr)` later without getting this `runnable`)?

src/main/host/syscall/handler/timerfd.rs
```diff
@@ -0,0 +1,244 @@
...
+        };
+
+        // Verify a valid range for new_time nanosecond vals.
+        if new_value.it_value.tv_nsec < 0
+            || new_value.it_value.tv_nsec >= 999_999_999
+            || new_value.it_interval.tv_nsec < 0
+            || new_value.it_interval.tv_nsec >= 999_999_999
+        {
+            return Err(Errno::EINVAL.into());
+        }
```
stevenengler, line 169: Optional: Instead of these checks, we could do the conversion to `SimulationTime` here instead. For example:

```rust
let value: SimulationTime = new_value.it_value.try_into().or(Err(Errno::EINVAL))?;
let interval: SimulationTime = new_value.it_interval.try_into().or(Err(Errno::EINVAL))?;

...

if value.is_zero() {
    ...
```

The `try_into` conversions already check the values of `tv_sec` and `tv_nsec`. (The existing checks here don't check of negative `tv_sec` values.)

src/main/host/syscall/handler/timerfd.rs
stevenengler: See comment above. These could panic on negative `tv_sec` values.

stevenengler (approved):
Looks good! There's a bunch of comments, but they're all minor things.

src/main/host/descriptor/timerfd.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    /// Returns the relative duration until the next expiration event occurs if the timer is armed,
+    /// and `None` if the timer is disarmed.
+    pub fn get_timer_remaining(&self) -> Option<SimulationTime> {
+        self.timer.remaining_time()
+    }
+
+    /// Returns the relative duration over which the timer has been configured to periodically
+    /// expire, which may be zero if the timer is only configured to expire once.
+    pub fn get_timer_interval(&self) -> SimulationTime {
```
sporksmith, line 82: Maybe change this and `expire_interval` to return an `Option<SimulationTime>`. Or drop a TODO comment to do so and limit this PR to a more straightforward port of the existing C code.

src/main/host/descriptor/timerfd.rs
```diff
@@ -0,0 +1,275 @@
...
+        request: u64,
+        _arg_ptr: ForeignPtr<()>,
+        _memory_manager: &mut MemoryManager,
+    ) -> SyscallResult {
+        // The only timerfd-specific ioctl request is for `TFD_IOC_SET_TICKS`, which is available
+        // since Linux 3.17 but only if the kernel was configured with `CONFIG_CHECKPOINT_RESTORE`.
+        // See timerfd_create(2) for more details.
+        log::warn!("We do not yet handle ioctl request {} on TimerFds", request);
+        Err(Errno::EINVAL.into())
+    }
```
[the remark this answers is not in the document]
robgjansen: Also, it would be nicer here to return `ENOTSUP` when the ioctl request is `TFD_IOC_SET_TICKS ` and then `EINVAL`  otherwise. However, I'm not finding the `TFD_IOC_SET_TICKS` constant in the libc crate.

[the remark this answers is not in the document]
sporksmith: Yeah, I'd be more comfortable with just logging at debug level if we returned something like `ENOTSUP`, since `EINVAL` incorrectly indicates "you passed bad arguments". OTOH `ENOTSUP` isn't actually listed as a possible error in the `ioctl` man page, so that's not really "right" either.

I like the idea of returning `EINVAL` and logging a warning once... ideally for each distinct value of `request` that we don't support.

Maybe keep it as `warn` (and `EINVAL`) for now and add a `TODO` to make it warn-once when we implement that function?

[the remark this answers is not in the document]
sporksmith: I can see about exporting `TFD_IOC_SET_TICKS` (and `itimerspec` etc) in the `linux-api` crate.

src/main/host/syscall/handler/timerfd.rs
robgjansen: (Here is a good example of a place where the mem mgr public api docs could use a few more details.)

src/main/host/syscall/handler/timerfd.rs
stevenengler: Yeah the eventfd code was written a while ago, and also back before we got some nicer rust features like captured identifiers in format strings.

src/main/host/syscall/handler/timerfd.rs
sporksmith: My thinking when I wrote that is that we should just borrow the `runnable` once in a syscall and use it repeatedly, instead of going through the outer wrapper + borrowing + unwrapping every time.

Maybe the `runnable` ought to be what goes in the syscall handler context object in the first place instead of the `Process` itself. That'd save the overhead + make it clearer that handlers are always dealing with a runnable `Process`.

sporksmith: I'm fine with either way here for the moment, but I suppose just using `process.memory_borrow_mut` makes sense if we intend to change `process` itself to be the `runnable` later.

robgjansen: OK, `process.memory_borrow_mut` is much simpler so let's go that route with a plan to put the runnable in the ctx object later (and keep the syscall handlers cleaner).

src/main/host/descriptor/timerfd.rs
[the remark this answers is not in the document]
stevenengler: Okay sounds good to me. I think it can be merged as-is, and we can update it later if we want.

src/main/host/syscall/handler/timerfd.rs
```diff
@@ -54,17 +55,20 @@ impl SyscallHandler {
             .register_descriptor(desc)
             .or(Err(Errno::ENFILE))?;
 
-        log::trace!("timerfd_create() returning fd {}", fd);
+        log::trace!("timerfd_create() returning fd {fd}");
 
-        Ok(fd.val().into())
+        i32::try_from(fd.val()).or(Err(Errno::ENFILE.into()))
```
stevenengler: I think it would be better to either panic here, or unregister the fd. Otherwise this will leak a timerfd that will sit in the descriptor table forever.

PR #2999, opened 2023-06-06
shim: allow native syscalls from shim's whole .text segment

sporksmith:
This fixes a performance regression introduced in 789f9f247cf9a87c55518cf735725001891f4f5d: making a direct syscall instead of a libc function call in vasi-sync caused us to go through the slow path of getting stopped by seccomp, and have to retry in our seccomp-handling.                   
    
Allowing native syscalls made from the shim's whole `.text` segment includes direct syscalls made from anything linked statically with the shim. This fixes the performance regression and likewise allows us to continue moving from libc to direct syscalls in the shim without sacrificing performance.    
    
We currently don't have a use-case of *wanting* syscalls made directly from the shim to be intercepted. e.g. the closest would be when we call a managed-code-provided signal handler from shim code, and want its syscalls to be intercepted. Those syscalls will still be intercepted though, since they won't be made *directly* from the shim.

### DIFF
--- a/src/lib/shim/shim_seccomp.c
+++ b/src/lib/shim/shim_seccomp.c
@@ -49,6 +49,43 @@ static void _shim_seccomp_handle_sigsys(int sig, siginfo_t* info, void* voidUcon
     ctx->uc_mcontext.gregs[REG_RAX] = rv;
 }
 
+// TODO: dedupe this with `maps` parsing in `patch_vdso.c` and `proc_maps.rs`,
+// ideally into something that doesn't allocate or use libc.
+static void _getSectionContaining(const void* target, void** start, void** end) {
+    assert(start);
+    *start = NULL;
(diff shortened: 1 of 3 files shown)


sporksmith:
Started a tgen benchmark to validate that this fixes the perf regression: https://github.com/shadow/benchmark/actions/runs/5189972251

sporksmith:
Started a tor benchmark too, since the regression isn't as noticeable in the tgen results https://github.com/shadow/benchmark/actions/runs/5190222112

stevenengler (approved):

sporksmith:
Confirmed this brings tor performance back in line with 3.0.0-pre https://github.com/shadow/benchmark-results/blob/master/tor/2023-06-06-T15-07-30/plots/run_time.png

PR #3000, opened 2023-06-06
Add `log_once` macro

stevenengler:
This adds a `log_once_at_level` (and `warn_once_then_debug`/`warn_once_then_trace`) macro to log a message once at one level, and then at a different level every time after. The caller should indicate somewhere in the log message that it's only being logged once.

### DIFF
--- a/src/main/utility/macros.rs
+++ b/src/main/utility/macros.rs
@@ -7,6 +7,70 @@ macro_rules! debug_panic {
     };
 }
 
+/// Log a message once at level `lvl_once`, and any later log messages from this line at level
+/// `lvl_remaining`. A log target is not supported. It is recommended to also prepend "(LOG_ONCE)"
+/// to the log message to indicate that it will not be logged again at that level, but will be
+/// logged at a different level.
+///
+/// ```
+/// # use log::Level;
+/// # use shadow_rs::log_once_at_level;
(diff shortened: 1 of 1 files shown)


stevenengler:
> Log once per distinct set of params. e.g. it'd be nice if
> `warn_once!("Unsupported operation: {}", operation)` warned once for each distinct value of `operation`.

I think if we wanted to do anything more complex than what it does now, we'd need to essentially rewrite the log macro, including things like checking the static max level, formatting the string, and all that stuff, which would be easy to get wrong or introduce bugs.

sporksmith:
> Yeah, I guess we can always duplicate at the call sites.

Oops, that's to address the other problem, which I think we *could* handle pretty easily in the macro?

stevenengler:
> Oops, that's to address the other problem, which I think we _could_ handle pretty easily in the macro? If we're logging at debug every time that at least we can still find the distinct values in the logs if we're at debug level.

Yeah we can do this easily I think. So it should take two log levels, one for the first occurrence, and then once for every other occurrence. What should we name them in this case? Maybe `log_once_at_level`? I think it would be confusing to have a `log_once` macro that logs more than once. And then we could use `warn_once_then_debug`/`warn_once_then_trace` as more specific macros.

sporksmith:
Yup, sounds good

stevenengler:
@sporksmith Letting you take another look at it. I accidentally rebased instead of doing a fixup commit.

sporksmith (approved):

PR #3001, opened 2023-06-06
Move and rename pod

sporksmith:
We need to move the `Pod` trait out of shadow's `main` crate so that our other crates can implement the trait for their own types (instead of the `main` crate implementing `Pod` for ~foreign types from helper crates, which was a bit of fragile long-distance coupling).

This also renames the `Pod` trait to `AnyBitPattern`, since it implements the same concept as [`bytemuck::AnyBitPattern`](https://docs.rs/bytemuck/latest/bytemuck/trait.AnyBitPattern.html), and *not* the same concept as [`bytemuck::Pod`](https://docs.rs/bytemuck/latest/bytemuck/trait.Pod.html)

I think eventually we can *replace* our `AnyBitPattern` with bytemuck's, but we need to move away from using the libc struct types with the MemoryManager first, since we can't implement `bytemuck::AnyBitPattern` for them. We probably want to replace all of those usages with structs in the linux-api crate eventually, and *can* implement `bytemuck::AnyBitPattern` for *them*.

Once we do that, we'll still want this new `bytemuck-util` crate, since it provides some related helpers that `bytemuck` doesn't (though maybe we could upstream them to `bytemuck`).


sporksmith:
Hmm, while working on a follow-up I came across that [bytemuck::AnyBitPattern](https://docs.rs/bytemuck/latest/bytemuck/trait.AnyBitPattern.html) disallows pointers, which doesn't really work for us. It sounds like they added some other project-specific property they needed:

> * It is disallowed for types to contain pointer types, Cell, UnsafeCell, atomics, and any other forms of interior mutability.
> * More precisely: A shared reference to the type must allow reads, and only reads. RustBelt’s separation logic is based on the notion that a type is allowed to define a sharing predicate, its own invariant that must hold for shared references, and this predicate is the reasoning that allow it to deal with atomic and cells etc. We require the sharing predicate to be trivial and permit only read-only access.

It'd probably be ok in practice, but maybe better to just keep our own trait. I'll drop the commits that rename things and just keep the part where we move `pod` out to its own crate.

sporksmith:
It was easier to just redo this as a new PR than try to rewrite the history. Obsoleted by https://github.com/shadow/shadow/pull/3006. Sorry for the churn!

PR #3002, opened 2023-06-06
Make timer intervals optional

robgjansen:


src/main/host/timer.rs
```diff
@@ -158,10 +178,33 @@ impl Timer {
...
+        &mut self,
+        host: &Host,
+        expire_time: EmulatedTime,
+        expire_interval: Option<SimulationTime>,
+    ) {
         self.magic.debug_check();
         debug_assert!(expire_time >= Worker::current_time().unwrap());
 
+        // Zero is not a valid expire interval.
+        let expire_interval = match expire_interval {
```
sporksmith, line 200: A little more succinctly as:

```
if let Some(t) = expire_interval {
  debug_assert!(t.is_positive());
}
```

sporksmith (approved):

Issue #2790, opened 2023-03-16
shim fails to initialize with tor's --enable-fragile-hardening (asan)


sporksmith:
In Tor we're trying to run simulations with a tor binary built with [AddressSanitizer](https://github.com/google/sanitizers/wiki/AddressSanitizer); specifically using the `--enable-fragile-hardening` flag to tor's `configure` script.

Initially, the `asan` instrumentation in the tor binary inspects `LD_PRELOAD` and refuses to run because our `LD_PRELOAD`'d libraries might interfere with its instrumentation. In particular we get an error in the tor's `shimlog` -- since the shim's initialization hasn't run yet the write isn't interposed, and `STDOUT_FILENO` is the `shimlog` file.

We're able to cirumvent that check by setting `ASAN_OPTIONS=verify_asan_link_order=0` in the tor process's `environment`. https://github.com/google/sanitizers/issues/796#issuecomment-578073928

After doing that, the `tor` process runs, but all output goes to the `shimlog`; it appears that the shim is never getting initialized.

I'm able to reproduce this issue with tor-0.4.7.9 and shadow's `tor-minimal` test.

More debugging info follows...

sporksmith:
I'm unable to attach gdb to the tor process using shadow's `--debug-hosts` flag; when I try to attach the process has already exited. IIUC since the shim's initialization code never runs, it never gets blocked, and it runs and exits "normally" before we can attach.

I tried running shadow itself under gdb instead, setting a breakpoint on `_managedthread_fork_exec`, and setting `follow-fork-mode` to `child` when it's about to start the first `tor` process.

Initially I ran into an internal gdb error related to `vfork`. I locally changed the `vfork` in `ManagedThread` to `fork` instead as a workaround.

After doing that, and following the fork in `gdb`, the child `tor` process stops at `SIGABRT`. It looks like it's running into trouble running asan's global *destructor*; i.e. this is already after `tor` tried and failed to run without ever initializing properly:

```
Thread 2.1 "tor" received signal SIGABRT, Aborted.
(gdb) bt
#0  0x00007ffff6a96a7c in pthread_kill () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007ffff6a42476 in raise () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007ffff6a287f3 in abort () from /lib/x86_64-linux-gnu/libc.so.6
[...]

PR #3022, opened 2023-06-12
Hack things up to get tor-minimal-test working with asan'd tor

sporksmith:
~mergeable changes:

* Enable unblocked syscall latency in the tor minimal test. asan uses a spin loop during thread creation.

* Drop the sched_yield exception from our seccomp filter. asan uses an inlined sched_yield syscall in its spin loop which we otherwise weren't intercepting. Luckily we don't need this exception anymore.

Hacks:

* Hack up shim initialization to try again later if env variables don't work yet. (Real fix is to finish migrating off of env variable usage).

* Hard-code the shim *not* to use native thread local storage even if it appears to be available. asan intercepts internal function calls used to access thread-local-storage.

Maybe not needed:

* Change set_robust_list to return 0 instead of ENOSYS. I think this ended up not being needed.

* Hack up our posix_spawn call to force it *not* to use vfork. This was just to make gdb behave better while I was debugging.

* Start some processes slightly later. This was just for debugging so that the first process to start was a tor process, and nothing else tried to start simultaneously.

* Set "ASAN_OPTIONS: verify_asan_link_order=0" so that asan doesn't complain about the shadow shim being first in LD_PRELOAD order.

### DIFF
--- a/src/lib/shim/shim.c
+++ b/src/lib/shim/shim.c
@@ -247,8 +247,11 @@ static void _shim_parent_init_host_shm() {
     assert(shim_hostSharedMem());
 }
 
-static void _shim_parent_init_ipc() {
+static int _shim_parent_init_ipc() {
     const char* ipc_blk_buf = getenv("SHADOW_IPC_BLK");
+    if (!ipc_blk_buf) {
+        return -1;
+    }
     assert(ipc_blk_buf);
     bool err = false;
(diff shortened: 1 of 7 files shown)


sporksmith:
I think we can drop this PR now. The remaining issue(s) for supporting asan'd binaries are in https://github.com/shadow/shadow/issues/2790

Issue #2681, opened 2023-01-20
We don't escape sched_yield busy loops with use_preload_libc=false (or if the code makes a direct syscall)


sporksmith:
Our seccomp filter currently always passes through the `sched_yield` syscall to avoid recursing in the shim's own spin lock.

If we don't intercept `sched_yield` via LD_PRELOAD, e.g. because `use_preload_libc` is false or because the managed code makes a direct syscall, this causes us to execute the syscall natively instead of emulating it. In some cases this is ok, but it breaks our ability to escape busy loops using `model_unblocked_syscall_latency`.

Removing the sched_yield exception in our seccomp filter, in `shim_seccomp_init`, seemed to work OK for me locally, but I wouldn't be surprised if there was a race condition s.t. it caused occasional failures in practice.

The right fix is probably to add a better mechanism for `shadow_spinlock` to ensure its `sched_yield` executes natively.

PR #3023, opened 2023-06-13
shim_seccomp: stop allowing sched_yield

sporksmith:
We needed this exception before to avoid recursive syscall interception in our spinlock implementation. Now that we're allowing direct syscalls from the shim's whole `.text` segment, we don't need to allow `sched_yield` by syscall number.

Fixes https://github.com/shadow/shadow/issues/2681: Removing this exception allows us to intercept direct `sched_yield` syscalls that *aren't* made from the shim itself, thereby letting us escape spin loops that use them.

Progress on https://github.com/shadow/shadow/issues/2790: asan has such spin loops in its thread-creation wrappers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ path" for intercepting syscalls at the library level instead of via seccomp. The
 * Fixed a bug in `sched_getaffinity`. This bug was previously mostly latent due to an incorrectly generated libc syscall wrapper, though would have affected managed programs that
 made the syscall without going through libc.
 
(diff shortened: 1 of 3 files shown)


stevenengler (approved):

sporksmith:
Oh, I should've just used Rustix's `sched_yield` in that test. I should've read the coding guidelines about which crate to use when ;P

sporksmith:
actually maybe it's just as well to use `linux_api` here, since we want to be very sure of what's happening underneath

PR #3026, opened 2023-06-13
Convert clock_nanosleep syscall handler to Rust

robgjansen:
Extended and improved the `clock_nanosleep` Rust tests that we have in place and added a new `clock_nanosleep` Rust handler.

The C handler helper and C sleep tests are still in place until we finish converting `nanosleep`, but I wanted to split that into a separate PR since this one is already getting a bit large.


src/test/time/clock_nanosleep/test_clock_nanosleep.rs
```diff
@@ -0,0 +1,467 @@
...
+                rv: libc::EPERM,
+            }),
+        },
+        Arg::<libc::clockid_t> {
+            value: libc::CLOCK_BOOTTIME_ALARM,
+            validity: Validity::Invalid(Verify {
+                ord: perm_ord,
+                rv: libc::EPERM,
+            }),
+        },
```
robgjansen, line 117: `CLOCK_REALTIME_ALARM` and `CLOCK_BOOTTIME_ALARM` are actually valid clock ids for `clock_nanosleep` if the caller has the `CAP_WAKE_ALARM` capability, despite not being listed as valid in the man page. I verified this by switching these entries to `validity: Validity::Valid,` and running as root and the test passes.

So what should we do in Shadow? Should we
1. just ignore the need for the `CAP_WAKE_ALARM` and support these in Shadow, and then exclude these two clock ids from being tested when running the test on Linux since our tests are not normally run as root?
2. have Shadow return `EPERM` for these clock ids to match the case of Linux running the test without the capability?
3. Have Shadow return `ENOTSUP` for these clock ids, and then exclude the clock ids from being tested when running in Shadow?
4. Something else?

src/test/time/clock_nanosleep/test_clock_nanosleep.rs
```diff
@@ -0,0 +1,451 @@
...
+        tv_nsec: 0,
+    };
+    let rv = unsafe { libc::clock_gettime(clockid, &mut now) };
+    ensure_ord!(rv, ==, 0);
+    Ok(now)
+}
+
+fn timespec_to_duration(ts: libc::timespec) -> Duration {
+    // valid tv_nsec values are [0, 999999999], which fit in a u32
+    Duration::new(ts.tv_sec as u64, ts.tv_nsec as u32)
```
sporksmith, line 458: Optional:

* Using `Duration::from_secs` and `Duration::from_nanos` makes it clearer that the units are correct. 
* Using `try_into` ensures no surprise truncations etc

 `Duration::from_secs(ts.tv_sec.try_into().unwrap()) + Duration::from_nanos(ts.tv_nsec.try_into().unwrap())`

sporksmith: Alternatively: `Duration::from(nix::sys::time::TimeSpec::from_timespec(ts))`

src/test/time/clock_nanosleep/test_clock_nanosleep.rs
```diff
@@ -0,0 +1,451 @@
...
+    flags: libc::c_int,
+    relative_dur: Duration,
+) -> anyhow::Result<libc::timespec> {
+    let absolute_offset = match flags == libc::TIMER_ABSTIME {
+        true => clock_now_duration(clockid)?,
+        false => Duration::ZERO,
+    };
+
+    let request_dur = absolute_offset.checked_add(relative_dur).unwrap();
+    let request = libc::timespec {
```
sporksmith, line 434: Optional alternative: `libc::timespec::from(nix::sys::time::TimeSpec::from(relative_dur))`

src/lib/linux-api/src/errno.rs
```diff
@@ -234,6 +234,7 @@ impl Errno {
     pub const ENOTRECOVERABLE: Self = Self::from_u32_const(bindings::LINUX_ENOTRECOVERABLE);
     pub const ERFKILL: Self = Self::from_u32_const(bindings::LINUX_ERFKILL);
     pub const EHWPOISON: Self = Self::from_u32_const(bindings::LINUX_EHWPOISON);
+    pub const EINTR: Self = Self::from_u32_const(bindings::LINUX_EINTR);
```
sporksmith, line 237: Huh; how'd I miss that one? Could you add it to the `Debug` implementation match statement above too?

src/main/host/syscall/handler/time.rs
```diff
@@ -96,4 +97,105 @@ impl SyscallHandler {
...
+
+    fn nanosleep_helper(
+        ctx: &mut SyscallContext,
+        flags: std::ffi::c_int,
+        request_ptr: ForeignPtr<linux_api::time::timespec>,
+        remain_ptr: ForeignPtr<linux_api::time::timespec>,
+    ) -> Result<std::ffi::c_int, SyscallError> {
+        let request = ctx.objs.process.memory_borrow().read(request_ptr)?;
+        let request_time = SimulationTime::try_from(request).or(Err(Errno::EINVAL))?;
+        let flags = ClockNanosleepFlags::from_bits_truncate(flags);
```
sporksmith, line 153: I don't think we want to silently truncate unknown flags here. e.g.

```
let Ok(flags) = ClockNanosleepFlags::from_bits(flags) else {
  debug!("Invalid flags {flags}");
  return Err(Errno::EINVAL.into())
};
```

src/main/host/syscall/handler/time.rs
```diff
@@ -96,4 +97,105 @@ impl SyscallHandler {
...
+
+        let now = Worker::current_time().unwrap();
+
+        // The requested wakeup time may be absolute or relative.
+        let abs_wakeup_time = match flags.contains(ClockNanosleepFlags::TIMER_ABSTIME) {
```
sporksmith, line 158: An `if` would be more idiomatic here

src/main/host/syscall_condition.rs
```diff
@@ -126,6 +126,20 @@ impl SysCallCondition {
         }
     }
 
+    /// Create a new syscall condition that triggers a wakeup on the calling thread only after the
+    /// `abs_wakeup_time` has been reached.
+    ///
+    /// Panics if `abs_wakeup_time` is before the current emulated time.
+    pub fn new_from_time(abs_wakeup_time: EmulatedTime) -> Self {
```
sporksmith, line 133: `new_from_wakeup_time`?

sporksmith (changes requested):

src/main/host/syscall/handler/time.rs
robgjansen: I think the Linux implementation only checks if `TIMER_ABSTIME` is set, it does not complain or return EINVAL if other bits are set too. Your suggestion would break from that behavior and return EINVAL if bits not defined in `ClockNanosleepFlags` are set.

I should have added a comment to explain why the truncate, so I'll do that.

sporksmith: Yeah, looking at `kernel/time/posix-timers.c` I think you're right :)

src/main/host/syscall/handler/time.rs
```diff
@@ -96,4 +97,105 @@ impl SyscallHandler {
 
         Ok(0.into())
     }
+
+    #[log_syscall(/* clock_id */ linux_api::time::linux___kernel_clockid_t,
+        /* flags */ std::ffi::c_int, /* request */ *const std::ffi::c_void,
```
sporksmith, line 102: Optional - we could get nicer syscall logging by using the higher level types `ClockId` and `ClockNanosleepFlags` here (not in the signature itself). You would need to add a conversion and formatting rule for `ClockNanosleepFlags`  though. For an example see

```
main/host/syscall/type_formatting.rs:simple_debug_impl!(linux_api::time::ClockId);
```

and

```
lib/shadow-shim-helper-rs/src/syscall_types.rs:impl TryFrom<SysCallReg> for linux_api::time::ClockId {
```

Issue #3024, opened 2023-06-13
Consolidate logging


sporksmith:
Our logging infrastructure is a bit complicated due to the shim and shadow main each having specialized logging requirements, and our mix of C and Rust code in each. Additionally, logging in the shim uses libc, contributing to https://github.com/shadow/shadow/issues/2919, and doesn't log from Rust code at all.

# Current logging flow

```dot
digraph {
  log [label="`log` crate (from crates.io): Rust logging frontend"]
  logger [label="`logger`  crate: our C logging frontend"]
  shim_log [label="`ShimLogger`: `logger` back-end for use in shim"]
  shadow_log [label="Shadow logger: `log` back-end for use in shadow"]
  log_wrapper [label="RustLogger: `logger` backend in module `log_wrapper`"]
  c_lang [ label="C code" ]
  rust [ label="Rust code" ]   
  nowhere [ label="dropped" ]
  shadow_stdout [ label="shadow's stdout" ]
  shimlog_files [ label="*.shimlog" ]

  c_lang -> logger
  rust -> log
  shadow_log -> shadow_stdout
  shim_log -> shimlog_files
  log_wrapper -> log

  // in shadow
  log -> shadow_log [label="shadow"]
  logger -> log_wrapper [label="shadow"]

  // in shim
  log -> nowhere [label="shim"]
  logger -> shim_log [label="shim"]
}
```
![logging](https://github.com/shadow/shadow/assets/179146/74f4face-3942-4364-bda2-d9856683bc1d)

# Proposal

* Migrate `ShimLogger` from being a `logger` backend to being a `log` backend, while rewriting it in Rust. This will make it easier to make it not depend on libc, since there are more alternatives from Rust code, but isn't part of this issue itself.
* Merge the `logger` C logging frontend with the `RustLogger` back-end; i.e. make it no longer a generalized C logging framework, but instead always delegate to the Rust `log` crate. Getting rid of this extra "framework" layer will let us simplify the code. This will be the new crate `log-c2rust`.

```dot
digraph {
  log [label="`log` crate (from crates.io): Rust logging frontend"]
  log_c2rust  [label="log_c2rust: previously `logger`+`RustLogger`"]
  shim_log [label="`ShimLogger`: `log` back-end for use in shim"]
  shadow_log [label="Shadow logger: `log` back-end for use in shadow"]
  c_lang [ label="C code" ]
  rust [ label="Rust code" ]
  shadow_stdout [ label="shadow's stdout" ]
[...]

PR #3030, opened 2023-06-14
Build shim as a Rust crate

sporksmith:
This clears the way to start converting shim-only code to Rust, e.g. the `ShimLogger`, which is needed for https://github.com/shadow/shadow/issues/3024


src/lib/vasi-sync/src/scchannel.rs
```diff
@@ -160,6 +161,7 @@ impl Display for SelfContainedChannelError {
 /// microbenchmark; the latter is expected to be more representative of real
 /// large simulation runs (i.e. pinning should be enabled).
 #[cfg_attr(not(loom), derive(VirtualAddressSpaceIndependent))]
+#[repr(C)]
 pub struct SelfContainedChannel<T> {
     message: UnsafeCell<MaybeUninit<T>>,
     state: AtomicChannelState,
```
stevenengler, line 167: Shouldn't this have been enforced by the `_validate_stable_layout` in the `ShMemBlock`?

sporksmith: I think that might only catch if the top-level type has a non-stable repr.

sporksmith: which in this case is `IPCData`

sporksmith: Actually I wonder if we could get a recursive check by moving `_validate_stable_layout` into one of the marker traits. will experiment with that...

sporksmith: Adding the function to the trait doesn't work; trait methods can't take `self`, only `&self`, which doesn't forbid `Self` from being `[repr(rust)]`.

Maybe the derive macro could *inject* the `_validate_stable_layout` method?

stevenengler: I think it is recursive, but it's not working because it's generic.  For example the following successfully fails to build:

```rust
#[repr(C)]
struct Test(TestInner);

#[repr(C)]
struct TestInner(Vec<i32>);

#[deny(improper_ctypes_definitions)]
extern "C" fn _test(_x: Test) {}
```

But in `ShMemBlock` it uses a generic. For example the following does build without errors:

```
#[repr(C)]
struct Test(TestInner);

#[repr(C)]
struct TestInner(Vec<i32>);

#[deny(improper_ctypes_definitions)]
extern "C" fn test<T>(_x: T) {}

fn main() {
    test(Test(TestInner(Vec::new())));
}
```

stevenengler: > Maybe the derive macro could inject the `_validate_stable_layout` method?

This is the only way I can think of to make this work. I think we could add this to the `VirtualAddressSpaceIndependent` derive and specify the type in the signature explicitly, or maybe make a new derive macro named something like `AssertStableLayout` which does this, and then `ShmMemBlock` can require that `T: VirtualAddressSpaceIndependent + AssertStableLayout`.

Anyway this isn't needed for this PR, but we should probably fix this so that it does detect non-stable layouts in shared memory, or remove the existing `_validate_stable_layout` that doesn't seem to work.

.github/workflows/extra_tests.yml
```diff
@@ -118,6 +118,11 @@ jobs:
             -name target -prune \
             -or -name '*.o' -prune \
             -or -print | xargs tar --no-recursion --append -f $TARFILE
+          # Above command skips the whole target directories, but we
+          # need .so outputs. Add those explicitly:
+          find shadow \
+            -path '*/target/*.so' \
+            | xargs tar --no-recursion --append -f $TARFILE
```
stevenengler, line 125: I think this should be scoped to just `build/src/target`? That way we don't accidentally include a target directory like `src/target` which may exist if we ran `rustfmt --check` or something.

src/CMakeLists.txt
```diff
@@ -121,6 +121,65 @@ add_test(NAME rust-unit-tests
...
+    rust-shadow-shim-project
+    PREFIX ${CMAKE_CURRENT_BINARY_DIR}
+    BINARY_DIR ${CMAKE_CURRENT_SOURCE_DIR}
+    BUILD_ALWAYS 1
+    DOWNLOAD_COMMAND ""
+    CONFIGURE_COMMAND ""
+    BUILD_COMMAND bash -c " \
+        ${CARGO_ENV_VARS} \
+        cargo build \
+        -p shadow-shim \
```
stevenengler, line 161: nit: maybe `--package` for readability

src/CMakeLists.txt
```diff
@@ -121,6 +121,65 @@ add_test(NAME rust-unit-tests
...
+add_library(shadow-shim SHARED IMPORTED GLOBAL)
+add_dependencies(shadow-shim rust-shadow-shim-project)
+set_target_properties(shadow-shim PROPERTIES IMPORTED_LOCATION_DEBUG ${TARGETDIR}/${RUST_TARGET}/debug/libshadow_shim.so)
+set_target_properties(shadow-shim PROPERTIES IMPORTED_LOCATION_RELEASE ${TARGETDIR}/${RUST_TARGET}/release/libshadow_shim.so)
+# Use this instead when we're able to bump the cmake min version to >= 3.21:
+# > install(IMPORTED_RUNTIME_ARTIFACTS shadow-shim LIBRARY DESTINATION lib)
+install(FILES ${TARGETDIR}/${RUST_TARGET}/debug/libshadow_shim.so DESTINATION lib CONFIGURATIONS Debug)
+install(FILES ${TARGETDIR}/${RUST_TARGET}/release/libshadow_shim.so DESTINATION lib CONFIGURATIONS Release)
+# Rust's std needs -ldl and -lpthread. Remove this when the shim is no_std.
+target_link_libraries(shadow-shim INTERFACE dl pthread)
```
stevenengler, line 181: Why does cmake even do any linking, shouldn't cargo handle that?

src/lib/libc_preload/libc_impls.c
```diff
@@ -33,10 +33,21 @@
...
+    long arg2 = va_arg(args, long);
+    long arg3 = va_arg(args, long);
+    long arg4 = va_arg(args, long);
+    long arg5 = va_arg(args, long);
+    long arg6 = va_arg(args, long);
     va_end(args);
-    return rv;
+
+    return shim_api_syscall(n, arg1, arg2, arg3, arg4, arg5, arg6);
 }
```
stevenengler, line 51: Won't this read uninitialized/invalid memory if they passed fewer than 6 arguments?

src/lib/shim/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+[build-dependencies]
+bindgen = { version = "0.65.1" }
+cbindgen = { version = "0.24.5" }
+cc = { version = "1.0", features = ["parallel"] }
+shadow-build-common = { path = "../shadow-build-common" }
+system-deps = "6.1"
+
+[package.metadata.system-deps]
+# Keep consistent with the minimum version number in src/CMakeLists.txt
+glib = { name = "glib-2.0", version = "2.32" }
```
stevenengler, line 27: I don't see this anywhere in `src/CMakeLists.txt`?

src/lib/shim/build.rs
```diff
@@ -0,0 +1,104 @@
...
+            "netdb.h".into(),
+            "stdarg.h".into(),
+            "sys/socket.h".into(),
+            "sys/types.h".into(),
+            "netdb.h".into(),
+        ],
+        after_includes: {
+            let mut v = base_config.after_includes.clone().unwrap_or_default();
+            // We have to manually create the vararg declaration.
+            v.push_str("long shim_api_syscall(long n, ...);\n");
```
stevenengler, line 48: Is there any guarantee that a vararg function has the same calling convention as a non-vararg function? And probably worth adding a comment in the bindings that passing more than 7 args will invoke UB (not that I expect anyone to here, but I think it's worth documenting when the function prototype doesn't match the function definition).

src/lib/shim/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// Pointers must be dereferenceable.
+    #[no_mangle]
+    pub unsafe extern "C" fn shim_api_getaddrinfo(
+        node: *const core::ffi::c_char,
+        service: *const core::ffi::c_char,
+        hints: *const libc::addrinfo,
+        res: *mut *mut libc::addrinfo,
+    ) -> i32 {
+        unsafe { bindings::shimc_api_getaddrinfo(node, service, hints, res) }
+    }
```
stevenengler, line 57: Why do these functions need to be re-exported from C back to C?

src/lib/shim/shim.c
```diff
@@ -414,7 +414,7 @@ static void _shim_child_init_preload() {
 
 // This function should be called before any wrapped syscall. We also use the
 // constructor attribute to be completely sure that it's called before main.
-__attribute__((constructor)) void _shim_load() {
+void _shim_load() {
```
stevenengler, line 417: Where does this get called from now if not a constructor?

stevenengler (approved):

.github/workflows/extra_tests.yml
sporksmith: I don't think it'd actually hurt much other than to make the artifact a bit bigger. I changed this to explicitly grab just the shim .so, though.

src/CMakeLists.txt
sporksmith: done

src/CMakeLists.txt
sporksmith: libinjector (a C target) links against the shim. Thinking about it some more though, I *think* we don't need this since the `.so` already has its runtime dependencies embedded in its metadata. Removed.

src/lib/libc_preload/libc_impls.c
sporksmith: These arguments are guaranteed to be passed in registers by the ABI. `va_start` will have initialized them with *some* register values, even if the caller didn't explicitly pass them.

I agree it feels a bit sketchy, but is also what e.g. glibc does and is basically the only way to implement `syscall` without mapping every syscall number to number of expected arguments. https://github.com/bminor/glibc/blob/d2fda60e7c4072180ba91df46bbbdacc0f4a133c/sysdeps/unix/sysv/linux/syscall.c#L23

src/lib/shim/Cargo.toml
sporksmith: Should be `/CMakelists.txt`. And was out of date. Updated version numbers and comments.

src/lib/shim/build.rs
sporksmith: > Is there any guarantee that a vararg function has the same calling convention as a non-vararg function?

I haven't been able to find super authoritative documentation to that effect, but we've used it elsewhere. Here's a somewhat related reference: https://www.gnu.org/software/libc/manual/html_node/Why-Variadic.html :

> Some functions such as open take a fixed set of arguments, but occasionally ignore the last few. Strict adherence to ISO C requires these functions to be defined as variadic; in practice, however, the GNU C compiler and most other C compilers let you define such a function to take a fixed set of arguments—the most it can ever use—and then only declare the function as variadic (or not declare its arguments at all!). 

> And probably worth adding a comment in the bindings that passing more than 7 args will invoke UB

Similar to above; this isn't UB. Extra arguments are just ignored.

src/lib/vasi-sync/src/scchannel.rs
sporksmith: I tried doing this locally but it didn't seem to catch the error. Oh well, I'll open another PR for follow-up

PR #3032, opened 2023-06-15
Update `UdpSocket` to buffer byte chunks rather than packets

stevenengler:


src/main/host/host.rs
```diff
@@ -601,9 +601,11 @@ impl Host {
         res
     }
 
-    pub fn get_next_packet_priority(&self) -> f64 {
+    pub fn get_next_packet_priority(&self) -> u64 {
         let res = self.packet_priority_counter.get();
-        self.packet_priority_counter.set(res + 1.0);
+        // do a saturating add since if we did ever reach this number, we wouldn't want new packets
+        // to have a better priority
+        self.packet_priority_counter.set(res.saturating_add(1));
```
sporksmith, line 608: What happens if we somehow reach this? Out of order packet delivery? Nondeterminism in the simulation?

I think `checked_add` with `unwrap` would be preferable. We really shouldn't be able to increment through all the u64s in practice.

src/main/host/descriptor/socket/inet/udp.rs
```diff
@@ -887,17 +933,43 @@ impl UdpSocket {
...
+/// Non-payload data for a message in the send buffer.
+#[derive(Debug)]
+struct MessageSendHeader {
+    /// The source address (typically the bind address). The application can theoretically use
+    /// `IP_PKTINFO` to set a per-message source address.
+    src: SocketAddrV4,
+    /// The destination address (for example the peer).
+    dst: SocketAddrV4,
+    /// The priority for the packet that we'll create in the future, given to us by the host.
+    packet_priority: u64,
```
sporksmith, line 945: Optional - maybe it's be worth creating a type for packet priorities? Maybe it's not worth the fuss while much of the code is still in C though.

sporksmith (approved):

src/main/host/host.rs
stevenengler: If we were to reach this, packets would still be in order and the simulation should still be deterministic (unless the fifo qdisc has a non-determinism bug). We already have packets that have duplicate packet priorities, such as control packets that all have a priority of 0.

I'll change this to a panic.

src/main/host/descriptor/socket/inet/udp.rs
stevenengler: Added a `FifoPacketPriority` typedef.

PR #3035, opened 2023-06-15
VirtualAddressSpaceIndependent: require FFI safety

sporksmith:
@stevenengler FYI here's my initial attempt. This doesn't have the `repr` fix from https://github.com/shadow/shadow/pull/3030, so *should* fail to build, but doesn't. I'll try to poke at this some more when I have some cycles, but feel free to also poke if you like :)

### DIFF
--- a/src/lib/vasi-macro/src/lib.rs
+++ b/src/lib/vasi-macro/src/lib.rs
@@ -8,18 +8,33 @@ use syn::{parse_quote, Attribute, GenericParam, Generics, Type};
 /// Implement `vasi::VirtualAddressSpaceIndependent` for the annotated type.
 /// Requires all fields to implement `vasi::VirtualAddressSpaceIndependent`.
 ///
-/// An empty struct trivially qualifies:
-/// ```
+/// An empty struct fails becase Rust doesn't consider fieldless structs to be
+/// FFI-safe:
+/// ```compile_fail
 /// use vasi::VirtualAddressSpaceIndependent;
 ///
 /// #[derive(VirtualAddressSpaceIndependent)]
+/// #[repr(C)]
 /// struct Foo {}
 /// ```
 ///
-/// A struct containing only `VirtualAddressSpaceIndependent` fields qualifies:
+/// FFI-safe structs containing only `VirtualAddressSpaceIndependent`
+/// fields qualify:
 /// ```
 /// use vasi::VirtualAddressSpaceIndependent;
 ///
+/// #[repr(C)]
+/// #[derive(VirtualAddressSpaceIndependent)]
+/// struct Foo {
+///   x: i32,
+/// }
+/// ```
+///
+/// `#[repr(transparent)]` is OK too.
+/// ```
+/// use vasi::VirtualAddressSpaceIndependent;
+///
+/// #[repr(transparent)]
 /// #[derive(VirtualAddressSpaceIndependent)]
 /// struct Foo {
 ///   x: i32,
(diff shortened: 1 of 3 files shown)


stevenengler:
I got it a little further with:

```rust
// a hopefully unique name
let fn_name = quote::format_ident!("{}_fn", name);

let gen = quote! {
    unsafe impl #impl_generics vasi::VirtualAddressSpaceIndependent for #name #ty_generics #where_clause {
        const IGNORE: () = {
            const fn check<T: ::vasi::VirtualAddressSpaceIndependent>() {}
            #calls_to_check
        };
    }

    #[deny(improper_ctypes_definitions)]
    #[allow(non_snake_case)]
    extern "C" fn #fn_name #impl_generics (_: #name #ty_generics) {}
};
```

which results in errors for `ChannelState`, `AtomicChannelState`, and `SelfContainedChannel<T>`; ex:

```
error: `extern` fn uses type `SelfContainedChannel<T>`, which is not FFI-safe
   --> lib/vasi-sync/src/scchannel.rs:163:12
    |
163 | pub struct SelfContainedChannel<T> {
    |            ^^^^^^^^^^^^^^^^^^^^^^^ not FFI-safe
    |
    = help: consider adding a `#[repr(C)]` or `#[repr(transparent)]` attribute to this struct
    = note: this struct has unspecified layout
note: the type is defined here
   --> lib/vasi-sync/src/scchannel.rs:163:1
    |
163 | pub struct SelfContainedChannel<T> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: the lint level is defined here
   --> lib/vasi-sync/src/scchannel.rs:162:30
    |
162 | #[cfg_attr(not(loom), derive(VirtualAddressSpaceIndependent))]
    |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: this error originates in the derive macro `VirtualAddressSpaceIndependent` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I'm not sure that this covers every case, but it does seem to catch types with generics (for example `SelfContainedChannel<T>`). Since the macro requires that all fields are also `VirtualAddressSpaceIndependent`, I think that means that all fields will also be ffi-safe?

src/lib/vasi-macro/src/lib.rs
```diff
@@ -259,6 +284,10 @@ fn impl_derive_virtual_address_space_independent(ast: syn::DeriveInput) -> proc_
                 #calls_to_check
             };
         }
+        #[deny(improper_ctypes_definitions)]
+        const _: () = {
+            extern "C" fn _vasi_validate_ffi_safe #impl_generics (_: #name #ty_generics) #where_clause {}
+        };
```
stevenengler, line 290: maybe add a comment about why this exists?

src/lib/vasi-macro/src/lib.rs
```diff
@@ -8,28 +8,43 @@ use syn::{parse_quote, Attribute, GenericParam, Generics, Type};
...
 ///
 /// #[derive(VirtualAddressSpaceIndependent)]
-/// struct Foo {}
+/// #[repr(C)]
+/// struct FooC {}
+///
+/// #[repr(transparent)]
+/// #[derive(VirtualAddressSpaceIndependent)]
+/// struct FooTransparent {}
 /// ```
```
stevenengler, line 22: Why does this doc comment fail to compile? They're empty structs, one is repr(C) and the other is repr(transparent)? Or are zero-sized types not ffi safe?

stevenengler (approved):

src/lib/vasi-macro/src/lib.rs
sporksmith:
```
error: `extern` fn uses type `FooC`, which is not FFI-safe
 --> src/lib.rs:17:8
  |
7 | struct FooC {}
  |        ^^^^ not FFI-safe
  |
  = help: consider adding a member to this struct
  = note: this struct has no fields
```

Tweaked and broke up this example a bit for clarity

src/lib/vasi-macro/src/lib.rs
sporksmith: done

PR #3037, opened 2023-06-16
Support `SIOCGSTAMP` for inet sockets

stevenengler:
This is only supported by Linux for udp sockets, so tcp has been updated to only return `ENOENT`.

### DIFF
--- a/src/test/socket/ioctl/test_ioctl.rs
+++ b/src/test/socket/ioctl/test_ioctl.rs
@@ -81,8 +81,12 @@ fn get_tests() -> Vec<test_utils::ShadowTest<(), String>> {
                 test_utils::ShadowTest::new(
                     &append_args("test_siocgstamp"),
                     move || test_siocgstamp(init_method, sock_type),
-                    // TODO: this isn't supported yet in shadow
-                    set![TestEnv::Libc],
+                    // TODO: this isn't supported yet in shadow for unix sockets
+                    if init_method.domain() == libc::AF_UNIX {
+                        set![TestEnv::Libc]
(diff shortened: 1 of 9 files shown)


src/test/socket/ioctl/test_ioctl.rs
```diff
@@ -184,3 +194,112 @@ fn test_fionread(init_method: SocketInitMethod, sock_type: libc::c_int) -> Resul
...
+            _ => unimplemented!(),
+        };
+        match expected_err {
+            None => {
+                // the receive time reported by the kernel
+                let recv_time = ioctl_siocgstamp(fd_peer).unwrap();
+                let recv_time = Duration::from_secs(recv_time.tv_sec.try_into().unwrap())
+                    + Duration::from_micros(recv_time.tv_usec.try_into().unwrap());
+
+                // get the difference (we can't know if the receive or send time will be smaller)
```
sporksmith, line 289: How could the receive time be before the send time?

sporksmith (approved):

src/test/socket/ioctl/test_ioctl.rs
stevenengler: I expanded the comment here. We measure the send time as soon as the `send()` call returns, but the message may have already been received before we can take that measurement. For example:

1. `send()` call
  a. since it's localhost, the packet is given directly to the receiving socket within the `send()` call
  b. receive time is recorded by the kernel here
2. shadow records the send time with `let send_time = std::time::SystemTime::now()`

In this case the send time is later than the receive time.

PR #3042, opened 2023-06-16
Migrate ShimLogger to Rust

sporksmith:
* Adds a new crate `formatting-nostd` to support string formatting in `no_std` (i.e. no allocation).
* Also adds our own vsnprintf implementation that doesn't allocate, though does currently still call the libc implementation.
* Moves `TimeParts` from crate `main` to `shadow-shim-helper-rs`
* Reimplements the `ShimLogger` in Rust instead of C. This version of the `ShimLogger` is a backend to Rust's `log` crate instead of our C `logger`, which simplifies the flow and allows us to log from Rust code in the shim as well.

Progress on https://github.com/shadow/shadow/issues/3024


src/lib/shim/src/shimlogger.rs
```diff
@@ -0,0 +1,148 @@
...
+                    + Duration::from_nanos(now.tv_nsec.try_into().unwrap());
+
+                let start = Duration::from_micros(m.log_start_time_micros.try_into().unwrap());
+                let elapsed = now - start;
+                let parts = TimeParts::from_nanos(elapsed.as_nanos());
+                core::fmt::write(
+                    &mut writer,
+                    format_args!("{} ", parts.fmt_hr_min_sec_nano()),
+                )
+                .unwrap();
```
stevenengler, line 100: nit: just `write!(&mut writer, "{} ", parts.fmt_hr_min_sec_nano())`? (and for other writes below)

stevenengler (approved):

src/lib/shim/src/lib.rs
```diff
@@ -12,9 +14,21 @@ mod bindings {
     include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
 }
 
+pub fn simtime() -> Option<SimulationTime> {
+    SimulationTime::from_c_simtime(unsafe { bindings::shim_sys_get_simtime_nanos() })
+}
+
+pub fn manager_shmem() -> Option<&'static ManagerShmem> {
+    unsafe { bindings::shim_managerSharedMem().as_ref() }
+}
```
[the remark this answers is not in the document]
sporksmith: done

src/lib/shim/src/shimlogger.rs
sporksmith: ah, yeah that's nicer

sporksmith (commented):
Thanks! Lots of excellent catches and suggestions as usual!

PR #3045, opened 2023-06-17
Sleep rafactor

robgjansen:
This PR refactors and moves a bunch of functionality written for the `clock_nanosleep` out to the test_utils lib so that it can be re-used in other tests.

### DIFF
--- a/src/test/test_utils.rs
+++ b/src/test/test_utils.rs
@@ -16,6 +16,7 @@ use nix::sys::signal;
 use nix::sys::time::TimeVal;
 
 pub mod socket_utils;
+pub mod time;
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
 pub enum TestEnvironment {
@@ -345,6 +346,20 @@ pub fn nop_sig_handler() -> nix::sys::signal::SigHandler {
     nix::sys::signal::SigHandler::Handler(nop_handler)
 }
 
+pub fn install_nop_signal_handler() -> anyhow::Result<()> {
+    unsafe {
+        nix::sys::signal::sigaction(
+            nix::sys::signal::SIGUSR1,
+            &nix::sys::signal::SigAction::new(
+                nop_sig_handler(),
+                nix::sys::signal::SaFlags::empty(),
+                nix::sys::signal::SigSet::empty(),
+            ),
+        )?
+    };
+    Ok(())
+}
+
 /// Convenience wrapper around `anyhow::ensure` that generates useful error messages.
 ///
 /// Example:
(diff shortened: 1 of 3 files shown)


src/test/test_utils.rs
```diff
@@ -538,3 +557,23 @@ pub fn verify_syscall_result(
...
+
+/// Run a function and check that it returns within an expected duration.
+pub fn check_fn_exec_duration<F>(
+    expected: Duration,
+    tolerance: Duration,
+    f: F,
+) -> anyhow::Result<()>
+where
+    F: FnOnce() -> anyhow::Result<()>,
+{
```
stevenengler, line 569: You can make this function a bit more flexible using:

```rust
pub fn check_fn_exec_duration<F, E>(
    expected: Duration,
    tolerance: Duration,
    f: F,
) -> anyhow::Result<()>
where
    F: FnOnce() -> Result<(), E>,
    anyhow::Error: From<E>,
{
```

This way `F` can return almost any error (including those from `result_assert()`), not just anyhow errors.

src/test/test_utils.rs
```diff
@@ -445,3 +479,101 @@ where
...
+    check.sort();
+
+    if let Some(error) = check.first() {
+        // Should have been error, which are returned in rv (not in errno)
+        if let Some(expected_rv) = error.rv {
+            ensure_ord!(expected_rv, ==, rv);
+        }
+        if let Some(expected_errno) = error.errno {
+            ensure_ord!(expected_errno, ==, errno);
+        }
```
stevenengler, line 553: Does this comment mean this code is specifically for libc calls that return errors in the rv rather than errno?

src/test/test_utils.rs
```diff
@@ -445,3 +479,101 @@ where
...
+    Fifth,
+    Sixth,
+}
+
+#[derive(Debug, Copy, Clone, Eq, Ord, PartialEq, PartialOrd)]
+pub struct Verify {
+    pub order: VerifyOrder,
+    pub rv: Option<libc::c_int>,
+    pub errno: Option<libc::c_int>,
+}
```
stevenengler, line 498: (Optional) I don't completely understand this code, but instead of having an abstract "First", "Second", etc, maybe it would be nicer if `Verify` was generic over `T: Debug + Copy + Clone + Eq + Ord + PartialEq + PartialOrd` and used `pub order: T` instead of `pub order: VerifyOrder`. That way the calling code could call `Verify::new()` with its own enum (ex: `enum ClockVerifyOrder { ClockId, Request, Flags }`, or even just `u8`) and use its own ordering instead. But maybe having just "first", "second", etc is clearer. Either way is fine with me, just with some more documentation.

stevenengler (approved):

PR #3046, opened 2023-06-19
Implement clock_getres syscall handler in Rust

robgjansen:


src/test/time/clock_getres/test_clock_getres.rs
```diff
@@ -0,0 +1,135 @@
...
+    let resolutions: Vec<FuzzArg<*mut libc::timespec>> = vec![
+        FuzzArg::new(
+            &mut libc::timespec {
+                tv_sec: 0,
+                tv_nsec: 0,
+            },
+            Ok(()),
+        ),
+        FuzzArg::new(std::ptr::null_mut(), Ok(())),
+    ];
```
stevenengler, line 66: I think this is UB. It's constructing a pointer to a temporary stack object, but this object will immediately go out of scope.

stevenengler (changes requested):

src/test/time/clock_getres/test_clock_getres.rs
robgjansen: Good find, thanks!

stevenengler (approved):

PR #3050, opened 2023-06-21
linux-api: add rt_sigaction and rt_sigprocmask syscall functions

sporksmith:
We need these for a `no_std`Rust implementation of the shim's signal handling. Rustix doesn't provide them.


src/lib/linux-api/src/signal.rs
```diff
@@ -942,10 +947,39 @@ pub enum SignalHandler {
...
+///
+/// Normally libc's implementation of the `sigaction` function injects a similar
+/// restorer function.
+///
+/// # Safety
+///
+/// This function is only intended for use as a `restorer` in `sigaction`.
+/// Do not call this function directly.
+//
+// TODO: use the language-provided `naked` attribute once stabilized.
```
stevenengler, line 960: nit: link to the rfc or stabilization PR to make it easier to check if it's been stabilized?

src/lib/linux-api/src/signal.rs
```diff
@@ -942,10 +947,39 @@ pub enum SignalHandler {
...
 
+/// Restorer suitable for use with `sigaction`.
+///
+/// Normally libc's implementation of the `sigaction` function injects a similar
+/// restorer function.
+///
+/// # Safety
+///
+/// This function is only intended for use as a `restorer` in `sigaction`.
+/// Do not call this function directly.
```
stevenengler, line 958: Can you expand the documentation a bit to explain what this does and why/when it's used? For example what does the sigaction restorer do, and why is it marked as a `naked_function::naked`?

src/lib/linux-api/src/signal.rs
```diff
@@ -960,6 +994,58 @@ impl sigaction {
...
+    ) -> Self {
+        let handler = match handler {
+            SignalHandler::Handler(h) => Some(h),
+            SignalHandler::Action(a) => {
+                assert!(flags.contains(SigActionFlags::SA_SIGINFO));
+                Some(unsafe {
+                    core::mem::transmute::<
+                        unsafe extern "C" fn(i32, *mut siginfo_t, *mut core::ffi::c_void),
+                        unsafe extern "C" fn(i32),
+                    >(a)
```
stevenengler, line 1013: Can this use the `SignalHandlerFn` and `SignalActionFn` like in the transmute below? And maybe a comment explaining why this is safe?

src/lib/linux-api/src/signal.rs
```diff
@@ -960,6 +994,58 @@ impl sigaction {
...
+                })
+            }
+            SignalHandler::SigIgn => Some(unsafe {
+                core::mem::transmute::<usize, unsafe extern "C" fn(i32)>(Self::SIG_IGN)
+            }),
```
stevenengler, line 1018: Is it allowed by rust to transmute from a non-pointer to a `fn()`? I guess there's no other way to construct a C sigaction struct without doing this, but it would be nice to have some comments here explaining why it's believed to be safe.

src/lib/linux-api/src/signal.rs
```diff
@@ -1097,6 +1183,142 @@ pub fn kill(pid: i32, sig: Option<Signal>) -> Result<(), Errno> {
...
+            SigActionFlags::empty(),
+            sigset_t::EMPTY,
+        );
+
+        // Signal that we'll be using.
+        let signal = Signal::SIGUSR2;
+
+        // Install our handler.
+        let mut old_action = zeroed();
+        unsafe { rt_sigaction(signal, &action, Some(&mut old_action)) }.unwrap();
```
stevenengler, line 1324: Maybe add a comment for this test that this handler is set globally across all tests, so if another test sets its own signal handler in another thread for this signal, there would be a race condition.

stevenengler (approved):

src/lib/linux-api/src/signal.rs
```diff
@@ -1,3 +1,8 @@
+// The `naked` macro requires this import, but the compiler
+// sees it as unused.
+#[allow(unused)]
+use core::arch::asm;
```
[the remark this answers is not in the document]
sporksmith: Ah, yeah it turns out this isn't needed. It seems that the `naked` macro expects to see `asm!` inside the function, but rewrites it...

src/lib/linux-api/src/signal.rs
sporksmith: done. Mostly referenced and pasted `sigreturn(2)`.

src/lib/linux-api/src/signal.rs
sporksmith: yeah, I'd tried chasing it down and sort of gave up. I added a comment pointing to https://github.com/rust-lang/rfcs/blob/master/text/1201-naked-fns.md as a starting point, but the status is unclear.

src/lib/linux-api/src/signal.rs
sporksmith: done

src/lib/linux-api/src/signal.rs
sporksmith: Yeah, unclear but there's no other way. Added some explanation and extra checks to help ensure it gets transmuted back to the correct type when interpreted.

src/lib/linux-api/src/signal.rs
sporksmith: done

PR #3052, opened 2023-06-22
Fix strace logging for time syscall handlers

stevenengler:
Fixed for `clock_getres`, `clock_nanosleep`, and `nanosleep`.

### DIFF
--- a/src/main/host/syscall/handler/time.rs
+++ b/src/main/host/syscall/handler/time.rs
@@ -52,7 +52,8 @@ impl SyscallHandler {
         Ok(0.into())
     }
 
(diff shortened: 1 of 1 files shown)


robgjansen (approved):
Sorry that I messed these up. Do we or should we have a way to test this to make sure we're doing it correctly in future PRs?

stevenengler:
> Sorry that I messed these up. Do we or should we have a way to test this to make sure we're doing it correctly in future PRs?

No problem, I was just confused at why the strace log was showing `clock_nanosleep` return `CLOCK_REALTIME` :) And not really because we don't have a ground-truth to compare to. Ideally I think I'd like to get rid of this macro completely and have a separate rust module that has this list of arguments for all syscalls (including syscalls we don't yet support). But for now I don't think there's anything better we can do here.

PR #3053, opened 2023-06-22
Improve formatting of bitflags types in strace logging

stevenengler:
A newer bitflags version changed the debug formatting to include the type name (for example `ClockNanosleepFlags(TIMER_ABSTIME)` instead of `TIMER_ABSTIME`) which adds noise to the strace log, so instead we use more explicit formatting.

### DIFF
--- a/src/main/host/syscall/type_formatting.rs
+++ b/src/main/host/syscall/type_formatting.rs
@@ -78,6 +78,36 @@ macro_rules! simple_debug_impl {
     };
 }
 
+/// Implement `SyscallDisplay` for a bitflags type. The type must implement `TryFromSyscallReg`.
+macro_rules! bitflags_impl {
+    ($type:ty, $($types:ty),+) => {
+        bitflags_impl!($type);
+        bitflags_impl!($($types),+);
+    };
+    ($type:ty) => {
+        impl SyscallDisplay for SyscallVal<'_, $type> {
+            fn fmt(
+                &self,
+                f: &mut std::fmt::Formatter<'_>,
+                _options: FmtOptions,
+                _mem: &MemoryManager,
+            ) -> std::fmt::Result {
+                match <$type>::try_from_reg(self.reg) {
+                    Some(x) => {
+                        if x.is_empty() {
+                            write!(f, "(empty)")
+                        } else {
+                            bitflags::parser::to_writer(&x, f)
+                        }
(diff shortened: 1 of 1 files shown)


src/main/host/syscall/type_formatting.rs
```diff
@@ -78,6 +78,36 @@ macro_rules! simple_debug_impl {
...
+    };
+    ($type:ty) => {
+        impl SyscallDisplay for SyscallVal<'_, $type> {
+            fn fmt(
+                &self,
+                f: &mut std::fmt::Formatter<'_>,
+                _options: FmtOptions,
+                _mem: &MemoryManager,
+            ) -> std::fmt::Result {
+                match <$type>::try_from_reg(self.reg) {
```
sporksmith, line 95: Optional - A bit tangential, but maybe the `TryFrom<SysCallReg>` implementations of these ought to use `Self::from_bits_retain` instead of `Self::from_bits`. I imagine the formatted result is a little nicer than just the raw integer. Hopefully it's the formatted known flags + the "leftover" int.

sporksmith (approved):

src/main/host/syscall/type_formatting.rs
sporksmith: P.S. When making the `linux-api` types, I debated omitting the "type prefixes" of individual flags, but left them so that it's easier to cross-reference them with C code, man pages, etc. I suppose if we didn't leave them then the new `bitflags` formatting behavior would be a bit less redundant, but oh well.

stevenengler: > Optional - A bit tangential, but maybe the `TryFrom<SysCallReg>` implementations of these ought to use `Self::from_bits_retain` instead of `Self::from_bits`. I imagine the formatted result is a little nicer than just the raw integer. Hopefully it's the formatted known flags + the "leftover" int.

I was planning on doing this in another PR, and probably get rid of the `TryFromSyscallReg` trait.

> P.S. When making the `linux-api` types, I debated omitting the "type prefixes" of individual flags, but left them so that it's easier to cross-reference them with C code, man pages, etc. I suppose if we didn't leave them then the new `bitflags` formatting behavior would be a bit less redundant, but oh well.

I do personally like the current linux-api types, with the names matching the C names.

PR #3055, opened 2023-06-23
Reimplement shim's TLS (thread local storage) in Rust

sporksmith:
Reimplements shim's thread local storage in Rust, while removing calls into libc.

Adds unregistration and memory-reuse to the "fallback path" of thread local storage. While the previous implementation had a hard limit of number of threads that could be created during a managed process's lifetime, this is now a limit to the number of *simultaneous* threads a managed process can have.

Adds a new fast fallback mode for thread local storage that relies on libc thread local storage having been set up, but *doesn't* require linking libc. This has unit tests, and I've locally run shadow tests with this enabled, but is currently disabled at compile-time in favor of the old behavior of using native thread local storage when available. I'd like to leave the question of when and how to enable this for a future PR.

Adds the `noinline` attribute to `_shim_clone`. Other changes just happened to trigger an existing bug here.

Moves all thread-locals in the shim to Rust so that we can delete the old C thread local storage implementation, and not have to expose the new Rust implementation to C.

`linux-api`: adds `sigaltstack`

`SelfContainedMutex`: adds a const initializer

`FormatBuffer`: support creation of `&CStr`

Adds a `no_std` implementation of `LazyLock`.


sporksmith:
Started with a somewhat naive, slow implementation. Running a benchmark to see if it makes much difference overall. https://github.com/shadow/benchmark/actions/runs/5360614491

sporksmith:
Sorry, this is another big one, and the history is a bit of a mess. It's mostly new code though, so I think it should be fairly straightforward to review by crate or module @ HEAD of the PR.

In particular I iterated on the design a fair bit and would rather not squash that down just yet. If it's helpful I could do it and keep the original history in a backup branch, but I think I'd end up roughly end up just making a commit for each crate touched here, so I don't think the review experience is much different.

In particular I started with a simpler design for thread local storage, but found it to be very significantly slower. I went back to roughly the design in C and iterated a bit on further improving performance and safety.

sporksmith:
> Started with a somewhat naive, slow implementation. Running a benchmark to see if it makes much difference overall. https://github.com/shadow/benchmark/actions/runs/5360614491

I ended up killing that run because I started to run some local benchmarks while it was running and found that initial implementation was *much* slower.

A lot of the iteration on this PR was on performance based on locally checking the tor-minimal benchmark locally, and the `go-preempt` benchmark which seems to especially exercise this path.

Here's a more recent run of the full tor benchmark on this PR; there's ~no difference from the previous nightly https://github.com/shadow/benchmark-results/blob/master/tor/2023-06-25-T02-31-26/plots/run_time.png

src/lib/vasi-sync/src/lazy_lock.rs
```diff
@@ -0,0 +1,176 @@
...
+        Ref {
+            val: self.value.get(),
+            _phantom: PhantomData,
+        }
+    }
+
+    #[inline]
+    pub fn initd(&self) -> bool {
+        InitState::try_from(self.init_state.load(atomic::Ordering::Acquire)).unwrap()
+            == InitState::Initd
```
stevenengler, line 148: I think this can just be "relaxed" since this check isn't being used to protect anything (any accesses to the internals need to go through `force` which uses the proper ordering).

src/lib/vasi-sync/src/lazy_lock.rs
```diff
@@ -0,0 +1,176 @@
...
+        {
+            unsafe { self.value.get_mut().deref().assume_init_drop() }
+        }
+    }
+}
+
+pub struct Ref<'a, T, Init> {
+    val: ConstPtr<MaybeUninit<T>>,
+    _phantom: PhantomData<&'a LazyLock<T, Init>>,
```
stevenengler, line 164: I think this should be just `PhantomData<&'a T>`? I think `PhantomData<&'a LazyLock<T, Init>>` would be overly restrictive. For example it should be safe to send a `Ref<'a, T>` to other threads (if `T` is `Sync`), even if `Init` is not `Sync`.

src/lib/vasi-sync/src/lazy_lock.rs
```diff
@@ -0,0 +1,176 @@
...
+    fn drop(&mut self) {
+        if InitState::try_from(self.init_state.load(atomic::Ordering::Acquire)).unwrap()
+            == InitState::Initd
+        {
+            unsafe { self.value.get_mut().deref().assume_init_drop() }
+        }
+    }
+}
+
+pub struct Ref<'a, T, Init> {
```
stevenengler, line 162: Derive `Clone` (or `Copy` if `ConstPtr` allows it)?

src/lib/vasi-sync/src/lazy_lock.rs
```diff
@@ -0,0 +1,176 @@
...
+    val: ConstPtr<MaybeUninit<T>>,
```
stevenengler, line 163: Why is this a `ConstPtr<MaybeUninit<T>>`? Since `T` is initialized, can't it just be `ConstPtr<T>`?

src/lib/vasi-sync/src/sync.rs
```diff
@@ -182,33 +188,68 @@ impl<T: ?Sized> MutPtr<T> {
...
 unsafe impl<T: ?Sized> Send for MutPtr<T> where T: Send {}
 
+#[cfg(not(loom))]
+pub struct ConstPtr<T: ?Sized>(*const T);
+#[cfg(not(loom))]
+impl<T: ?Sized> ConstPtr<T> {
+    /// # Safety
+    ///
+    /// See `loom::cell::ConstPtr::deref`.
+    #[allow(clippy::mut_from_ref)]
```
stevenengler, line 198: Why is this clippy lint needed? I don't see any `mut`?

src/lib/vasi-sync/src/lazy_lock.rs
```diff
@@ -0,0 +1,176 @@
...
+#[repr(i8)]
+enum InitState {
+    Uninitd,
+    Initializing,
+    Initd,
+}
+
+/// Analogous to `std::sync::LazyLock`, but works on stable Rust, is no_std,
+/// only makes direct syscalls, etc.
+pub struct LazyLock<T, Init = fn() -> T> {
```
stevenengler, line 37: This needs to be `#[repr(C)]` if it is supposed to be `VirtualAddressSpaceIndependent`. I think the vasi-sync types also need to be made `#[repr(C)]`, such as `UnsafeCell`.

src/lib/vasi-sync/src/lazy_lock.rs
```diff
@@ -0,0 +1,176 @@
...
+    value: UnsafeCell<MaybeUninit<T>>,
+    init_state: AtomicI8,
+    initializer: UnsafeCell<Option<Init>>,
+}
+unsafe impl<T, Init> VirtualAddressSpaceIndependent for LazyLock<T, Init>
+where
+    T: VirtualAddressSpaceIndependent,
+    Init: VirtualAddressSpaceIndependent,
+{
+}
```
stevenengler, line 47: Why implement it directly instead of using the derive? For example `Option` is not `VirtualAddressSpaceIndependent` and doesn't have a stable layout.

stevenengler (changes requested):
I haven't looked through the shim code yet, so just leaving comments for what I have so far.

src/lib/vasi-sync/src/lazy_lock.rs
```diff
@@ -0,0 +1,176 @@
...
+    fn init(&self, mut state: InitState) {
+        loop {
+            match state {
+                InitState::Uninitd => {
+                    match self.init_state.compare_exchange(
+                        InitState::Uninitd.into(),
+                        InitState::Initializing.into(),
+                        atomic::Ordering::Acquire,
+                        atomic::Ordering::Relaxed,
+                    ) {
```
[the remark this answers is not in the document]
sporksmith: Nice catch; I'd forgotten to run loom after the latest changes.

Unfortunately the test still doesn't pass under loom. I think your suggestions are correct, but for completeness tried with and without - it fails either way. I'll try to debug some more tomorrow.

src/lib/vasi-sync/src/lazy_lock.rs
```diff
@@ -0,0 +1,176 @@
...
+    ///
+    /// std's LazyLock returns an actual reference here and implements Deref.
+    /// We can't because of our usage of loom's UnsafeCell, which doesn't support
+    /// getting a raw reference that's not borrowed from an intermediate temp object.
+    #[inline]
+    pub fn force(&self) -> Ref<T, Init> {
+        let state = InitState::try_from(self.init_state.load(atomic::Ordering::Acquire)).unwrap();
+        if state != InitState::Initd {
+            self.init(state);
+        }
```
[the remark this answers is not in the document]
sporksmith: My thinking here was to not mark the `init` function `#[inline]` since it's relatively complex, and then try to avoid calling it on the fast path. After further thought I think it probably makes sense to just mark it `#[inline]` too, call it here unconditionally, and let the compiler decide how much inlining to actually do.

src/lib/vasi-sync/src/lazy_lock.rs
```diff
@@ -0,0 +1,176 @@
...
+        Ref {
+            val: self.value.get(),
+            _phantom: PhantomData,
+        }
+    }
+
+    #[inline]
+    pub fn initd(&self) -> bool {
```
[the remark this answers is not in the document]
sporksmith: done

src/lib/vasi-sync/src/lazy_lock.rs
sporksmith: done

src/lib/vasi-sync/src/lazy_lock.rs
sporksmith: Can't; `ConstPtr` implements neither. https://docs.rs/loom/latest/loom/cell/struct.ConstPtr.html

src/lib/vasi-sync/src/lazy_lock.rs
sporksmith: `UnsafeCell` only provides access to a `ConstPtr<MaybeUninit<T>>`. There is no, e.g. `ConstPtr::map`.

src/lib/vasi-sync/src/lazy_lock.rs
sporksmith: Good point; done

src/lib/vasi-sync/src/sync.rs
sporksmith: Oops, yeah copy-pasta

src/lib/vasi-sync/src/lazy_lock.rs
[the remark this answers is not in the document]
sporksmith: Ah, I was accidentally using `core::atomic` instead of `crate::sync::atomic`, so the loom tests weren't recognizing that sync was happening at all. should have a fix in a bit...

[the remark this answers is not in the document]
sporksmith: fixed. I ended up adding blocking via futex as well to get the loom tests to pass.

sporksmith:
Noticed that the tls unit tests were nondeterministically failing last night. The problem turned out to be in the tests themselves I needed some extra synchronization. Fixed that and made a pass over fixing and documenting some of the atomic operation orderings there.

sporksmith:
Kicked off a tgen benchmark and new tor benchmark since there's been a bit of iteration:

https://github.com/shadow/benchmark/actions/runs/5400574841
https://github.com/shadow/benchmark/actions/runs/5400579973

sporksmith:
Sorry for yet more churn here - I had the start of a microbenchmark for the lazylock on a local branch, and after rebasing it on the latest changes noticed a significant performance hit. Most of it seemed to be from the inlining change. Added the microbenchmarks here and reverted the inlining change.

src/lib/vasi-sync/src/lazy_lock.rs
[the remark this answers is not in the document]
stevenengler: Now that loom is working, if you switch the compare_exchange back to the original version does loom say anything? I'm curious to know whether it would catch that.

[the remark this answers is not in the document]
sporksmith: Yup, it catches it :)

src/lib/linux-api/src/signal.rs
```diff
@@ -1393,20 +1394,23 @@ pub type stack_t = linux_stack_t;
...
+pub const LINUX_SS_AUTODISARM: u32 = 1 << 31;
 bitflags::bitflags! {
+    // While the Linux header defines the only current value of this flag as
+    // u32, the field in `stack_t::ss_flags` where this type is used is i32.
     #[repr(transparent)]
     #[derive(Copy, Clone, Debug, Default, Eq, PartialEq)]
     pub struct SigAltStackFlags: i32 {
-        const SS_AUTODISARM = LINUX_SS_AUTODISARM;
+        // The raw u32 value wraps around to a negative i32.
+        const SS_AUTODISARM = i32_from_u32_allowing_wraparound(LINUX_SS_AUTODISARM);
```
stevenengler: Should `LINUX_SS_AUTODISARM` be non-pub if we're exposing it through `SigAltStackFlags`?

sporksmith: It's currently exposed for our re-export to C

src/lib/vasi-sync/src/lazy_lock.rs
```diff
@@ -112,9 +136,26 @@ where
...
                 InitState::Initializing => {
-                    // Another thread is initializing. Spin.
-                    // TODO: block?
-                    rustix::process::sched_yield();
+                    // `Acquire` ordering on both paths to potentially pair with initialization `Release`.
+                    state = match self.init_state.compare_exchange(
+                        InitState::Initializing.into(),
+                        InitState::InitializingWithSleepers.into(),
+                        atomic::Ordering::Acquire,
+                        atomic::Ordering::Acquire,
```
stevenengler, line 144: I don't think we need "acquire" in the success case, since there was no "release" to pair with if the state is still "initializing".

src/lib/shim/src/tls.rs
```diff
@@ -0,0 +1,610 @@
...
+                this_var_bytes as *const ShimTlsVarStorage<T>;
+            assert_eq!(this_var.align_offset(Self::align()), 0);
+            // SAFETY: The TLS bytes for each thread are initialized to 0, and
+            // all-zeroes is a valid value of `ShimTlsVarStorage<T>`.
+            //
+            // We've ensure proper alignment when calculating the offset,
+            // and verified in the assertion just above.
+            let this_var: &ShimTlsVarStorage<T> = unsafe { &*this_var };
+            this_var
+        };
```
stevenengler, line 501: I think most of this code would be nicer as a method of `ShimTlsVarStorage` instead. It's essentially casting an `&UnsafeCell<ShimThreadLocalStorage>` with an offset to a `&ShimTlsVarStorage<T>`, and I think this would be better as a self-contained function that explains exactly why this is safe.

src/lib/shim/src/tls.rs
```diff
@@ -0,0 +1,610 @@
...
+
+        // SAFETY: This thread has exclusive access to the underlying storage,
+        // and we don't let references to `initd` escape this method,
+        // so we have exclusive access.
+        let initd: &mut i8 = unsafe { &mut *this_var.initd.get() };
+        if *initd == 0 {
```
stevenengler, line 507: If `initd` used a `Cell` instead, I don't think `unsafe` would be needed here?

src/lib/shim/src/tls.rs
```diff
@@ -0,0 +1,610 @@
...
+            log::debug!("Allocating fallback thread-local storage");
+            rustix::mm::mmap_anonymous(
+                core::ptr::null_mut(),
+                core::mem::size_of::<StoragesType>(),
+                ProtFlags::READ | ProtFlags::WRITE,
+                MapFlags::PRIVATE,
+            )
+            .unwrap()
+        };
+        unsafe { &*(ptr as *const StoragesType) }
```
stevenengler, line 269: What is the alignment of the pointer returned from `mmap_anonymous`, and does it have a valid alignment for `StoragesType`? (If so, add some documentation?)

src/lib/shim/src/tls.rs
```diff
@@ -0,0 +1,610 @@
...
+    }
+
+    let current = FastThreadId::current();
+    if let Some(position) = global_thread_ids::get(current) {
+        // Zero for the next thread.
+        // TODO: Consider running drop impls?
+        global_storages::get(position)
+            .get()
+            .write(core::mem::zeroed());
+        global_thread_ids::remove(position);
```
stevenengler, line 322: It's not possible to safely run a drop impl. Since the thread-locals return `'static` references, you could put one thread-local's `'static` reference in another thread-local. Then if the first thread local is dropped first, it will leave a dangling reference in the other thread-local, which is UB. So I think it should be documented here to never run drop impls on thread-locals.

src/lib/shim/src/tls.rs
```diff
@@ -0,0 +1,610 @@
...
+// storage of exited threads, so this is the max *concurrent* threads per
+// process.
+const TLS_FALLBACK_MAX_THREADS: usize = 1000;
+
+// This is `pub` for export to C, where we can allocate
+// native thread local storage.
+#[repr(C, align(16))]
+pub struct ShimThreadLocalStorage {
+    _bytes: [u8; BYTES_PER_THREAD],
+}
```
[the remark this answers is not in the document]
sporksmith: Done

src/lib/shim/src/tls.rs
```diff
@@ -0,0 +1,610 @@
...
+impl lazy_lock::Producer<usize> for OffsetInitializer {
+    // Finds and assigns the next free and suitably aligned offset within
+    // thread-local-storage for a value of type `T`, initialized with function
+    // `F`.
+    fn initialize(self) -> usize {
+        static NEXT_OFFSET: AtomicUsize = AtomicUsize::new(0);
+        let mut next_offset_val = NEXT_OFFSET.load(atomic::Ordering::Relaxed);
+        loop {
+            let overhang = next_offset_val % self.align;
+            let this_var_offset = next_offset_val + self.align - overhang;
```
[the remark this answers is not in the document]
sporksmith: done

src/lib/shim/src/tls.rs
```diff
@@ -0,0 +1,610 @@
...
+
+impl<T, F> ShimTlsVar<T, F>
+where
+    F: Fn() -> T,
+{
+    /// Create a variable that will be uniquely instantiated for each thread.
+    /// Typically this should go in a `static`.
+    pub const fn new(f: F) -> Self {
+        Self {
+            offset: LazyLock::const_new(OffsetInitializer::new::<T>()),
```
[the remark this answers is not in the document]
sporksmith: Good catch; added more tests to ensure this is caught. It wasn't getting caught before because of some combination of none of the tests allocating more than one thread local variable, that variable type having a large alignment, and the extra space getting wasted in the alignment calculation.

sporksmith (commented):
Addressed some of these; will continue tomorrow.

src/lib/shim/src/tls.rs
```diff
@@ -0,0 +1,806 @@
...
+    }
+
+    #[test_log::test]
+    fn test_minimal() {
+        static MY_VAR: ShimTlsVar<u32> = ShimTlsVar::new(|| 0);
+        test_each_mode(|| {
+            assert_eq!(*MY_VAR.get(), 0);
+            unsafe { unregister_current_thread() };
+        });
+    }
```
stevenengler, line 704: All of these tests will be running in parallel, so I don't see how the `test_each_mode` works properly.

src/lib/shim/src/tls.rs
```diff
@@ -0,0 +1,610 @@
...
+            FastThreadId::None => 0,
+        }
+    }
+}
+
+/// Tracks the storage indexes for registered threads.
+struct FastThreadIds([AtomicUsize; TLS_FALLBACK_MAX_THREADS]);
+impl FastThreadIds {
+    /// Find the storage index of `id`, if it's already registered.
+    pub fn position(&self, id: FastThreadId) -> Option<usize> {
```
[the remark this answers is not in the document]
sporksmith: done

src/lib/shim/src/tls.rs
```diff
@@ -0,0 +1,610 @@
...
+        let start_pos = FastThreadId::hash_idx(&id);
+        let mut idxs = (start_pos..TLS_FALLBACK_MAX_THREADS).chain(0..start_pos);
+        let id_as_usize = usize::from(id);
+        idxs.find(|idx| self.0[*idx].load(core::sync::atomic::Ordering::Relaxed) == id_as_usize)
+    }
+
+    /// Find a free storage index and assign it to `id`.
+    pub fn insert(&self, id: FastThreadId) -> usize {
```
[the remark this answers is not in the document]
sporksmith: done

src/lib/shim/src/tls.rs
stevenengler: Oops I missed the mutex here somehow.

src/lib/shim/src/tls.rs
```diff
@@ -0,0 +1,610 @@
...
+        let mut idxs = (start_pos..TLS_FALLBACK_MAX_THREADS).chain(0..start_pos);
+        let none_as_usize = usize::from(FastThreadId::None);
+        let id_as_usize = usize::from(id);
+        idxs.find(|idx| {
+            let res = self.0[*idx].compare_exchange(
+                none_as_usize,
+                id_as_usize,
+                // Acquire the storage for this index, released in
+                // `global_thread_ids::remove`.
+                core::sync::atomic::Ordering::Acquire,
```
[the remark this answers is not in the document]
sporksmith: Done. Moved this type into the `global_thread_ids` module and documented it there.

src/lib/shim/src/tls.rs
```diff
@@ -0,0 +1,610 @@
...
+        })
+        .unwrap()
+    }
+}
+
+pub mod global_thread_ids {
+    use super::*;
+
+    static THREAD_IDS: LazyLock<FastThreadIds> =
+        LazyLock::const_new(|| FastThreadIds(core::array::from_fn(|_| AtomicUsize::new(0))));
```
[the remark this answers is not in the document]
sporksmith: Ah, that's a neat trick. Done.

sporksmith (commented):
Sync; still working through comments

src/lib/shim/src/lib.rs
```diff
@@ -24,9 +34,191 @@ pub fn manager_shmem() -> Option<&'static ManagerShmem> {
...
+    // signal handlers, to let the OS reclaim the (now-popped) signal handler stack
+    // frames.
+    const SHIM_SIGNAL_STACK_MIN_USABLE_SIZE: usize = 1024 * 100;
+    const SHIM_SIGNAL_STACK_SIZE: usize =
+        SHIM_SIGNAL_STACK_GUARD_OVERHEAD + SHIM_SIGNAL_STACK_MIN_USABLE_SIZE;
+
+    /// # Safety
+    ///
+    /// See `sigaltstack(2)`
+    pub unsafe fn init() {
```
stevenengler, line 76: Is there some more-specific safety requirements? For example this function doesn't take any arguments, so it's not clear what from sigaltstack needs to be upheld.

src/lib/shim/src/lib.rs
```diff
@@ -24,9 +34,191 @@ pub fn manager_shmem() -> Option<&'static ManagerShmem> {
...
+            linux_api::signal::sigaltstack(Some(&stack_descriptor), None).unwrap();
+        }
+    }
+
+    /// # Safety
+    ///
+    /// The current thread must exit before:
+    /// * Another thread runs.
+    /// * That thread also frees its stack.
+    /// * This thread runs again on the signal stack (e.g. by handling a new signal).
```
stevenengler, line 120: These safety requirements aren't clear to me. Maybe it should say something like "After calling `free()`, the current thread must exit before"? Then what does it mean by "the current thread must exit before this thread runs again on the signal stack" (how can the thread run if it's already exited)?

src/lib/shim/src/lib.rs
```diff
@@ -24,9 +34,191 @@ pub fn manager_shmem() -> Option<&'static ManagerShmem> {
...
+            }
+            .unwrap();
+        }
+    }
+}
+
+mod tls_ipc {
+    use super::*;
+    static IPC_DATA_BLOCK: ShimTlsVar<RefCell<Option<ShMemBlockAlias<IPCData>>>> =
+        ShimTlsVar::new(|| RefCell::new(None));
```
stevenengler, line 154: Comment that this is used to communicate / pass messages with the main shadow process?

src/lib/shim/src/lib.rs
```diff
@@ -24,9 +34,191 @@ pub fn manager_shmem() -> Option<&'static ManagerShmem> {
...
+    }
+
+    /// # Safety
+    ///
+    /// Environment variable SHADOW_IPC_BLK must contained a serialized block of
+    /// type `IPCData`, which outlives the current thread.
+    pub unsafe fn set_from_env() {
+        use core::fmt::Write;
+        let mut envname = FormatBuffer::<20>::new();
+        write!(&mut envname, "SHADOW_IPC_BLK").unwrap();
```
stevenengler, line 179: (optional) maybe simpler as just:

```rust
let envname = CStr::from_bytes_with_nul(b"SHADOW_IPC_BLK\0").unwrap();
let ipc_blk = unsafe { libc::getenv(envname.as_ptr()) };
```

src/lib/shim/src/lib.rs
```diff
@@ -24,9 +34,191 @@ pub fn manager_shmem() -> Option<&'static ManagerShmem> {
...
+    pub unsafe fn set(blk: &ShMemBlockSerialized) {
+        let blk: ShMemBlockAlias<ThreadShmem> = unsafe { Serializer::global().deserialize(blk) };
+        assert!(THREAD_SHMEM.get().replace(Some(blk)).is_none());
+    }
+}
+
 mod shimlogger;
 pub use shimlogger::export as shimlogger_export;
 
+mod tls;
```
stevenengler, line 220: optional nit: I think it's nicer when the `mod` statements that link to other files are at the top of the file so they're all in one place, and it's easy to add more.

src/lib/shim/src/lib.rs
```diff
@@ -36,6 +228,41 @@ extern crate shadow_shim_helper_rs;
...
+        //
+        // TODO: This shouldn't be necessary once we've gotten rid of all
+        // calls to libc from the shim's initialization.
+        return;
+    }
+    STARTED_THREAD_INIT.get().force();
+
+    static THREAD_INIT: ShimTlsVar<LazyLock<()>> = ShimTlsVar::new(|| {
+        LazyLock::const_new(|| {
+            static GLOBAL_INIT: LazyLock<()> = LazyLock::const_new(|| ());
```
stevenengler, line 252: Maybe add comments here as well to make the code easier to follow? For example that `THREAD_INIT` only allows the closure to run once per thread, and `GLOBAL_INIT` only allows the code in the if statement to run once per process?

stevenengler (changes requested):
Okay I'm done :)

src/lib/formatting-nostd/src/format_buffer.rs
```diff
@@ -23,21 +26,28 @@ use core::mem::MaybeUninit;
 /// ```
 pub struct FormatBuffer<const N: usize> {
     buffer: [MaybeUninit<u8>; N],
+    // Does *not* include NULL byte.
     used: usize,
```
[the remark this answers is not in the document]
stevenengler: Not really (maybe the IDE tooltip will include documentation about things?), but sometimes I build shadow docs with `--document-private-items` to get an overview of some code.

sporksmith:
I'm a little confused how the loom CI test is running, since it's not in this branch. One way or another I'll have to reduce `LOOM_MAX_PREEMPTIONS` from 3 to 2 for one of the lazylock tests to get the loom test to pass, but I'll wait for approval and rebase before trying to mess with it.

src/lib/shim/src/tls.rs
```diff
@@ -0,0 +1,610 @@
...
+    // initialized. `bool` is documented as *casting* to 0 or 1, but it's
+    // possible (if unlikely) that the underlying representation in memory isn't
+    // actually 0 and 1.
+    //
+    // Likewise we can't use an `Option<T>` for `value`, since we can't guarantee
+    // that the "0 value" will be a valid representation of `None`.
+    //
+    // We need `UnsafeCell` for both so that we can initialize, despite this being
+    // stored in a larger array over which we don't have exclusive access.
+    initd: UnsafeCell<i8>,
```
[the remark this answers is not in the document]
stevenengler: I'm still not clear why `Cell` can't be used here. `Cell<T>` has the same representation as `T`, so wouldn't `Cell<T>` also be allowed to be initialized as zeroes?

> `Cell<T>` has the same memory layout and caveats as `UnsafeCell<T>`. In particular, this means that `Cell<T>` has the same in-memory representation as its inner type T.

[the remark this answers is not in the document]
sporksmith: Oh cool, I didn't realize that. Making the change...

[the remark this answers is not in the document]
sporksmith: done

sporksmith:
hmm, looks like a performance regression caused `go_preempt-shadow` to timeout in a debug build. This one pretty heavily stresses this code path and timed out before I optimized. Investigating...

sporksmith:
> hmm, looks like a performance regression caused `go_preempt-shadow` to timeout in a debug build. This one pretty heavily stresses this code path and timed out before I optimized. Investigating...

Found a bug I'd introduced in 30fdb9136cf58b491126271eecb8a754470cf6c6 that resulted in shadow effectively using `NativeTlsId` instead of `Native`. While `NativeTlsId` is *close* to `Native` performance in release builds, it is a bit of a more significant penalty in debug builds, and this particular test stresses this path a lot.

Now that it's fixed this test is still a bit slower than `main` on my machine in debug builds (~8s vs ~6s), but pretty close in release builds (i think 1.1s vs 1.0s). I'll kick off another Tor test to make sure performance there still looks good.

stevenengler:
> I'm a little confused how the loom CI test is running, since it's not in this branch.

Yeah this has always been confusing to me. I think the CI is merging your branch with the main branch, and running the workflow from that. In the workflow we use `ref: ${{ github.event.pull_request.head.sha }}` to avoid testing the merged code, but this code checkout in the workflow occurs after github has already started running the merged workflow file.

sporksmith:
> I'll kick off another Tor test to make sure performance there still looks good.

Performance looks ok to me.

I ran 3 trials; the aggregate result graph looks slightly slower: https://github.com/shadow/benchmark-results/blob/master/tor/2023-06-29-T23-59-06/plots/run_time.png

But the individual trial runtimes were [7989](https://github.com/shadow/benchmark-results/blob/master/tor/2023-06-29-T23-59-06/tornet-0.15-1/tornet.plot.data/resource_usage.json), [7397](https://github.com/shadow/benchmark-results/blob/master/tor/2023-06-29-T23-59-06/tornet-0.15-2/tornet.plot.data/resource_usage.json), [7386](https://github.com/shadow/benchmark-results/blob/master/tor/2023-06-29-T23-59-06/tornet-0.15-3/tornet.plot.data/resource_usage.json), compared to the last nightly @ [7414](https://github.com/shadow/benchmark-results/blob/master/tor/2023-06-29-T04-10-57/tornet-0.15-1/tornet.plot.data/resource_usage.json). i.e. i think we just had one trial where one of the nondeterministic stalls happened.

src/lib/shim/src/tls.rs
```diff
@@ -128,179 +230,213 @@ impl From<FastThreadId> for usize {
...
-// SAFETY: This thread must not access thread local storage again before exiting.
+/// Should be called by every thread that accesses thread local storage, before
+/// it makes the `exit` syscall and after it is done accessing thread local storage.
+///
+/// This is a no-op when using native thread-local storage, but lets us reclaim memory
+/// otherwise, and is necessary for correctness when [`Mode::NativeTlsId`] is enabled.
+///
+/// # Safety
+///
+/// This thread must not access thread local storage again before exiting.
```
stevenengler, line 439: Maybe add ", and must not have any existing references to thread local storage" (I think just "must not access thread local storage" could make it easy to forget about things like drop impls or other methods of indirectly accessing thread locals)?

src/lib/shim/src/tls.rs
```diff
@@ -309,105 +445,123 @@ pub unsafe fn unregister_current_thread() {
...
-impl<T, F> OffsetInitializer<T, F> {
-    pub const fn new() -> Self {
+impl OffsetInitializer {
+    pub const fn new<T>() -> Self {
         Self {
-            _phantom_t: PhantomData,
-            _phantom_f: PhantomData,
+            // FIXME: should be of storage.
+            align: core::mem::align_of::<ShimTlsVarStorage<T>>(),
+            size: core::mem::size_of::<ShimTlsVarStorage<T>>(),
```
stevenengler, line 530: Should this FIXME be removed?

src/lib/shim/src/tls.rs
```diff
@@ -309,105 +445,123 @@ pub unsafe fn unregister_current_thread() {
...
-{
+impl lazy_lock::Producer<usize> for OffsetInitializer {
     // Finds and assigns the next free and suitably aligned offset within
     // thread-local-storage for a value of type `T`, initialized with function
     // `F`.
     fn initialize(self) -> usize {
+        // The alignment we ensure here is an offset from the base of a [`ShimThreadLocalStorage`].
+        // It won't be meaningful if [`ShimThreadLocalStorage`] has a smaller alignment requirement
+        // than this variable.
+        assert!(self.align <= core::mem::align_of::<ShimThreadLocalStorage>());
```
stevenengler, line 543: I still don't understand this part. What happens if the alignment were larger than `ShimThreadLocalStorage`?

stevenengler (approved):

src/lib/shim/src/tls.rs
```diff
@@ -441,33 +606,27 @@ impl<T, F> ShimTlsVar<T, F>
...
 
     /// Access the inner value.
     ///
-    /// The returned wrapper can't be sent to or shared with other threads, to
-    /// allow the implementation to be extended in the future to deallocate
-    /// and/or reuse storage when a thread exits.
+    /// The returned wrapper can't be sent to or shared with other threads,
+    /// since the underlying storage is invalidated when the originating thread
+    /// calls [`unregister_current_thread`].
     pub fn get(&self) -> TLSVarRef<T> {
```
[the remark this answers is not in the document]
sporksmith: I was thinking of moving the `exit` syscall *into* `unregister_current_thread`, and calling it something like `unregister_current_thread_and_exit` or just `exit_thread`.

The sequence would go:

* `sigprocmask` to block all* signals (see below), to ensure no signal handlers can run on the current thread
* unregister the thread
* `exit` the thread

I'll have to check what  happens with seccomp when `SIGSYS` is blocked; my hope is that it would kill the process, but I don't know for sure. We also can't block `SIGKILL` or `SIGSTOP` but neither of those are allowed to have signal handlers, either.

If we do that, I think that removes most of the concern here in `get`? I think the unregistration function could also no longer be `unsafe`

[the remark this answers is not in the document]
stevenengler: Yeah I think this would be enough to make it safe. If we make a direct `exit` syscall and there aren't any signal handlers, I can't think of any other way that shim code could run between the unregistering/zeroing and the exit syscall.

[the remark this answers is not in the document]
stevenengler: I guess it depends on if it would be considered UB for a dangling rust reference to exist even briefly (which I think is technically UB), but I don't think there's a solution to that other than not allowing `'static` references to thread-locals. But I think having the `unregister_current_thread` also exit would be better than having them separate.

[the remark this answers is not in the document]
stevenengler: > I don't think there's a solution to that other than not allowing 'static references to thread-locals.

Actually I don't think this is right either, since you could still call `unregister_current_thread` while holding a non-static reference to a thread-local. Maybe a solution could be a global atomic counter, and when a `TLSVarRef` is created it increases and when it's dropped it decreases, and then the `unregister_current_thread` makes sure there are no references. But I don't know what the performance implication is, and I don't think that's needed for this PR.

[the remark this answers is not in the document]
sporksmith: I *think* the counter could just be a `Cell` in `ShimThreadLocalStorage`, but yeah I'll probably leave that for a separate PR.

src/lib/shim/src/tls.rs
```diff
@@ -0,0 +1,610 @@
...
+/// An opaque, per-thread identifier. These are only guaranteed to be unique for
+/// *live* threads; in particular `FastThreadId::Fs` of a live thread can have
+/// the same value as a previously seen dead thread. See
+/// [`unregister_current_thread`].
+#[derive(Debug, Eq, PartialEq, Copy, Clone, Hash)]
+pub(crate) enum FastThreadId {
+    None,
+    Fs(usize),
+    NativeTid(Pid),
+}
```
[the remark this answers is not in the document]
sporksmith: We're not permitted to implement `From<Option<FastThreadId>> for usize`, but on further consideration I think it is better to move this outside and just turn that conversion into a method instead of using the `From` trait.

Issue #3041, opened 2023-06-16
Support reading/writing socket control/auxiliary data


stevenengler:
Shadow supports `sendmsg()` and `recvmsg()`, but doesn't implement reading or writing any control/auxiliary data. This auxiliary data is accessible in the socket file's `sendmsg()` and `recvmsg()` methods (for example `UdpSocket::recvmsg(...)`) through the `args: RecvmsgArgs` argument (for example `args.control_ptr`).

This auxiliary data can be written and read to this control buffer using the `CMSG` libc macros documented in [cmsg(3)](https://man7.org/linux/man-pages/man3/cmsg.3.html). These would typically be used by an application accessing the kernel through libc, but we can probably reuse these in shadow.

This issue is for supporting any code required to read and write this auxiliary data. Once we have code that can do this, we can use it to support socket options that depend on it, such as `SO_TIMESTAMP` in #3040. Some other options that use this auxiliary data are `IP_RECVORIGDSTADDR` (return the original destination of a received packet), `IP_PKTINFO` (return information about the incoming udp packet), `MSG_ERRQUEUE` (return error information during recv), `IP_RECVTOS` (return packet TOS field), and `SCM_RIGHTS` (send a fd over a unix socket).

PR #3059, opened 2023-06-28
Fake support of TOS

FlyearthR:
Some QUIC implementations won't run if the system does not support Explicit Congestion Notification.
Type Of Service is used by Linux to implement ECN. As long as ECN is not implemented in Shadow, no process will receive a congestion notification. So, there should not be any harm with a fake ECN support.
Process using specific TOS features could be impacted by the scheduling of messages not taking into account TOS, but, as a warning is issued, there is no difference with no support of TOS at all.

### DIFF
--- a/src/main/host/descriptor/socket/inet/legacy_tcp.rs
+++ b/src/main/host/descriptor/socket/inet/legacy_tcp.rs
@@ -1226,6 +1226,14 @@ impl LegacyTcpSocket {
                 // TODO: implement this, pkg.go.dev/net uses it
                 log::trace!("setsockopt SO_BROADCAST not yet implemented");
             }
+            (libc::IPPROTO_IP, libc::IP_TOS) => {
+                // this is used by linux to encode ecn
+                log::trace!("setsockopt IP_TOS not yet implemented");
+            }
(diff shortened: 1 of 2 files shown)


stevenengler:
Thanks for the PR. We discussed this and decided not to merge this at this time. We do have some exceptions for socket options that are commonly used by applications but that shadow doesn't support such as `SO_REUSEADDR`, but these don't have any impact on network performance. Our main concern is that if the application used the `IP_TOS` option, this would normally affect the network performance as Linux queuing disciplines change their behaviour due to this option. We wouldn't want for example someone to run two simulations, one where the application changes the TOS settings and one where it doesn't, and make incorrect conclusions from the network performance. Even with a warning, it can be easy to miss them since Shadow logs a lot of messages.

Until we can better understand the effect of `IP_TOS` and preferrably emulate that in shadow, and support [auxiliary data](#3041) for `IP_RECVTOS`, it would probably be best to keep this as a local patch for now. But we would like to support things like ECN, TOS, and QUIC.

Could you open an issue about this? Preferably with the exact options you need, what applications/libraries use them, and any other details you have?

PR #3063, opened 2023-06-30
Create a specialized hash table for the shim's TLS

sporksmith:
This should be able to replace the ad-hoc hash table implemented in `shadow_shim::tls`.

It still has somewhat scary safety requirements for users, but should ultimately be an improvement over the current implementation. It also adds refcounting to let us catch some issues.

### DIFF
--- a/src/lib/vasi-sync/Cargo.toml
+++ b/src/lib/vasi-sync/Cargo.toml
@@ -10,6 +10,7 @@ num_enum = { version = "0.6.1", default-features=false }
 rustix = { version = "0.37.19", default-features = false, features=["fs", "thread", "process"] }
 static_assertions = "1.1.0"
 vasi = { path = "../vasi" }
+rustc-hash = { version = "1.1.0", default-features=false }
 
 [dev-dependencies]
 criterion = "0.5.1"
--- a/src/lib/vasi-sync/src/atomic_tls_map.rs
+++ b/src/lib/vasi-sync/src/atomic_tls_map.rs
@@ -1,0 +1,290 @@
+use crate::sync::{atomic, AtomicUsize, Cell, ConstPtr, UnsafeCell};
+
+use core::hash::{BuildHasher, Hasher};
+use core::marker::PhantomData;
+use core::mem::MaybeUninit;
+use core::num::NonZeroUsize;
+use core::ops::Deref;
+
+/// A lockless, no_std, no-alloc hash table. Allows insertion and removal from
+/// an immutable reference, but does not support getting mutable references to
+/// internal values, and requires that a particular key is only ever accessed
+/// from the thread that inserted it, until that thread removes it.
+///
+/// Uses linear probing, and doesn't support resizing.  Lookup is `Θ(1)`
+/// (average case) if the key is present and the key hasn't been forced far away
+/// from its "home" location, but is `O(N)` worst case. Lookup of a non-present
+/// key is always `O(N)`; we need to scan the whole table.
+///
+/// This is designed mostly for use by `shadow_shim::tls` to help implement
+/// thread-local storage.
+pub struct AtomicTlsMap<const N: usize, V, H = core::hash::BuildHasherDefault<rustc_hash::FxHasher>>
+where
+    H: BuildHasher,
+{
+    keys: [AtomicOptionNonZeroUsize; N],
+    values: [UnsafeCell<MaybeUninit<V>>; N],
+    // This lets us enforce that a still-referenced key is not removed.
+    // TODO: Consider storing `RefCell<V>` in `values` instead. That'd be a bit
+    // more idiomatic, and is probably a better layout for cache performance.
+    refcounts: [Cell<usize>; N],
+    build_hasher: H,
+}
+/// Override default of `UnsafeCell` and `Cell` not being `Sync`.  We
+/// synchronize access to these (if partly by requiring users to guarantee no
+/// parallel access to a given key from multiple threads).
+/// Likewise `V` only needs to be `Send`.
+unsafe impl<const N: usize, V, H> Sync for AtomicTlsMap<N, V, H>
+where
+    // Requires for the Drop implementation to be able to drop values that were
+    // inserted by a different thread. Also if we want to support values being
+    // accessed by multiple threads with some kind of external synchronization,
+    // but I don't think we do.
+    //
+    // Alternatively we could only have this bound on the `Drop` implemenation,
+    // and document that the final contents aren't dropped if `V` isn't send. Or
+    // just remove the `Drop` impementation altogether.
+    V: Send,
+    H: Sync + BuildHasher,
+{
+}
+
+/// Adapter for `Option<NonZeroUsize>` around `AtomicUsize`
+struct AtomicOptionNonZeroUsize(AtomicUsize);
+impl AtomicOptionNonZeroUsize {
+    fn to_usize(val: Option<NonZeroUsize>) -> usize {
+        val.map(NonZeroUsize::get).unwrap_or(0)
+    }
+
+    fn from_usize(val: usize) -> Option<NonZeroUsize> {
+        NonZeroUsize::new(val)
+    }
+
+    pub fn new(val: Option<NonZeroUsize>) -> Self {
+        Self(AtomicUsize::new(Self::to_usize(val)))
+    }
+
+    pub fn load(&self, order: atomic::Ordering) -> Option<NonZeroUsize> {
+        Self::from_usize(self.0.load(order))
+    }
+
+    pub fn store(&self, val: Option<NonZeroUsize>, order: atomic::Ordering) {
+        self.0.store(Self::to_usize(val), order)
+    }
+
+    pub fn compare_exchange(
+        &self,
+        current: Option<NonZeroUsize>,
+        new: Option<NonZeroUsize>,
+        success: atomic::Ordering,
+        failure: atomic::Ordering,
+    ) -> Result<Option<NonZeroUsize>, Option<NonZeroUsize>> {
+        self.0
+            .compare_exchange(
+                Self::to_usize(current),
+                Self::to_usize(new),
+                success,
+                failure,
+            )
+            .map(Self::from_usize)
+            .map_err(Self::from_usize)
+    }
+}
+
+impl<const N: usize, V, H> AtomicTlsMap<N, V, H>
+where
+    H: BuildHasher,
+{
+    pub fn new_with_hasher(build_hasher: H) -> Self {
+        Self {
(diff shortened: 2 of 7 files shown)


src/lib/vasi-sync/Cargo.toml
```diff
@@ -10,6 +10,7 @@ num_enum = { version = "0.6.1", default-features=false }
 rustix = { version = "0.37.19", default-features = false, features=["fs", "thread", "process"] }
 static_assertions = "1.1.0"
 vasi = { path = "../vasi" }
+zwohash = { version = "0.1.2", default-features=false }
```
stevenengler, line 13: Is there a reason for using zwohash rather than fxhash (rustc-hash)? We already depend on rustc-hash and it has a lot more users (40,226,224 downloads vs 7,210).

sporksmith: fxhash isn't no_std :(

I don't have any particular attachment to zwohash though, if there's another no_std hash we ought to consider

src/lib/vasi-sync/src/atomic_tls_map.rs
```diff
@@ -0,0 +1,273 @@
...
+    ptr: ConstPtr<MaybeUninit<V>>,
+    refcount: &'a Cell<usize>,
+    _phantom: PhantomData<&'a V>,
+}
+
+impl<'a, V> Ref<'a, V> {
+    /// # Safety
+    ///
+    /// Current thread must be the only one to access `refcount`
+    unsafe fn new(ptr: ConstPtr<MaybeUninit<V>>, refcount: &'a Cell<usize>) -> Self {
```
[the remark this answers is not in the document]
sporksmith: It's primarily to detect if `remove` is called on a key that's still borrowed.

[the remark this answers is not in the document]
sporksmith: It occurred to me later that it might be a little more intuitive to make the `values` field an array of `RefCell<V>` instead of just `V`. That'd also build-in a method of doing mutable borrows, though I don't think we need that for TLS. I'm leaning slightly towards leaving it as-is, but am fine with either way.

src/lib/vasi-sync/Cargo.toml
stevenengler: It looks like it is if you specify `default-features=false`?

https://github.com/rust-lang/rustc-hash/blob/5ff7e6ae50e770381cb7bb393bfd22e016f2b8a6/src/lib.rs#L26
https://github.com/rust-lang/rustc-hash/blob/5ff7e6ae50e770381cb7bb393bfd22e016f2b8a6/Cargo.toml#L13

sporksmith: Ah, I was looking at the [fxhash](https://crates.io/crates/fxhash) crate. Sure, I'll switch it over.

stevenengler: Sounds good, I don't have a preference either if there's an advantage using a different hash function, but all other things equal I think it would probably be best to choose the popular (and rust-lang-owned) one.

src/lib/vasi-sync/src/atomic_tls_map.rs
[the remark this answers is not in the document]
sporksmith: Added a comment to the `remove` method

src/lib/vasi-sync/src/atomic_tls_map.rs
```diff
@@ -0,0 +1,271 @@
...
+    /// any outstanding references for `key`.
+    ///
+    /// # Safety
+    ///
+    /// *No thread* must have outstanding references to the value at `key`, or perform
+    /// parallel operations on `self` with `key`.
+    pub unsafe fn remove(&self, key: NonZeroUsize) -> Option<V> {
+        let idx = self.idx(key)?;
+        assert_eq!(self.refcounts[idx].get(), 0);
+        let value = self.values[idx].get_mut().with(|value| {
```
stevenengler, line 192: I can't tell what's going on here. `self.values` is a `[UnsafeCell<MaybeUninit<V>>; N]`, so `self.values[idx]` should be an `UnsafeCell<MaybeUninit<V>>`. But `UnsafeCell::get_mut()` requires a mutable reference, but `self` is a shared reference, so I must be reading this wrong.

src/lib/vasi-sync/src/atomic_tls_map.rs
```diff
@@ -0,0 +1,271 @@
...
+            let value = unsafe { &mut *value };
+            unsafe { value.assume_init_read() }
+        });
+        // Syncs with `Acquire` on insertion
+        self.keys[idx].store(None, atomic::Ordering::Release);
+        Some(value)
```
stevenengler, line 198: Maybe comment that anything between the `assume_init_read` above and the `store` must not panic, otherwise we'll drop the value twice.

src/lib/vasi-sync/src/atomic_tls_map.rs
```diff
@@ -0,0 +1,271 @@
...
+                    let value = unsafe { &mut *value };
+                    // SAFETY: We know the value is initialized.
+                    unsafe { value.assume_init_drop() }
+                })
+            }
+        }
+    }
+}
+
+pub struct Ref<'a, V> {
```
stevenengler, line 234: static assert that `Ref<()>` is not `Send` or `Sync`, to make sure even if `V` is send or sync that `Ref` isn't?

src/lib/vasi-sync/tests/atomic-tls-map-tests.rs
```diff
@@ -0,0 +1,213 @@
...
+                );
+                assert_eq!(
+                    *table.get_or_insert_with(NonZeroUsize::try_from(2).unwrap(), || 1000),
+                    12
+                );
+
+                // Likewise `get` should return the inserted values.
+                assert_eq!(
+                    table
+                        .get(NonZeroUsize::try_from(NonZeroUsize::try_from(1).unwrap()).unwrap())
```
stevenengler, line 51: Is the extra `try_from` needed here? It seems to be converting a `NonZeroUsize` to a `NonZeroUsize`?

stevenengler (approved):

src/lib/vasi-sync/src/atomic_tls_map.rs
```diff
@@ -0,0 +1,271 @@
...
+    refcounts: [Cell<usize>; N],
+    build_hasher: H,
+}
+/// Override default of `UnsafeCell` and `Cell` not being `Sync`.  We
+/// synchronize access to these (if partly by requiring users to guarantee no
+/// parallel access to a given key from multiple threads).
+/// Likewise `V` only needs to be `Send`.
+unsafe impl<const N: usize, V, H> Sync for AtomicTlsMap<N, V, H>
+where
+    V: Send,
```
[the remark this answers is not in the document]
sporksmith: It is now. Done.

src/lib/vasi-sync/src/atomic_tls_map.rs
```diff
@@ -0,0 +1,271 @@
...
+            // Relaxed because of requirement that only one thread ever accesses
+            // a given key at once.
+            self.keys[*idx].load(atomic::Ordering::Relaxed) == Some(key)
+        })
+    }
+
+    /// # Safety
+    ///
+    /// No other thread must access `self` with `key` in parallel, including
+    /// having outstanding references.
```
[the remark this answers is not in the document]
sporksmith: Changed to "The value at `key`, if any, must have been inserted by the current thread."

As noted above, this is stronger than strictly necessary for the implementation, but that's ok.

src/lib/vasi-sync/src/atomic_tls_map.rs
```diff
@@ -0,0 +1,271 @@
...
+    #[inline]
+    pub unsafe fn get(&self, key: NonZeroUsize) -> Option<Ref<V>> {
```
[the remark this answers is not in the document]
sporksmith: Reworded all the safety comments to simplify/strengthen the requirements.

src/lib/vasi-sync/src/atomic_tls_map.rs
```diff
@@ -0,0 +1,271 @@
...
+        Some(unsafe { Ref::new(ptr, &self.refcounts[idx]) })
+    }
+
+    /// Insert `(key, value)`.
+    ///
+    /// If `key` is already present in `self`, the previous value would shadow
+    /// the newly inserted value. We don't expose this function in the public
+    /// API since this behavior would be confusing.
+    ///
+    /// Returns a reference to the newly inserted value.
```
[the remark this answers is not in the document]
sporksmith: Done

src/lib/vasi-sync/src/atomic_tls_map.rs
sporksmith: Added a method comment on `sync::UnsafeCell`

```
    /// Note that this has a different signature from the method
    /// of the same name in `core::cell::UnsafeCell`.
    /// See https://docs.rs/loom/latest/loom/#handling-loom-api-differences
```

src/lib/vasi-sync/src/atomic_tls_map.rs
sporksmith: done

src/lib/vasi-sync/src/atomic_tls_map.rs
```diff
@@ -0,0 +1,271 @@
...
+
+impl<const N: usize, V, H> Drop for AtomicTlsMap<N, V, H>
+where
+    H: BuildHasher,
+{
+    fn drop(&mut self) {
+        for idx in 0..N {
+            // Relaxed because we have exclusive access to `self`. Any other threads
+            // that have accessed the table must have synchronized with the current
+            // thread to be able to drop `self` in the first place.
```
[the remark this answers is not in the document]
sporksmith: Right. Reworded to try to clarify.

src/lib/vasi-sync/src/atomic_tls_map.rs
sporksmith: Done

src/lib/vasi-sync/tests/atomic-tls-map-tests.rs
sporksmith: Oops, done

sporksmith:
It occurred to me that I didn't actually have a test for "cross thread" drops. After adding one, the [mysterious loom crash](https://github.com/tokio-rs/loom/issues/316) resurfaced. I reduced the number of `--test-threads` to 1 when testing under loom, which seems to avoid it more reliably.

PR #3066, opened 2023-07-06
Shim: Move shared memory objects into Rust

sporksmith:
We need at least the Process shared memory in Rust to be able to migrate the signal handling to Rust. Might as well move the others too.

### DIFF
--- a/src/lib/shim/src/lib.rs
+++ b/src/lib/shim/src/lib.rs
@@ -7,7 +7,7 @@ use core::ffi::CStr;
 use crate::tls::ShimTlsVar;
 
 use shadow_shim_helper_rs::ipc::IPCData;
-use shadow_shim_helper_rs::shim_shmem::{ManagerShmem, ThreadShmem};
+use shadow_shim_helper_rs::shim_shmem::{ManagerShmem, ThreadShmem, HostShmem};
 use shadow_shim_helper_rs::simulation_time::SimulationTime;
 use shadow_shmem::allocator::{Serializer, ShMemBlockAlias, ShMemBlockSerialized};
 use vasi_sync::lazy_lock::LazyLock;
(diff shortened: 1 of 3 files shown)


src/lib/shim/src/lib.rs
```diff
@@ -225,6 +219,125 @@ mod tls_thread_shmem {
...
+
+    pub fn try_get(
+    ) -> Option<impl core::ops::Deref<Target = ShMemBlockAlias<'static, ManagerShmem>> + 'static>
+    {
+        if !SHMEM.initd() {
+            None
+        } else {
+            Some(get())
+        }
+    }
```
stevenengler, line 259: This seems kind of confusing to me. I would expect a `try_get` to try to get the object, and return `None` if it couldn't. But `try_get` doesn't even try to get the object. The serialized object may be ready in `INITIALIZER`, but `try_get` won't even try deserializing it. You must call `get` first before `try_get` will ever work. I guess this is relying on the fact that `set()` calls `SHMEM.force()` (which is a `get()`). I think the documentation here could be improved a bit.

src/lib/shim/src/lib.rs
```diff
@@ -225,6 +219,125 @@ mod tls_thread_shmem {
...
+    });
+
+    /// # Safety
+    ///
+    /// `blk` must contained a serialized block of type `ManagerShmem`, which
+    /// outlives the process.
+    pub unsafe fn set(blk: &ShMemBlockSerialized) {
+        assert!(INITIALIZER.lock().replace(*blk).is_none());
+        SHMEM.force();
+    }
```
stevenengler, line 243: optional: maybe also assert that `!SHMEM.initd()`?

src/lib/shim/src/lib.rs
```diff
@@ -225,6 +219,125 @@ mod tls_thread_shmem {
...
+    // It uses `INITIALIZER` to do its one-time init.
+    static SHMEM: LazyLock<ShMemBlockAlias<ManagerShmem>> = LazyLock::const_new(|| {
+        let serialized = INITIALIZER.lock().take().unwrap();
+        unsafe { Serializer::global().deserialize(&serialized) }
+    });
+
+    /// # Safety
+    ///
+    /// `blk` must contained a serialized block of type `ManagerShmem`, which
+    /// outlives the process.
```
stevenengler, line 239: nit: and add that the block must outlive the process?

src/lib/shim/src/lib.rs
```diff
@@ -433,4 +546,67 @@ pub mod export {
...
+    }
+
+    #[no_mangle]
+    pub extern "C" fn shim_managerSharedMem(
+    ) -> *const shadow_shim_helper_rs::shim_shmem::export::ShimShmemManager {
+        let rv = global_manager_shmem::try_get();
+        rv.map(|x| {
+            let rv: &shadow_shim_helper_rs::shim_shmem::export::ShimShmemManager = &x;
+            rv as *const _
+        })
```
stevenengler, line 567: This seems a bit confusing since it's returning a reference to something passed by value to the closure (the closure takes an `x` and returns `&x`). I'm guessing that `deref` is being called here somewhere implicitly? Maybe make this more explicit?

stevenengler (approved):

src/lib/shim/src/lib.rs
sporksmith: Right. Added comments and an explicit call to `deref`

PR #3069, opened 2023-07-11
Convert shim signal handling to Rust

sporksmith:
For now this is basically a transcription of the C implementation. I plan to work on cleaning up `process_signals` in particular to get rid of the usage of libc's swapcontext and see if I can clean up the control flow.


src/lib/shim/src/signals.rs
```diff
@@ -0,0 +1,363 @@
...
+            unsafe {
+                *process
+                    .protected
+                    .borrow_mut(&host_lock.root)
+                    .signal_action_mut(sig) = sigaction::new_with_default_restorer(
+                    SignalHandler::SigDfl,
+                    SigActionFlags::empty(),
+                    sigset_t::EMPTY,
+                )
+            };
```
stevenengler, line 111: leave a comment why this is safe (is `SigDfl`)?

stevenengler (approved):

src/lib/shim/src/signals.rs
sporksmith: Done

PR #3071, opened 2023-07-12
Fix and reenable clone tests

sporksmith:
These tests *appear* to no longer be flaky with the new implementation of shim thread local storage. I did have to make some changes to avoid smashing the relatively small stack we use in this test though:

* Enable minimal optimizations in debug builds. Rust has a tendency to put more temporaries on the stack than in C code. Enabling some optimizations lets us optimize many of those away, e.g. when initializing an object that will ultimately live in static or heap memory.
* Reduced the max threads supported by shim TLS from 1000 to 100. 1000 was from when this was a maximum number of threads *ever* for a process. In the new implementation this limit is for *simultaneous* threads per process, so 1000 is probably excessive. This reduces the size of the corresponding table, which seems to end up as a temporary on the stack even with optimizations enabled.

I verified that neither of those alone were sufficient to fix the test.

We could theoretically increase the stack size in the test from 16 KB instead of one or both of those  changes, but I think it's worth avoiding a requirement for a relatively large stack for initialization. e.g. on my machine libc defines `MINSIGSTKSZ` as 2048 and `SIGSTKSZ` as 8192, so requiring 16 KB could already be pushing it.

I also reenabled shim logging in these tests; that should no longer risk trying to touch native thread local storage now that logging has been migrated to no-std no-libc rust.

### DIFF
--- a/src/Cargo.toml
+++ b/src/Cargo.toml
@@ -20,3 +20,16 @@ members = [
     "lib/vasi-macro",
     "lib/vasi-sync",
 ]
+
+[profile.dev]
+# Without this, the shim requires a relatively large stack, especially during
+# initialization.  Rust makes it difficult to avoid putting objects temporarily
+# on the stack, even if it'll ultimately live in heap or static memory, but
+# opt-level 1 is enough to optimize a lot of those away.
+# 
+# See https://stackoverflow.com/questions/25805174/creating-a-fixed-size-array-on-heap-in-rust/68122278#68122278
+#
+# In our test suite the main place we see failures without this is in the
+# `clone-shadow` test, but it could also affect golang or other environments
+# with "light" user-space threading.
+opt-level = 1 
--- a/src/lib/shim/src/tls.rs
+++ b/src/lib/shim/src/tls.rs
@@ -96,7 +96,7 @@ pub const BYTES_PER_THREAD: usize = 1024;
 // Max threads for our slow TLS fallback mechanism.  We support recycling
 // storage of exited threads, so this is the max *concurrent* threads per
 // process.
-const TLS_FALLBACK_MAX_THREADS: usize = 1000;
+const TLS_FALLBACK_MAX_THREADS: usize = 100;
 
 /// An ELF thread pointer, as specified in
 /// <https://www.akkadia.org/drepper/tls.pdf)>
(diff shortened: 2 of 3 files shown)


sporksmith:
> requiring 16 KB could already be pushing it.

To be fair, we use 4 KB as a guard page, so it's really a 12 KB stack. I verified locally that allocating 12 KB (giving an 8 KB stack) also works, but I'm happy to leave a *little* breathing room here to keep this test from becoming flaky. (Allocating 8 KB to give a 4 KB stack fails)

src/Cargo.toml
```diff
@@ -20,3 +20,11 @@ members = [
...
     "lib/vasi-sync",
 ]
+
+[profile.dev]
+# shim/src/tls.rs, and maybe other places, needs this
+# to avoid smashing the stack with a large immediate temporary.
+# opt-level 1 is enough to optimize away the offending temporary.
+# 
+# See https://stackoverflow.com/questions/25805174/creating-a-fixed-size-array-on-heap-in-rust/68122278#68122278
+opt-level = 1
```
stevenengler, line 30: I think this will also affect things like stack traces in debug builds, which might make it more difficult to debug issues. Which part of the code is overflowing the stack? I only see things in the code that look like they're at most 24 KiB, which I think wouldn't cause a problem? Are there any other options, like requiring the type to be `Copy`, passing references around, and using `ptr.write()`? Or constructing it in a `const` context so that it gets embedded into the binary instead?

stevenengler (approved):

src/Cargo.toml
stevenengler: Why is the stack so small (16 KB)?

sporksmith: > Which part of the code is overflowing the stack?

I don't know of a good way to identify for sure, other than trial and error or poking through assembly/gdb looking for `rsp` decrements. https://crates.io/crates/cargo-call-stack looked promising, but I couldn't get it working.

> Why is the stack so small (16 KB)?

This test uses a small stack mostly to push how small of a stack we can run on; ideally it'd be a separate test I suppose.

As mentioned in the PR description, `MINSIGSTKSZ` and `SIGSTKSZ` (described in `sigaltstack(2)`) are even smaller, so it'd be nice not to require a stack a lot larger than that.

I'd rather try things this way for now, and if the opt level makes debugging too onerous falling back to increasing this test stack size.

stevenengler: Maybe add a comment to the `opt-level = 1` to point it at this test, if it's the only place we use a stack small enough to run into this issue?

sporksmith: done

PR #3077, opened 2023-07-17
Migrate shim clone code to Rust and make some improvements

sporksmith:
* Migrates clone-handling code to Rust
* This gets rid of some C references to the shared memory allocator types
* Transfer the serialized IPC block for the new thread through the new thread's stack instead of through a global static + spinlock
* Delete the now unused specialized spinlock and semaphore


sporksmith:
Incidentally I wouldn't be surprised if this spinlock was related to the mysterious pauses we see sometimes at simulation start. I've never been able to confirm it, and theoretically it shouldn't create *much* contention, but it definitely resulted in there being at least some over-subscription of CPUs; it's not too difficult to imagine there occasionally being some especially bad corner case

src/lib/shadow-shim-helper-rs/build.rs
```diff
@@ -62,9 +62,4 @@ fn main() {
     let build_common = ShadowBuildCommon::new(Path::new("../../.."), Some(deps));
 
     run_cbindgen(&build_common);
-
-    build_common
-        .cc_build()
-        .files(&["shadow_sem.c", "shadow_spinlock.c"])
-        .compile("shim_c");
```
stevenengler, line 69: 🎉

src/lib/shadow-shim-helper-rs/src/shim_event.rs
```diff
@@ -25,6 +25,19 @@ pub struct ShimEventSyscallComplete {
 #[repr(C)]
 pub struct ShimEventAddThreadReq {
     pub ipc_block: ShMemBlockSerialized,
+    // Params for clone syscall
+    pub flags: libc::c_ulong,
+    pub child_stack: UntypedForeignPtr,
+    pub ptid: UntypedForeignPtr,
+    pub ctid: UntypedForeignPtr,
```
stevenengler, line 32: nit: it should be obvious based on the context, but maybe doc-comment which address space the pointers are valid in since they're in shared memory.

src/lib/shadow-shim-helper-rs/src/syscall_types.rs
```diff
@@ -99,6 +99,16 @@ impl<T> ForeignPtr<T> {
...
+
+    #[inline]
+    pub fn into_raw(self) -> *const T {
+        self.val as *const T
+    }
+
+    #[inline]
+    pub fn into_raw_mut(self) -> *mut T {
+        self.val as *mut T
+    }
```
stevenengler, line 111: nit: maybe doc comment to remind the caller to make sure the `ForeignPtr` is valid in this address space, otherwise it creates a dangling pointer

src/lib/shim/src/clone.rs
```diff
@@ -0,0 +1,253 @@
...
+    unsafe { core::ptr::write(child_ipc_blk, event.ipc_block) };
+
+    // Ensure stack is 16-aligned so that we can safely make function calls.
+    let child_current_rsp = unsafe { align_down(child_current_rsp, 16) };
+
+    let rv: i64;
+    // SAFETY:
+    // * There's currently no way to tell Rust than an asm block "returns twice",
+    //   so the child does not return from this asm block. It instead *jumps*
+    //   to the point where the clone syscall war originally made.
```
stevenengler, line 176: since this can create a new thread with the same address space, doesn't this lead to the same issues as [vfork](https://github.com/rust-lang/libc/issues/1596)? does the jump avoid the same memory corruption issue?

stevenengler (approved):

src/lib/shim/src/clone.rs
sporksmith: Yes that's the idea - we're jumping to where the context was created, which should be where the managed code was trying to make a `clone` syscall, which should be a place where it's already safe to return twice. Thanks for the reference  - I was meaning to look that up again and add it here.

sporksmith: Expanded this comment and  added the link to the Rust issue.

sporksmith:
benchmark full results: https://github.com/shadow/benchmark-results/tree/master/tor/2023-07-19-T16-01-27

wall time:
image

Appears to be on the faster side of recent runs. Could be an improvement and/or solve the pause problem, but hard to be certain without more runs. As expected though, I think we can be fairly confident it at least doesn't hurt performance.

PR #3080, opened 2023-07-19
Extra tests workflow: tweaks to sccache

sporksmith:
Fetch of sccache release was failing: https://support.github.com/ticket/personal/0/2258460

Was going to disable it in this PR, but it looks like the issue has been resolved. Still worth merging the other couple small tweaks I made to the related code though, I think.

### DIFF
--- a/.github/workflows/extra_tests.yml
+++ b/.github/workflows/extra_tests.yml
@@ -53,6 +53,8 @@ jobs:
           # See https://github.com/shadow/shadow/issues/2166
           ref: ${{ github.event.pull_request.head.sha }}
           persist-credentials: false
+      - name: Set Rust toolchain
+        run: ln -s shadow/ci/rust-toolchain-stable.toml rust-toolchain.toml
       - name: Install dependencies
         run: |
           cd shadow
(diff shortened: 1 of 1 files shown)


.github/workflows/extra_tests.yml
```diff
@@ -53,6 +53,8 @@ jobs:
           # See https://github.com/shadow/shadow/issues/2166
           ref: ${{ github.event.pull_request.head.sha }}
           persist-credentials: false
+      - name: Set Rust toolchain
+        run: ln -s shadow/ci/rust-toolchain-stable.toml rust-toolchain.toml
```
robgjansen, line 57: Are you sure we don't want the `rust-toolchain.toml` to be sitting in the CWD where we initiate the shadow build, which I think is the shadow base dir? (We `cd` into `shadow/` before running the build script.)

robgjansen (approved):

.github/workflows/extra_tests.yml
sporksmith: Yes, the ancestor directories are also searched for `rust-toolchain.toml`, so this affects the `shadow` directory as well.

PR #3082, opened 2023-07-20
Rm shutdown helper

sporksmith:
Add reference counts for `EpollWatch` and `Descriptor`, and remove `DescriptorTable::shutdown_helper`.

Calling `DescriptorTable::shutdown_helper`, which unregisters all epoll descriptors' listeners, would be incorrect if there are still open descriptors to the same epoll file description (e.g. in a forked process).

### DIFF
--- a/src/main/host/descriptor/descriptor.c
+++ b/src/main/host/descriptor/descriptor.c
@@ -270,14 +270,6 @@ void legacyfile_removeFlags(LegacyFile* descriptor, gint flags) {
     descriptor->flags &= ~flags;
 }
 
-void legacyfile_shutdownHelper(LegacyFile* legacyDesc) {
-    MAGIC_ASSERT(legacyDesc);
-
-    if (legacyDesc->type == DT_EPOLL) {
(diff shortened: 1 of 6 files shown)


sporksmith:
@robgjansen  Shadow added the behavior of explicitly clearing epoll listeners on shutdown in your 069e715a827d51a3251ab082e28beae5b36983ff. At that point it was during Host shutdown, since descriptors were then per-Host instead of per-Process, and `DescriptorTable` didn't exist yet.

My best guess is that this was meant to break some circular reference, which would otherwise cause a memory leak. In the current code I don't see how that would happen though, and afaict there are no leaks or other problems after removing it @ head.

Do you happen to remember anything else about why this was added or know of any other potential issues with removing it?

robgjansen (approved):
I believe the hack was to make sure refcounts were zero at the end of the sim due to some circular refs, but I'm happy to have this hack removed if our better designed Rust / File handles are now taking care of this.

I guess there is some small chance that we only reach the corner case that the shutdown helper was designed to fix in a larger simulation, but if that happens )e.g. in the next nightly) we can revisit.

PR #3083, opened 2023-07-20
Fix shell scripts

sporksmith:
We had a subtle issue in our tor minimal test validation script causing it to ignore some errors.

trinity @ Tor pointed out that shellcheck would've caught it. Tor uses shellcheck in their CI in general.

So I also added a shellcheck lint to our CI and fixed various warnings. I don't think any of these were currently causing issues, but they could potentially have caused issues in the future (e.g. if an unquoted variable value later ended up having spaces in it).

I locally tested `ci/run.sh` since some of the related scripts were being modified. It turns out there were some issues with the rust toolchain files from a previous commit, so I also fixed those.

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -71,7 +71,7 @@ run_one () {
     TAG="$REPO:$CONTAINER_FOR_TAG-$CC-$BUILDTYPE"
 
     if [ "${DRYTAG}" == "1" ]; then
-        echo $TAG
+        echo "$TAG"
         return 0
     fi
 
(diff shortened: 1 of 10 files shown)


src/test/tor/minimal/verify.sh
```diff
@@ -18,7 +18,11 @@ check_host () {
     local NAME=$1
     local MINIMUM_STREAMS=$2
     local TOTAL_STREAMS=$3
-    local stream_count=$(grep -c stream-success hosts/$NAME/tgen.*.stdout)
+    # Careful to declare the local separately from assigning the value;
+    # otherwise we miss errors from `grep` such as if the files don't exist.
+    # https://unix.stackexchange.com/a/343259
+    local stream_count
```
robgjansen, line 24: Rename to `stream_success_count`?

ci/run.sh
```diff
@@ -98,14 +98,14 @@ run_one () {
         COPY ci/container_scripts/install_extra_deps.sh /root/install_extra_deps.sh
         # Set the default rust toolchain to be installed by the script.
         #
-        # Copying just this config file before we copy the whole shadow source
+        # Copying just this config files before we copy the whole shadow source
```
robgjansen, line 101: `this` -> `these`

robgjansen (approved):
Cool, nice to have automated processes helping us avoid mistakes :)

src/test/tor/minimal/verify.sh
sporksmith: done

ci/run.sh
sporksmith: done

PR #3086, opened 2023-08-01
Remove `updatePacketHeader` function

stevenengler:
@robgjansen You had mentioned a long time ago why this `updatePacketHeader` function existed, but I can't remember anymore and I can't find an issue for it. Regardless, I don't think it's needed anymore as its own separate function since it's currently only called immediately after a `pullOutPacket()` and we can just merge them into a single operation (just `pullOutPacket`).

Since this is only relevant to the C TCP sockets, `tcp_networkInterfaceIsAboutToSendPacket` is now called from `LegacyTcpSocket::pull_out_packet`.

(This was added in 840234d0a3a92292.)

### DIFF
--- a/src/main/host/descriptor/compat_socket.c
+++ b/src/main/host/descriptor/compat_socket.c
@@ -174,23 +174,3 @@ Packet* compatsocket_pullOutPacket(const CompatSocket* socket, const Host* host)
 
     utility_panic("Invalid CompatSocket type");
 }
-
-void compatsocket_updatePacketHeader(const CompatSocket* socket, const Host* host, Packet* packet) {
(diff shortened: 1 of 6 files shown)


robgjansen (approved):
The point of the function was just to make sure that we set the packet header fields to the most recent state that we have available at the time that the packet is actually being "sent" into the network. Prior to this function, we set the packet header fields when the packet was created (which we do as soon as the application writes to the socket) but then by the time the packet was sent the header fields would be stale.

In any case, I agree that merging this with the `pull_out_packet` function makes sense, and I think in the new rust TCP code you are planning to not even create the header until right before the packet gets sent, which is a better design.

PR #3091, opened 2023-08-03
Use std::io::IsTerminal instead of nix::unistd::isatty

sporksmith:
`isatty` can return surprising errors on some file systems. `IsTerminal` interprets such errors as "not a terminal", which is generally the behavior we want. In particular this should fix a crash we're getting in our tor and tgen benchmarks.

Background:
* https://docs.kernel.org/driver-api/ioctl.html#return-code
* https://github.com/containers/podman/issues/120700

Crash example:
* https://github.com/shadow/benchmark/actions/runs/5746502577/job/15576071390

  ```
  thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: ENOSYS', main/core/main.rs:146:64
  fatal runtime error: failed to initiate panic, error 5
  ```

### DIFF
--- a/src/main/utility/mod.rs
+++ b/src/main/utility/mod.rs
@@ -347,6 +347,8 @@ mod tests {
 }
 
 mod export {
+    use std::io::IsTerminal;
+
     #[no_mangle]
     pub unsafe extern "C" fn utility_handleErrorInner(
         file_name: *const libc::c_char,
@@ -394,7 +396,8 @@ mod export {
             **ABORTING**"
         );
 
-        if !nix::unistd::isatty(libc::STDOUT_FILENO).unwrap_or(true) {
+        // TODO: Why do we only print this if stdout isn't a terminal?
+        if !std::io::stdout().lock().is_terminal() {
             println!("{error_msg}");
         }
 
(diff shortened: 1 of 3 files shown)


src/main/utility/mod.rs
```diff
@@ -394,7 +396,8 @@ mod export {
             **ABORTING**"
         );
 
-        if !nix::unistd::isatty(libc::STDOUT_FILENO).unwrap_or(true) {
+        // TODO: Why do we only print this if stdout isn't a terminal?
+        if !std::io::stdout().lock().is_terminal() {
             println!("{error_msg}");
         }
```
stevenengler, line 402: I think this was intended to be the same reason as what we're doing with error log messages. We always do `eprintln!("{error_msg}");` below, so there's no reason to also do a `println!("{error_msg}");` if they're both being written to a terminal.

stevenengler (approved):

src/main/utility/mod.rs
sporksmith: ah right, I missed the unconditional `eprintln` below

PR #3092, opened 2023-08-04
Add epoll bindings to linux-api

robgjansen:


src/lib/linux-api/README.md
```diff
@@ -0,0 +1,23 @@
...
+You probably want to pin your version of bindgen to keep the diff small.
+Check the current comment at the top of `src/bindings.rs` to see the
+version used to generate the bindings last time. Then, force your local
+tooling to match versions like:
+
+    cargo install --force --version 0.65.1 bindgen-cli
+
+Then the magix you want to regenerate src/bindings.rs is:
+
+    bash ./gen-kernel-bindings.sh && rustfmt src/bindings.rs
```
sporksmith, line 12: optional - we could probably add the `rustfmt` step to `gen-kernel-bindings.sh` itself. I think I'd considered running `cargo fmt` but couldn't figure out a way to get it to only format one file, and forgot we could just call `rustfmt` directly

sporksmith (approved):

PR #3095, opened 2023-08-07
Fix sending packets from sockets bound to 0.0.0.0

### LINKED ISSUE #3088: UDP sockets bound to 0.0.0.0 cannot send packets
When a UDP socket informs the host that it has packets to send, it calls the host's `notify_socket_has_packets` method. This method takes an IP address, but it's not clear what this IP address is supposed to be. In `notify_socket_has_packets`, the IP address is passed to the network namespace's `interface_borrow` method, which returns a network interface for that IP address. The `notify_socket_has_packets` then notifies that returned network interface that there are packets to send. This means that the provided IP address must be either 127.0.0.1 or the host's public IP address. If the IP address is anything else, then the `notify_socket_has_packets` returns without doing anything. If a UDP socket is bound to 0.0.0.0 and gives this wildcard IP address to `notify_socket_has_packets`, nothing happens since that isn't one of these two IP addresses. It's unclear what IP address the UDP socket is supposed to give to `notify_socket_has_packets`. Should it call `notify_socket_has_packets` twice, once with each IP address? This seems like it could be bad since both network interfaces will be trying to send packets at the same time. And if the loopback interface tried to send a non-localhost packet, that would be a problem. So it should probably call `notify_socket_has_packets` once with the public IP address. But maybe `notify_socket_has_packets` should handle 0.0.0.0 instead. All of our integration tests send UDP messages only over localhost, so this bug wasn't caught by them.

stevenengler:
I don't think this is a proper fix since it means that localhost packets from sockets bound to 0.0.0.0 will be sent over the internet interface rather than the loopback interface. But I don't think there's anything else we can do here. The C TCP code does [something](https://github.com/shadow/shadow/issues/3088#issuecomment-1668389960) slightly different, but I don't think it's better than using just the internet interface.

Closes #3088.

### DIFF
--- a/src/main/host/network/namespace.rs
+++ b/src/main/host/network/namespace.rs
@@ -145,9 +145,16 @@ impl NetworkNamespace {
         &self,
         addr: Ipv4Addr,
     ) -> Option<impl Deref<Target = NetworkInterface> + '_> {
+        // Notes:
+        // - The `is_loopback` matches all loopback addresses, but shadow will only work correctly
+        //   with 127.0.0.1. Using any other loopback address will lead to problems.
+        // - If the address is 0.0.0.0, we return the `internet` interface. This is not ideal if a
+        //   socket bound to 0.0.0.0 is trying to send a localhost packet and uses this method to
+        //   get the network interface, since the packet will be sent on the internet interface
+        //   instead of loopback. It's not clear if this will lead to bugs.
         if addr.is_loopback() {
             Some(self.localhost.borrow())
-        } else if addr == self.default_ip {
+        } else if addr == self.default_ip || addr.is_unspecified() {
             Some(self.internet.borrow())
         } else {
             None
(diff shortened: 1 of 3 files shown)


src/main/host/network/namespace.rs
```diff
@@ -147,7 +147,7 @@ impl NetworkNamespace {
     ) -> Option<impl Deref<Target = NetworkInterface> + '_> {
         if addr.is_loopback() {
             Some(self.localhost.borrow())
-        } else if addr == self.default_ip {
+        } else if addr == self.default_ip || addr.is_unspecified() {
```
sporksmith, line 150: Maybe worth a comment here that this is a (temporary?) workaround with not-quite-right behavior?

src/main/host/network/namespace.rs
```diff
@@ -162,7 +162,7 @@ impl NetworkNamespace {
     ) -> Option<impl Deref<Target = NetworkInterface> + DerefMut + '_> {
         if addr.is_loopback() {
             Some(self.localhost.borrow_mut())
-        } else if addr == self.default_ip {
+        } else if addr == self.default_ip || addr.is_unspecified() {
```
sporksmith, line 165: same here

sporksmith (approved):

src/main/host/network/namespace.rs
stevenengler: Added some comments.

src/main/host/network/namespace.rs
stevenengler: Added the same comments as above.

PR #3096, opened 2023-08-07
Add better support for recursive RootedRc

sporksmith:
This is in preparation for moving the `DescriptorTable` into a `RootedRc`. Since the object containing the `DescriptorTable` (currently `Process`, but soon to be `Thread`) is also in a `RootedRc`, safely dropping becomes a bit complex.

In the past when we've run into this, we've solved it by having the `Drop` impl of the "outer" type get the `Root` from the thread-local `Worker`, but in general we're trying to move away from implicit thread-local accesses in favor of explicit parameter passing.

This first adds `RootedRc::into_inner`. This lets us extract the inner value when dropping the last strong reference of a `RootedRc`. The caller is then able to recursively safely destruct the inner value. This has the downside though of requiring any call site that wants to destruct a `RootedRc<T>` to also know the details about how to safely destruct `T`.

This next adds a trait `ExplicitDrop`, so that types can provide an explicit drop method through a uniform interface. This also lets us add `RootedRc<T>::explicit_drop_recursive`  where `T` implements `ExplicitDrop`, which takes care of safely destructing `T` when the last strong reference to it is dropped.

### DIFF
--- a/src/lib/shadow-shim-helper-rs/src/rootedcell/rc.rs
+++ b/src/lib/shadow-shim-helper-rs/src/rootedcell/rc.rs
@@ -3,6 +3,8 @@ use std::{
     ptr::NonNull,
 };
 
+use crate::explicit_drop::ExplicitDrop;
+
 use super::{Root, Tag};
 
 struct RootedRcInternal<T> {
@@ -160,10 +162,12 @@ unsafe impl<T: Sync + Send> Sync for RootedRcCommon<T> {}
 /// the owner is required to prove ownership of the associated [Root]
 /// to perform any sensitive operations.
 ///
-/// Instances must be destroyed using [RootedRc::safely_drop], which validates
-/// that the [Root] is held before manipulating reference counts, etc.
-/// Failing to call [RootedRc::safely_drop] results in a `panic` in debug builds,
-/// or leaking the object in release builds.
+/// Instances must be destroyed explicitly, using [`RootedRc::explicit_drop`],
+/// [`RootedRc::explicit_drop_recursive`], or [`RootedRc::into_inner`].  These
+/// validate that the [Root] is held before manipulating reference counts, etc.
+///
+/// Dropping `RootedRc` without calling one of these methods results in a
+/// `panic` in debug builds, or leaking the object in release builds.
 pub struct RootedRc<T> {
     common: RootedRcCommon<T>,
 }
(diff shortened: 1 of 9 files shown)


src/lib/shadow-shim-helper-rs/src/rootedcell/rc.rs
```diff
@@ -202,15 +204,32 @@ impl<T> RootedRc<T> {
...
-    /// safely cleaned up. In debug builds this will result in a `panic`.
-    /// Otherwise the underlying reference count will simply not be decremented,
-    /// ultimately resulting in the enclosed value never being dropped.
+    /// Drop the `RootedRc`, and return the inner value if this was the last
+    /// strong reference.
     #[inline]
-    pub fn safely_drop(self, root: &Root) {
+    pub fn into_inner(this: Self, root: &Root) -> Option<T> {
+        this.common.safely_drop(root, RefType::Strong)
+    }
```
stevenengler, line 212: Do we have anything in a `RootedRc` in Shadow that relies on the fact that it doesn't (well didn't previously) move? For example I think I remember the host and host shmem lock having circular references with unsafe, relying on the fact that the lock would never move? I don't have any specific issue in mind, but just wanted to leave a comment to maybe double check that this change doesn't break any assumption of that code.

src/lib/shadow-shim-helper-rs/src/rootedcell/rc.rs
```diff
@@ -202,15 +204,32 @@ impl<T> RootedRc<T> {
...
+
+    /// Drops `self`, and if `self` was the last strong reference, call
+    /// `ExplicitDrop::explicit_drop` on the internal value.
+    pub fn explicit_drop_recursive<I, O>(self, root: &Root, param: &I) -> Option<O>
+    where
+        T: ExplicitDrop<ExplicitDropParam = I, ExplicitDropResult = O>,
+    {
+        Self::into_inner(self, root).map(|val| val.explicit_drop(param))
+    }
+}
```
stevenengler, line 222: Does this need the `I` and `O` generics? I think you could just use `<T as ExplicitDrop>::ExplicitDropParam` and `<T as ExplicitDrop>::ExplicitDropResult` wherever `I` and `O` are used to simplify the type inference?

stevenengler (approved):

src/lib/shadow-shim-helper-rs/src/rootedcell/rc.rs
sporksmith: Yup! Good catch!

PR #3098, opened 2023-08-08
Add 'tcp' crate

stevenengler:
This adds the 'tcp' crate as of commit dd77c2ca. In this PR it's not used at all by shadow, and that will come as a future PR.

This TCP code is a first draft of the TCP code. It contains the states and can do the usual socket things (connect, listen, send, recv, etc), but it's not really TCP yet since it doesn't support retransmissions, send/recv windows, congestion control, etc. There are a lot of TODO comments, and a lot of things that are only meant to be temporary (like the send buffer). There are also tests which try to mimic how this TCP code would be used in Shadow, and test things like active/simultaneous open, active/passive close, listening/connecting/accepting, sending/receiving, etc. Some markdown notes are included for things that are too long to document in the code directly.


src/lib/tcp/src/tests/util/time.rs
```diff
@@ -0,0 +1,404 @@
...
+use crate::Dependencies;
+
+/// Like [`std::time::Instant`], but the time is arbitrary and is controlled by the
+/// [`Dependencies`].
+// time is internally represented in nanoseconds
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Instant(u64);
+
+/// Like [`std::time::Duration`], but the time is arbitrary and is controlled by the
+/// [`Dependencies`].
```
sporksmith, line 14: I don't follow why we can't use `std::time::Duration` here. Unlike `std::time::Instant` we *can* create arbitrary `Duration` values, so it's probably related to another reason that applies to both that I'm not following :)

src/lib/tcp/src/tests/util/time.rs
```diff
@@ -0,0 +1,404 @@
+//! Time-related types for our unit tests. The types represent simulated times.
+
+// this is only used in doc comments.
+#[allow(unused)]
+use crate::Dependencies;
+
+/// Like [`std::time::Instant`], but the time is arbitrary and is controlled by the
+/// [`Dependencies`].
```
sporksmith, line 8: I'm not sure what  "controlled by `Dependencies`" means. I can imagine needing this for the ability to create arbitrary values though, which `std::time::Instant` doesn't allow.

src/lib/tcp/src/util/mod.rs
```diff
@@ -0,0 +1,73 @@
+pub mod time;
+
+/// An array that derefs to a variable-length slice. Useful for storing variable-length data in a
+/// struct without allocating. The generic `N` must be a valid `u8`, so can only store up to
+/// `u8::MAX` items.
+#[derive(Copy, Clone, Debug)]
+pub struct SmallArrayBackedSlice<const N: usize, T> {
```
sporksmith, line 7: Maybe the existing [SmallVec](https://docs.rs/smallvec/latest/smallvec/struct.SmallVec.html) would work for our purposes?

src/lib/tcp/src/util/mod.rs
sporksmith, line 1: Maybe worth adding some unit tests for these?

sporksmith (approved):

sporksmith:
Very cool!

src/lib/tcp/src/lib.rs
```diff
@@ -0,0 +1,632 @@
...
+//   mutable references. It contains a non-public `TcpStateEnum`.
+// - `TcpStateEnum` — An enum of all individual TCP state types (ex: `ListeningState`,
+//   `EstablishedState`). It implements the `TcpStateTrait` trait, so its methods usually take owned
+//   objects and return owned objects.
+// - `TcpStateTrait` — A trait implemented by each individual TCP state type, as well as the
+//   `TcpStateEnum` enum that encapsulates all individual states. Its methods usually take owned
+//   state objects and return owned `TcpStateEnum` objects.
+
+// ignore dead code throughout
+#![allow(dead_code)]
```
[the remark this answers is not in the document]
stevenengler: It's only needed by the tests, so I moved it to `src/tests/mod.rs`. I also added a TODO comment to remove this later.

src/lib/tcp/src/lib.rs
```diff
@@ -0,0 +1,632 @@
...
+    /// should be updated during the [`finalize`](AcceptedTcpState::finalize) call (on the
+    /// [`AcceptedTcpState`] returned from [`accept`](TcpState::accept)) to run callbacks directly
+    /// on this state instead, and the callbacks should be given `TimerRegisteredBy::Parent` (the
+    /// child state has effectively become a parent state). This `Dependencies` object should also
+    /// make sure that all existing timers from before the state was accept()ed are also updated to
+    /// run callbacks directly on the state.
+    fn fork(&self) -> Self;
+}
+
+/// Specifies whether the callback is meant to run on the child state or a child state. For example
```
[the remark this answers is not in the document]
stevenengler: Fixed.

src/lib/tcp/src/lib.rs
```diff
@@ -0,0 +1,632 @@
...
+    X: Dependencies,
+    TcpStateEnum<X>: From<Self>,
+{
+    /// Start closing this socket. It may or may not close immediately depending on what state the
+    /// socket is currently in.
+    fn close(self) -> (TcpStateEnum<X>, Result<(), CloseError>) {
+        (self.into(), Err(CloseError::InvalidState))
+    }
+
+    // TODO: this shouldn't be publicly exposed
```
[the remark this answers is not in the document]
stevenengler: Good catch, this was actually fixed and I forgot to remove the comment. The `TcpStateTrait` used to be public, but isn't public anymore now that the `TcpState` wrapper exists, and `TcpState` doesn't expose this method so it's now private.

src/lib/tcp/src/lib.rs
```diff
@@ -0,0 +1,632 @@
...
+/// // as_impl!(as_init, Init, InitState);
+/// #[inline]
+/// pub fn as_init(&self) -> Option<&InitState<X>> {
+///     match self {
+///         Self::Init(x) => Some(x),
+///         _ => None,
+///     }
+/// }
+/// ```
+#[allow(unused_macros)]
```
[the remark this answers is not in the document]
stevenengler: Added.

src/lib/tcp/src/tests/util/time.rs
stevenengler: There isn't really a good reason to not use a `std::time::Duration` here in the tests. My main reason was that I think having a separate type makes it more difficult to accidentally use a `std::time::Duration` derived from a `std::time::Instant` where we would actually want a `std::time::Duration` derived from a `tests::util::Instant`. We would be using `std::time::Duration` to represent durations in two different reference frames. For example in a unit test it might be easy to accidentally create a `std::time::Duration` from a `std::time::Instant` and then add that to a `tests::util::Instant`.

But I changed this to use `std::time::Duration` instead since it's probably not worth having this extra code until we actually find that we need it.

src/lib/tcp/src/tests/util/time.rs
stevenengler: Yeah this wording is confusing. I removed the "controlled by Dependencies" and added some more comments.

src/lib/tcp/src/util/mod.rs
stevenengler: I do think we'll need `SmallVec` eventually in some places, but an advantage of `SmallArrayBackedSlice` (which could use a better name) is that it can't grow past the buffer size. For example a TCP packet can have up to 4 selective acknowledgements, so `SmallArrayBackedSlice` allows us to enforce that at the type level, while `SmallVec` has no upper limit on the size.

src/lib/tcp/src/util/mod.rs
stevenengler: Added some unit tests.

PR #3099, opened 2023-08-08
Migrate most bare clone tests to Rust

sporksmith:
This is in preparation to add additional clone tests as we extend clone functionality.
    
Since we're not setting up thread local storage in these tests, we can't safely call libc functions from the child threads. Using Rust lets use no_std alternatives that don't use thread local storage.

### DIFF
--- a/src/test/clone/test_clone.rs
+++ b/src/test/clone/test_clone.rs
@@ -1,0 +1,189 @@
+use core::ffi::c_void;
+use std::error::Error;
+use std::num::NonZeroI32;
+use std::sync::atomic::{self, AtomicU32};
+
+use linux_api::errno::Errno;
+use linux_api::ldt::linux_user_desc;
+use linux_api::sched::CloneFlags;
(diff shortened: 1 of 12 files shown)


src/test/clone/test_clone.rs
```diff
@@ -0,0 +1,187 @@
...
+            &mut tls,
+        )
+    };
+    assert!(child > 0);
+    THREAD_DONE_CHANNEL.receive().unwrap();
+    Ok(())
+}
+
+fn test_clone_clear_tid() -> Result<(), Box<dyn Error>> {
+    extern "C" fn thread_fn(_param: *mut c_void) -> i32 {
```
stevenengler, line 94: Should the thread function for each test be documented to say that can't call libc functions, as in the commit comment?

stevenengler (approved):

src/test/clone/test_clone.rs
sporksmith: Done

PR #3103, opened 2023-08-10
Add a basic fork test

sporksmith:
Doesn't pass in shadow yet.

This also includes some additions and tweaks to linux-api. Some of this ended up being unused for now; I intended to write an analogue of `wait_for_thread_exit` for a child process, but it appears that `kill(child_pid, 0)` "succeeds" (i.e. doesn't return `ESRCH`) if the child hasn't been reaped, even if no exit signal was specified. (I also don't want the first version of this test to use `waitpid`, since that implementation is a little further down the road)

### DIFF
--- a/src/lib/linux-api/src/posix_types.rs
+++ b/src/lib/linux-api/src/posix_types.rs
@@ -41,3 +41,17 @@ impl Pid {
         self.0
     }
 }
+
+#[cfg(feature = "rustix")]
+impl From<rustix::process::Pid> for Pid {
+    fn from(value: rustix::process::Pid) -> Self {
+        Pid(value.as_raw_nonzero())
+    }
+}
+
+#[cfg(feature = "rustix")]
+impl From<Pid> for rustix::process::Pid {
+    fn from(value: Pid) -> Self {
+        rustix::process::Pid::from_raw(value.as_raw_nonzero().into()).unwrap()
+    }
+}
--- a/src/lib/linux-api/src/posix_types.rs
+++ b/src/lib/linux-api/src/posix_types.rs
@@ -1,3 +1,5 @@
+use core::num::NonZeroI32;
+
 use crate::bindings;
 
 pub use bindings::linux___kernel_pid_t;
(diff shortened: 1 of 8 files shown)


src/lib/linux-api/src/posix_types.rs
```diff
@@ -15,3 +17,26 @@ pub type kernel_ulong_t = bindings::linux___kernel_ulong_t;
...
+        if pid < 0 {
+            None
+        } else {
+            Some(Self(NonZeroI32::new(pid).unwrap()))
+        }
+    }
+
+    pub fn as_raw(this: Option<Self>) -> kernel_pid_t {
+        this.map(|x| kernel_pid_t::from(x.0)).unwrap_or(0)
+    }
```
stevenengler, line 37: nit: Maybe a little confusing since the value will not necessarily be the same after a round trip (for example `Pid::as_raw(Pid::from_raw(-1)) == 0`). I think this should at least document that 0 is returned for `None`).

src/lib/linux-api/src/sched.rs
```diff
@@ -50,3 +53,63 @@ pub fn sched_yield() -> Result<(), Errno> {
...
+/// # Safety
+///
+/// Too many requirements to list here. See `clone(2)`.
+pub unsafe fn clone_raw(
+    flags: core::ffi::c_long,
+    stack: *mut core::ffi::c_void,
+    parent_tid: *mut kernel_pid_t,
+    child_tid: *mut kernel_pid_t,
+    tls: core::ffi::c_ulong,
+) -> Result<core::ffi::c_long, Errno> {
```
stevenengler, line 66: nit: maybe use the same types as the syscall [implementation](https://github.com/torvalds/linux/blob/2ccdd1b13c591d306f0401d98dedc4bdcd02b421/kernel/fork.c#L3039) (`unsigned long, unsigned long, int*, int*, unsigned long`)?

src/lib/linux-api/src/signal.rs
```diff
@@ -107,6 +107,10 @@ impl Signal {
         assert!(rv.0 > Self::STANDARD_MAX.0);
         rv
     }
+
+    pub fn as_raw(this: Option<Self>) -> i32 {
```
stevenengler, line 111: Document that `None` returns 0?

stevenengler (approved):

src/lib/linux-api/src/posix_types.rs
sporksmith: Done

src/lib/linux-api/src/sched.rs
sporksmith: done

src/lib/linux-api/src/signal.rs
sporksmith: Done

PR #3105, opened 2023-08-11
Add new TCP socket implementation

stevenengler:
This adds a new TCP socket using the tcp library from #3098. This has the same current limitations of the tcp library (no retransmission, congestion control, etc). This also doesn't support ioctl(), shutdown(), most socket options, and some other things.

This TCP socket can be enabled with the experimental `--use-new-tcp true` flag. A few tests have been added that use this flag to test the new TCP socket. Some other tests are close to passing, but need retransimssion support before they can be enabled.

Current passing socket tests:

- socket-new-tcp-shadow
- bind-new-tcp-shadow
- accept-new-tcp-shadow
- tcp-nonblocking-epoll-lossless-new-tcp-shadow
- tcp-nonblocking-select-lossless-new-tcp-shadow
- tcp-blocking-loopback-new-tcp-shadow
- tcp-nonblocking-poll-lossless-new-tcp-shadow
- tcp-nonblocking-poll-loopback-new-tcp-shadow
- tcp-nonblocking-epoll-loopback-new-tcp-shadow
- tcp-nonblocking-select-loopback-new-tcp-shadow
- tcp-blocking-lossless-new-tcp-shadow
- tcp-iov-lossless-new-tcp-shadow
- tcp-iov-loopback-new-tcp-shadow

### DIFF
--- a/src/main/host/descriptor/socket/inet/mod.rs
+++ b/src/main/host/descriptor/socket/inet/mod.rs
@@ -339,11 +339,16 @@ impl InetSocketRefMut<'_> {
         -> Result<(), SyscallError>
     );
 
-    pub fn accept(&mut self, cb_queue: &mut CallbackQueue) -> Result<OpenFile, SyscallError> {
+    pub fn accept(
+        &mut self,
+        net_ns: &NetworkNamespace,
+        rng: impl rand::Rng,
+        cb_queue: &mut CallbackQueue,
+    ) -> Result<OpenFile, SyscallError> {
         match self {
-            Self::LegacyTcp(socket) => socket.accept(cb_queue),
-            Self::Tcp(socket) => socket.accept(cb_queue),
-            Self::Udp(socket) => socket.accept(cb_queue),
+            Self::LegacyTcp(socket) => socket.accept(net_ns, rng, cb_queue),
+            Self::Tcp(socket) => socket.accept(net_ns, rng, cb_queue),
+            Self::Udp(socket) => socket.accept(net_ns, rng, cb_queue),
         }
     }
 
--- a/src/main/host/descriptor/socket/inet/mod.rs
+++ b/src/main/host/descriptor/socket/inet/mod.rs
@@ -22,14 +22,17 @@ use crate::utility::sockaddr::SockaddrStorage;
 use crate::utility::HostTreePointer;
 
 use self::legacy_tcp::LegacyTcpSocket;
+use self::tcp::TcpSocket;
 use self::udp::UdpSocket;
 
 pub mod legacy_tcp;
+pub mod tcp;
 pub mod udp;
 
 #[derive(Clone)]
 pub enum InetSocket {
     LegacyTcp(Arc<AtomicRefCell<LegacyTcpSocket>>),
+    Tcp(Arc<AtomicRefCell<TcpSocket>>),
     Udp(Arc<AtomicRefCell<UdpSocket>>),
 }
 
@@ -37,6 +40,7 @@ impl InetSocket {
     pub fn borrow(&self) -> InetSocketRef {
         match self {
             Self::LegacyTcp(ref f) => InetSocketRef::LegacyTcp(f.borrow()),
+            Self::Tcp(ref f) => InetSocketRef::Tcp(f.borrow()),
             Self::Udp(ref f) => InetSocketRef::Udp(f.borrow()),
         }
     }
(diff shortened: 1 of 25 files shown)


src/main/host/descriptor/socket/inet/tcp.rs
```diff
@@ -60,54 +101,226 @@ impl TcpSocket {
...
+        cb_queue: &mut CallbackQueue,
+        f: impl FnOnce(&mut tcp::TcpState<TcpDeps>) -> T,
+    ) -> T {
+        let rv = f(&mut self.tcp_state);
+
+        // we may have mutated the tcp state, so update the socket's file state and notify listeners
+
+        // if there are packets to send, notify the host
+        if self.tcp_state.wants_to_send() {
+            if let Some(socket) = self.socket_weak.upgrade() {
```
sporksmith, line 116: Maybe worth a comment about why this might fail and/or a log message?

src/main/host/descriptor/socket/inet/tcp.rs
```diff
@@ -168,32 +468,261 @@ impl TcpSocket {
...
+        }
+
+        // if not an inet socket address
+        let Some(peer_addr) = peer_addr.as_inet() else {
+            return Err(Errno::EINVAL.into());
+        };
+
+        let mut peer_addr: std::net::SocketAddrV4 = (*peer_addr).into();
+
+        // https://stackoverflow.com/a/22425796
```
sporksmith, line 579: A short comment about what info is expected to be at this URL would be helpful, e.g. in case the URL breaks in the future.

src/main/network/packet.rs
```diff
@@ -431,3 +556,54 @@ fn display_udp_bytes(packet: *const c::Packet, mut writer: impl Write) -> std::i
...
+        match flag {
+            tcp::TcpFlags::FIN => new_flags |= c::ProtocolTCPFlags_PTCP_FIN,
+            tcp::TcpFlags::SYN => new_flags |= c::ProtocolTCPFlags_PTCP_SYN,
+            tcp::TcpFlags::RST => new_flags |= c::ProtocolTCPFlags_PTCP_RST,
+            tcp::TcpFlags::PSH => panic!("Unsupported TCP flag: {flag:?}"),
+            tcp::TcpFlags::ACK => new_flags |= c::ProtocolTCPFlags_PTCP_ACK,
+            tcp::TcpFlags::URG => panic!("Unsupported TCP flag: {flag:?}"),
+            tcp::TcpFlags::ECE => panic!("Unsupported TCP flag: {flag:?}"),
+            tcp::TcpFlags::CWR => panic!("Unsupported TCP flag: {flag:?}"),
+            _ => unreachable!(
```
sporksmith, line 573: Maybe log the value in case this is somehow reached? e.g. maybe just `panic!("Invalid TCP flag: {flag:?}")`

sporksmith (approved):

src/main/host/descriptor/socket/inet/mod.rs
```diff
@@ -429,10 +429,13 @@ impl InetSocketWeak {
 /// non-zero port will be chosen. The final local address will be returned. If the peer address is
 /// unspecified and has a port of 0, the socket will receive packets from every peer address. The
 /// socket will be automatically disassociated when the returned [`AssociationHandle`] is dropped.
+/// If `check_less_specific` is true, the association will also fail if there is already a socket
```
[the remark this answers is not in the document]
stevenengler: Yeah I'm not super happy with this, but I don't really know a better way. I think ideally we'd get rid of the `is_interface_available` function altogether. I'm not sure what a better name is here, I think because this part of the code isn't really correct in the first place.

> I also wonder if there's a more descriptive name we could use for this parameter? Or if we could eliminate the bool parameter completely by having 2 different functions here (which might call an internal function with a bool parameter), or an enum?

We could make it two functions, but I don't know how to name those either. And these are functions that get passed through about 4 layers of things (inet module -> network namespace -> rust network interface -> C network interface) and I'm not sure where we'd make the 2 separate functions to 1 function with a bool distinction.

There's also a bigger problem where this part of the network interface is not working as intended anyway. This code in the network interface is supposed to prevent associating a wildcard (0.0.0.0) and non-wildcard peer address at the same time, but currently this check only works if you associate the wildcard address first, and then try associating the specific address. If you do it the other way around (associate a socket with the specific peer address and then associate a socket with a wildcard peer address), then the check doesn't work and you can incorrectly get two socket associations with the same local address when you shouldn't have.

So I don't want to make too many changes here, but if you think of a better name, or a nicer way to do this without restructuring too much of the code, let me know.

src/main/host/descriptor/socket/inet/tcp.rs
stevenengler: Added.

stevenengler: Changed this to a panic instead and added a comment.

src/main/host/descriptor/socket/inet/tcp.rs
stevenengler: Added.

src/main/network/packet.rs
stevenengler: Added.

PR #3108, opened 2023-08-15
Reorganize `networkinterface_isAssociated`

stevenengler:
I changed `networkinterface_isAssociated` to check only a single 4-tuple. This moves the complexity further up, and calling code can choose whether or not to also check the more-generic association (peer 0.0.0.0:0). This doesn't fix any bugs in the network interface code.

This is a follow up from https://github.com/shadow/shadow/pull/3105#discussion_r1293687008.


sporksmith (approved):
Cool, yeah this is an improvement :)

PR #3109, opened 2023-08-16
AtomicTlsMap extensions

sporksmith:
Extracted these from a larger local WIP PR implementing fork.

We need `AtomicTlsMap::clear` to reset TLS after forking. (After fork, the child is the only thread in a new process; the TLS from the parent process will be reinserted into the map).

Also changed it to use `RefCell` internally. I did it with the intent to add support for mutation, to make it easier to reset the thread-local IPC, thread shmem, and process shmem objects after forking. I'm reconsidering that approach since the changes in the next layer up are a bit complex, but think it's a worthwhile cleanup anyway.

FWIW the alternative I'm considering is to replace e.g. `ShimTlsVar<IPCData>` with `ShimTlsVar<RefCell<IPCData>>`. Since it's an extra layer of reference objects I'll probably need to replace `get` methods with `with` methods, which will lead to some awkwardness, but maybe preferable vs further complicating the TLS implementation itself.

### DIFF
--- a/src/lib/vasi-sync/src/atomic_tls_map.rs
+++ b/src/lib/vasi-sync/src/atomic_tls_map.rs
@@ -72,6 +72,10 @@ impl AtomicOptionNonZeroUsize {
         self.0.store(Self::to_usize(val), order)
     }
 
+    pub fn swap(&self, val: Option<NonZeroUsize>, order: atomic::Ordering) -> Option<NonZeroUsize> {
+        Self::from_usize(self.0.swap(Self::to_usize(val), order))
+    }
+
     pub fn compare_exchange(
         &self,
         current: Option<NonZeroUsize>,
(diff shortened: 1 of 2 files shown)


src/lib/vasi-sync/src/atomic_tls_map.rs
```diff
@@ -217,6 +221,29 @@ where
...
+                .is_some()
+            {
+                self.values[idx].get_mut().with(|value| {
+                    assert_eq!(self.refcounts[idx].get(), 0);
+                    // SAFETY: Caller has guaranteed that we effectively have exclusive access
+                    // to `self`.
+                    let value = unsafe { &mut *value };
+                    // SAFETY: The `some` value for the key indicates that this value
+                    // is initialized.
+                    unsafe { value.assume_init_drop() }
```
stevenengler: The `AtomicTlsMap` doesn't provide any synchronization for writes since it's assumed that an entry is only accessed by a single thread. So I don't see how we can drop the values here in a different thread without synchronization.

stevenengler (changes requested):

src/lib/vasi-sync/src/atomic_tls_map.rs
sporksmith: Good catch.

I have an alternate idea; will probably make a new PR

sporksmith:
Obsoleted by https://github.com/shadow/shadow/pull/3113

PR #3112, opened 2023-08-18
Implement fork

sporksmith:


src/main/host/process.rs
```diff
@@ -510,6 +528,71 @@ impl RunnableProcess {
...
+        new_tgl: RootedRc<RootedRefCell<Thread>>,
+    ) -> RootedRc<RootedRefCell<Process>> {
+        let new_tgl_tid;
+        let native_pid;
+        {
+            let new_tgl = new_tgl.borrow(host.root());
+            new_tgl_tid = new_tgl.id();
+            native_pid = new_tgl.native_pid();
+        }
+        let pid = ProcessId::from_thread_group_leader_tid(new_tgl_tid);
```
stevenengler, line 543: Does the thread itself keep track of if it's a thread group leader or not? Could we assert here that the tid does refer to a thread group leader so that we know `ProcessId::from_thread_group_leader_tid` is a valid operation?

src/main/host/process.rs
```diff
@@ -510,6 +528,71 @@ impl RunnableProcess {
...
+            &host.shim_shmem_lock_borrow().unwrap().root,
+            host.shim_shmem().serialize(),
+            host.id(),
+            strace_logging.as_ref().map(|x| x.file.borrow().as_raw_fd()),
+        );
+        let shim_shared_mem_block = shadow_shmem::allocator::shmalloc(shim_shared_mem);
+
+        let runnable_process = RunnableProcess {
+            common,
+            expected_final_state: None,
```
stevenengler, line 579: What does it mean to have an expected final state of `None`?

stevenengler, line 579: Maybe update the documentation to explain that `expected_final_state` only applies to the process in the config, and not any forked processes? And explain how the final state of forked processes is handled? (Is it ignored?)

src/main/host/managed_thread.rs
```diff
@@ -337,12 +327,41 @@ impl ManagedThread {
             other => panic!("Unexpected result from shim: {other:?}"),
         };
 
+        let native_pid = if flags & CloneFlags::CLONE_THREAD.bits() == 0 {
+            nix::unistd::Pid::from_raw(child_native_tid)
+        } else {
+            self.native_pid
+        };
+
+        if flags & CloneFlags::CLONE_THREAD.bits() == 0 {
```
stevenengler, line 348: Convert flags to `CloneFlags` and use `flags.contains(CloneFlags::CLONE_THREAD)`? (And above)

stevenengler (approved):

src/main/host/process.rs
sporksmith: It means that Shadow won't enforce any particular final state. (Already documented at the field definition)

PR #3113, opened 2023-08-18
Add AtomicTlsMap::forget_all

sporksmith:
Intended for use after fork; we can't safely drop values belonging to other threads, but can safely forget about and overwrite them.

### DIFF
--- a/src/lib/vasi-sync/src/atomic_tls_map.rs
+++ b/src/lib/vasi-sync/src/atomic_tls_map.rs
@@ -217,6 +217,28 @@ where
         self.keys[idx].store(None, atomic::Ordering::Release);
         Some(value)
     }
+
+    /// Resets metadata in the map to mark all entries vacant, without dropping
+    /// the values.
+    ///
+    /// Intended for use after `fork`, after which entries belonging to other threads
+    /// are not guaranteed to be in any consistent state (so can't be dropped), but
+    /// the threads owning those entries no longer exist in the child, so they *can*
+    /// be safely overwritten.
+    ///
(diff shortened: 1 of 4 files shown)


src/lib/vasi-sync/src/atomic_tls_map.rs
```diff
@@ -217,6 +217,27 @@ where
...
+    /// the threads owning those entries no longer exist in the child, so they *can*
+    /// be safely overwritten.
+    ///
+    /// # Safety
+    ///
+    /// Any outstanding references from `self` (e.g. obtained via Self::get)
+    /// must not be accessed again. e.g. references held by other threads before
+    /// `fork` are OK, since those threads do not exist in the current process,
+    /// and so will not access the child's copy of this table.
+    pub unsafe fn forget_all(&self) {
```
stevenengler, line 235: I think it should be required that not only must you not access any existing references, but any references also must not exist. Both for safety purposes (don't want dangling references), but also since existing `Ref` objects will still decrement the reference count when dropped, which would be bad when we've reset the reference count to 0. After a fork, those `Ref` objects were essentially `std::mem::forget` forgotten, which I think would satisfy this "any references also must not exist" condition.

stevenengler (approved):

src/lib/vasi-sync/src/atomic_tls_map.rs
sporksmith: Clarified that any outstanding references also can't be dropped, but "forgotten" references (whether by mem::forget or fork) are ok.

PR #3116, opened 2023-08-21
cargo clippy --fix

sporksmith:
`cargo clippy --fix` can be helpful for fixing clippy warnings. There are currently a lot of things in our code it changes though, making it awkward to use in a PR since it'll change code unrelated to the PR.

I'm not sure why, but `cargo clippy --fix` warns about and fixes things that just `cargo clippy` doesn't complain about. Maybe it has a different default policy set.

This PR includes some manual fixes as well for things that `clippy --fix` fixes suboptimally, or complains about without fixing.

### DIFF
--- a/src/lib/shim/src/tls.rs
+++ b/src/lib/shim/src/tls.rs
@@ -815,7 +815,7 @@ mod test {
                         my_i16.get().store(i as i16, atomic::Ordering::Relaxed);
 
                         assert_eq!(my_i16.get().load(atomic::Ordering::Relaxed), i as i16);
-                        assert_eq!(my_i32.get().load(atomic::Ordering::Relaxed), { i });
+                        assert_eq!(my_i32.get().load(atomic::Ordering::Relaxed), i);
                         assert_eq!(my_i8.get().load(atomic::Ordering::Relaxed), i as i8);
                         unsafe { tls.unregister_current_thread() };
                     })
--- a/src/lib/shim/src/tls.rs
+++ b/src/lib/shim/src/tls.rs
@@ -810,12 +810,12 @@ mod test {
                         assert_eq!(my_i16.get().load(atomic::Ordering::Relaxed), 0);
 
                         // Order shouldn't matter here, but change it from above anyway.
-                        my_i32.get().store(i as i32, atomic::Ordering::Relaxed);
+                        my_i32.get().store(i, atomic::Ordering::Relaxed);
                         my_i8.get().store(i as i8, atomic::Ordering::Relaxed);
                         my_i16.get().store(i as i16, atomic::Ordering::Relaxed);
 
                         assert_eq!(my_i16.get().load(atomic::Ordering::Relaxed), i as i16);
-                        assert_eq!(my_i32.get().load(atomic::Ordering::Relaxed), i as i32);
+                        assert_eq!(my_i32.get().load(atomic::Ordering::Relaxed), { i });
                         assert_eq!(my_i8.get().load(atomic::Ordering::Relaxed), i as i8);
                         unsafe { tls.unregister_current_thread() };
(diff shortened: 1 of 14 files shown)


src/lib/shim/src/tls.rs
```diff
@@ -742,12 +742,12 @@ mod test {
...
 
                         // Order shouldn't matter here, but change it from above anyway.
-                        my_i32.get().store(i as i32, atomic::Ordering::Relaxed);
+                        my_i32.get().store(i, atomic::Ordering::Relaxed);
                         my_i8.get().store(i as i8, atomic::Ordering::Relaxed);
                         my_i16.get().store(i as i16, atomic::Ordering::Relaxed);
 
                         assert_eq!(my_i16.get().load(atomic::Ordering::Relaxed), i as i16);
-                        assert_eq!(my_i32.get().load(atomic::Ordering::Relaxed), i as i32);
+                        assert_eq!(my_i32.get().load(atomic::Ordering::Relaxed), { i });
```
stevenengler, line 750: why the braces?

src/lib/vasi-sync/tests/scchannel-tests.rs
```diff
@@ -31,9 +31,7 @@ mod scchannel_tests {
                     channel.send(42);
                 })
             };
-            let reader = {
-                sync::thread::spawn(move || channel.receive())
-            };
+            let reader = { sync::thread::spawn(move || channel.receive()) };
```
stevenengler: I think we can remove the braces?

src/lib/vasi-sync/tests/scchannel-tests.rs
```diff
@@ -98,9 +96,7 @@ mod scchannel_tests {
             };
             // Simulate a separate watchdog thread that detects that the writer process
             // has exited, and closes the channel in parallel with the other operations.
-            let watchdog = {
-                sync::thread::spawn(move || channel.close_writer())
-            };
+            let watchdog = { sync::thread::spawn(move || channel.close_writer()) };
```
stevenengler: I think we can remove the braces?

src/main/network/router/codel_queue.rs
```diff
@@ -406,32 +406,29 @@ mod tests {
         // Not above target so interval is not set.
         assert!(cdq.interval_end.is_none());
         let now = start + TARGET - one;
-        assert_eq!(cdq.process_standing_delay(&now, TARGET - one), false);
+        assert!(!cdq.process_standing_delay(&now, TARGET - one));
         assert!(cdq.interval_end.is_none());
```
stevenengler, line 410: iirc I think this used the true/false for readability. (It's easier to see the difference between `assert_eq(long_string, true)` and `assert_eq(long_string, false)` compared to `assert(long_string)` and `assert(!long_string)`.) My opinion leans slightly towards the true/false, but it doesn't really matter to me.

stevenengler (approved):

src/main/network/router/codel_queue.rs
sporksmith: Ah right. I don't have a preference of one vs the other, so lean towards it not being worth fighting clippy/convention here

src/lib/shim/src/tls.rs
sporksmith: Removed

src/lib/vasi-sync/tests/scchannel-tests.rs
sporksmith: Removed

src/lib/vasi-sync/tests/scchannel-tests.rs
sporksmith: Removed

PR #3118, opened 2023-08-21
Fix clippy warnings in rust 1.72 beta

stevenengler:


stevenengler:
```
error: unknown lint: `clippy::tuple_array_conversions`
```

It looks like maybe we'll have to wait until rust 1.72 is released in a few days before merging.

sporksmith (approved):

stevenengler:
It turns out that I was using an older beta, and the most recent beta doesn't check `tuple_array_conversions` by default. I've dropped the commit from this PR that ignores the `tuple_array_conversions` lint.

Issue #2989, opened 2023-06-01
Implement wait4 (syscall for waitpid, wait)


sporksmith:
Required for https://github.com/shadow/shadow/issues/1987

sporksmith:
Some notes on the status integer set by `wait4` etc.

Theoretically this is part of the kernel ABI, but I haven't been able to find anything referencing this integer in the kernel user-space headers or ABI docs.

[glibc's waitstatus.h](https://github.com/bminor/glibc/blob/4290aed05135ae4c0272006442d147f2155e70d7/bits/waitstatus.h#L55) has macros for examining the status integer with decent comments, and macros for *constructing* this status. Not as authoritative as the kernel code, but a pretty useful hint:

```
/* Macros for constructing status values.  */
#define	__W_EXITCODE(ret, sig)	((ret) << 8 | (sig))
#define	__W_STOPCODE(sig)	((sig) << 8 | 0x7f)
#define __W_CONTINUED		0xffff
#define	__WCOREFLAG		0x80
```

The kernel code for constructing the status is a bit scattered around. e.g. I think this is the line where it's set for a normal process exit, and looks consistent with above: https://github.com/torvalds/linux/blob/89bf6209cad66214d3774dac86b6bbf2aec6a30d/kernel/exit.c#L1313

```
wo->wo_stat = (exit_code << 8) | 0x7f;
```

and for signal death: https://github.com/torvalds/linux/blob/89bf6209cad66214d3774dac86b6bbf2aec6a30d/kernel/exit.c#L1193

```
	status = (p->signal->flags & SIGNAL_GROUP_EXIT)
		? p->signal->group_exit_code : p->exit_code;
	wo->wo_stat = status;
```

PR #3119, opened 2023-08-21
Implement parent pid and related functionality

sporksmith:
Progress on #2989


src/main/host/syscall/handler/unistd.rs
```diff
@@ -478,4 +478,11 @@ impl SyscallHandler {
             }
         }
     }
+
+    #[log_syscall(/* rv */ linux_api::posix_types::kernel_pid_t)]
+    pub fn getppid(
+        ctx: &mut SyscallContext,
+    ) -> Result<linux_api::posix_types::kernel_pid_t, SyscallError> {
+        Ok(ctx.objs.process.parent_id().into())
+    }
```
stevenengler, line 487: Maybe add a small test in `src/test/unistd/test_unistd.rs`? There's a test for `getpid`, so could probably just copy-paste.

stevenengler (approved):

src/main/host/syscall/handler/unistd.rs
sporksmith: Done

PR #3122, opened 2023-08-22
Add process group IDs and session IDs

sporksmith:
Motivated mostly because the `wait4` syscall supports waiting on a process group. I think shells tend to manipulate process groups as well (e.g. putting processes in a pipeline within a single process group), so we'll want it for that too.

Not sure how important implementing the session IDs is, but it was relatively easy to do while I was already reading about how process groups work; Every process belongs to a single process group, and every process group belongs to a single session.

Also fixed some handling for process cleanup that I came across while testing the new syscalls - now that processes can have parents other than shadow, we need to defer freeing zombie processes until their parent has reaped them or died.

Progress on #2989


src/main/host/syscall/handler/unistd.rs
```diff
@@ -480,9 +481,101 @@ impl SyscallHandler {
...
+        let _process_borrow;
+        let process: &Process;
+        if pid == 0 || pid == kernel_pid_t::from(ctx.objs.process.id()) {
+            _processrc_borrow = None;
+            _process_borrow = None;
+            process = ctx.objs.process;
+        } else {
+            let pid = ProcessId::try_from(pid).map_err(|_| Errno::EINVAL)?;
+            let Some(pbrc) = ctx.objs.host.process_borrow(pid) else {
+                return Err(Errno::ESRCH.into())
```
stevenengler, line 525: I'm surprised this compiles without a semicolon.

src/main/host/host.rs
```diff
@@ -458,18 +458,53 @@ impl Host {
...
+                if process.parent_id() != pid {
+                    // Not a child of the current process
+                    false
+                } else {
+                    process.set_parent_id(ProcessId::INIT);
+                    let is_zombie = process.borrow_zombie().is_some();
+                    is_zombie
+                }
+            })
+            .map(|(pid, _process)| *pid)
```
stevenengler, line 493: nit: `filter_map` instead? (I think could replace `false` with `None` and `is_zombie` with `is_zombie.then_some(pid)`.)

src/test/clone/test_fork.rs
```diff
@@ -116,6 +116,110 @@ fn test_clone_parent(set_clone_parent: bool) -> Result<(), Box<dyn Error>> {
...
+
+            assert_eq!(rustix::io::write(&writer, &[0]), Ok(1));
+            unsafe { libc::exit(0) };
+        }
+        CloneResult::CallerIsParent(child_pid) => child_pid,
+    };
+
+    // Close our copy of the writer-end of the pipe, so that parent hang trying
+    // to read from the pipe if the child exited abnormally.
+    drop(writer);
```
stevenengler, line 154: I don't follow this. By closing this process' writer, doesn't this make this process' `read` *not* hang if the child exited abnormally? I'm not sure if just the comment is wrong here. If there are no writers left, then a `read(reader, &buf, 1)` should return 0.

src/test/clone/test_fork.rs
```diff
@@ -116,6 +116,110 @@ fn test_clone_parent(set_clone_parent: bool) -> Result<(), Box<dyn Error>> {
...
```
stevenengler, line 207: same as above

stevenengler (approved):

PR #3125, opened 2023-08-23
Convert Epoll descriptor and syscall handlers to Rust

robgjansen:
This PR is the initial migration of epoll to Rust. I'm explicitly deciding to keep the following next steps for future PRs to keep this one smaller:
  - The old C epoll object is still currently still used in the C syscall handler, where it is only used in the `poll` syscall handlers. The `poll` handlers are relatively simple, so I think it would be easier to convert those to Rust and use the Rust epoll rather than trying to add a C API to the Rust epoll descriptor object.
    - Once `poll` is converted, we can remove the `epoll.c/.h` descriptor object.
  - Some of our epoll integration tests in `src/test/epoll` are
    - written in C and could be converted to Rust,
    - incomplete and could be expanded.


robgjansen:
@stevenengler For the failed tgen test, only this one of the 40 tests are failing: `tgen-duration-1mbit_300ms-1000streams-shadow` with the result:
```
1mbit_300ms net expected 106250 got client=105188 server=117284
```

In other words, this is the configuration with the least bandwidth and the longest latency with the most number of parallel streams.

Our test is defined as:
```
#   1Mbit/s = 125000 bytes/s
#   125000 bytes/s * 0.85 = 106250 bytes/s
if [[ "${network}" == "1mbit"* ]]; then
    expected=106250
...
```

in other words our result (105188) is meeting an 84% threshold (105000) but not an 85% threshold (106250). Do we think this is a legitimate problem with the PR, or does this just the simulation changed and the performance threshold is noisy? (I seem to remember you had similar issues a while back?)

stevenengler:
> Do we think this is a legitimate problem with the PR, or does this just the simulation changed and the performance threshold is noisy?

I'm not sure. We've lowered the threshold a few times in the past, so I don't see an issue with lowering it a bit again.

It might be good to run a tor benchmark and see what that looks like first.

robgjansen:
I ran a Tor benchmark. I think performance looks OK, but there is a large increase in memory usage that we'll have to figure out.



See all graphs here: https://github.com/shadow/benchmark-results/tree/master/tor/2023-08-25-T21-09-03

I think a code review would help make further progress so I'll request that now. (I suggest just looking at the collective changes rather than commit-by-commit since I ended up changing some of the same code multiple times and didn't bother cleaning up.)

robgjansen:
Oh I think the problem with the memory usage is that the new epoll is not automatically deleting an entry when a File is closed. I think we can do that by removing from the `epoll->monitoring` map when we notice the `FileState::Closed` flag getting set on the file.

robgjansen:
Here is the close behavior in our current C impl of epoll, inside of the listener notify callback function

https://github.com/shadow/shadow/blob/03bf478c5b2b2dc177cc7225ab24a28cea70b68c/src/main/host/descriptor/epoll.c#L761-L767

robgjansen:
Adding the fix for removing closed Files from the monitor map seems to have fixed the memory usage problem:

ram_simtime

All benchmark graphs here:
https://github.com/shadow/benchmark-results/tree/master/tor/2023-08-26-T20-00-46

src/main/host/syscall/handler/epoll.rs
```diff
@@ -0,0 +1,423 @@
...
+    events_ptr: ForeignPtr<linux_api::epoll::epoll_event>,
+) -> Result<(), SyscallError> {
+    let events_ptr = ForeignArrayPtr::new(events_ptr, ready.len());
+    let mut mem_ref = mem.memory_ref_mut(events_ptr)?;
+
+    for (i, (ev, data)) in ready.iter().enumerate() {
+        let plugin_ev = &mut mem_ref.deref_mut()[i];
+        plugin_ev.events = ev.bits();
+        plugin_ev.data = *data;
+    }
```
stevenengler, line 418: nit: what about something like the following to avoid the array indexing and possible panic?

```rust
for ((ev, data), plugin_ev) in ready.iter().zip(mem_ref.deref_mut().iter_mut())
```

stevenengler (changes requested):

src/main/host/syscall/handler/epoll.rs
```diff
@@ -0,0 +1,423 @@
...
+        let (events, data) = if op == EpollCtlOp::EPOLL_CTL_DEL {
+            // epoll_ctl(2): Since Linux 2.6.9, the event pointer is ignored and can be specified as
+            // NULL when using EPOLL_CTL_DEL.
+            (EpollEvents::empty(), 0)
+        } else {
+            let mem = ctx.objs.process.memory_borrow();
+            let ev = mem.read(event_ptr)?;
+
+            let Some(mut events) = EpollEvents::from_bits(ev.events) else {
+                log::debug!("Invalid epoll_ctl events: {}", { ev.events });
```
stevenengler, line 154: nit: the braces can be removed

robgjansen: I'll add a comment to the code to reference `rustc --explain E0793`.

stevenengler: Ah I never realized this is a packed struct on x86-64.

src/main/host/descriptor/epoll/key.rs
```diff
@@ -56,7 +56,7 @@ impl Eq for PriorityKey {}
 
 impl PartialEq for PriorityKey {
     fn eq(&self, other: &Self) -> bool {
-        self.key.eq(&other.key)
+        self.pri.eq(&other.pri)
```
stevenengler: Add a comment abut why you only need to compare by the priority and not the key? I think it's because no two `PriorityKey`s should ever have the same priorities? Maybe add this restriction to the docs for `PriorityKey::new` as well?

stevenengler (approved):

src/test/tgen/fixed_duration/verify.sh
```diff
@@ -29,23 +29,28 @@ client_Bps_read="$(grep -r --include="tgen.*.stdout" "driver-heartbeat" ./hosts/
...
+# overhead includes non-payload bytes from packet headers and bytes from retransmitted packets. The
+# latter tends to increase on very slow congested networks due to retransmission, meaning that TCP
+# is less efficient when retransmitting many packets. Thus, our validation test progressively
+# requires higher efficiency on faster networks where we expect fewer retransmissions.
 expected=0
 if [[ "${network}" == "1mbit"* ]]; then
-    expected=106250
+    # 1 Mbit/s = 125000 bytes/s
+    # 125000 bytes/s * 0.84 = 105000 bytes/s
+    expected=105000
```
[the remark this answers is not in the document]
robgjansen: No luck on raising the limit.

```
** Shadow completed successfully
TGen configuration: network=1mbit_300ms client=1stream
Server bytes per second=119740
Client bytes per second=88716
Verification failed: 1mbit_300ms net expected 105000 got client=88716 server=119740
<end of output>
```

So for now we'll keep the 70% threshold (`# 125000 bytes/s * 0.7 = 87500 bytes/s`)

robgjansen:
Latest Tor benchmark still looks reasonable
https://github.com/shadow/benchmark-results/tree/master/tor/2023-09-26-T00-22-03

PR #3126, opened 2023-08-24
Improve process exit handling

sporksmith:
Have child processes signal their parents on death.

Add additional checks for when a dead process should be kept around as a zombie (e.g. *not* when the parent process created the child with no exit signal, or has themselves blocked the child exit signal).

### DIFF
--- a/src/main/core/support/configuration.rs
+++ b/src/main/core/support/configuration.rs
@@ -1036,12 +1036,24 @@ impl<'de> serde::Deserialize<'de> for ProcessArgs {
     }
 }
 
+// TODO: use linux_api's Signal internally, which we control and which supports
+// realtime signals. We need to implement conversion to and from strings to do
+// so, while being careful that the conversion is compatible with nix's so as
+// not to be a breaking change to our configuration format.
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub struct Signal(nix::sys::signal::Signal);
 
 impl From<nix::sys::signal::Signal> for Signal {
     fn from(value: nix::sys::signal::Signal) -> Self {
         Self(value)
+    }
+}
+
+impl TryFrom<linux_api::signal::Signal> for Signal {
+    type Error = <nix::sys::signal::Signal as TryFrom<i32>>::Error;
+    fn try_from(value: linux_api::signal::Signal) -> Result<Self, Self::Error> {
+        let signal = nix::sys::signal::Signal::try_from(value.as_i32())?;
+        Ok(Self(signal))
     }
 }
 
--- a/src/main/host/process.rs
+++ b/src/main/host/process.rs
@@ -154,6 +154,9 @@ struct Common {
     // Session id, as returned e.g. by `getsid`.
     session_id: Cell<ProcessId>,
 
+    // Signal to send to parent on death.
+    exit_signal: Option<Signal>,
+
     // unique id of the program that this process should run
     name: CString,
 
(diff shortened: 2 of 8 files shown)


src/main/host/process.rs
```diff
@@ -111,7 +111,7 @@ impl From<ThreadId> for ProcessId {
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub enum ExitStatus {
     Normal(i32),
-    Signaled(nix::sys::signal::Signal),
+    Signaled(Signal),
```
stevenengler, line 114: Do we show this value to the user? For example if the process fails with an unexpected exit state? If so, should we add a nicer debug impl to the linux-api Signal to show the signal name and not just the signal number?

src/test/clone/test_fork.rs
```diff
@@ -220,6 +222,168 @@ fn test_child_change_group() -> Result<(), Box<dyn Error>> {
...
+            f();
+            assert_eq!(rustix::io::write(&writer, &[0]), Ok(1));
+            unsafe { libc::exit(0) };
+        }
+        CloneResult::CallerIsParent(child_pid) => child_pid,
+    };
+
+    // Close our copy of the writer-end of the pipe, so that parent hang trying
+    // to read from the pipe if the child exited abnormally.
+    drop(writer);
```
stevenengler, line 251: Same as comment in other PR.

src/test/clone/test_fork.rs
```diff
@@ -220,6 +222,168 @@ fn test_child_change_group() -> Result<(), Box<dyn Error>> {
...
+    ensure_ord!(rustix::io::read(&reader, &mut exit_code), ==, Ok(1));
+    ensure_ord!(exit_code[0], ==, 0);
+
+    Ok(())
+}
+
+fn test_exit_signal_normal_exit(exit_signal: nix::sys::signal::Signal) -> anyhow::Result<()> {
+    run_test_in_subprocess(|| {
+        static mut SIGNO: c_int = 0;
+        static mut INFO: MaybeUninit<siginfo_t> = MaybeUninit::uninit();
```
stevenengler, line 266: Why not `Option` to avoid any chance of UB (I guess other than the UB inherent in `static mut` if the signal handler writes to it while the test is in the middle of reading it)?

stevenengler (approved):

src/main/core/support/configuration.rs
```diff
@@ -1045,6 +1045,13 @@ impl From<nix::sys::signal::Signal> for Signal {
     }
 }
 
+impl From<linux_api::signal::Signal> for Signal {
+    fn from(value: linux_api::signal::Signal) -> Self {
+        let signal = nix::sys::signal::Signal::try_from(value.as_i32()).unwrap();
+        Self(signal)
+    }
+}
```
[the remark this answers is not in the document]
sporksmith: Done, though it just moves the unwrap up to the one point where it's used :). Probably makes more sense there, though.

We currently only use this to convert from a signal that a process actually died of, into the config format for display. The conversion currently can't fail since we don't support sending realtime signals in the first place.

It's probably be better to change the config format to use linux_api's Signal, but then we'd need to implement conversions to/from string while being careful to preserve compatibility. Added a comment.

src/main/host/process.rs
sporksmith: We don't. We convert to the configuration's `ProcessFinalState` and display that, so that the actual final state is displayed in a way that's consistent with how the expected final state is specified.

src/test/clone/test_fork.rs
sporksmith: Done

src/test/clone/test_fork.rs
sporksmith: Done

PR #3127, opened 2023-08-24
run_tests workflow: add sccache

sporksmith:
* Set up and use sccache in the "run_tests" workflow. This causes it to be about as fast as the "run_tests_incremental" workflow. It brings the total run time from [2.5h](https://github.com/shadow/shadow/actions/runs/5967281286/usage) to about [1.5h](https://github.com/shadow/shadow/actions/runs/5968070092/usage)
* Delete the "run_tests_incremental" workflow. It no longer has a speed advantage over the "run_tests" workflow, required periodically generating new pre-baked Docker images (which currently involved manually running the workflow), and is conceptually more complicated.
* Disable rust incremental builds in the extra_tests workflow. I think they may be causing the cache to be incompatible with the cache of the same system config in the "run_tests" workflow, and incremental builds are unhelpful overhead in this case anyway.


sporksmith:
@robgjansen you might also want to take a look to verify this makes sense for "maintenance mode"

the sccaches are around 400 MB each, and I think the account limit is 10 GB, so you might occasionally need to manually delete some caches if too many new versions end up getting saved before they get GC'd

sporksmith:
10 GB limit: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy

robgjansen:
Hmm it looks like we're already over our cache limit and for now Github is doing the smart thing:

Screenshot 2023-08-24 at 7 24 16 PM

So I think we're a 👍  for maintenance mode. :)

.github/workflows/extra_tests.yml
```diff
@@ -38,7 +38,9 @@ jobs:
       # sccache doesn't work on incrementally compiled crates, and incremental
       # compilation doesn't help us here anyway.
       CARGO_INCREMENTAL: false
-      # SCCACHE_RECACHE: 1 # Uncomment this to clear cache, then comment it back out
+      # Increment to force eviction of previous caches.
+      # Keep in sync with same variable in `run_tests.yml`.
+      SCCACHE_CACHE_VERSION: 1
```
stevenengler, line 43: Add a comment about when you might want to do this?

stevenengler (approved):

.github/workflows/run_tests.yml
```diff
@@ -96,6 +108,39 @@ jobs:
...
+          # TODO: Consider whether we can safely drop e.g. CC, CONTAINER, and/or BUILDTYPE,
+          # merging those caches.
+          #
+          # sccache won't cache crates that "invoke the system linker". If that means it
+          # won't cache crates that link with platform libraries, then I think we can merge
+          # the caches. I'm not sure whether or not it does, though.
+          # https://github.com/mozilla/sccache#rust
+          key: sccache-${{ steps.get-month.outputs.month }}-${{ steps.get-rustv.outputs.rustv }}-${{ env.CONTAINER }}-${{ env.CC }}-${{ env.BUILDTYPE }}-${{ hashFiles('src/Cargo.lock') }}
+          restore-keys: |
+                        sccache-${{ steps.get-month.outputs.month }}-${{ steps.get-rustv.outputs.rustv }}-${{ env.CONTAINER }}-${{ env.CC }}-${{ env.BUILDTYPE }}-
```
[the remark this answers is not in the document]
sporksmith: done

.github/workflows/extra_tests.yml
sporksmith: done

PR #3130, opened 2023-08-25
Add support for Debian Bookworm

sporksmith:


sporksmith:
Oops jumped the gun, apparently. A previous iteration passed the tests; not sure how it's segfaulting now. Investigating.

sporksmith:
Fixed

ci/container_scripts/install_deps.sh
```diff
@@ -12,9 +12,10 @@ APT_PACKAGES=(
   libglib2.0-0
   libglib2.0-dev
   make
+  netbase
```
stevenengler, line 15: Add to the installation docs required dependencies?

src/test/resolver/test_getaddrinfo.c
```diff
@@ -168,6 +168,7 @@ bool addrinfo_equals(const struct addrinfo* lhs, const struct addrinfo* rhs) {
                    getaddrinfo_rv_string(buf2_##__LINE__, rv##__LINE__),                           \
                    (rv##__LINE__ == EAI_SYSTEM) ? strerror(errno) : "N/A");                        \
             g_test_fail();                                                                         \
+            return;                                                                                \
```
stevenengler, line 171: should this also be added to the `assert_addrinfo_equals` above?

stevenengler (approved):

ci/container_scripts/install_deps.sh
sporksmith: Done

src/test/resolver/test_getaddrinfo.c
sporksmith: We need to return from this one because it indicates that the output pointer is invalid, so the rest of the test can't run. The same isn't necessarily true for `assert_addrinfo_equals`.

Added a comment.

robgjansen:
FYI, I just added `debian:12-slim` as a required test in the repo settings.

PR #3132, opened 2023-08-26
Add support for sequence numbers, acks, and windows to new tcp sockets

stevenengler:
Not everything added is handled correctly, but it's an improvement. The buffers will probably need to change more and the current versions are inefficient, but good enough for now. There might be a nicer way to do the range intersection check.


src/lib/tcp/src/seq.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    #[inline]
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Returns `true` if the sequence number is contained within this half-open range.
+    #[inline]
+    pub fn contains(&self, seq: Seq) -> bool {
+        seq - self.start < self.end - self.start
```
sporksmith, line 115: Took me a little to see how this works. Optional, maybe clearer as:

```
seq - self.start < self.len()
```
or
```
SeqRange::new(self.start, seq).len() < self.len()
```

src/lib/tcp/src/seq.rs
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    /// Returns the intersecting range if there is a single intersecting range. Returns `None` if
+    /// there is no intersection, or if there are two intersections. A range will always intersect
+    /// with itself, which also holds for empty ranges. An empty range will intersect with a
+    /// non-empty range if the empty range is contained within the non-empty range.
+    pub fn intersection(&self, other: &Self) -> Option<SeqRange> {
```
sporksmith, line 122: Optional - since we are using a custom range type here, I wonder if it'd be simpler to disallow empty intervals. Maybe by making it a closed interval instead of half-open.

sporksmith (approved):
:tada:

src/lib/tcp/src/seq.rs
stevenengler: I did think about this initially, but empty ranges end up being commonplace in the tcp code, and I found it simpler just to handle empty and non-empty ranges the same rather than have different code paths to handle the two cases. Using a closed interval also adds other questions like what to do for `[0, u32::MAX].len()`, which would be `u32::MAX + 1` which can't be represented as a u32.

src/lib/tcp/src/seq.rs
stevenengler: Added the last version.

PR #3134, opened 2023-08-28
Fix bug in rust tcp sequence number

stevenengler:
I missed this before. It was previously skipping sequence number ranges and somehow still worked.

### DIFF
--- a/src/lib/tcp/src/connection.rs
+++ b/src/lib/tcp/src/connection.rs
@@ -240,7 +240,16 @@ impl<I: Instant> Connection<I> {
                 #[cfg(debug_assertions)]
                 debug_assert!(wants_to_send);
 
-                break 'packet (self.send.buffer.next_seq(), TcpFlags::empty(), Bytes::new());
+                // use the sequence number of the next unsent message if we have one buffered,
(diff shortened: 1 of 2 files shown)


src/lib/tcp/src/connection.rs
```diff
@@ -240,7 +240,16 @@ impl<I: Instant> Connection<I> {
                 #[cfg(debug_assertions)]
                 debug_assert!(wants_to_send);
 
-                break 'packet (self.send.buffer.next_seq(), TcpFlags::empty(), Bytes::new());
+                // use the sequence number of the next unsent message if we have one buffered,
```
sporksmith, line 243: Optional - maybe it's worth writing additional unit/regression tests that'd catch this?

sporksmith (approved):

src/lib/tcp/src/connection.rs
stevenengler: Added a unit test that tries to cover this.

stevenengler: I also had to disable miri for this new test, otherwise it times-out.

PR #3141, opened 2023-09-05
Partially upgrade nix to 1.27.1

stevenengler:
The latest nix version makes three main changes:
- no default cargo features
- old epoll api deprecated
- many (but not all) functions take `AsFd` types instead of `RawFd` (an integer)

This third point is particularly painful in tests where we use integer file handles everywhere. I've left the tests on nix 1.26 for now, but it seems worthwhile to get this upgrade out of the way for the actual shadow code. If we're going to need to update a lot of tests, it's probably worth changing them to rustix or linux-api instead.

Also I added nix cargo features to the `shadow-shim-helper`, `shadow-rs`, and `shadow-tests` crates. Ideally we should add all of the nix features to every crate that requires them, but this would probably quickly become out of date anyways so I haven't bothered with that. Instead we take advantage of the fact that features enabled in one crate enable it in all other crates for the current build.

### DIFF
--- a/src/Cargo.lock
+++ b/src/Cargo.lock
@@ -948,16 +948,27 @@ dependencies = [
 
 [[package]]
 name = "nix"
-version = "0.26.2"
+version = "0.26.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


stevenengler:
Made an issue for the nix upgrade in the tests: #3265

PR #3144, opened 2023-09-05
Reduce the number of allocations needed by rust tcp

stevenengler:
When an application makes many small writes to a tcp socket, the rust tcp socket creates a `Bytes` object for each write. This requires an allocation for each `Bytes` object, which can mean many small allocations. Instead we now pre-allocate a larger `BytesMut` buffer (2 KB) and "break off" new `Bytes` objects from it. So instead of many small allocations, we have fewer large allocations. This does not change the number of `Bytes` objects we have and each write will still have its own `Bytes` object. It's just that these `Bytes` should be allocated more efficiently.

### DIFF
--- a/src/lib/tcp/src/buffer.rs
+++ b/src/lib/tcp/src/buffer.rs
@@ -17,6 +17,7 @@ pub(crate) struct SendQueue<T: Instant> {
     // exclusive
     end_seq: Seq,
     fin_added: bool,
+    unused: BytesMut,
 }
 
 impl<T: Instant> SendQueue<T> {
@@ -28,6 +29,7 @@ impl<T: Instant> SendQueue<T> {
             start_seq: initial_seq,
             end_seq: initial_seq,
             fin_added: false,
+            unused: BytesMut::new(),
         };
 
         queue.add_syn();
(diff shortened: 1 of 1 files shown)


src/lib/tcp/src/buffer.rs
```diff
@@ -48,21 +50,49 @@ impl<T: Instant> SendQueue<T> {
...
         mut len: usize,
     ) -> Result<(), std::io::Error> {
-        // this value shouldn't affect the tcp behaviour, only how the underlying bytes are
-        // allocated
-        const MAX_BYTES_PER_CHUNK: usize = 10_000;
+        // These values shouldn't affect the tcp behaviour, only how the underlying bytes are
+        // allocated. The numbers are chosen arbitrarily.
+        const MAX_BYTES_PER_ALLOC: usize = 10_000;
+        const MIN_BYTES_PER_ALLOC: usize = 2000;
+        assert!(MIN_BYTES_PER_ALLOC <= MAX_BYTES_PER_ALLOC);
```
sporksmith, line 57: nit - https://docs.rs/static_assertions/latest/static_assertions/macro.const_assert_eq.html

sporksmith (approved):

src/lib/tcp/src/buffer.rs
stevenengler: Added.

PR #3149, opened 2023-09-06
Implement waitpid and wait4 syscalls

### LINKED ISSUE #2989: Implement wait4 (syscall for waitpid, wait)

sporksmith:
Fixes #2989

### DIFF
--- a/src/main/host/host.rs
+++ b/src/main/host/host.rs
@@ -498,14 +498,19 @@ impl Host {
                 let process = processrc.borrow(&self.root);
                 if process.parent_id() != pid {
                     // Not a child of the current process
+                    return None;
+                }
+                process.set_parent_id(ProcessId::INIT);
+                let Some(z) = process.borrow_zombie() else {
+                    // Not a zombie
+                    return None;
+                };
+                if z.reaper(self).is_some() {
+                    // Not an orphan
                     None
                 } else {
-                    process.set_parent_id(ProcessId::INIT);
-                    process
-                        .borrow_zombie()
-                        .and_then(|z| z.reaper(self))
-                        .is_none()
-                        .then_some(*other_pid)
+                    // Is a zombie orphan child
+                    Some(*other_pid)
                 }
             })
             .collect();
--- a/src/main/host/host.rs
+++ b/src/main/host/host.rs
@@ -539,6 +539,21 @@ impl Host {
         self.processes.borrow_mut().remove(&id)
     }
 
+    /// Borrow the set of processes. Generally this should only be used to
+    /// iterate over the set of processes. e.g. fetching a specific process
+    /// should be done via via `process_borrow`.
+    // TODO: It would be preferable to return an iterator instead of the
+    // collection itself. There has to be an intermediate object though since we
+    // need both the borrowed map of processes, and an iterator that borrows
+    // from that. I suppose we could create an abstract "Iterator factory" and
+    // return that here instead of exposing BTreeMap type.
+    #[track_caller]
+    pub fn processes_borrow(
+        &self,
+    ) -> impl Deref<Target = BTreeMap<ProcessId, RootedRc<RootedRefCell<Process>>>> + '_ {
+        self.processes.borrow()
+    }
+
     pub fn cpu_borrow(&self) -> impl Deref<Target = Cpu> + '_ {
         self.cpu.borrow()
     }
--- a/src/main/host/host.rs
+++ b/src/main/host/host.rs
@@ -533,6 +533,12 @@ impl Host {
         Ref::filter_map(self.processes.borrow(), |processes| processes.get(&id)).ok()
     }
 
+    /// Remove the given process from the Host, if it exists.
+    #[track_caller]
+    pub fn process_remove(&self, id: ProcessId) -> Option<RootedRc<RootedRefCell<Process>>> {
+        self.processes.borrow_mut().remove(&id)
+    }
+
     pub fn cpu_borrow(&self) -> impl Deref<Target = Cpu> + '_ {
         self.cpu.borrow()
     }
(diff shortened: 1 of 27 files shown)


src/main/host/host.rs
```diff
@@ -533,6 +538,27 @@ impl Host {
...
+    // collection itself. There has to be an intermediate object though since we
+    // need both the borrowed map of processes, and an iterator that borrows
+    // from that. I suppose we could create an abstract "Iterator factory" and
+    // return that here instead of exposing BTreeMap type.
+    #[track_caller]
+    pub fn processes_borrow(
+        &self,
+    ) -> impl Deref<Target = BTreeMap<ProcessId, RootedRc<RootedRefCell<Process>>>> + '_ {
+        self.processes.borrow()
+    }
```
stevenengler, line 560: Could you do something like `impl Deref<Target = IntoItererator<Item = (ProcessId, RootedRc<RootedRefCell<Process>>)>` instead?

stevenengler (commented):
Still looking through it, I plan to continue tomorrow morning so just leaving these now.

src/main/host/host.rs
sporksmith: I don't think so, because the iterator needs to *borrow from* the returned object (`Ref<BTreeMap>`), but can't contain both that borrowed object and the internal iterator (without jumping through `unsafe` circular reference hoops). `IntoIterator` won't work here because it consumes `self` instead of borrowing from it.

sporksmith: Oh interesting. I can't quite make it work though: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1fa9cdb85ffdbee7b151af1860e86b1b

src/main/host/syscall/handler/wait.rs
```diff
@@ -0,0 +1,247 @@
...
+            // ECHILD: The process specified by pid (waitpid()) or idtype and id
+            // (waitid()) does not exist or is not  a  child  of  the calling
+            // process.
+            return Err(Errno::ECHILD.into());
+        }
+
+        if !options.contains(WaitFlags::WEXITED) {
+            // The other events that can be waited for (WUNTRACED, WSTOPPED,
+            // WCONTINUED) currently can't happen under Shadow.
+            // TODO: If and when those things *can* happen, check for them here.
```
stevenengler, line 109: Is this worth a log-once message?

src/test/clone/test_fork.rs
```diff
@@ -383,6 +384,537 @@ fn test_exit_signal_with_fatal_signal(exit_signal: nix::sys::signal::Signal) ->
...
+                core::ptr::null_mut(),
+            )
+        }
+        .unwrap()
+        {
+            CloneResult::CallerIsChild => {
+                unsafe { libc::exit(0) };
+            }
+            CloneResult::CallerIsParent(child_pid) => child_pid,
+        };
```
stevenengler, line 713: nit: maybe split up some of these to make them easier to read? for example maybe `let clone_rv = linux_api::sched::clone(...); let clone_child_no_signal = match clone_rv { ... };`

src/test/clone/test_fork.rs
```diff
@@ -446,6 +978,438 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+
+    for use_wclone in [true, false] {
+        tests.push(ShadowTest::new(
+            &format!(
+                "test_waitfn_honors_wclone:waitid:{}",
+                if use_wclone { "set" } else { "unset" }
+            ),
+            move || {
+                test_waitfn_honors_wclone(use_wclone, |options| {
+                    let mut info: siginfo_t = unsafe { std::mem::zeroed() };
```
stevenengler, line 1250: Maybe add some comments around the new tests? These are testing some pretty specific linux things that someone looking at the code might not know much about (myself included), so having more comments will help if this breaks in the future.

stevenengler (approved):
Looks good to me! I do think it would be good to use the rust event source object if possible rather than introduce new uses of the status listener since it will be really nice to get rid of the status listener eventually, but I don't think it needs to hold up this PR.

src/main/host/host.rs
stevenengler: Yeah I don't see how to make it work either. Was worth a try :)

src/main/host/status_listener.rs
```diff
@@ -0,0 +1,132 @@
...
+        let status_listener = unsafe {
+            cshadow::statuslistener_new(
+                Some(notify_fn),
+                callback_obj.cast::<std::ffi::c_void>(),
+                Some(object_free_fn),
+                std::ptr::null_mut(),
+                None,
+                host,
+            )
+        };
```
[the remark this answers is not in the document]
sporksmith: Yikes, good catch. Fixed.

src/main/host/syscall/handler/wait.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    pub fn from_waitpid_pid(current_process: &Process, pid: kernel_pid_t) -> Self {
+        match pid.cmp(&0) {
+            cmp::Ordering::Less => {
+                if pid == -1 {
+                    WaitTarget::Any
+                } else {
+                    WaitTarget::Pgid((-pid).try_into().unwrap())
+                }
+            }
```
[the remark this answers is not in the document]
sporksmith: Done

src/main/host/syscall/handler/wait.rs
sporksmith: Done

src/test/clone/test_fork.rs
sporksmith: done

src/test/clone/test_fork.rs
sporksmith: Done

PR #3152, opened 2023-09-10
Fix issue in verify.sh

mello29:
On Ubuntu 22.04.3 (6.2.0-32-generic), verification of Tor test fails:
```shell
user @ VirtualBox [11:07:05] [~/test/shadow/src/test/tor/minimal] [main *]
-> % ./verify.sh 
Bootstrapped count: 11/11
grep: hosts/client/tgen.*.stdout: No such file or directory
```

Edit line 25 of verify.sh to fix bug:
```shell
user @ VirtualBox [11:09:25] [~/test/shadow/src/test/tor/minimal] [main *]
-> % ./verify.sh
Bootstrapped count: 11/11
Successful client stream count: 10/10 (minimum 10)
Successful torclient stream count: 10/10 (minimum 10)
Successful torbridgeclient stream count: 10/10 (minimum 10)
Successful fileserver stream count: 30/30 (minimum 30)
Successful torhiddenclient stream count: 10/10 (minimum 5)
Successful hiddenserver stream count: 10/10 (minimum 5)
Verification succeeded: Yay :)
```

### DIFF
--- a/src/test/tor/minimal/verify.sh
+++ b/src/test/tor/minimal/verify.sh
@@ -22,7 +22,7 @@ check_host () {
     # otherwise we miss errors from `grep` such as if the files don't exist.
     # https://unix.stackexchange.com/a/343259
     local stream_success_count
-    stream_success_count=$(grep -c stream-success hosts/"$NAME"/tgen.*.stdout)
+    stream_success_count=$(find ./shadow.data/hosts/"$NAME"/ -name "tgen.*.stdout" -exec grep -c stream-success {} +)
     echo "Successful $NAME stream count: ${stream_success_count}/$TOTAL_STREAMS (minimum $MINIMUM_STREAMS)"
     if [ "${stream_success_count}" -lt "$MINIMUM_STREAMS" ]; then
         printf "Verification %bfailed%b: Not enough $NAME streams were successful :(\n" "$RED" "$NC"


sporksmith:
The current version of the script is correct for the way it's invoked by our cmake test framework (the proposed change [breaks it](https://github.com/shadow/shadow/actions/runs/6136329358/job/16654063596?pr=3152#step:18:29), but I see how it'd be confusing if you try to run the script directly. In particular it looks like our [documentation](https://shadow.github.io/docs/guide/getting_started_tor.html?highlight=tor-minim#simple-tor-network-example) for running this test directly is incorrect.

Maybe we should just delete the part showing how to run the test outside of the test framework; that also has it running the test inside the source tree instead of the build tree, which is also not great.

I'll have to think a bit more about the best way to fix the documentation here, but in the meantime the easiest way to run the test is probably something like:

```
$ ./setup build --extra --debug --test && ./setup test --extra tor-minimal
```

if you're interested in running it more directly, you can inspect `build/Testing/Temporary/LastTest.log` to see the shell command that this ultimately executes. On my machine it's:

```
Command: "/usr/bin/sh" "-c" "      rm -rf tor-minimal-shadow.data       &&  /home/jnewsome/projects/shadow/dev/build/src/main/shadow       --data-directory=tor-minimal-shadow.data       --log-level=info       --use-cpu-pinning true --parallelism 2 --strace-logging-mode off --template-directory shadow.data.template       /home/jnewsome/projects/shadow/dev/src/test/tor/minimal/tor-minimal.yaml       && (cd tor-minimal-shadow.data && /home/jnewsome/projects/shadow/dev/src/test/tor/minimal/verify.sh)       "
Directory: /home/jnewsome/projects/shadow/dev/build/src/test/tor/minimal
```

mello29:
Thanks for the additional info, I was not aware that the script was being used also by cmake. 
As you mentioned, I was just following the documentation; all things considered, I completely agree that the PR can be rejected and get to fix the documentation instead.

stevenengler:
I opened a related issue #3153 to test all documentation examples in the CI.

stevenengler:
Thanks for reporting the problem. I'm closing this in favour of a dedicated issue #3158.

Issue #2988, opened 2023-06-01
Implement execve


sporksmith:
Required for https://github.com/shadow/shadow/issues/1987

sporksmith:
Split execveat into https://github.com/shadow/shadow/issues/3166

PR #3156, opened 2023-09-13
Eliminate env usage in the shim

sporksmith:
Fixes https://github.com/shadow/shadow/issues/2848

This is also helpful for https://github.com/shadow/shadow/issues/2988, since otherwise we'd need to manipulate the environment of spawned processes as well (being careful not to leave duplicate or conflicting variables inherited from the original spawn).


src/lib/shim/src/lib.rs
```diff
@@ -543,22 +542,19 @@ pub mod export {
...
-        let ipc_blk = ShMemBlockSerialized::from_str(ipc_blk).unwrap();
+        let mut bytes = [0; core::mem::size_of::<ShMemBlockSerialized>()];
+        let bytes_read = rustix::io::read(
+            unsafe { rustix::fd::BorrowedFd::borrow_raw(libc::STDIN_FILENO) },
+            &mut bytes,
+        )
+        .unwrap();
+        // Implement looping? We should get it all in one read, though.
+        assert_eq!(bytes_read, bytes.len());
+        let ipc_blk = unsafe { core::mem::transmute(bytes) };
```
stevenengler, line 579: I don't think this has the correct alignment. `ShMemBlockSerialized` is also not ffi-safe so may have different layouts in the shim and shadow. Also I don't understand how this works at all since `ShMemBlockSerialized` contains a `PahtBuf` which is dynamically allocated.

stevenengler, line 579: Can we send just a string instead and not worry about alignment, layout, and other unsafe things?

src/lib/shadow-shim-helper-rs/src/shim_event.rs
```diff
@@ -52,6 +52,11 @@ pub struct ShimEventStartReq {
     pub thread_shmem_block_to_init: ForeignPtr<ShMemBlockSerialized>,
     /// Shim pointer to be initd by Shadow. Optional.
     pub process_shmem_block_to_init: ForeignPtr<ShMemBlockSerialized>,
+
+    /// Shim pointer to be initd by Shadow. Optional.
+    /// If set, shadow will write a null-terminated path.
+    pub initial_working_dir_to_init: ForeignPtr<u8>,
+    pub initial_working_dir_to_init_len: usize,
```
stevenengler, line 59: Why not put the working dir in the process shared memory rather than writing to the plugin's memory directly?

stevenengler (changes requested):

src/lib/shim/src/lib.rs
stevenengler: Oh I think we don't need to worry about alignment. And I see we're using a custom `PahtBuf` which is not dynamically allocated.

I think it would be nicer if the pod crate provided a safe transmute function that we called here instead. Such as `fn transmute(x: impl Pod) -> T`.

stevenengler: > I think it would be nicer if the pod crate provided a safe transmute function that we called here instead. Such as `pub fn transmute<T: Pod>(x: impl Pod) -> T`.

Maybe this isn't possible since `Pod` types can have uninitialized memory, so you can't convert a `Pod` to another `Pod`. Though you could probably do `pub fn transmute<const N: usize, T: Pod>(x: [u8; N]) -> T`.

stevenengler: I don't think this is worth doing if we have a nicer pod transmute function like the comment above.

src/lib/shadow-shim-helper-rs/src/shim_event.rs
stevenengler: We discussed this, and since the working dir needs to be ~4 KiB, it might not be worth putting in shared memory. So I think either way is fine, but should have some comments here explaining why it's done this way.

src/lib/shim/src/lib.rs
sporksmith: Nice; done

sporksmith: Agreed; resolving

src/lib/shadow-shim-helper-rs/src/shim_event.rs
sporksmith: Done

src/main/host/managed_thread.rs
```diff
@@ -687,6 +697,29 @@ impl ManagedThread {
...
+            let written = nix::errno::Errno::result(unsafe {
+                libc::write(
+                    stdin_writer,
+                    serialized_bytes.as_ptr().cast(),
+                    serialized_bytes.len(),
+                )
+            })
+            .unwrap();
+            // TODO: loop if needed. Shouldn't be in practice, though.
+            assert_eq!(written, isize::try_from(serialized_bytes.len()).unwrap());
```
stevenengler, line 718: If the pipe is blocking, Linux guarantees that the write will block until all `n` bytes have been written. So as long as the user hasn't set the max pipe size smaller than the struct size (in which case shadow would deadlock or maybe ENOSPC), this should be guaranteed (but no harm in checking the rv anyways).

src/main/host/managed_thread.rs
```diff
@@ -605,6 +603,18 @@ impl ManagedThread {
         let mut file_actions: libc::posix_spawn_file_actions_t = shadow_pod::zeroed();
         Errno::result(unsafe { libc::posix_spawn_file_actions_init(&mut file_actions) }).unwrap();
 
+        // Set up stdin
+        let (stdin_reader, stdin_writer) =
+            nix::unistd::pipe2(nix::fcntl::OFlag::O_CLOEXEC).unwrap();
```
stevenengler, line 608: maybe create the pipe in `O_DIRECT` mode to guarantee that the entire struct will be read in a single `read` call?

src/main/host/managed_thread.rs
```diff
@@ -219,6 +219,18 @@ impl ManagedThread {
...
+                    if !start_req.initial_working_dir_to_init.is_null() {
+                        // Write the working dir.
+                        let mut mem = ctx.process.memory_borrow_mut();
+                        let mut writer = mem.writer(ForeignArrayPtr::new(
+                            start_req.initial_working_dir_to_init,
+                            start_req.initial_working_dir_to_init_len,
+                        ));
+                        writer
+                            .write_all(ctx.process.current_working_dir().to_bytes_with_nul())
+                            .unwrap();
```
stevenengler, line 231: Should there be a flush() call somewhere?

stevenengler (approved):

src/main/host/managed_thread.rs
sporksmith: Done

src/main/host/managed_thread.rs
sporksmith: From my reading of the man page, I don't think it would make a difference. It seems more useful for a scenario where  we were writing multiple variable-sized objects and wanted to prevent reading e.g. all of the first object and some of subsequent objects. (In particular everything should already be done in one write call and one read call if the data is < PIPE_BUF bytes, and we'll end up failing otherwise)

PR #3159, opened 2023-09-14
Move preload injection to ManagedThread::spawn

sporksmith:
When using `ManagedThread::spawn` to help implement `execve`, we'll want to ensure that shadow's `LD_PRELOAD`'d libraries are injected into the new processes as well (in case the parent didn't pass through it's own `LD_PRELOAD` environment through to the child's environment).

Progress on #2988

### DIFF
--- a/src/main/core/manager.rs
+++ b/src/main/core/manager.rs
@@ -833,6 +833,12 @@ fn get_required_preload_path(libname: &str) -> anyhow::Result<PathBuf> {
 
     let libpath = libpath.ok_or_else(|| anyhow::anyhow!(format!("Could not library in rpath")))?;
 
+    let bytes = libpath.as_os_str().as_bytes();
+    if bytes.iter().any(|c| *c == b' ' || *c == b':') {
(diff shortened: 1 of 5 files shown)


src/main/core/manager.rs
```diff
@@ -684,65 +691,6 @@ impl<'a> Manager<'a> {
...
-        //   - preload values from LD_PRELOAD entries in the environment process option
-
-        let preload = self.make_preload_paths();
-
-        for path in &preload {
-            let path = path.as_os_str().as_bytes();
-            // these two characters separate paths and aren't valid in a preload path
-            assert!(!path.contains(&b':'));
-            assert!(!path.contains(&b' '));
-        }
```
stevenengler, line 701: I think we should keep this check somewhere.

stevenengler (approved):

src/main/core/manager.rs
sporksmith: Done

PR #3161, opened 2023-09-20
execve scaffolding

sporksmith:
Misc updates for supporting `execve`

Progress on #2988


src/main/utility/mod.rs
```diff
@@ -478,6 +478,29 @@ pub fn inject_preloads(mut envv: Vec<CString>, injected_preloads: &[PathBuf]) ->
...
+/// open longer than needed. Theoretically the subprocess could also operate on
+/// the leaked descriptor, causing difficult-to-diagnose issues, but this is
+/// unlikely in practice, especially since shadow's shim should prevent any
+/// native file operations from being executed from managed code in the first place.
+pub fn debug_assert_cloexec(file: &impl std::os::fd::AsRawFd) {
+    #[cfg(debug_assertions)]
+    {
+        let flags = nix::fcntl::fcntl(file.as_raw_fd(), nix::fcntl::FcntlArg::F_GETFD).unwrap();
+        let flags = nix::fcntl::FdFlag::from_bits_retain(flags);
+        debug_assert!(flags.contains(nix::fcntl::FdFlag::FD_CLOEXEC));
```
robgjansen, line 500: Is it worth printing a warning or error level message to help give some extra context before the assertion fails and dumps a stack trace?

robgjansen (approved):

src/lib/shadow-shim-helper-rs/Cargo.toml
```diff
@@ -15,6 +15,7 @@ nix = { version = "0.27.1", features = ["event", "fs", "socket"] }
 log = { version = "0.4", features = ["release_max_level_debug"] }
 logger = { path = "../logger" }
 once_cell = "1.18.0"
+shadow-pod = { path = "../pod" }
```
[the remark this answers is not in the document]
sporksmith: We've had some discussion about this before, but don't recall if we ever reached a consensus. I've gone back and forth and am probably responsible for both names in this case :sweat_smile: 

I've more recently been leaning towards hyphens since that seems to be the more established convention in rust crate names, but OTOH underscores don't have the "magical" transformation from hyphen to underscore as a rust identifier (e.g. package `shadow-pod` being imported as `use shadow_pod`)

I'm fine with picking one or the other and enshrining it in our [style guide](https://shadow.github.io/docs/guide/coding_style.html), but I think that'd make sense as a separate PR

src/main/utility/mod.rs
sporksmith: Done

PR #3171, opened 2023-09-25
Don't report core dumps

sporksmith:
Whether a signaled process generates a core dump depends on platform conditions outside of our control (see `core(5)`). To avoid different simulation results depending on whether a core was actually dumped, which in turn depends on the system configuration, we should always report that it hasn't been.

### DIFF
--- a/src/lib/linux-api/src/signal.rs
+++ b/src/lib/linux-api/src/signal.rs
@@ -1184,7 +1184,7 @@ impl Default for sigaction {
 }
 
 // Corresponds to default actions documented in signal(7).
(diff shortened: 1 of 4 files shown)


stevenengler (approved):
I'm not against this change, but I thought we weren't concerned with determinism across different machines. Is there a reason this flag is different?

sporksmith:
It's not just the machine, but the user's specific environment such as their current resource limits. There are other user-specific things that could affect the simulation too (such as the user's credentials affecting ability to access a given file), but it's easy enough to eliminate this one.

PR #3182, opened 2023-09-27
Pluggable transport

sporksmith:
Adds an obs4 transport bridge and client to the tor minimal test, and includes a small fix to epoll to get it working.

Also minimally updates the tor example documentation.


sporksmith:
I'm going to follow-up to do more of an overhaul of the tor example. I think I'm going to "fork" the simulation config into something simpler for the example that won't require obs4. Maybe it can drop the tgen dependency too and just use netcat or something.

That'll free us to further complicate the "minimal" tor test, e.g. to add arti clients :).

docs/getting_started_tor.md
```diff
@@ -5,19 +5,21 @@ transfer](getting_started_basic.md) and [traffic
...
-`~/.local/bin/tor` (see [the Tor install
-README](https://github.com/torproject/tor/blob/main/README)). You also need to
-install (or link) a TGen executable in `~/.local/bin/tgen` (see [the TGen
-installation guide](https://github.com/shadow/tgen)).
+This example requires that you have installed:
+* [`tor`](https://github.com/torproject/tor/blob/main/README); can typically be installed
+via your system package manager.
+* [`tgen`](https://github.com/shadow/tgen); will most likely need to be built from source.
+* [`obfs4proxy`](https://gitlab.com/yawning/obfs4); can typically be installed via your system package manager. The simulation will still functon without it, but the simulated
+hosts `ptbridge` and `torptbridgeclient` will have errors.
```
stevenengler, line 13: (Not blocking this PR) I think it would be better if the example didn't have additional dependencies. This example should probably be split out from the test anyway. Maybe later when you fork the tor minimal test, also update this example to use the simpler thing that has fewer dependencies.

.github/workflows/extra_tests.yml
```diff
@@ -301,6 +301,10 @@ jobs:
             libssl-dev \
             zlib1g
 
+      - name: Install obfs4proxy
+        run: |
+          apt-get install -y obfs4proxy
```
stevenengler, line 306: Should probably also add something to the `docs/extra_tests.md` page about this dependency.

stevenengler, line 306: Where do we install Go? Should this be installed in the same place?

src/main/host/descriptor/epoll/mod.rs
```diff
@@ -371,6 +371,9 @@ impl Epoll {
         // Add everything that is still ready back to the ready set.
         self.ready.extend(keep);
 
+        // We've mutated the ready list; we may need to trigger callbacks.
+        CallbackQueue::queue_and_run(|cb_queue| self.refresh_state(cb_queue));
```
stevenengler, line 375: I think the `CallbackQueue` should be passed in as an argument so that the closure will start before the `borrow_mut()` call. So `epoll_wait_helper` would have a:

```rust
let ready = CallbackQueue::queue_and_run(|cb_queue| epoll.borrow_mut().collect_ready_events(max_events, cb_queue));
```

src/test/tor/minimal/CMakeLists.txt
```diff
@@ -16,6 +16,8 @@ add_shadow_tests(BASENAME tor-minimal
                    --use-cpu-pinning true
                    --parallelism 2
                    --strace-logging-mode off
+                   # Disable to support fork
+                   --use-memory-manager false
```
stevenengler, line 20: Should the `shadow_config_spec.md` docs also say the memory manager must be disabled for fork?

src/test/tor/minimal/shadow.data.template/hosts/ptbridge/lock
stevenengler, line 1: I don't think you need to check in the lock file?

src/test/tor/minimal/tor-minimal.yaml
```diff
@@ -100,6 +100,19 @@ hosts:
...
       start_time: 60
       expected_final_state: running
+  ptbridge:
+    network_node_id: 0
+    ip_addr: 100.0.0.100
+    processes:
+    - path: tor
+      args: --Address ptbridge --Nickname ptbridge
+            --BridgeRelay 1
+            --ServerTransportPlugin "obfs4 exec /usr/bin/obfs4proxy --enableLogging --logLevel=DEBUG --unsafeLogging"
```
stevenengler, line 110: Will tor do $PATH-based lookups? If so, it would be nice to not hardcode a obfs4proxy path.

src/test/tor/minimal/shadow.data.template/hosts/ptbridge/keys/ed25519_master_id_secret_key
stevenengler, line 1: I wonder if you'll get an automated email about uploading a private key to github...

stevenengler (approved):

src/test/tor/minimal/tor-minimal.yaml
sporksmith: It won't. Even if it did, we don't set the `PATH`, or provide an easy way to propagate the user's `PATH` into managed processes.

We have the same issue with the `python3` in the fork+exec tests.

I'm still thinking about the best way to address it. For our tests the best solution I can think of that would be robust to odd layouts (e.g. nix, guix) would be to use CMake to locate binaries using the build-user's PATH, and then patch the fully qualified path into generated configs. It's a bit gross though :/.

Issue #3158, opened 2023-09-14
Fix example tor simulation documentation


stevenengler:
In #3152, @mello29 found that the tor example at [docs/getting_started_tor.md](https://github.com/shadow/shadow/blob/f83d18e593fb8f853b892dea22b7b02923d0f32c/docs/getting_started_tor.md) threw an error when running the verify script.

```text
user @ VirtualBox [11:07:05] [~/test/shadow/src/test/tor/minimal] [main *]
-> % ./verify.sh 
Bootstrapped count: 11/11
grep: hosts/client/tgen.*.stdout: No such file or directory
```

We should fix the example so that there are no errors.

Since the scripts in the `shadow/src/test/tor/minimal/` directory are also used in our tor minimal cmake test, we could modify the verify script so that it works for both the documentation example and the cmake test. But it might be better to add some similar bash code directly to the example that better demonstrates what we intend for users to learn from the example (for example code that greps the log files for "stream-success") rather than just running a script in the example.

Related: #3153

PR #3184, opened 2023-09-27
Simplify and test tor example

sporksmith:
We've gradually expanded the tor-minimal test to test more tor-network use-cases. It's probably no longer the best "minimal" example in our documentation for how to run tor in shadow.

This forks and simplifies the tor-minimal test into a simpler simulation. It also reworks it to test the actual commands we ask the user to run.

Fixes https://github.com/shadow/shadow/issues/3158


docs/getting_started_tor.md
```diff
@@ -9,28 +9,58 @@ This example requires that you have installed:
...
+```yaml
+{{#include ../examples/docs/tor/shadow.yaml}}
+```
+
+## Running the Simulation
+
+We run this example similarly as before. Here we use an additional command-line
+flag `--template-directory` to copy a template directory layout containing each
+host's `tor` configuraton files into its host directory before the simulation
+begins.
```
stevenengler, line 30: Maybe add a link to where they can find this template directory if they want to run it on their own?

stevenengler (approved):

docs/getting_started_tor.md
sporksmith: Done

Issue #2980, opened 2023-05-30
Support `AF_NETLINK` address family


cohosh:
**Describe the issue**
Trying out snowflake simulations on the newest version of shadow and ran into some trouble with go library calls again. Seems like Snowflake's requirements have changed since I last tried a year ago :)

When I run a [minimal snowflake example](https://github.com/cohosh/shadow-snowflake-minimal), all of the tgen streams fail and up inspecting the logs from the snowflake client, I see the following log messages:

```
2000/01/01 00:01:30 snowflake-client 2.5.1
2000/01/01 00:01:30 Started SOCKS listener at 127.0.0.1:9000.
2000/01/01 00:02:00 SOCKS accepted: {127.0.0.1:8080   map[]}
2000/01/01 00:02:00 


 --- Starting Snowflake Client ---
2000/01/01 00:02:00 Using ICE servers:
2000/01/01 00:02:00 url: stun:stun:3478
2000/01/01 00:02:00 Rendezvous using Broker at: http://broker:8080
2000/01/01 00:02:00 ---- SnowflakeConn: begin collecting snowflakes ---
2000/01/01 00:02:00 ---- SnowflakeConn: starting a new session ---
2000/01/01 00:02:00 WebRTC: Collecting a new Snowflake. Currently at [0/1]
2000/01/01 00:02:00 snowflake-9b4b05eedfb883bf  connecting...
2000/01/01 00:02:00 WebRTC: DataChannel created
2000/01/01 00:02:00 Failed to prepare offer failed to create network: route ip+net: netlinkrib: address family not supported by protocol
2000/01/01 00:02:00 WebRTC: closing DataChannel
2000/01/01 00:02:00 WebRTC: closing PeerConnection
2000/01/01 00:02:00 WebRTC: Closing
2000/01/01 00:02:00 WebRTC: failed to create network: route ip+net: netlinkrib: address family not supported by protocol  Retrying...
2000/01/01 00:02:00 redialing on same connection
```

These presumably correspond to the following warnings in `shadow.log`:

```
00:00:00.428894 [154689:shadow-worker] 00:02:00.000000001 [WARN] [snowflakeclient:11.0.0.3] [socket.c:1109] [syscallhandler_socket] unsupported socket domain "16", we only support AF_INET
```

The same thing appears to be happening with the proxy (which also uses WebRTC).

**To Reproduce**
Run a minimal Snowflake experiment: https://github.com/cohosh/shadow-snowflake-minimal

[...]

PR #3198, opened 2023-10-03
Netlink sockets

ppopth:
Following https://github.com/shadow/shadow/issues/2980

This PR implements Netlink sockets. Please look at netlink(7) and rtnetlink(7) before reviewing the PR so that you can understand it better.

I use the `ip addr` command as a baseline. This PR implements anything that is required by that command: `capget`, `catpset`, `setsockopt`, and `getsockname`.

## How it works

When the user calls `sendmsg` we check only the metadata, not the content of the message, like the destination address, the supported sendmsg flags. If it passes all the checks, we simply put the message into the buffer.

When the user calls `recvmsg` we get the message from the buffer, process the message, and return the response. Currently we support only `RTM_GETLINK` and `RTM_GETADDR`. For each message, we return a bunch of `RTM_NEWLINK`s and `RTM_NEWADDR`s respectively.

## Commit messages

Please also look at the commit messages. I think it explains things in detail.

commit 26927f6be415ab9d5a6eaeccfda5b2bd49d3c6ec
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Wed Oct 4 21:22:45 2023 +0700

    Install libcap and ip to the CI
    
    This commit installs libcap and ip to the CI because capget,capset
    require libcap and ip is used to test the Netlink sockets.
    
    In Debian/Ubuntu, the corresponding packages are libcap-dev and
    iproute2.
    
    In Fedora, the corresponding packages are libcap-devel and iproute.

commit b79a26c639728409ceb48e7b755901e5271dd96a
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Mon Oct 2 14:41:30 2023 +0700

    Test Netlink send buffer limit
    
    The user can set the send buffer limit by setting SO_SNDBUF using
    setsockopt. We test this for Netlink sockets by setting this option and
    send a lot of messages to the socket.

commit f4c33f7cbe1d732171b04cb94a82f7d87da61c21
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Sun Oct 1 23:11:19 2023 +0700

    Test Netlink with `ip addr`
    
    Since `ip addr` uses a netlink socket to get the information of the
    interfaces and addresses, we use it to test our Netlink implementation.
    
    We check that the output of the `ip addr` command matches the
    ip-addr.stdout file.
    
    We mark the test as an extra because it requirs the iproute2 package.

commit 9778bd35302b014d9ce4e89b281b4b1966fc8025
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Fri Sep 22 13:23:04 2023 +0700

    Implement Netlink setsockopt syscall
    
    Currently we handle only (SOL_SOCKET, SO_SNDBUF) and (SOL_SOCKET,
    SO_RCVBUF). In SO_SNDBUF, we set the send buffer limit, while,
    in SO_RCVBUF, we do nothing.

commit 0cd5678ca0f99892cf2748a41056fc7440cadebc
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Sat Sep 30 17:38:50 2023 +0700

    Test capget,capset
    
    This commits adds both the linux tests and the shadow tests for both
    capget and capset.
    
    We assign them as extra tests because they require the local machine to
    install libcap-dev.

commit ec4d84338ab1a35fe1a779d9595d9c5abcd3602b
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Wed Sep 20 14:59:19 2023 +0700

    Implement capget,capset
    
    In capget, we will always return the empty capability bitfield because
    we don't allow any capability in the plugin.
    
    In capset, we will allow to set with the empty capability bitfield.

commit 25223ea6754b4bbc1a1a4f78106aead07d6977d5
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Wed Sep 20 16:59:05 2023 +0700

    Implement Netlink getsockname syscall
    
    Previously we have only the is_bound field in InitialState which is used
    to indicate whether the socket is already bound or not. Now we want to
    implement getsockname which requires us to return the bound address, so
    we need to change from the boolean is_bound to the concrete bound_addr
    field.

commit 65ea91abf965d834f5a9ee58f95ee45a6083e96e
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Fri Sep 22 12:55:43 2023 +0700

    Handle RTM_GETLINK
    
    Previously we handle only RTM_GETADDR, this commit now enables us to
    handle RTM_GETLINK.
    
    This commit implements the RTM_GETLINK handling part in a method
    called handle_ifinfomsg.
    
    In handle_ifinfomsg, it works as follows:
    - Check that ifi_family of the message is either AF_UNSPEC or AF_INET
      because we support only IPv4 addresses
    - Check that the rest of the fields (except ifi_change) are zeroes
      because they are unsupported
    - Send the message containing all the interfaces with the following
      attributes: IFLA_IFNAME, IFLA_TXQLEN, and IFLA_MTU
    
    We can also send the attributes IFLA_ADDRESS and IFLA_BROADCAST for the
    MAC address, but we mark it as a todo instead.
    
    The ifi_flags of the loopback interface is IFF_UP, IFF_LOOPBACK, and
    IFF_RUNNING.
    
    The ifi_flags of the ethernet interface is IFF_UP, IFF_BROADCAST,
    IFF_RUNNING, and IFF_MULTICAST.
    
    All the flags listed above are from what I observe in the `ip addr`
    command.

commit 15a20736f6d5dce9a81360b7ae2d4b6334364942
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Tue Sep 26 00:01:25 2023 +0700

    Support MSG_PEEK in Netlink recvmsg
    
    Since we implemented peek_packet in SharedBuf already, we can now
    support MSG_PEEK in the recvmsg syscall for Netlink sockets.

commit 7c8b20adc1a7500cb12928cf7769359563ab99cc
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Mon Sep 25 22:23:09 2023 +0700

    Peek the packets in SharedBuf
    
    This commit allows us to peek the packets stored inside the buffer
    without removing them from the buffer.

commit 9d84b69d03c0a759a82e5f81b1142d6a624c3b4f
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Fri Sep 22 20:03:38 2023 +0700

    Support Netlink msg without dst addresss
    
    Previously we check that the message sent in sendmsg has the destination
    address, but we won't check it anymore because we will just assume that
    it will be sent to the kernel.

commit 8db9d965d3512479fd883f9fa2c68dd6d7087c8e
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Tue Sep 19 11:11:16 2023 +0700

    Handle unsupported Netlink syscalls
    
    We currently don't support some syscalls for Netlink sockets: listen,
    connect, getpeername, getsockname, and accept. This commit implements
    dummy handlers for such syscalls, i.e. it just logs the warning and
    returns the error.

commit 01385f32f618c68d45dc487cc2f349e9801cecd2
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Sat Sep 16 23:43:10 2023 +0700

    Implement Netlink close syscall
    
    This commit adds the logic that calls add_reader to the shared buffer
    when the socket is created and calls remove_reader to the shared buffer
    when the socket is closed.

commit 70068ae5799d74ed93a024dccaa8351516ac35cc
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Thu Sep 14 14:11:31 2023 +0700

    Implement Netlink recvmsg syscall
    
    This commit implements the recvmsg syscall.
    
    The syscall handlers works as follows.
    - Retreive the received message from the buffer
    - Check that the message type is RTM_GETADDR because the only currently
      supported one is RTM_GETADDR
    - Check that ifa_family of the message is either AF_UNSPEC or AF_INET
      because we support only IPv4 addresses
    - Check that the rest of the fields are zeroes because they are
      unsupported
    - Send the message containing all the interfaces
    
    Each node in Shadow has only two interfaces: lo and eth0. The lo one has
    the prefix of 8 and the eth0 one has the prefix of 24.
    
    The detail of the interfaces is similar to the one in the getifaddrs
    syscall handler.

commit 5a40f1c823831d2b3df3233c9a55aff77308178f
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Mon Aug 14 20:03:57 2023 +0700

    Implement Netlink sendmsg syscall
    
    This commit implements the senmsg syscall.
    
    Only the following parameters are supported:
    - The destination pid must be kernel (pid=0), since the only entity that
      will reply the message is the kernel.
    - There are no destination groups. Since the only destination is the
      kernel, multicasting does not make sense.
    - The only supported flags are MSG_DONTWAIT, MSG_NOSIGNAL, and
      MSG_TRUNC.
    
    Rate limiting is done by limiting the number of in-flight bytes by
    NETLINK_SOCKET_DEFAULT_BUFFER_SIZE.
    
    When the syscall is succesfully, the socket pushes the message into the
    buffer and emits an event by using the StateEventSource instance.

commit 7464ab5b4a5879bee329187ed88cd804abf6965e
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Sat Aug 12 04:22:35 2023 +0700

    Implement Netlink bind syscall
    
    According to netlink(7), there are two issues worth mentioning.
    
    If the pid of the address is zero, the kernel takes care of assigning it, but we will
    leave it untouched at the moment. We can implement the assignment later when we want
    to support it.
    
    If the groups of the address is non-zero, it means that the socket wants to listen to
    some groups. Since we don't support broadcasting to groups yet, we will emit the error
    intead.

commit 7de08d3add7fc9b0ddb93fdbeeca16c51e1063a0
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Fri Aug 11 13:24:02 2023 +0700

    Socket boilerplate for Netlink
    
    This commit adds Netlink to SocketRef and SocketRefMut, adds all the
    boilerplate methods for NetlinkSocket, and adds the protocol states to
    NetlinkSocket.
    
    There are two protocol states in NetlinkSocket: Initial and Closed,
    which is similar to the stateless Unix domain socket because, as
    specified in netlink(7), Netlink is a datagram-oriented service.

commit 4825e9a6e0b5b487e563c3786262c7258402bfbd
Author: Suphanat Chunhapanya <haxx.pop@gmail.com>
Date:   Thu Aug 10 14:37:51 2023 +0700

    Implement Netlink socket syscall
    
    Parse the socket type and the netlink family from the socket syscall
    parameters. The only supported socket types are SOCK_RAW and SOCK_DGRAM
    and the only supported netlink family is NETLINK_ROUTE.


robgjansen:
Hi @ppopth! Wow, thanks for submitting the PR!

It might take us a bit to review this, but before we do I think we will want you to rebase against the main branch so that (1) you can take care of fixing the conflicts, and (2) then we can run the shadow CI tests against it to help us during review.

Once you rebase, fix the conflicts, and have the tests passing locally, go ahead and force-push the update onto the same branch which should auto-update this PR. Then we'll find some time to take a look more closely at it.

Thanks again!

src/main/host/descriptor/socket/netlink.rs
```diff
@@ -0,0 +1,1185 @@
...
+            // Get the IP address of the host
+            let default_ip = Worker::with_active_host(|host| host.default_ip()).unwrap();
+            // All the interface configurations are the same as in the getifaddrs function handler
+            let mut interfaces = Vec::new();
+            interfaces.push(Interface {
+                address: Ipv4Addr::LOCALHOST,
+                label: String::from("lo"),
+                prefix_len: 8,
+                if_type: Arphrd::Loopback,
+                mtu: 65536,
```
stevenengler, line 67: I don't think shadow uses a different MTU for loopback and eth0 (loopback should be 1500 as well). But maybe also worth opening an issue to increase the MTU on loopback if that's what Linux does.

ppopth:
I rebased it to the main branch already. Every commit after this comment is a fix-up commit.

src/main/host/descriptor/socket/netlink.rs
ppopth: > But maybe also worth opening an issue to increase the MTU on loopback if that's what Linux does.

Yes, that's what Linux does. Try running `ip addr`.

ppopth: Resolved by changing both to `cshadow::CONFIG_MTU`

stevenengler: Created #3200.

ppopth:
After seeing the lint errors, I decided to change many things. It's better to close this and I will open it again soon.

ppopth:
Reopened this PR, since I have already changed what I think I need to. I think the code is now far better. It's now ready for review. Thanks for your patience.

robgjansen:
Hi @ppopth ! Thanks for the updates. I see that you added tests which is awesome! 

We will do a more thorough review, but for now two quick points about the tests:
- We are strongly preferring that new tests are written in Rust rather than C. I know we still have a lot of C tests sitting around, and that's probably what you modeled your tests after, but we have been slowly trying to rewrite those in Rust so that we can get rid of as much C code as possible. If you think it would be pretty simple, I think now would best time to rewrite the tests because you are the most familiar and the logic is fresh in your mind. Otherwise they will sit as C tests for a long time and someone else will have to relearn to convert to Rust.
- There are two types of integration tests: one where we just test that a particular function works when called, and one where we try all possible combination of functionalities and input options to make sure we understand how it works on Linux, and then enable only the ones that pass when testing in Shadow. You have implemented the former type, but we strongly prefer that latter more rigorous type of test. Good examples of these are in `test/socket/*` and `test/time/*`. It is important to us to test all of the ways a particular functionality can be used because we noticed that otherwise it tends to cause many issues and bugs down the road when some other application tries to use that function but in a different way. Doing the more rigorous tests makes Shadow more stable and makes it much easier to tell why it fails if an application exercises some unsupported functionality.

This PR is already big enough, so I wouldn't change anything here. But would you be willing to work on the above two points in follow-up PRs to really help solidify NetLink as a more roundly supported and better-tested socket type?

sporksmith:
> What I thought is that it's good to write the test in C because all of the libc and POSIX syscall APIs are documented and fully supported in C, and all the examples in the man pages are in C as well.

In places where we're specifically trying to invoke a specific syscall or libc function under test, we usually do it through the [libc](https://docs.rs/libc/latest/libc/) crate for this reason. We still take advantage of Rust's higher-level APIs for other things, like setup and teardown of state, threads, etc

robgjansen:
Just building a bit on what @sporksmith wrote:

Rust's libc crate provides the sort of "direct syscall access" that is available to C code. So if you are writing a test that is designed to test say `epoll_create`, then in Rust you would use the libc crate to make that specific low-level syscall.

But there are many other places in a test where you just want some functionality and you don't necessarily care which syscall is being invoked, because that specific test case isn't trying to test the ancillary functionality. For that we found it much nicer (easier to write the code and reason about its correctness) when using the higher level Rust APIs.

We also found it much easier to write the more rigorous tests that I described above (let's call them fuzz tests) using Rust simply because the language has more ready-to-use constructs and crates that help make it easier. We already have some nice helper functions in rust that can make the fuzz-type testing easier. I wouldn't want us to have to write and maintain those helpers for every language.

robgjansen:
Oh also, thanks for starting this conversation about testing! ❤️ You are helping me realize that we don't have enough documentation about our expectations around testing, where to look for examples of good fuzz testing practices, etc. I will work on updating our documentation to be clear about our expectations to help clear this up for future work too :)

src/main/Cargo.toml
```diff
@@ -23,6 +23,7 @@ crossbeam = "0.8.2"
 gml-parser = { path = "../lib/gml-parser" }
 libc = "0.2"
 linux-api = { path = "../lib/linux-api", features = ["std"] }
+linux-raw-sys = "0.5.2"
```
robgjansen, line 26: Avoid.

src/main/host/descriptor/socket/netlink.rs
```diff
@@ -0,0 +1,1215 @@
...
+use linux_api::ioctls::IoctlRequest;
+use neli::consts::nl::{NlmF, NlmFFlags, Nlmsg};
+use neli::consts::rtnl::{
+    Arphrd, Ifa, IfaF, IfaFFlags, Iff, IffFlags, Ifla, RtAddrFamily, RtScope, Rtm,
+};
+use neli::nl::{NlPayload, Nlmsghdr};
+use neli::rtnl::{Ifaddrmsg, Ifinfomsg, Rtattr};
+use neli::types::{Buffer, RtBuffer};
+use neli::{FromBytes, ToBytes};
+use nix::sys::socket::{MsgFlags, NetlinkAddr, Shutdown};
```
robgjansen, line 16: I think we are also trying to stay away from `nix` usage like this, even though we had been using this a lot in other parts of our code. The [coding section in the docs](https://shadow.github.io/docs/guide/coding.html#platform-libc-and-linux-crates) explains a bit more.

I might be wrong, but I _think_ the use case here is to handle args ultimately coming from a syscall hander. In these cases I think we want to add these to our `linux-api` crate too. We only started building up the `linux-api` crate recently so it doesn't have everything we need yet. But[the time module](https://github.com/shadow/shadow/blob/main/src/lib/linux-api/src/time.rs) shows an example of what goes there.

@sporksmith can you confirm?

src/main/host/descriptor/socket/netlink.rs
```diff
@@ -0,0 +1,1215 @@
...
+            .refresh_file_state(&mut self.common, cb_queue)
+    }
+
+    pub fn shutdown(
+        &mut self,
+        _how: Shutdown,
+        _cb_queue: &mut CallbackQueue,
+    ) -> Result<(), SyscallError> {
+        // We follow the same approach as UnixSocket
+        log::warn!("shutdown() syscall not yet supported for netlink sockets; Returning ENOSYS");
```
robgjansen, line 154: We have been using one of the macros `warn_once_then_debug()` or `warn_one_then_trace()` for code paths where Shadow does not yet support all features that the application expects to be available in Linux, so that we don't end up spamming the log file if an application calls an unsupported function over and over but then falls back to something else that does work.

Example here:

https://github.com/shadow/shadow/blob/ab8f38e7f5b394035c6c7f3e12a7d66df100ac21/src/main/host/syscall/handler/fcntl.rs#L175-L178

Notice that we tend to prefix these messages with `(LOG_ONCE) ` to make it clear in the logs that this particular message is not going to be repeated.

Note that `log::warn!()` should still be used in places where an application is legitimately sending an invalid arg or something.

I noticed several places throughout your PR where the warn once pattern should be applied, but I'm only going to comment here once to avoid the extra spam. Please scan through your usage of `log::warn!()` to see which should be changed to the warn once pattern.

src/main/host/descriptor/socket/netlink.rs
```diff
@@ -0,0 +1,1215 @@
...
+
+    pub fn listen(
+        _socket: &Arc<AtomicRefCell<Self>>,
+        _backlog: i32,
+        _net_ns: &NetworkNamespace,
+        _rng: impl rand::Rng,
+        _cb_queue: &mut CallbackQueue,
+    ) -> Result<(), SyscallError> {
+        // We follow the same approach as UnixSocket
+        log::warn!("We do not yet handle listen request on netlink sockets");
```
robgjansen, line 292: Do we know if some of these syscalls like `listen()` are actually valid for netlink sockets? If we're not sure and we need to figure out what Linux does first (e.g., via fuzz-style testing), adding a comment to that effect would be helpful. (The same goes for the others below.)

src/main/host/descriptor/socket/netlink.rs
```diff
@@ -0,0 +1,1215 @@
...
+        // After sending the messages with the NLM_F_MULTI flag set, we need to send the NLMSG_DONE message
+        let done_msg = {
+            let len = None;
+            let nl_type = Nlmsg::Done;
+            let flags = NlmFFlags::new(&[NlmF::Multi]);
+            // Use the same sequence number as the request
+            let seq = Some(nlmsg.nl_seq);
+            let pid = None;
+            // Linux also emits 4 bytes of zeroes after the header. See `strace ip addr`
+            let payload: NlPayload<Nlmsg, u32> = NlPayload::Payload(0);
```
robgjansen, line 791: This is strange. I wonder if there is an extended header and we're missing the suffix part of it or something like that? If you could find out what the linux kernel source code is doing, it would be helpful to have a link to that here as a comment.

src/main/host/descriptor/socket/netlink.rs
```diff
@@ -0,0 +1,1215 @@
...
```
robgjansen, line 887: Just a note that if you find a linux source link for line 790, to also add here.

src/main/host/syscall/handler/clone.rs
```diff
@@ -376,8 +376,74 @@ impl SyscallHandler {
...
         Ok(kernel_pid_t::from(ctx.objs.thread.id()))
     }
+
+    #[log_syscall(/* rv */ std::ffi::c_int,
+        /* hdrp */ *const std::ffi::c_void,
+        /* datap */ *const std::ffi::c_void)]
+    pub fn capget(
+        ctx: &mut SyscallContext,
+        hdrp: ForeignPtr<linux_raw_sys::general::__user_cap_header_struct>,
+        datap: ForeignPtr<[linux_raw_sys::general::__user_cap_data_struct; 2]>,
```
robgjansen, line 390: Note that the structs here (and in `capset`) would change to `linux-api` versions (as would the `LINUX_CAPABILITY_VERSION_3` constant below).

robgjansen (changes requested):
Overall this looks great!

I'm going to mark as "request changes" for now and then I'll re-review the updates.

I did not repeat my prior testing-related comments here, because I assume we'll handle Rust testing and fuzz testing in separate PRs.

I'm hoping we can get @sporksmith's thoughts to help clear up some of my uncertainty around my `linux-api`-related comments to make sure we are doing the right thing by preferring to add parts to `linux-api` rather than adding a dependency on the `linux-raw-sys` crate.

src/lib/pod/src/lib.rs
```diff
@@ -325,3 +325,7 @@ unsafe impl Pod for libc::utmpx {}
 unsafe impl Pod for libc::utsname {}
 unsafe impl Pod for libc::winsize {}
 unsafe impl Pod for libc::clone_args {}
+
+// linux-raw-sys types
+unsafe impl Pod for linux_raw_sys::general::__user_cap_header_struct {}
+unsafe impl Pod for linux_raw_sys::general::__user_cap_data_struct {}
```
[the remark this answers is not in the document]
sporksmith: Yup, agreed we should try to have everything use `linux-api` for consistency. For some types `linux-api` creates safe(r) wrappers around the raw bindings, but IMO adding and exposing the raw binding types is probably adequate for this PR.

At some point it might make sense for `linux-api` to use  `linux-raw-sys` instead of generating bindings itself, but we currently rely on the name mangling done in our binding-generation for compatibility with C code. (Even if we don't need to expose these new types to C code, I think for now we want everything to go through this path for consistency)

Hopefully the docs and examples in `linux-api` are sufficient, but LMK if you have questions!

src/main/host/descriptor/socket/netlink.rs
sporksmith: Confirmed

src/main/host/descriptor/socket/netlink.rs
```diff
@@ -0,0 +1,1215 @@
...
+            let address = interface.address.octets();
+            let broadcast = if interface.prefix_len == 32 {
+                address
+            } else {
+                Ipv4Addr::from(
+                    (0xffff_ffff_u32) >> u32::from(interface.prefix_len)
+                        | u32::from(interface.address),
+                )
+                .octets()
+            };
```
stevenengler, line 746: minor nit: I think simpler as just the following, which doesn't need the check for `prefix_len >= 32` to avoid the panic:

```rust
let broadcast = Ipv4Addr::from(
    0xffff_ffff_u32
        .checked_shr(u32::from(interface.prefix_len))
        .unwrap_or(0)
        | u32::from(interface.address),
)
.octets();
```

src/main/utility/sockaddr.rs
```diff
@@ -588,6 +649,41 @@ mod tests {
         assert_eq!(u32::from_be(addr.sin_addr.s_addr), addr_original.ip());
     }
 
+    /// Convert from a `sockaddr_nl` to a `SockaddrStorage` to a `NetlinkAddr`.
+    #[test]
+    fn netlink_addr_from_libc() {
+        let mut addr_nl: libc::sockaddr_nl = unsafe { std::mem::zeroed() };
+        addr_nl.nl_family = libc::AF_NETLINK as u16;
+        addr_nl.nl_pid = 0x38deb915;
+        addr_nl.nl_groups = 0xf229a8ea;
```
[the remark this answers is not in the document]
ppopth: They are just magic numbers used for the test. They can be anything, but I will add the comment anyway.

src/main/utility/sockaddr.rs
```diff
@@ -588,6 +649,41 @@ mod tests {
...
+        let addr = addr.as_netlink().unwrap();
+
+        assert_eq!(addr.pid(), u32::from_le(addr_nl.nl_pid));
+        assert_eq!(addr.groups(), u32::from_le(addr_nl.nl_groups));
+    }
+
+    /// Convert from a `NetlinkAddr` to a `SockaddrStorage` to a `sockaddr_nl`.
+    #[test]
+    fn netlink_addr_to_libc() {
+        let addr_original = nix::sys::socket::NetlinkAddr::new(0x38deb915, 0xf229a8ea);
```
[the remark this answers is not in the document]
ppopth: They are just magic numbers used for the test. They can be anything, but I will add the comment anyway.

ci/container_scripts/install_deps.sh
```diff
@@ -39,6 +39,8 @@ esac
 APT_CI_PACKAGES=(
   curl
   rsync
+  libcap-dev
+  iproute2
```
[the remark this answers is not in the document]
ppopth: Yes, they would. Is there anything I have to do more for this?

src/main/host/descriptor/socket/netlink.rs
ppopth: I think, no, because the only supported socket types of netlink sockets are SOCK_DGRAM and SOCK_RAW (which is the same as SOCK_DGRAM see netlink(7)).

So `listen`, `accept` are not the things, but `connect` is still a valid syscall just (just like in Datagram-oriented protocol like UDP). For `ioctl`, I don't know much about it and I can't find any doc for netlink like tcp(7) or udp(7).

src/main/host/descriptor/socket/netlink.rs
ppopth: The unix socket is my main reference for implementing the netlink socket. I see what I did a lot in the unix socket https://github.com/shadow/shadow/blob/97d99/src/main/host/descriptor/socket/unix.rs

Is it supposed to be changed later in the unix socket as well?

src/lib/pod/src/lib.rs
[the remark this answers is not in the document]
ppopth: I saw this `linux-api` thing when I implemented this, but, since it requires some work, I used `linux-raw-sys` first and see what you guys think and yeah let's go with `linux-api`.

src/main/host/descriptor/socket/netlink.rs
robgjansen: Yes, we would ideally update other usages inside the Shadow codebase too. (We only just recently added the "log once" functionality in 82e59bc but we didn't thoroughly go through and fix all log lines yet.)

src/main/host/descriptor/socket/netlink.rs
ppopth: Yeah I was surprised by this as well. Yeah it's a good idea to dig into the linux source code.

ci/container_scripts/install_deps.sh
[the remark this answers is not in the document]
robgjansen: Yes, I think these deps should be moved into the `APT_PACKAGES` and `RPM_PACKAGES` sections and removed from the `APT_CI_PACKAGES` and `RPM_CI_PACKAGES` sections. 

But actually, now that I looked again I am reminded that the tests you added are only enabled in the `EXTRA` configuration. I see that it's not clear how to add dependencies that only apply to EXTRA tests. @stevenengler do we have a way to add deps only for EXTRA tests? Or what is the preferred way to handle this?

src/main/host/descriptor/socket/netlink.rs
ppopth: I saw that `MsgFlags` and `Shutdown` are also used in `socket/inet/tcp.rs` and `NetlinkAddr` is analogous to `SockaddrIn` in `socket/inet/tcp.rs`.

If I shouldn't use `nix`, I don't have an example to follow and I guess it's a new thing for Shadow as well.

They are not used only in tcp.rs but also udp.rs and unix.rs as well.

In addition, if I want to remove `Shutdown`, the socket interface has to be changed as well. See https://github.com/shadow/shadow/blob/main/src/main/host/descriptor/socket/mod.rs#L315

src/main/host/descriptor/socket/netlink.rs
ppopth: Yeah, good idea

ci/container_scripts/install_deps.sh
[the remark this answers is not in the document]
stevenengler: I would say do whatever we do for Go programs which also only run under the `--extra` configuration.

So since we list the `golang-go` package under `APT_PACKAGES`, I would put `libcap-dev` and `iproute2` there as well instead of under `APT_CI_PACKAGES` (and the same for the RPM packages below). The `APT_CI_PACKAGES` was originally meant for things used by the CI workflow yaml and bash scripts.

But I think that's all that needs to be done. Since they're "extra" tests, we don't need to list these dependencies anywhere else.

src/main/host/descriptor/socket/netlink.rs
stevenengler: Yeah while we are trying to get rid of nix, it's probably best to use nix for at least `Shutdown` which is part of the "socket interface", and you can't change it here without also needing to change it in all other socket types as well.

For `NetlinkAddr`, it would be nice to move away from nix as well (for example see `SockaddrUnix` in `main/utility/sockaddr.rs`; this was written before linux-api existed). But re-implementing `NetlinkAddr` in linux-api might be a bit of work, so I don't think it's worth holding up this PR over.

For `MsgFlags`, it would probably be straightforward to create this as a bitflags type in `linux-api`, but I don't have much opinion on this. If you want to do this, the steps are probably (1) Add `#include <linux/socket.h>` to `lib/linux-api/bindings-wrapper.h`, (2) run `lib/linux-api/gen-kernel-bindings.sh` to generate new bindings for the kernel api, (3) add a `lib/linux-api/src/socket.rs` file, and (4) create a new `MsgFlags` bitflags following an example like `OFlag` from `lib/linux-api/src/fcntl.rs`. You could find the `MsgFlags` flags by searching `lib/linux-api/src/bindings.rs` for constants that begin with `MSG_`.

stevenengler: > For `MsgFlags`, it would probably be straightforward to create this as a bitflags type in `linux-api`

Actually Linux doesn't expose the `MSG_` flags. I don't think we have decided on a way to expose things in `linux-api` that the Linux headers don't expose. So probably best to keep using `nix::MsgFlags` unless someone else here gives better advice about how to handle this.

ppopth: I guess the conclusion is that we should use `nix` for all these three things.

sporksmith: Yeah, continuing to use `nix` for these probably makes sense for this PR.

> Actually Linux doesn't expose the MSG_ flags. I don't think we have decided on a way to expose things in linux-api that the Linux headers don't expose.

For when we do want to tackle this - the way `linux-raw-sys` handles this is by having some manually created header files with the missing definitions. https://github.com/sunfishcode/linux-raw-sys/blob/main/gen/modules/net.h

We've done the same for some things, currently just in the single "wrapper" header file: https://github.com/shadow/shadow/blob/main/src/lib/linux-api/bindings-wrapper.h#L28

Eventually we may want to have `linux-api` just be high-level wrappers around the `linux-raw-sys` bindings instead of generating our own, but we currently still need our custom symbol name mangling to support some C code.

ppopth:
> But there are many other places in a test where you just want some functionality and you don't necessarily care which syscall is being invoked, because that specific test case isn't trying to test the ancillary functionality. For that we found it much nicer (easier to write the code and reason about its correctness) when using the higher level Rust APIs.

Does this mean that I should remove the tests for capset/capget? since they are only invoked by `ip addr` and we don't really care much about it.

In addition, it also seems hard for me to find a library in Rust or Go that invokes those syscalls directly.

robgjansen:
Sorry for the delay. I have a deadline on Thursday, but should be able to review this PR after that. Thanks for your patience!

stevenengler:
> Does this mean that I should remove the tests for capset/capget? since they are only invoked by ip addr and we don't really care much about it.

I think if we add new features to Shadow, we should test them. Even if they're not very important for netlink support, something else may try to use them in the future. And if Shadow supports the syscalls, then we want to have tests to make sure they're handled correctly.

> In addition, it also seems hard for me to find a library in Rust or Go that invokes those syscalls directly.

Yeah it would be nice if there was a libc wrapper. Instead I think you could add `capget` and `capset` functions to the `linux-api` library to make the syscalls. For example see [`linux_api::signal::kill_raw`](https://github.com/shadow/shadow/blob/efaf075ef5e5753e3e6989f777a16d4778b2805a/src/lib/linux-api/src/signal.rs#L1248). The Linux headers expose `cap_user_header_t` and `cap_user_data_t` in `linux/capability.h`, so you can add `linux/capability.h` in `lib/linux-api/bindings-wrapper.h` and regenerate the bindings to access them.

src/main/host/descriptor/socket/netlink.rs
```diff
@@ -554,7 +554,6 @@ impl InitialState {
                 "Attempted to bind netlink socket to an address with non-zero groups {}",
                 addr.groups()
             );
-            return Err(Errno::EINVAL.into());
```
robgjansen: @stevenengler what is our policy for shadow allowing unsupported features to silently succeed? I guess probably if we don't return the error here we should at least add a comment that says in what case the error should be returned again if more support is added in future work.

robgjansen (commented):
Almost ready I think! I have a couple of minor questions for Steve.

src/main/host/descriptor/socket/netlink.rs
```diff
@@ -207,7 +207,9 @@ impl NetlinkSocket {
                 // the send buffer, process it, and return the response immediately to the caller
             }
             _ => {
-                log::warn!("setsockopt called with unsupported level {level} and opt {optname}");
+                warn_once_then_debug!(
```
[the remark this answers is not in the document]
stevenengler: The `warn_once` macro will only log at "warn" level once ever, even if the message changes. So even if `level` or `optname` change, it won't log at "warn" again with these new values. So it would log as "debug" with the new values.

We already do this in a few places though. For example:

```text
$ rg "warn_once_then_debug"
main/host/descriptor/timerfd.rs
209:        warn_once_then_debug!("We do not yet handle ioctl request {request:?} on TimerFds");

main/host/syscall/handler/time.rs
42:            warn_once_then_debug!("Timer type {which:?} unsupported");
69:            warn_once_then_debug!("Timer type {which:?} unsupported");
169:            warn_once_then_debug!("Clock id {clock_id:?} unsupported in Shadow.",);

main/host/syscall/handler/mman.rs
150:                warn_once_then_debug!("Invalid mmap flags: {flags}");

...
```

src/lib/linux-api/src/bindings.rs
```diff
@@ -2222,6 +2488,803 @@ fn bindgen_test_layout_rseq() {
...
...NUX_CLONE_SETTLS is set, the tls descriptor\n                is set to tls.\n @set_tid:      Pointer to an array of type *pid_t. The size\n                of the array is defined using @set_tid_size.\n                This array is used to select PIDs/TIDs for\n                newly created processes. The first element in\n                this defines the PID in the most nested PID\n                namespace. Each additional element in the array\n                defines the PID in the parent PID namespace of\n                the original PID namespace. If the array has\n                less entries than the number of currently\n                nested PID namespaces only the PIDs in the\n                corresponding namespaces are set.\n @set_tid_size: This defines the size of the array referenced\n                in @set_tid. This cannot be larger than the\n                kernel's limit of nested PID namespaces.\n @cgroup:       If LINUX_CLONE_INTO_CGROUP is specified set this to\n                a file descriptor for the cgroup.\n\n The structure is versioned by size and thus extensible.\n New struct members must go at the end of the struct and\n must be properly 64bit aligned."]
```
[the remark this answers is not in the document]
stevenengler: If you mean the `#[doc = ...]` line, this is created by bindgen so I think it's okay. It takes the code comments from the linux header files and adds the text as rustdoc comments for the rust structure. I'm not sure why they use `#[doc = "a really long line"]` rather than using a multi-line `///` comment, but I'm guessing it's to handle whitespace formatting more consistently.

src/main/host/descriptor/socket/netlink.rs
ppopth: > @ppopth Is there a reason you removed this error? Does this cause some application to fail?

Yes, it doesn't work with rust-libp2p if we returns an error. The application that I implemented in https://github.com/shadow/shadow/pull/3242 doesn't work if we return an error because it actually binds to multicast groups.

> The man page says:
> 
> > Only processes with an effective UID of 0 or the CAP_NET_ADMIN capability may send or listen to a netlink multicast group.
> 
> I think it would be reasonable to return `EPERM` here? Most applications shouldn't expect to be able to use multicast groups anyways and should probably expect permission errors.

Yeah, I think Linux returns a successful one, when it tries to bind. Please try running the application in https://github.com/shadow/shadow/pull/3242.
```bash
$ strace ./build/src/target/debug/test_libp2p_ping 2>&1 | grep "bind.*AF_NETLINK"
bind(10, {sa_family=AF_NETLINK, nl_pid=0, nl_groups=0x000110}, 12) = 0
```

stevenengler: Okay I've had a chance to look into this. So when the application binds to the multicast netlink address, it's essentially subscribing to network-related OS events. Since Shadow doesn't support these, I think it should return an error for unknown bit flags. But if the application is only subscribing to IP address add/remove/change events (`RTMGRP_IPV4_IFADDR` and `RTMGRP_IPV6_IFADDR`), I think it's fine to return success since we know that IP addresses are never added/removed/changed in Shadow, so we know we're handling these flags correctly (by not doing anything at all).

My proposal is to expose the `RTMGRP_*` flags from the Linux headers in linux-api. So this means adding `linux/rtnetlink.h` to the includes in `lib/linux-api/bindings-wrapper.h`, regenerating the bindings with `(cd src/lib/linux-api/ && ./gen-kernel-bindings.sh)` and then creating `lib/linux-api/src/rtnetlink.rs` which exposes these constants. Then here in `main/host/descriptor/socket/netlink.rs`, we can check if `addr.groups()` contains only these `RTMGRP_IPV4_IFADDR` (0x10) or `RTMGRP_IPV6_IFADDR` (0x100) flags, and if so return 0. This will allow the libp2p example to work, which uses `nl_groups=0x000110`. If `addr.groups()` contains other flags that we haven't investigated, then we can return an error.

@ppopth Does this seem reasonable to you?

ppopth: @stevenengler That's a great idea. Thank you for figuring out what those flags really are.

ppopth:
Since the merge commit is quite large and I don't know why, I think it's helpful to put the conflicting files here to help you review.
```
Unmerged paths:
  (use "git add/rm <file>..." as appropriate to mark resolution)
	both modified:   src/main/host/descriptor/socket/mod.rs
	both modified:   src/main/host/syscall/handler/clone.rs
	both modified:   src/main/host/syscall/handler/mod.rs
	deleted by them: src/main/host/syscall_handler.c
```

ppopth:
@robgjansen @stevenengler I change all my C tests into Rust already. I also started implementing fuzz tests. Please look at `src/test/netlink/socket` to see it.

I implemented the fuzz tests only for bind. I'm implementing more now. Please let me know what you think.

stevenengler (commented):
> @robgjansen @stevenengler I changed all my C tests into Rust already. I also started implementing fuzz tests. Please look at `src/test/netlink/socket` to see it. I implemented the fuzz tests only for bind. I'm implementing more now. Please let me know what you think.

Sorry for the delay. Thanks for doing the additional work on the tests, the bind tests look good. This PR is hard for us to review since we don't have much netlink experience, and some of the most complicated parts (sendto and recvfrom) don't really have any tests. But if you can add some tests in a new PR for these additional syscalls, that would be really helpful.

But for now, I've left a few more comments, and after that I think you can rebase and we can approve and merge this PR (I think @robgjansen is in agreement here). This PR is already very large, so after this PR if you could add the new tests you're working on to a new PR, we'd appreciate it.

> I decided to have a new directory `src/test/netlink/socket` instead of amending the existing one `src/test/socket` because I think the existing one is too large already and there is too much difference between the netlink sockets and the inet sockets.

👍

src/lib/linux-api/src/capability.rs
```diff
@@ -0,0 +1,62 @@
...
+// Somehow this is not automatically generated in bindings.rs
+#[repr(C)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct __user_cap_data_struct {
+    pub effective: bindings::linux___u32,
+    pub permitted: bindings::linux___u32,
+    pub inheritable: bindings::linux___u32,
+}
+
+unsafe fn capget_raw(hdrp: *const user_cap_header, datap: *mut user_cap_data) -> Result<(), Errno> {
```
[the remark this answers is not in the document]
ppopth: I don't quite understand this comment. What is `SYSCALL_DEFINE2`? `user_cap_header` and `user_cap_data` are just `__user_cap_header_struct` and `__user_cap_data_struct`. I changed the name because the Linux names look quite bad; they have the `__` prefix and the `_struct` suffix.

Does this one https://github.com/shadow/shadow/pull/3198/commits/1de8496c7e63563f36e2f9dd5b72e10d2ff0d137 look good for you?

src/main/host/syscall/handler/clone.rs
```diff
@@ -411,8 +412,74 @@ impl SyscallHandler {
...
+        }
+
+        let datap: [_; 2] = ctx.objs.process.memory_borrow().read(datap)?;
+        for data in &datap {
+            // We don't allow the plugin to set any capability
+            if data.effective != 0 || data.permitted != 0 || data.inheritable != 0 {
+                warn_once_then_debug!(
+                    "The version of Linux capabilities is not supported ({})",
+                    hdrp.version
+                );
```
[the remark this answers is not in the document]
ppopth: Yeah, it was wrong. Thanks.

src/main/host/descriptor/socket/netlink.rs
```diff
@@ -0,0 +1,1239 @@
...
+            addr: Some(src_addr),
+            msg_flags: 0,
+            control_len: 0,
+        })
+    }
+
+    fn handle_error(&self, bytes: &[u8]) -> Vec<u8> {
+        // If we can't get the pid, set it to zero
+        let nlmsg_seq = match bytes.get(8..12) {
+            Some(x) => u32::from_le_bytes(x.try_into().unwrap()),
```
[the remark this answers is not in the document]
ppopth: It's the offset of the `nlmsg_seq` field in the C struct `nlmsghdr`, see `man 7 netlink`. But that struct is not defined in Rust, so I cannot use `memoffset::span_of!`. What should I use instead? Do I need to define `nlmsghdr` in linux-api?

src/main/host/descriptor/socket/netlink.rs
```diff
@@ -0,0 +1,1239 @@
...
+
+            let reader = IoVecReader::new(iovs, mem);
+            let reader = reader.take(len.try_into().unwrap());
+
+            // send the packet directly to the buffer of the socket so that it will be
+            // processed when the socket is read.
+            self.buffer
+                .borrow_mut()
+                .write_packet(reader, len, cb_queue)
+                .map_err(|e| Errno::try_from(e).unwrap())?;
```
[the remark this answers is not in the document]
ppopth: Yeah, that makes sense.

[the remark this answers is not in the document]
ppopth: Resolved in [f7a0358](https://github.com/shadow/shadow/pull/3198/commits/f7a0358d582b50fd7c9b01eb1d881d6e0badf011)

ppopth:
@stevenengler Is it okay if I will squash all the commits in this PR into a single commit? since I think it's already hard to squash all the fix-up commits.

src/lib/linux-api/src/capability.rs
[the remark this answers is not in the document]
stevenengler: Ah sorry I should have been more clear. The `SYSCALL_DEFINE2` is the actual linux syscall implementation.

https://github.com/torvalds/linux/blob/b401b621758e46812da61fa58a67c3fd8d91de0d/kernel/capability.c#L141

Linux casts the capget syscall arguments as `header: cap_user_header_t` and `dataptr: cap_user_data_t`. Since your wrapper in the linux-api library uses `header: *const user_cap_header` and `dataptr: *mut user_cap_data`, I was looking for a comment that explains the relationship between those types. Typically we'd want to use the exact same type in our wrapper as what Linux uses, but it's fine in this case since they are the same because `linux/capability.h` defines:

```c
typedef struct __user_cap_header_struct {
	__u32 version;
	int pid;
} __user *cap_user_header_t;

struct __user_cap_data_struct {
        __u32 effective;
        __u32 permitted;
        __u32 inheritable;
};
typedef struct __user_cap_data_struct __user *cap_user_data_t;
```

So your `*user_cap_header` is a typedef of `*__user_cap_header_struct`, which is also typedefed as `cap_user_header_t`.

> Does this one 1de8496 look good for you?

Looks good to me, but the `SYSCALL_DEFINE2` lines should be an exact copy-paste of the Linux source code. So just needs the following changes:

https://github.com/torvalds/linux/blob/b401b621758e46812da61fa58a67c3fd8d91de0d/kernel/capability.c#L141

```diff
-SYSCALL_DEFINE2(capget, cap_user_header_t, hdrp, cap_user_data_t, datap)
+SYSCALL_DEFINE2(capget, cap_user_header_t, header, cap_user_data_t, dataptr)
```

and:

https://github.com/torvalds/linux/blob/b401b621758e46812da61fa58a67c3fd8d91de0d/kernel/capability.c#L220

```diff
-SYSCALL_DEFINE2(capset, cap_user_header_t, hdrp, cap_user_data_t, datap)
+SYSCALL_DEFINE2(capset, cap_user_header_t, header, const cap_user_data_t, data)
```

src/main/host/descriptor/socket/netlink.rs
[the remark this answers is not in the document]
stevenengler: > Do I need to define `nlmsghdr` in linux-api?

Ideally yes (`nlmsghdr` is in `linux/netlink.h`), but to avoid blocking this PR any longer, you can just put a comment here instead with a "TODO: " which says to replace the byte offsets with the span of the `nlmsghdr.nlmsg_seq` field, and with a perma-link to this comment (https://github.com/shadow/shadow/pull/3198#discussion_r1491978336).

src/main/host/descriptor/socket/netlink.rs
```diff
@@ -0,0 +1,1239 @@
...
+
+        // We set the source address as the netlink address of the kernel
+        let src_addr = SockaddrStorage::from_netlink(&NetlinkAddr::new(0, 0));
+
+        let buffer = (|| {
+            if packet_buffer.len() < 16 {
+                log::warn!("The processed packet is too short");
+                return self.handle_error(&packet_buffer[..]);
+            }
+            let nlmsg_type = u16::from_le_bytes((&packet_buffer[4..6]).try_into().unwrap());
```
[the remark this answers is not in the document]
ppopth: > I'm guessing the 16 comes from the size of `nlmsghdr`.)

Yes.

stevenengler:
> @stevenengler Is it okay if I will squash all the commits in this PR into a single commit? since I think it's already hard to squash all the fix-up commits.

I'll let @robgjansen decide on that. (Since the PR is ~3,500 lines changed, my thinking is that if you could at least split it into 3 commits: one for the linux-api changes, one for the netlink socket implementation, and one for the tests.)

ppopth:
> > @stevenengler Is it okay if I will squash all the commits in this PR into a single commit? since I think it's already hard to squash all the fix-up commits.
> 
> I'll let @robgjansen decide on that. (Since the PR is ~3,500 lines changed, my thinking is that if you could split it into 3 commits: one for the linux-api changes, one for the netlink socket implementation, and one for the tests.)

It turned out that it's not as hard as I thought. I squashed all the fix-up commits already. See this branch https://github.com/ppopth/shadow/tree/netlink2

Please let me know when you are ready to bring it to this PR.

src/test/netlink/socket/bind/test_bind.rs
```diff
@@ -59,11 +59,15 @@ fn get_tests() -> Vec<test_utils::ShadowTest<(), String>> {
...
                 test_utils::ShadowTest::new(
                     &append_args("test_supported_groups"),
                     move || test_supported_groups(sock_type, flag),
                     set![TestEnv::Shadow],
                 ),
+                // This test is tested only in Shadow because Linux supports all the groups, while
+                // Shadow does not.
                 test_utils::ShadowTest::new(
                     &append_args("test_unsupported_groups"),
                     move || test_unsupported_groups(sock_type, flag),
```
robgjansen: I think we want to write our tests such that the behavior that Shadow does implement is tested in both Linux AND Shadow, to make sure that Shadow's behavior is consistent with Linux. For behavior that Shadow does not support, that can be tested in just Linux.

It looks like here we are writing only Shadow-specific tests, but we have no way of comparing Shadow behavior to Linux behavior. Am I missing something?

ppopth: > For behavior that Shadow does not support, that can be tested in just Linux.

`test_unsupported_groups` makes sure that, when the unsupported features are used, Shadow will emit an error.

I think testing the unsupported features only in Linux and not caring what Shadow does with it make less sense.

What do you think?

> It looks like here we are writing only Shadow-specific tests, but we have no way of comparing Shadow behavior to Linux behavior. Am I missing something?

Linux is supposed to support all the groups. Perhaps, I should do `test_supported_groups` in both Shadow and Linux, but do `test_unsupported_groups` only in Shadow.

robgjansen: OK, got it. Yeah normally we want to write tests that pass in Linux, with the goal of (eventually) making them also pass in Shadow (ie so Shadow matches Linux). So we would write something like a `test_group` function to be modular and take a group, and then specify which groups should pass in linux (probably all of them) and which should pass in Shadow.

In this case you've already separated the ones that should pass in Shadow. So I agree that you should do `test_supported_groups` in both Shadow and Linux.

For `test_unsupported_groups`, you check that Shadow is returning the correct error, which is a good start. We still don't know what Linux is doing in those cases, but I guess that's OK for now since we're not planning to add support for those in Shadow yet. If we do eventually add support in Shadow, then we could add the Linux+Shadow tests at that time.

robgjansen (approved):
Thanks for the hard work and for sticking with us during the long review process!

robgjansen:
I think you can go ahead and rebase to squash the fixups now. If you want to go further and squash into just 3 commits as suggested by @stevenengler, I think that would make sense. The code from some of the earlier commits was removed in later commits, so squashing down to 3 commits would keep things cleaner. After that, force push the cleaned branch and we'll work on merging. Thanks!

Issue #3151, opened 2023-09-07
Update rust ffi code to use `C-unwind`


stevenengler:
Most of our rust-C ffi code uses `pub extern "C" fn`, but these functions are not allowed by rust to panic (a panic is not allowed to cross the ffi boundary). While Shadow does do this and this is technically UB, it has seemed to work fine for us. But this may break in the future, so we should switch to `pub extern "C-unwind" fn` instead.

See:

- https://rust-lang.github.io/rfcs/2945-c-unwind-abi.html
- https://github.com/rust-lang/rust/issues/74990
- https://github.com/rust-lang/rust/pull/106075/
- https://github.com/rust-lang/rust/releases/tag/1.71.0

stevenengler:
I'm going to close this since in #3210 it's good enough for now. The long-term goal is to get rid of the C code, and if a new rust version in the future causes issues we can revisit. Shadow already seems to SIGABORT when panicking so our unwinding might not be working correctly anyways.

PR #3210, opened 2023-10-25
Change many function signatures from "C" to "C-unwind"

stevenengler:
I don't completely understand when to use "C" vs "C-unwind", but it sounds like the "C-unwind" will be the future-proof version of the current "C" behaviour.

From what I understand, function signatures should use "C-unwind" if they may unwind. Most of our exported rust functions should be "C-unwind" since they may panic. Our exported C functions should also be "C-unwind" since they may call C++ or rust functions that may throw an exception or panic. The linux-api code was left as "C" and not "C-unwind".

I'm not sure that this is correct everywhere, but hopefully is an improvement.

See #3151.


stevenengler:
@sporksmith A ping in case you have any objections.

robgjansen (approved):

sporksmith:
Thanks - no objection :)

> The linux-api code was left as "C" and not "C-unwind".

Any particular reason? I suppose it's relatively easy there to verify that those functions don't panic?

stevenengler:
> Thanks - no objection :)
> 
> > The linux-api code was left as "C" and not "C-unwind".
> 
> Any particular reason? I suppose it's relatively easy there to verify that those functions don't panic?

I'm not really sure what we should use, but my thinking was that since these functions/structs are for the kernel interface, they probably shouldn't be functions that unwind. For example a signal handler probably shouldn't unwind, so I think `linux_api::signal::SignalHandlerFn` should be `extern "C" fn(i32)` rather than `extern "C-unwind" fn(i32)`.

I should clarify that I did change the rust functions we export to C, for example I changed:

```diff
-    pub extern "C" fn linux_sigemptyset() -> linux_sigset_t {
+    pub extern "C-unwind" fn linux_sigemptyset() -> linux_sigset_t {
```

but things like `linux_api::signal::SignalHandlerFn` and `pub unsafe extern "C" fn sigaction_restorer()` were left as "C" and not "C-unwind".

sporksmith:
> I should clarify that I did change the rust functions we export to C, for example I changed:

Ah ok, yeah that makes sense

PR #3213, opened 2023-10-30
Migrate mmap-related syscall handlers (mman.h) to rust

stevenengler:
Hopefully I didn't mess anything up. Could be cleaned up a bit, but tried to stay similar to the existing code.

I also fixed what I think is a bug, where the fd we open in the plugin is not closed in a specific code path.


src/main/host/syscall/handler/mman.rs
```diff
@@ -109,13 +117,325 @@ impl SyscallHandler {
...
+        ctx: &mut ThreadContext,
+        fd: std::ffi::c_ulong,
+        file: *mut c::RegularFile,
+    ) -> Result<i32, ()> {
+        assert!(!file.is_null());
+
+        log::trace!("Trying to open file {fd} in the plugin");
+
+        // TODO: make sure we don't open special files like /dev/urandom, /etc/localtime etc. in the
+        // plugin via mmap
```
robgjansen, line 288: Would it be easy enough to at least detect when this is happening and log a warning?

src/main/host/syscall/handler/mman.rs
```diff
@@ -239,16 +240,26 @@ impl SyscallHandler {
...
+            let mmap_result = thread.native_mmap(
+                &process_ctx,
                 addr,
                 len,
                 prot,
                 flags,
                 plugin_fd.unwrap_or(-1),
                 offset,
-            )?);
+            );
```
robgjansen, line 255: Do the tests still pass if we remove the extra native syscall try?

robgjansen (approved):

src/main/host/syscall/handler/mman.rs
stevenengler: I added a check for the few cases that shadow supports.

src/main/host/syscall/handler/mman.rs
stevenengler: I removed this code and added an assertion that the result is not `SyscallError::Native`. I'll wait to get your approval before I merge.

robgjansen (approved):

PR #3214, opened 2023-10-30
Migrate prctl and prlimit64 syscall handlers to rust

stevenengler:
All of the tests pass, but we also don't have any tests for prctl or prlimit64 :) Tor does call `prctl(PR_CAPBSET_READ, ...)` which shadow correctly handles natively. The nginx example calls `prlimit64(0, 7, ...)` which shadow also correctly handles natively.

Edit: Added a few tests for prctl().


src/test/prctl/test_prctl.rs
```diff
@@ -0,0 +1,101 @@
+/*
+ * The Shadow Simulator
+ * See LICENSE for licensing information
+ */
+
```
robgjansen, line 5: I don't think we've generally been including this license header on new rust files, right?

robgjansen (approved):
Looks great, thank you!

src/test/prctl/test_prctl.rs
stevenengler: Removed.

PR #3219, opened 2023-11-06
Migrate remaining syscall handlers from unistd.c and shadow.c

stevenengler:
Some more low-hanging fruit. All syscalls in unistd.c and shadow.c are now migrated to rust. There are still some C syscall handlers in unistd.c (for example `read` and `write`), but they don't need to be migrated. They exist only for the C `RegularFile` type and can eventually be deleted when `RegularFile` is migrated to rust.


robgjansen (approved):
Nice!

PR #3220, opened 2023-11-07
Rust boilerplate for select.h and futex.h syscalls

stevenengler:
This doesn't migrate anything to rust, but just adds rust wrappers with the correct types to make it easier to migrate some syscall handlers in the future (and also gives nicer strace logs).


src/main/host/syscall/handler/futex.rs
```diff
@@ -20,4 +21,27 @@ impl SyscallHandler {
...
+
+    #[log_syscall(/* rv */ std::ffi::c_int, /* pid */ std::ffi::c_int,
+                  /* head_ptr */ *const std::ffi::c_void, /* len_ptr */ *const libc::size_t)]
+    pub fn get_robust_list(
+        _ctx: &mut SyscallContext,
+        _pid: std::ffi::c_int,
+        _head_ptr: ForeignPtr<ForeignPtr<linux_api::futex::robust_list_head>>,
+        _len_ptr: ForeignPtr<libc::size_t>,
+    ) -> Result<(), SyscallError> {
+        log::debug!("get_robust_list was called but we don't yet support it");
```
robgjansen, line 33: Is this a case where we should use `warn_once_then_debug`? (Also line 44 below.)

robgjansen (approved):
Looks good, thank you!

src/main/host/syscall/handler/futex.rs
stevenengler: Added.

PR #3221, opened 2023-11-08
Remove the "(LOG_ONCE)" requirement from `warn_once_*` macros

stevenengler:
Now the "(LOG_ONCE)" will be prepended automatically. I think this is an improvement so that you don't need to remember to add it, and they will stay consistent.


robgjansen (approved):
Good idea :)

PR #3222, opened 2023-11-09
Rust boilerplate for poll.h, signal.h, file.h, and fileat.h syscalls

stevenengler:
Added more rust syscall handler wrappers. The `poll.h` and `signal.h` wrappers include all of the arguments, but the `file.h` and `fileat.h` wrappers don't since it would take a while to figure out the proper arguments and add linux-api bindings for all of them.


robgjansen (approved):
😎

PR #3223, opened 2023-11-09
Move more of `syscall_handler.c` to rust

stevenengler:
This moves a large chunk of `syscallhandler_make_syscall` to rust.

There are a few changes from the original code. Previously the "pre" and "post" syscall code would only be run for syscalls that had explicit handlers in C or rust (`HANDLE_C` or `HANDLE_RUST`). This meant that syscalls that are unsupported or handled natively would not count towards the syscall counters or syscall timers. In the new code, this "pre" and "post" syscall code runs regardless of what syscall it is. I did it this way mostly because it was easiest, but we can change this if there's any reason not to.

Some of the log lines have also changed. For example the previous lines:

```
SYSCALL_HANDLER_PRE(test_pipe,pid=1000): handling syscall 257 openat
SYSCALL_HANDLER_POST(test_pipe,pid=1000): syscall 257 openat result: state=DONE val=3(n/a)
```

are now:

```
SYSCALL_HANDLER_PRE: openat (257) — (testnode.test_pipe.1000, tid=1000)
SYSCALL_HANDLER_POST: openat (257) result 3 — (testnode.test_pipe.1000, tid=1000)
```

The log lines are just a personal preference, and can be changed back if anyone dislikes them.

There might be some other minor changes as well that I'm not aware of, hopefully nothing important.


src/main/host/syscall/handler/mod.rs
```diff
@@ -70,6 +70,18 @@ impl SyscallHandler {
             }};
         }
 
+        macro_rules! unsupported {
+            ($name: literal) => {{
+                log::warn!(
```
robgjansen, line 107: Should we make this `warn_once_then_debug`?

I'm not sure but I think maybe yes because getting to this code path is not a syscall routing programming mistake, but rather could result from valid application usage calling syscalls that we don't yet support (but the app might have a valid fallback path). (By comparison, I think the `shim_only!` macro is a syscall routing programming mistake so it makes sense to panic in that case.)

robgjansen (approved):
🤩 

I like the new syscall log line format that you used. I think the main thing is that we want it to be easy to grep out specific syscalls from that line, ie to help to start debugging at the syscall entry point. I think grepping is OK with the new format too.

src/main/host/syscall/handler/mod.rs
stevenengler: Changed this to `warn_once_then_debug`.

PR #3228, opened 2023-11-10
Add `TZ=:/usr/share/zoneinfo/UTC` env to all processes

stevenengler:
This only exists for benchmarking, and will be closed.


robgjansen:
Did this change have the desired effect?

stevenengler:
> Did this change have the desired effect?

There doesn't appear to be any difference on the benchmark runner, but since the runner was just updated, I don't know if it still makes all of the extra syscalls by default.

run_time

PR #3229, opened 2023-11-13
Fixes and improvements to syscall handling code

stevenengler:
Misc fixes and improvements.

- Improved debug formatting of `SysCallReg`
- Added missing string repr for `Errno::EACCES`
- Simplified some syscall logging code
- Removed specific "unsupported syscalls" in syscall handling
- Simplified the syscall mapping
- Added linux-api bingings for `NR_*` syscall numbers
- Updated `SyscallHandler` to use `SyscallNum`
- Only warn once for all unsupported syscalls
- Change a `warn` to `warn_once_then_debug` for unix sockets


src/main/host/syscall/handler/mod.rs
```diff
@@ -319,7 +321,25 @@ impl SyscallHandler {
...
-                log::warn!("Detected unsupported syscall {} ({}) called from thread {} in process {} on host {}",
+                // only show a warning the first time we encounter this unsupported syscall
+                static WARNED_SET: Mutex<Option<HashSet<SyscallNum>>> = Mutex::new(None);
+
+                // `insert()` returns `false` if the syscall num was already in the set
+                let has_already_warned = !WARNED_SET
+                    .lock()
+                    .unwrap()
+                    .get_or_insert_with(HashSet::new)
+                    .insert(syscall);
```
robgjansen, line 332: I assume this is global across all shadow worker threads? I guess maybe we don't care because we expect that unsupported syscalls won't happen super frequently so lock contention will not be of practical concern?

src/main/host/syscall/handler/mod.rs
```diff
@@ -319,7 +321,25 @@ impl SyscallHandler {
...
+
+                let level = if has_already_warned {
+                    log::Level::Debug
+                } else {
+                    log::Level::Warn
+                };
+
+                log::log!(
```
robgjansen, line 340: Maybe add a comment here saying that we can't use the `warn_once_then_debug` macro because that warns once per specific log message whereas here we need `warn_once` behavior based on the syscall number not the log message itself.

robgjansen (approved):
Glad to see we are filling out linux_api and removing some reliance on libc 🧙

src/main/host/syscall/handler/mod.rs
stevenengler: Yeah this is global across all threads, and yeah I'm assuming that it will be uncommon for two threads to be handing an unsupported syscall at the same time. The lock is only held for the hash set lookup (and possibly insertion), so assuming that unsupported syscalls aren't called very often, and a hash set lookup is fast, I don't expect there to be much if any contention. I could change this to an `RwLock`, but it would make the code a bit more complicated (first try acquiring the read lock, then possibly try acquiring the write lock).

src/main/host/syscall/handler/mod.rs
stevenengler: I'll add this. And just to clarify, it logs once per line of code (or more specifically, once per macro call), not per specific log message. So if you had `log_once_then_debug!("Foo {}", x)`, it would only ever log that line once, even if `x` changed.

src/main/host/syscall/handler/mod.rs
robgjansen: OK, thanks for the explanation! No need to change the code I think.

stevenengler: I changed this to an `RwLock` anyways, I'm not really sure if an `RwLock` is slower in the general case, but at least we won't have to worry about contention.

PR #3234, opened 2023-11-21
Move more syscall handler code to rust

stevenengler:


robgjansen (approved):
Thanks! The syscall handler is getting a lot nicer :)

stevenengler:
> Thanks! The syscall handler is getting a lot nicer :)

It's getting there, but there will probably still be a lot of the rust code that could be cleaned up further once more code has moved over.

PR #3239, opened 2023-11-26
Clean up network interface and avoid a hashmap linear search

stevenengler:
This cleans up the network interface and removes a linear hashmap search in the network interface's priority queue.

The network interface used to need to deal with both rust and C sockets, so it stored tagged pointers for `CompatSocket` objects. Now that the TCP socket is the only remaining C socket, and it has a rust wrapper `LegacyTcpSocket`, C sockets aren't used in the network interface anymore. The network interface now only stores boxed `InetSocket` objects. This also allows us to remove tagged pointers from our C code.

The hashmap in the priority queue used by the fifo qdisc now uses custom hash and eq implementations, which allows us to avoid a linear search through the hashmap which used the canonical handles to identify the same socket.

The new hash function is more expensive since it uses a real hash function ([`DefaultHasher`](https://doc.rust-lang.org/std/collections/hash_map/struct.DefaultHasher.html) as used by the stdlib `HashMap`), whereas the original glib code doesn't seem to actually use a hash function:

https://docs.gtk.org/glib/type_func.HashTable.new.html

> If `hash_func` is NULL, `g_direct_hash()` is used.

https://github.com/GNOME/glib/blob/40081f9b627904c2506b6d3837334e4ce05052da/glib/ghash.c#L2483-L2487

```c
guint
g_direct_hash (gconstpointer v)
{
  return GPOINTER_TO_UINT (v);
}
```

This PR contains a lot of changes along the rust/C boundary, but everything seems to still work so hopefully it's correct.

This may fix #3238.

**Edit:**

There doesn't seem to be much (if any) of a performance regression. The simulation results have remained unchanged. The [following](https://github.com/shadow/benchmark-results/blob/master/tor/2023-11-27-T03-09-35/README.md) are from a slightly older version of this PR, but I don't expect it would make any difference.

run_time
transfer_time_5242880 exit
ram_simtime

**Edit 2:** One commit was split off into #3240.


src/main/host/descriptor/socket/inet/mod.rs
```diff
@@ -79,6 +79,9 @@ impl InetSocket {
         }
     }
 
+    /// Useful for getting a unique integer handle for a socket, or when we need to compare a C
+    /// `LegacySocket` to a rust `InetSocket` (which may internall point to the same
```
robgjansen, line 83: typo: `internall` -> `internally`

src/main/host/descriptor/socket/inet/mod.rs
```diff
@@ -221,6 +221,27 @@ impl PartialEq for InetSocket {
...
+    /// Returns a hash for the socket based on its address, and not the socket's state. Two
+    /// different sockets with the same state will return different hashes, and the same socket will
+    /// return the same hash even after being mutated.
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        // To match the `Eq` behaviour of `InetSocket`, the hashes of two sockets *must* be equal if
+        // the types are the same and the Arc's pointers are equal, and the hashes *should not* be
+        // equal if the two types are not the same or the Arc's pointers are not equal. We do return
+        // the same hash if the two types are different but the pointers are equal, but this is
+        // allowed by the `Hash` trait (it's just a hash collision) and we should never run into
+        // this without a variant containing a ZST, which wouldn't make sense in the context of
```
robgjansen, line 234: Dumb question: what is "ZST"?

robgjansen (approved):
🚀

src/main/host/descriptor/socket/inet/mod.rs
stevenengler: Changed "ZST" to "zero-sized type". What this is trying to say is that usually two different objects will have different memory addresses. For example with `let x = Box::new(0)` and `let y = Box::new(0)`, the two integers will have different memory allocations and different addresses. But this isn't true in the case of zero-sized types. For example:

```rust
let x = Box::new(());
let y = Box::new([0u8; 0]);
println!("{:p}, {:p}", &*x as *const _, &*y as *const _);
```

prints "0x1, 0x1", which would indicate that the data stored by each `Box` have the same address even though they aren't the same objects or the same type.

PR #3246, opened 2023-12-10
Remove more CMake code

stevenengler:
Removes some CMake code, and moves other CMake code to the Cargo build script.

New output of `shadow --show-build-info`:

```text
Shadow v1.13.2-6508-g1c135ade5 2023-12-10--04:10:16
GLib v2.64.6
Shadow was built on 2023-12-10--23:04:15 UTC from branch rm-cmake-templates with PROFILE=debug, OPT_LEVEL=1, DEBUG=true, RUSTFLAGS="-Cforce-frame-pointers=y", CFLAGS="-std=gnu11 -O0 -ggdb -fno-omit-frame-pointer -Wreturn-type -Wswitch -DDEBUG"
For more information, visit https://shadow.github.io or https://github.com/shadow
```

(I think it's using the wrong version tag since it's using my own git repo which probably doesn't have the latest tags.)

I think there are some further improvements that can be made to the build process, but I'll leave that for a possible future PR.


src/main/build.rs
```diff
@@ -390,6 +391,46 @@ fn build_shadow_c(build_common: &ShadowBuildCommon) {
...
+    else {
+        return None;
+    };
+
+    if !git_date.status.success() {
+        return None;
+    }
+
+    let git_describe = std::str::from_utf8(&git_describe.stdout).unwrap().trim();
+    let git_date = std::str::from_utf8(&git_date.stdout).unwrap().trim();
```
robgjansen, line 436: Nit: could maybe use a default string in case we don't get the stdout output as expected. Or maybe we want it to panic if we don't get the stdout output?

src/main/build.rs
```diff
@@ -434,6 +435,56 @@ fn git_version() -> Option<String> {
...
+                    .unwrap()
+                    .trim()
+                    .to_string(),
+            );
+        }
+    };
+
+    let git_branch = git_branch.unwrap_or("<unknown>".to_string());
+
+    // Note that the CFLAGS aren't necessairily the flags that the C code is built with. The `cc`
```
robgjansen, line 473: > necessairily

typo

robgjansen (approved):

src/main/build.rs
stevenengler: Changed these to return `None` if git returns invalid utf-8.

PR #3247, opened 2023-12-11
Move git and timestamp info to new crate

stevenengler:
This improves the incremental build time if you're updating code that is *not* in `src/main/`. This improves incremental compilation in this case by about 8 seconds on my laptop since it skips the build script in `src/main/`. It still needs to (incrementally) rebuild `src/main/` though, which takes about 4 seconds.

This also makes small formatting changes to the `shadow --show-build-info` text.

### DIFF
--- a/src/main/build.rs
+++ b/src/main/build.rs
@@ -1,7 +1,4 @@
-use std::process::Command;
 use std::{env, path::PathBuf};
-
-use time::OffsetDateTime;
 
 use shadow_build_common::{CBindgenExt, ShadowBuildCommon};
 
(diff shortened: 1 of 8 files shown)


src/main/build.rs
```diff
@@ -399,7 +399,9 @@ fn build_info() -> String {
     let profile = std::env::var("PROFILE").unwrap();
     let opt_level = std::env::var("OPT_LEVEL").unwrap();
     let debug = std::env::var("DEBUG").unwrap();
-    let rflags = std::env::var("CARGO_ENCODED_RUSTFLAGS").unwrap();
+    let rflags = std::env::var("CARGO_ENCODED_RUSTFLAGS")
+        .unwrap()
+        .replace('\u{1f}', " ");
```
robgjansen, line 404: maybe add comment for what the replace is doing here?

robgjansen (approved):

Issue #2095, opened 2022-04-26
Move Shadow's main function from C to Rust


stevenengler:
Shadow's main function is a small C wrapper around a Rust function:

https://github.com/shadow/shadow/blob/117bbf40f5afa2dbb72a4e997abc21888eb906a7/src/main/main.c#L9

https://github.com/shadow/shadow/blob/117bbf40f5afa2dbb72a4e997abc21888eb906a7/src/main/core/main.rs#L234-L238

It would be nice to remove the C main function and build the Rust version directly.

Shadow currently uses CMake to build a C library `shadow-c`, a Rust library `shadow-rs`, and then builds the final C binary:

```cmake
add_library(shadow-c-and-rs INTERFACE)
target_link_libraries(shadow-c-and-rs INTERFACE "-Wl,--start-group" shadow-rs shadow-c "-Wl,--end-group")

add_executable(shadow main.c)
target_link_libraries(shadow shadow-c-and-rs)
set_target_properties(shadow PROPERTIES LINK_FLAGS "-Wl,--no-as-needed")
```

If we built the main Shadow binary as a Cargo project, we could reduce this to two steps: the C library and a Rust binary. This would also allow us to skip the final link step, which would save us a few seconds of build time.

One issue is that we set the runpath to the shim library directories and shadow [reads this runpath](https://github.com/shadow/shadow/blob/fcdfd4bbe462220ffc0844acf24b62fc613e64b8/src/main/core/manager.c#L156) to use the correct shim libraries. For example, after building Shadow, the runpath will be set to `/abs-path/shadow/build/src/lib/shim:/abs-path/shadow/build/src/lib/injector_preload:/abs-path/shadow/build/src/lib/libc_preload:/abs-path/shadow/build/src/lib/openssl_preload:`. After installing Shadow, the runpath will be changed to `/home/user/.local/lib`. We will either need to use a different mechanism for specifying the correct path of shim libraries, or use something like patchelf to manually modify the runpath as part of the build and installation steps (CMake does this for us, but we'd have to do it manually if building the binary with Cargo). And instead of the runpath, we could maybe use a different custom elf section instead.

PR #3248, opened 2023-12-12
[RFC] Embed preload libs in the shadow binary

stevenengler:
I wanted to try this and see how it could work. This embeds the shim/preload libraries in the shadow binary. I'm fine with leaving Shadow how it currently works (if it's not broken, don't fix it), but this might make the transition away from CMake easier, and maybe would be nicer for users. Pinging @robgjansen and @sporksmith since you might have opinions on this.

Advantages:

- Shadow becomes a single binary. You can move it anywhere without breaking links to any preload libraries. In current shadow you can move the shadow binary if you installed it as long as you move the libraries at the same time, but you cannot move a non-installed shadow binary even if you also move the libraries.
- No longer need to set shadow's rpath to locate the preload libraries. This means we don't need to `patchelf` shadow when installing it (which involves moving the preload libraries). CMake does this for us now, but in a future without CMake we would need to do this manually if we continue using the rpath.
- Unblocks #2095.
- Shadow will work when run in a directory containing spaces or colons. In `LD_PRELOAD` space/colon characters are path separators, so current shadow does not work if the preload libs have any of these characters in their paths.
- Don't need to worry about shadow using old/wrong shim/preload libs since the libs will be bundled with shadow at build time.

Disadvantages:

- Possibly more difficult to debug when something goes wrong since the preload libraries don't exist somewhere on disk to debug. (See below for workaround.)
- Requires setting `LD_LIBRARY_PATH` on the managed process to a temporary directory.
- It's weird.

How it works:

At build time we include the contents of the shim/preload libraries in the shadow binary. At runtime we create several memfds and write the contents of the libraries to these memfds. When we spawn the managed process, we use the paths to the memfds in `LD_PRELOAD`. So for example it looks like `LD_PRELOAD=/proc/self/fd/6:/proc/self/fd/7:/proc/self/fd/8`. The annoying piece is the shim, which is loaded by the preload libraries and must be named `libshadow_shim.so`. So we create a temporary directory containing a symlink from `libshadow_shim.so` to `/proc/self/fd/X` and set `LD_LIBRARY_PATH` for the managed process as the path to that temporary directory containing the symlink.

To help with debugging, this also adds two experimental options `--dump-preload-libs <directory>` and `--with-preload-libs <directory>`. The "dump" option writes the embedded preload libraries to individual files in the given directory. The "with" option skips all of the memfd stuff and uses the preload libraries from the given directory instead.


sporksmith:
Neat! Have you checked whether gdb works as expected? In particular if it still symbolizes stack frames from the libraries correctly? I'd think it *probably* would, but seems worth checking.

>  Like, what if we statically compile a private copy of the shim directly into each preload lib (keeping the symbols in the shim private so that they do not cause unintentional function interceptions)?

I don't think that'd work out. The shim has global state that we wouldn't want duplicated into private copies. Also relative to the other weirdness, the temp directory doesn't seem that bad.

In general I'm a bit on the fence whether it's worth the "weirdness". Not having to deal with patchelf and rpath is compelling, though :). I think I'm slightly in favor of going forward with it as long as gdb works as expected...

robgjansen:
>> Like, what if we statically compile a private copy of the shim directly into each preload lib (keeping the symbols in the shim private so that they do not cause unintentional function interceptions)?

> I don't think that'd work out. The shim has global state that we wouldn't want duplicated into private copies. 

I'm confused. I understand that .text segment data is shared across multiple managed processes that are loading the shim. And any global state that is changed inside a managed process will be visible to other threads of that managed process. But global state that is changed will not be visible across processes, right? If the shim uses shared memory to share state with shadow or other managed processes, that same method could be used if the shim.so was made private too, right? What am I missing?

stevenengler:
> Neat! Have you checked whether gdb works as expected? In particular if it still symbolizes stack frames from the libraries correctly? I'd think it probably would, but seems worth checking.

So I think that using the `--with-preload-libs` option would make gdb work the same as it did before. If you didn't use this option, I'd expect gdb to fail to load the libraries at `/proc/self/...`. But we could change the code to use shadow's actual pid instead. So we'd have `LD_PRELOAD=/proc/<shadow-pid>/fd/6:/proc/<shadow-pid>/fd/7:/proc/<shadow-pid>/fd/8`, which I think gdb should be able to access.

The problem I have when trying to test this, even with current Shadow, is that I'm not able to set a breakpoint in the shim, so I have no idea how to even use gdb to debug the libraries. If I use `set follow-fork-mode child`, `break _shim_parent_init_preload` and say "yes" to "Make breakpoint pending on future shared library load?", it never breaks. I've tried options like `symbol-file /path/to/libshadow_shim.so` but this causes the loader to segfault. I've also tried introducing a segfault into the shim, but shadow's signal handler catches it so I can't use that either. So I'm not really sure how to test this.

> I'm confused. I understand that .text segment data is shared across multiple managed processes that are loading the shim. And any global state that is changed inside a managed process will be visible to other threads of that managed process. But global state that is changed will not be visible across processes, right? If the shim uses shared memory to share state with shadow or other managed processes, that same method could be used if the shim.so was made private too, right? What am I missing?

I think what was meant was that each preload library would end up with its own shim state (the shim state would be duplicated across each preload library). For example if you statically link shim.c to each of libshadow_openssl_crypto.so, libshadow_libc.so, etc, then each of those preload libraries will get their own copies of the shim's global state. So for a given managed process, if libshadow_libc.so changes the shim state, that change won't be reflected in libshadow_openssl_crypto.so. You'd be effectively running multiple shims in the same managed process, which wouldn't work.

robgjansen:
> You'd be effectively running multiple shims in the same managed process, which wouldn't work.

Right, OK, thanks for expanding. We might be able to move all shim state to shared mem to effectively share it across the different preloads, but as @sporksmith says maybe the temp directory method isn't that bad anyway so it wouldn't be worth it.

stevenengler:
> I wonder if there might be a way we could avoid doing the thing with the temporary directory to handle libshadow_shim.so.

I think another option could be to add the shim to the preload list and remove the injector lib, but restrict the visibility of the shim symbols to only the `shim_api_*` functions using the `-fvisibility=none` compile option. I think this would work if the shim was entirely in C, but in rust there doesn't seem to be a way to limit what symbols are exported. There's some related discussion at https://github.com/rust-lang/rust/issues/73958, but I don't think there's any solution for us, and you're forced to export all symbols.

An even weirder idea could be to name the shim library as a number like `12345`, then dup2() the shim's memfd to a new fixed fd of 12345. Then we could set the `LD_LIBRARY_PATH` as `/proc/self/fd/` and when the preload libs search for the library "12345", it would find the shim. But this would be too weird for me.

robgjansen:
> but in rust there doesn't seem to be a way to limit what symbols are exported ... rust forces you to export all symbols

I don't think we need to export C function symbols that are called by Rust code. I think we would only have to export symbols defined in Rust code that need to be called from the C code. The other Rust symbols would be private and would not cause unintended interceptions. Here is the example I made to test this: https://github.com/robgjansen/symexp

The above assumes that we only have a single `cdylib` preload library though, built from cargo. It gets more complicated if we keep multiple preload libs as we have now, and that all load the same shim...

stevenengler:
> > but in rust there doesn't seem to be a way to limit what symbols are exported ... rust forces you to export all symbols
> 
> I don't think we need to export C function symbols that are called by Rust code. I think we would only have to export symbols defined in Rust code that need to be called from the C code. The other Rust symbols would be private and would not cause unintended interceptions. Here is the example I made to test this: https://github.com/robgjansen/symexp

I had thought that non-static C functions were exported automatically, but from your repo it looks like they aren't, so that's good.

One issue is that all `#[no_mangle] pub extern "C" fn` functions are exported, not just ones defined in the main program. For example if you add a dependency named "linux-api" to "preload", and in linux-api you add a function `#[no_mangle] pub extern "C-unwind" fn linux_sigemptyset() -> linux_sigset_t`, this will be exported in the preload shared library. We [use this kind of code](https://github.com/shadow/shadow/blob/348bc4c23309042f27a281c9e703e9d2b1f144fc/src/lib/linux-api/src/signal.rs#L1541) extensively already so that the shim's C code can call into shadow's linux-api code. This would be better when the entire shim is in rust, but we also can't control external dependencies, which may export any functions and we can't control that.

stevenengler:
I'm closing this since I don't have any immediate plans to look into this further. But if we want something like it in the future, then the code will still exist to revisit.

PR #3250, opened 2023-12-13
Support listening for file signal events

stevenengler:
We currently allow things to listen for state changes on files (ex: not readable -> readable), but there are cases where we want to listen for things that aren't state changes. For example edge-triggered epoll_wait()s will sometimes return if more bytes are written to a file's input buffer, even if the file's state didn't change (readable -> readable).

This adds the ability to emit signals from files, and to listen for them in the same way that you listen for state changes. This PR doesn't actually define any signals. For example listener callbacks were previously:

```rust
let listen_state = FileState::READABLE;
let filter = StateListenerFilter::Always;

let handle = file.add_listener(listen_state, filter, move |state, changed, cb_queue| {
    todo!();
});
```

You can now write:

```rust
let listen_state = FileState::READABLE;
let listen_signals = FileSignals::INPUT_BUFFER_CHANGED; // this signal doesn't actually exist
let filter = StateListenerFilter::Always;

/// listener will run when either the READABLE state changes, or when INPUT_BUFFER_CHANGED is emitted
let handle = file.add_listener(listen_state, listen_signals, filter, move |state, changed, signals, cb_queue| {
    todo!();
});
```

You emit signals using:

```rust
event_source.notify_listeners(file_state, states_changed, signals, cb_queue);
```


robgjansen:
👍 This design seems pretty elegant and should make it easier for us to implement some of epoll's api, and maybe other things too.

src/main/host/descriptor/mod.rs
```diff
@@ -139,6 +139,15 @@ impl From<FileState> for c::Status {
     }
 }
 
+bitflags::bitflags! {
+    /// File-related signals that listeners can watch for.
+    #[derive(Default, Copy, Clone, Debug)]
+    #[repr(transparent)]
+    pub struct FileSignals: u32 {
```
ppopth, line 146: Are you sure you want to make this a plural?

stevenengler: This is plural since the bitflags type represents a collection of signals. For example if we had `pub struct FileSignals: u32 { const A = 1, const B = 2 }`, then `A` and `B` would each be an individual signal, and `let foo: FileSignals = A | B` would be a collection of both signals `A` and `B`. Combining two independent signals doesn't make them a single signal (at least in the case of Shadow).

The `FileState` bitflags type is not plural since the state itself is a collection of flags. For example if a file is readable and writable (`READABLE | WRITABLE`), then both of those flags make up the file's "state". Maybe it could have been named `FileStateFlags` instead which would have been plural, and then the "state" would be a collection of "flags".

ppopth:
I have a concern on this design. Without signals, it's fine to accidentally add more than one listeners to the event source, but, with them, it's not. Let's look at the `refresh_listener` function of epoll.
```rust
    fn refresh_listener(&mut self, weak_self: Weak<AtomicRefCell<Epoll>>, key: Key) {
        let Some(entry) = self.monitoring.get_mut(&key) else {
            return;
        };

        // Check what state we need to listen for this entry.
        // We always listen for closed so we know when to stop monitoring the entry.
        let listen = entry.get_listener_state().union(FileState::CLOSED);
        let filter = StateListenerFilter::Always;

        // Set up a callback so we get informed when the file changes.
        let file = key.file().clone();
        let handle =
            file.borrow_mut()
                .add_listener(listen, filter, move |state, changed, cb_queue| {
                    if let Some(epoll) = weak_self.upgrade() {
                        epoll
                            .borrow_mut()
                            .notify_entry(&key, state, changed, cb_queue);
                    }
                });
        entry.set_listener_handle(Some(handle));
    }
```

This seems to me that I can call `refresh_listener` as many times as I want and the number of listeners added is the same as the number of times I call. Without signals, it's fine because you will just get excessive notifications, but with signals, you will be too many signals.

Can you make sure that all the listeners of all the event sources actually call `add_listener` only once? If not, I think we will probably need a new design.

stevenengler:
> Without signals, it's fine because you will just get excessive notifications, but with signals, you will be too many signals.

Each time `refresh_listener` is called, it adds a new listener and removes the old listener. When the handle returned by `add_listener` is dropped, the listener for that handle is removed. So when `entry.set_listener_handle(Some(handle))` is called, it replaces the old handle with the new handle, dropping the old handle and removing the old listener. So it should only receive a single notification, even if you call `refresh_listener` many times on the same key/entry.

ppopth:
> Each time `refresh_listener` is called, it adds a new listener and removes the old listener. When the handle returned by `add_listener` is dropped, the listener for that handle is removed. So when `entry.set_listener_handle(Some(handle))` is called, it replaces the old handle with the new handle, dropping the old handle and removing the old listener. So it should only receive a single notification, even if you call `refresh_listener` many times on the same key/entry.

Alright, got it. Thanks.

> I'm not completely on clear on the concern about multiple listeners receiving the same signal. And is the original "INPUT_BUFFER_PARITY" any different, I think you'd also see that state flag change on all listeners as well?

Yeah, it's probably not different. So, never mind :)

ppopth:
I think this PR is great. I will do the edged-triggered epoll after this PR is merged.

src/main/host/descriptor/mod.rs
```diff
@@ -241,35 +254,48 @@ impl StateEventSource {
...
+
+                let notify = match filter {
+                    // at least one monitored bit is on, and at least one has changed
+                    StateListenerFilter::OffToOn => flipped && on,
+                    // all monitored bits are off, and at least one has changed
+                    StateListenerFilter::OnToOff => flipped && !on,
+                    // at least one monitored bit has changed
+                    StateListenerFilter::Always => flipped,
+                    StateListenerFilter::Never => false,
+                };
```
robgjansen, line 285: I wonder if we could have just extended the `StateListenerFilter` enum to have e.g. a `StateListenerFilter::OffToOff` and a `StateListenerFilter::OnToOn` enum variants rather than separating signals into a separate construct. But perhaps separating the signals as you have done is more flexible and/or easier to understand in the code. I'm still fine merging as-is, but wanted to bring this up in case you hadn't considered it before but think it might be a simpler design to deal with in the code going forward.

robgjansen (approved):

src/main/host/descriptor/mod.rs
stevenengler: My reason against this was that it doesn't tell you anything about why the listener was notified. If the listener was called and given a state change of "nothing changed", you still wouldn't know why you were notified in the first place. You could interpret some meaning from it, but then you could accidentally add a seemingly inconsequential state change in a socket that then gets interpreted incorrectly. For example something like:

```rust
// should already be readable, but set the readable state flag here just to be sure
self.update_state(FileState::READABLE, true);
```

But now we'd be sending an "OnToOn" file state event, which might get interpreted as having some certain meaning in a listener, but the calling code didn't intend this.

Also the event source doesn't have a way to set an individual flag. So if you were only monitoring the "READABLE" flag with the "OffToOff" filter, but then the file changes only its "WRITABLE" flag, the listener watching the "READABLE" flag will still run even though the update has nothing to do with the "READABLE" flag since the state was technically updated.

robgjansen: Ok, makes sense to me - thanks for thinking this through!

PR #3258, opened 2023-12-30
Add futex helpers

stevenengler:
This adds a `libc_futex` helper to the simple latch code. Originally I was going to use the linux-api library so I added futex support to that, but then remembered that it won't work with miri. I included the linux-api code anyways since it might be useful in the future, and is a little nicer to use than the rustix equivalent since it takes a `&AtomicU32` instead of a `*mut u32`.

### DIFF
--- a/src/lib/linux-api/src/futex.rs
+++ b/src/lib/linux-api/src/futex.rs
@@ -1,6 +1,110 @@
-use crate::bindings;
+use core::sync::atomic::AtomicU32;
+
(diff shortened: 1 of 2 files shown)


robgjansen (approved):

robgjansen:
I've marked the CI test `tor-0.4.8.10` as optional since that has yet to be merged into main, and to avoid blocking the merging of this PR.

PR #3260, opened 2023-12-31
update profiling documentation

jtracey:
The documentation on profiling was very out of date, including references to build options that no longer exist, plugins, and elf-loader. I went ahead and re-wrote most of it to reflect good strategies for the current architecture.

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -23,6 +23,7 @@
     - [Performance Tuning]()
         - [Disabling Side-channel Mitigations](sidechannels.md)
         - [Parallel Simulations](parallel_sims.md)
+        - [Profiling](profiling.md)
     - [Stability Guarantees](semver.md)
     - [Non-goal: Security](security.md)
     - [Known limitations and workarounds](limitations.md)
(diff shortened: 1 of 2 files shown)


robgjansen (approved):
This looks really great! Thanks for taking the time to write this up. I definitely learned a few things and think this will be a good reference going forward :)

robgjansen:
@jtracey if you can rebase on main and force-push the update to your branch, I'll merge ASAP.

stevenengler:
> One limitation to note is that Shadow relies on spinlocks in barriers for some of its operation. Especially when running with many threads, these spinlocks will show as consuming most of the CPU anytime the simulation is bottlenecked on few simulated processes. Telling when this is happening can be difficult in these tools, because no symbol information is available.

Do you know which part of the code is spinning? As far as I know the only spinlock shadow still has in in the scheduler, but it would be good to know if there are any others. Also have you tried disabling the experimental [`use_worker_spinning`](https://shadow.github.io/docs/guide/shadow_config_spec.html#experimentaluse_worker_spinning) option? Does disabling it make profiling any easier?

robgjansen:
FYI, I'll take over the merging assignment now.

PR #3262, opened 2023-12-31
Fix bumped clang version check

stevenengler:
This seems like it was accidentally bumped during the shadow version bump in 172f88c7b.


robgjansen (approved):
Oops!

PR #3271, opened 2024-01-03
Reorganize some code

stevenengler:
Various files/modules have been moved around. These changes are all opinionated, so feel free to veto them.

Summary:

- removed the `main/core/support` directory (the contents are moved elsewhere)
- moved `main/core/main.rs` to `main/shadow.rs` (this contains the de facto main function, so I think it makes sense to have it in the top-most directory of the main shadow code)
- moved `main/host/syscall_foo` files to `main/host/syscall/foo`
- moved `SUID_DUMP_*` values to rust
- moved listener-related code in `host::descriptor` to a `host::descriptor::listener` module
- moved C syscall handlers in `main/host/syscall` to `main/host/syscall/handler`
- other small changes


robgjansen (approved):
I'm fine with the changes. This is a decent time to reorganize so if you notice other things you want to move too feel free :)

(I was actually thinking earlier today about moving all of the remaining C code from src/main into its own directory just to get it all in one place, but figured it would be too unpopular and we'd probably lose some context that we get from the file structure so stopped myself.)

stevenengler:
> (I was actually thinking earlier today about moving all of the remaining C code from src/main into its own directory just to get it all in one place, but figured it would be too unpopular and we'd probably lose some context that we get from the file structure so stopped myself.)

I don't have strong opinions on this, but we do still have a fair amount of C code and a lot of it is still intertwined with the rust code, so it might be worth waiting until we can better isolate the C code. It would be nice if we could just build a single C library and use it from rust without the C code also needing to access rust code.

PR #3273, opened 2024-01-03
Move the scheduler to its own library

stevenengler:
This moves the scheduler (both thread-per-core and thread-per-host) to its own library and adds some unit tests to the thread-per-host scheduler. This also involved making the thread-per-host scheduler generic, and moving the synchronization primitives in `main/utility/synchronization` to the new library.

The thread-per-host scheduler performance [didn't change](https://github.com/shadow/benchmark-results/blob/master/tor/2024-01-04-T03-31-42/README.md) (and maybe even became slightly better because of the `const` performance optimization).

image

The thread-per-core scheduler performance also [didn't change](https://github.com/shadow/benchmark-results/blob/master/tor/2024-01-04-T08-05-06/README.md).

run_time


src/main/core/scheduler/mod.rs
```diff
@@ -64,7 +64,7 @@ impl Scheduler {
 }
 
 pub enum SchedulerScope<'sched, 'pool, 'scope> {
-    ThreadPerHost(thread_per_host::SchedulerScope<'pool, 'scope>),
+    ThreadPerHost(thread_per_host::SchedulerScope<'pool, 'scope, Box<Host>>),
     ThreadPerCore(thread_per_core::SchedulerScope<'sched, 'pool, 'scope, Box<Host>>),
```
robgjansen: Just curious: why does `ThreadPerCore` need the `'sched` but `ThreadPerHost` does not?

src/Cargo.lock
```diff
@@ -1290,6 +1290,10 @@ dependencies = [
  "winapi-util",
 ]
 
+[[package]]
+name = "scheduler"
+version = "0.0.0"
```
robgjansen, line 1295: This is probably intended, but the version does not match the scheduler version `0.1.0` from `Cargo.toml`

src/lib/scheduler/src/sync/simple_latch.rs
```diff
@@ -1,7 +1,7 @@
 use std::sync::atomic::{AtomicU32, Ordering};
 use std::sync::Arc;
 
-use linux_api::errno::Errno;
+use nix::errno::Errno;
```
robgjansen, line 4: Isn't this going backwards? Ie, we have been wanting to remove dependency on nix. Though I know [our guidance](https://shadow.github.io/docs/guide/coding.html#platform-libc-and-linux-crates) is not strict. Not sure but would `rustix` make more sense than `nix` or `linux-api`?

robgjansen (approved):
❤️ I really (really!) like the added modularity and that we are able to add more unit tests (I feel like we don't have enough of those given Shadow's unique architecture).

src/main/core/scheduler/mod.rs
stevenengler: The thread-per-core scheduler stores the hosts in queues (which are owned by the scheduler), and the thread-per-host scheduler stores hosts in thread-locals (which have a static lifetime). So the thread-per-host scheduler doesn't need the scheduler lifetime.

```rust
// thread-per-core
pub struct SchedulerScope<'sched, 'pool, 'scope, HostType: Host>
where
    'sched: 'scope,
{
    thread_hosts: &'sched Vec<ArrayQueue<HostType>>,
    ...
```

```rust
// thread-per-host
pub struct SchedulerScope<'pool, 'scope, HostType: Host> {
    ...
    host_storage: &'static LocalKey<RefCell<Option<HostType>>>,
}
```

src/lib/scheduler/src/sync/simple_latch.rs
stevenengler: I don't think we necessarily want to remove shadow's dependency on nix, it's just that we shouldn't be using libc types (which nix wraps) in syscall handlers (since libc and syscall types are often different) or in the shim (because of re-entrant functions). But it's still okay to use in other parts of shadow.

> For code that runs in the shadow process, where we are acting as a "normal" program that wants to interact with the kernel, generally prefer the highest-level interface that provides the necessary functionality. e.g. when creating worker threads in Rust, we generally use std::thread; there's no reason to use one of the lower level crates. Occasionally we need some functionality not provided in std though, in which case it makes sense to drop down to one of the lower level crates.

Here the rust standard library doesn't really have a nice way to work with errno values, so the next highest-level type is the nix `Errno` type.

In this case we need to use libc for futexes anyway (to support miri), and I think it makes sense to have a libc-based errno implementation (the one in nix) since we're making the syscall through libc. We also already use nix for other parts of the scheduler (such as setting CPU affinity). Switching the errno type to nix from libc-api also lets us remove linux-api as a dependency of the scheduler since we don't use it for anything else.

src/Cargo.lock
stevenengler: This wasn't intended, but is fixed now.

src/lib/scheduler/src/sync/simple_latch.rs
robgjansen: OK, makes sense! Thanks for the explanations!

PR #3277, opened 2024-01-07
Don't update the file state in socket.c

### LINKED ISSUE #3274: Partial read triggers an event in Shadow, but not Linux
**Describe the issue**
When using epoll with EPOLLET flag set and the file under epoll is partially read (some is left in the buffer), the result of epoll_wait in the real Linux is different from the one in Shadow. This issue is a different one from https://github.com/shadow/shadow/issues/2673

**To Reproduce**
Add the following test to `src/test/epoll/test_epoll.rs` and run the test.
```rust
diff --git a/src/test/epoll/test_epoll.rs b/src/test/epoll/test_epoll.rs
index df0e67db2..f6b9849ce 100644
--- a/src/test/epoll/test_epoll.rs
+++ b/src/test/epoll/test_epoll.rs
@@ -4,6 +4,7 @@ use nix::errno::Errno;
 use nix::sys::epoll::{self, EpollFlags};
 use nix::unistd;
 
+use test_utils::socket_utils::{socket_init_helper, SocketInitMethod};
 use test_utils::{ensure_ord, set, ShadowTest, TestEnvironment};
 
 #[derive(Debug)]
@@ -329,6 +330,59 @@ fn test_ctl_invalid_op() -> anyhow::Result<()> {
     })
 }
 
+fn test_write_then_read() -> anyhow::Result<()> {
+    let (fd_client, fd_server) = socket_init_helper(
+        SocketInitMethod::Inet,
+        libc::SOCK_STREAM,
+        libc::SOCK_NONBLOCK,
+        /* bind_client = */ false,
+    );
+    let epollfd = epoll::epoll_create()?;
+
+    test_utils::run_and_close_fds(&[epollfd, fd_client, fd_server], || {
+        let mut event = epoll::EpollEvent::new(EpollFlags::EPOLLET | EpollFlags::EPOLLIN, 0);
+        epoll::epoll_ctl(
+            epollfd,
+            epoll::EpollOp::EpollCtlAdd,
+            fd_server,
+            Some(&mut event),
+        )?;
+
+        let timeout = Duration::from_millis(100);
+
+        let thread = std::thread::spawn(move || {
+            vec![
+                do_epoll_wait(epollfd, timeout, /* do_read= */ false),
+                // The second one is supposed to timeout.
+                do_epoll_wait(epollfd, timeout, /* do_read= */ false),
+            ]
+        });
+
+        // Wait for readers to block.
+        std::thread::sleep(timeout / 3);
+
+        // Make the read-end readable.
+        unistd::write(fd_client, &[0, 0])?;
+
+        // Wait and read some, but not all, from the buffer.
+        std::thread::sleep(timeout / 3);
+        unistd::read(fd_server, &mut [0])?;
+
+        let results = thread.join().unwrap();
+
+        // The first wait should have received the event
+        ensure_ord!(results[0].epoll_res, ==, Ok(1));
+        ensure_ord!(results[0].duration, <, timeout);
+        ensure_ord!(results[0].events[0], ==, epoll::EpollEvent::new(EpollFlags::EPOLLIN, 0));
+
+        // The second wait should have timed out with no events received.
+        ensure_ord!(results[1].epoll_res, ==, Ok(0));
+        ensure_ord!(results[1].duration, >=, timeout);
+
+        Ok(())
+    })
+}
+
 fn main() -> anyhow::Result<()> {
     // should we restrict the tests we run?
     let filter_shadow_passing = std::env::args().any(|x| x == "--shadow-passing");
@@ -340,6 +394,7 @@ fn main() -> anyhow::Result<()> {
     let mut tests: Vec<test_utils::ShadowTest<(), anyhow::Error>> = vec![
         ShadowTest::new("threads-edge", test_threads_edge, all_envs.clone()),
         ShadowTest::new("threads-level", test_threads_level, all_envs.clone()),
+        ShadowTest::new("write-then-read", test_write_then_read, all_envs.clone()),
         // in Linux these two tests have a race condition and don't always pass
         ShadowTest::new(
             "threads-level-with-late-read",

```
In Shadow, the second epoll_wait receives an event, but, in Linux, it doesn't.

 - OS and version: Ubuntu 22.04.3 LTS  
 - Kernel version: Linux thinkpad-t14 6.2.0-39-generic # 40~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Nov 16 10:53:04 UTC 2 x86_64 x86_64 x86_64 GNU/Linux  

 - Version: commit c133925f097edde86ef18e8f4835a00c9f696ce6
 - Which processes you are trying to run inside the Shadow simulation: the tests

**Additional context**

stevenengler:
In the old code when adding or removing data from a socket buffer, a socket.c function would update the file state (called the "status" in the C code), then a tcp.c function might also update the file state. For example `tcp_receiveUserData` would call `legacysocket_removeFromInputBuffer` which would update the state, and then `tcp_receiveUserData` would later also update the state. This would lead to situations where socket.c would remove the READABLE flag, and then tcp.c would immediately add it back. While the end result is that the flag would effectively appear unchanged, this would cause edge-triggered epoll to think that the file's state changed twice even though it effectively didn't, leading to #3274.

Since the TCP code is the only C socket code left, now the socket.c code no longer updates the file state (with an exception of `legacysocket_pullOutPacket`) and the tcp.c code is responsible for updating the file state.

The added test was authored by @ppoth in #3274.

We should also be testing other socket/file types, but I think this will be easier once the new `test_epoll_edge.rs` tests in #3243 are merged. I did test udp and unix sockets manually and the new test passes for them as well.

Closes #3274.

**Edit:** The runtime performance is equivalent, but the simulation results change, so I want to take another look at this. https://github.com/shadow/benchmark-results/blob/master/tor/2024-01-07-T22-11-55/README.md

**Edit 2:** I think it's expected that the simulation results would change. Before we would remove the `READABLE` flag and then add it back again, which would reset the epoll entry's priority each time a packet was pushed to a TCP socket. Now the priority is reset only when the event is collected. This means that epoll events will be processed in a different order than they previously were.

### DIFF
--- a/src/main/host/descriptor/socket.c
+++ b/src/main/host/descriptor/socket.c
@@ -20,6 +20,8 @@
 #include "main/routing/packet.h"
 #include "main/utility/utility.h"
 
+gsize _legacysocket_getOutputBufferSpaceIncludingTCP(LegacySocket* socket);
+
 static LegacySocket* _legacysocket_fromLegacyFile(LegacyFile* descriptor) {
     utility_debugAssert(legacyfile_getType(descriptor) == DT_TCPSOCKET);
     return (LegacySocket*)descriptor;
@@ -176,7 +178,19 @@ void legacysocket_dropPacket(LegacySocket* socket, const Host* host, Packet* pac
 /* functions implemented by socket */
 
 Packet* legacysocket_pullOutPacket(LegacySocket* socket, const Host* host) {
-    return legacysocket_removeFromOutputBuffer(socket, host);
+    Packet* packet = legacysocket_removeFromOutputBuffer(socket, host);
+
+    /* if packet was NULL, then the buffer shouldn't have changed so we can skip the check */
+    if (packet != NULL) {
+        /* we are writable if we now have space */
+        gsize space = _legacysocket_getOutputBufferSpaceIncludingTCP(socket);
+        gboolean is_active = legacyfile_getStatus((LegacyFile*)socket) & STATUS_FILE_ACTIVE;
+        if (space > 0 && is_active) {
+            legacyfile_adjustStatus((LegacyFile*)socket, STATUS_FILE_WRITABLE, TRUE, 0);
+        }
+    }
+
+    return packet;
 }
 
 Packet* legacysocket_peekNextOutPacket(const LegacySocket* socket) {
@@ -365,11 +379,6 @@ gboolean legacysocket_addToInputBuffer(LegacySocket* socket, const Host* host, P
             tracker, &compatSocket, socket->inputBufferLength, socket->inputBufferSize);
     }
 
-    /* we just added a packet, so we are readable */
-    if(socket->inputBufferLength > 0) {
-        legacyfile_adjustStatus((LegacyFile*)socket, STATUS_FILE_READABLE, TRUE, 0);
-    }
-
     return TRUE;
 }
 
@@ -394,11 +403,6 @@ Packet* legacysocket_removeFromInputBuffer(LegacySocket* socket, const Host* hos
             CompatSocket compatSocket = compatsocket_fromLegacySocket(socket);
             tracker_updateSocketInputBuffer(
                 tracker, &compatSocket, socket->inputBufferLength, socket->inputBufferSize);
-        }
-
-        /* we are not readable if we are now empty */
-        if(socket->inputBufferLength <= 0) {
-            legacyfile_adjustStatus((LegacyFile*)socket, STATUS_FILE_READABLE, FALSE, 0);
         }
     }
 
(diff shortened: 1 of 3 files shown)


stevenengler:
Hopefully this PR is correct. I looked it over and I think it's okay. The simulation runtime performance is the same, but the [results](https://github.com/shadow/benchmark-results/blob/master/tor/2024-01-11-T22-10-20/README.md) change slightly. I think this is expected.

run_time
ram_simtime
transfer_time_5242880 exit

robgjansen (approved):
Looks good! I agree that the epoll events should be processed in a different order than they previously were, which would cause the sim results to be nondeterministic.

PR #3281, opened 2024-01-12
Clean up file state type (`FileState` in rust, and previously `Status` in C)

stevenengler:
The rust `FileState` and C `Status` types represented the same thing but were defined separately (I don't think cbindgen supported bitflags types when `FileState` was first added). With this commit there's now only `FileState`.

(Context: The C code originally called it "status", but when first adding rust support to Shadow I named it "state" instead to be consistent with how Linux calls it. Now that we don't have as much C code, it's easier to move more C code to use "state" instead of "status".)


robgjansen (approved):
🙌

PR #3282, opened 2024-01-13
Move the `FutexTable` to rust

stevenengler:
More low-hanging fruit. The new futex table is ~40 lines of code plus ~90 lines of C compatibility code :)


robgjansen (approved):
Moving in the right direction though! Looking forward to a future PR that contains only deletion of a bunch of this type of legacy C compatibility code. :)

Issue #3275, opened 2024-01-07
The 'fork-linux' test is flaky


stevenengler:
I've noticed that the `fork-linux` test fails occasionally in the CI, so just making this issue to record the backtrace.

It might make sense to give this is the "flaky" label in CMake so that it's excluded in the CI. Or there might be an easy fix, I haven't looked into it.

```text
14/253 Testing: fork-linux
14/253 Test: fork-linux
Command: "/usr/bin/sh" "-c" "../../target/debug/test_fork --libc-passing"
Directory: /__w/shadow/shadow/build/src/test/clone
"fork-linux" start time: Jan 07 00:18 UTC
Output:
----------------------------------------------------------
Testing fork_via_clone_syscall-fork_runs... ✓
Testing fork_via_fork_syscall-fork_runs... ✓
Testing fork_via_libc-fork_runs... ✓
Testing clone-parent=true... ✓
Testing clone-parent=false... ✓
Testing test_child_change_session... ✓
Testing test_child_change_group... ✓
Testing test_exit_signal_normal_exit-SIGCHLD... ✓
Testing test_exit_signal_with_fatal_signal-SIGCHLD... ✓
Testing test_exit_signal_normal_exit-SIGUSR1...panic: PanicInfo { payload: Any { .. }, message: Some(Unexpected nanosleep result: Ok), location: Location { file: "test/clone/test_fork.rs", line: 323, col: 22 }, can_unwind: true, force_no_backtrace: false }
 ✗ (!(Ok(0) == Ok(1)))
Error: !(Ok(0) == Ok(1))

Stack backtrace:
   0: anyhow::error::<impl anyhow::Error>::msg
             at /github/home/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.78/src/error.rs:83:36
   1: test_fork::run_test_in_subprocess
             at /__w/shadow/shadow/src/test/clone/test_fork.rs:269:5
   2: test_fork::test_exit_signal_normal_exit
             at /__w/shadow/shadow/src/test/clone/test_fork.rs:276:5
   3: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/alloc/src/boxed.rs:2021:9
   4: test_utils::ShadowTest<T,E>::run
             at /__w/shadow/shadow/src/test/test_utils.rs:56:9
   5: test_utils::run_tests
[...]

PR #3283, opened 2024-01-16
test_fork: fix race condition under Linux

sporksmith:
Fixes https://github.com/shadow/shadow/issues/3275


sporksmith:
I wasn't able to reproduce the flaky failure locally, but I'm fairly confident this will fix it. It at least doesn't seem to hurt, other than slowing `fork-linux` down a bit.

Ideally we'd instead use something like `sigwaitinfo(2)` to atomically unblock the signal and wait for it to arrive instead of trying to ensure that the signal arrives after the parent has started sleeping, but we haven't implemented that in shadow yet.

robgjansen (approved):

Issue #3290, opened 2024-01-30
crash: rootedcell "Dropped without calling `safely_drop`"


jtracey:
**Describe the issue**
I can consistently crash shadow by running any tokio program with enough threads. I've seen this crash a few times on actual experiments with more modest thread counts, but it was infrequent enough that I hadn't been able to make a stable reproducer and so couldn't tell if it wasn't just a bad memory module or some such. Once I realized I could cause it with enough tokio threads, reproducing is easy, on multiple machines.

**To Reproduce**

Run any shadow experiment with the following Rust program:
```Rust
#[tokio::main(worker_threads = 20000)]
async fn main() {
    println!("Hello, world!");
}
```

where `20000` is just an arbitrarily huge number of threads. I've made a tiny reproducer branch, but it's basically just the above with the simplest shadow config:
https://github.com/jtracey/actions-sandbox/tree/shadow-bug

Outside Shadow, it takes a moment to run, but does run. Inside Shadow, it consistently crashes. Should be able to reproduce with `cargo install --path . && shadow shadow.config.yaml`.

**Operating System (please complete the following information):**
 - OS and version: post the output of `lsb_release -d`  
`Debian GNU/Linux 12 (bookworm)` (Qubes), `Ubuntu 22.04.2 LTS` (ripple's clonk)
 - Kernel version: post the output of `uname -a`  
 `6.1.57-1.qubes.fc37.x86_64`

**Shadow (please complete the following information):**
 - Version and build information: post the output of `shadow --show-build-info`
```
Shadow 3.1.0 — v3.1.0-99-gde06d5240 2024-01-26--14:49:37
GLib 2.74.6
Built on 2024-01-30--01:16:24
Built from git branch main
Shadow was built with PROFILE=debug, OPT_LEVEL=1, DEBUG=true, RUSTFLAGS="-C force-frame-pointers=y", CFLAGS="-std=gnu11 -O0 -ggdb -fno-omit-frame-pointer -Wreturn-type -Wswitch -DDEBUG"
```
 - Which processes you are trying to run inside the Shadow simulation:
above tokio toy ("shadow-bug")

**Additional context**
<details><summary>RUST_BACKTRACE=1 shadow.log</summary>

```
** Starting Shadow 3.1.0
[...]

PR #3291, opened 2024-01-31
Ensure clone handler drops RootedRc's on early exit

sporksmith:
Fixes https://github.com/shadow/shadow/issues/3290


robgjansen (approved):
Looks great, and much more failsafe :)

PR #3296, opened 2024-02-08
Use rust 1.76 features

stevenengler:
Makes some changes for the [1.76](https://blog.rust-lang.org/2024/02/08/Rust-1.76.0.html) release.

The biggest change is from switching `&foo as *const _` and `&mut foo as *mut _` to `std::ptr::from_ref(&foo)` and `std::ptr::from_mut(&mut foo)`. I think the functions will help to avoid errors when casting references to pointers, and make it more clear about what the code is doing. I may have missed some of them, but this should be most. We use a lot of reference to pointer and pointer to pointer casts, so hopefully I got these all right (the tests pass at least).

This also makes some other small changes and upgrades the rust versions used in the CI.

### DIFF
--- a/src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
+++ b/src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
@@ -444,7 +444,7 @@ pub mod export {
     ) -> *mut ShimShmemHostLock {
         let host = unsafe { host.as_ref().unwrap() };
         let mut guard = host.protected().lock();
-        let lock = &mut *guard as *mut _;
+        let lock = std::ptr::from_mut(&mut *guard);
         guard.disconnect();
         lock
     }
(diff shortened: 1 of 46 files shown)


src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
```diff
@@ -444,7 +444,7 @@ pub mod export {
     ) -> *mut ShimShmemHostLock {
         let host = unsafe { host.as_ref().unwrap() };
         let mut guard = host.protected().lock();
-        let lock = &mut *guard as *mut _;
+        let lock = std::ptr::from_mut(&mut *guard);
```
robgjansen, line 447: maybe double-check this one 😵

robgjansen (approved):
This wasn't a great task for a human, so hopefully the compiler and/or tests would have caught any mistakes.

src/lib/shadow-shim-helper-rs/src/shim_shmem.rs
stevenengler: Yeah this one made me think for a minute :) But the nice thing is that it wouldn't compile if any of the types were incorrect. I'll add some type annotations to better show what's going on.

PR #3301, opened 2024-02-23
Minor socket improvements

stevenengler:
More use of linux-api, and a change from `from_le_bytes` to `from_ne_bytes` in the netlink socket.


robgjansen (approved):
LGTM!

PR #3304, opened 2024-03-11
Reduce the build time of the tests

stevenengler:
When building the tests we also build extra stuff that we don't need, for example the linux-api C bindings. This PR reduces the build time of the tests from 66 seconds to 42 seconds (207 compilation units to 150) when built with `./setup build --test`, as documented in the report from `cargo build --timings`. The next most problematic crates are the linux-api, vasi-sync, formatting-nostd/toml-edit, and neli crates due to their use of syn and related proc macros, but there's not much we can do about those.

Partial graph of the time taken by the test crates after this PR:

1710128104_grim

### DIFF
--- a/src/lib/linux-api/build.rs
+++ b/src/lib/linux-api/build.rs
@@ -1,5 +1,7 @@
+#[cfg(feature = "c_bindings")]
 use shadow_build_common::ShadowBuildCommon;
 
+#[cfg(feature = "c_bindings")]
 fn run_cbindgen(build_common: &ShadowBuildCommon) {
     let base_config = build_common.cbindgen_base_config();
     let crate_dir = std::env::var("CARGO_MANIFEST_DIR").unwrap();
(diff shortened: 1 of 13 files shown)


src/lib/linux-api/build.rs
```diff
@@ -1,35 +1,38 @@
-use shadow_build_common::ShadowBuildCommon;
+fn main() {
```
sporksmith, line 1: Slight preference to keep `main` small by instead have the `cfg` guard separately for the `run_cbindgen` definition and the call-site in `main`

sporksmith (approved):
Nice!

src/lib/linux-api/build.rs
stevenengler: I added the `cfg` to the `use`, `run_cbindgen`, and scoped contents of `main`.

PR #3307, opened 2024-03-12
Show warning if Go 1.21.x is installed

### LINKED ISSUE #3267: Go tests segfault with Go 1.21 (Fedora 38 and 39)
On fedora 39, programs linked with `-linkshared` (which is required by shadow) segfault on simple io function calls. For example:

```bash
sudo docker run --rm -it fedora:39
```

```bash
dnf update -y
dnf install -y golang
go install -buildmode=shared -linkshared std
cd /tmp
cat <<EOF >> foo.go
package main

import (
    "fmt"
)

func main() {
    fmt.Println("hello")
}
EOF
go build -linkshared -o foo foo.go
./foo
```

```text
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x7f4ce37148b3]

goroutine 1 [running]:
os.(*File).write(...)
	/usr/lib/golang/src/os/file_posix.go:46
os.(*File).Write(0x0, {0xc000014088?, 0x6, 0x7f4ce3620145?})
	/usr/lib/golang/src/os/file.go:183 +0x53
fmt.Fprintln({0x403b68, 0x0}, {0xc000354720, 0x1, 0x1})
	/usr/lib/golang/src/fmt/print.go:305 +0x6f
fmt.Println(...)
	/usr/lib/golang/src/fmt/print.go:314
```

If built without the `-linkshared` flag, then it completes successfully.

For the shadow tests, this results in:

```text
The following tests FAILED:
	 60 - simple_http-shadow (Failed)
	 61 - gc-linux (Failed)
	 62 - gc-shadow (Failed)
	 64 - goroutines-linux (Failed)
	 65 - goroutines-shadow (Failed)
```

Adding an `os` import magically makes the example above work again:

```go
import (
    _ "os"
    "fmt"
)
```

stevenengler:
This adds a CMake warning if Go 1.21.x is installed when building Shadow. I don't see a nice way of enabling this check only when the `--extra` flag is used, so we need to always perform this check even if the `--extra` flag isn't used. For this reason I've made it a warning rather than an error. For example:

```text
$ ./setup build --test --extra
2024-03-12 23:39:35,595 [INFO] running 'cmake /shadow -DCMAKE_INSTALL_PREFIX=/root/.local -DCMAKE_BUILD_TYPE=Release -DSHADOW_TEST=ON -DCMAKE_PREFIX_PATH=' from '/shadow/build'
-- System name: Linux
-- System version: 5.15.0-91-generic
-- System processor: x86_64
-- CMAKE_MODULE_PATH = /shadow/cmake/
CMake Warning at CMakeLists.txt:82 (MESSAGE):
  Go version '1.21.7' is not supported.  Some extra tests will likely fail.
  See issue #3267.


-- SHADOW_VERSION_FULL=3.1.0
-- SHADOW_VERSION_MAJOR=3
-- SHADOW_VERSION_MINOR=1
-- SHADOW_VERSION_PATCH=0
```

This check is also fine if Go is not installed. In this case you won't see any message in the CMake output. For example:

```text
$  ./setup build --test --extra
2024-03-12 23:38:36,345 [INFO] running 'cmake /shadow -DCMAKE_INSTALL_PREFIX=/root/.local -DCMAKE_BUILD_TYPE=Release -DSHADOW_TEST=ON -DCMAKE_PREFIX_PATH=' from '/shadow/build'
-- System name: Linux
-- System version: 5.15.0-91-generic
-- System processor: x86_64
-- CMAKE_MODULE_PATH = /shadow/cmake/
-- SHADOW_VERSION_FULL=3.1.0
-- SHADOW_VERSION_MAJOR=3
-- SHADOW_VERSION_MINOR=1
-- SHADOW_VERSION_PATCH=0
```

Closes #3267.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,6 +77,13 @@ include(FindPkgConfig)
 # Keep consistent with Cargo files that pull in glib: `git grep glib -- '*/Cargo.toml'`
 pkg_check_modules(GLIB REQUIRED IMPORTED_TARGET glib-2.0>=2.58)
 
+## show a warning if Go is installed and is version 1.21.x
+execute_process(COMMAND bash -c "go version | { read _ _ v _; echo \${v#go}; }" OUTPUT_VARIABLE GO_VERSION OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_QUIET)
+if(GO_VERSION MATCHES "^1.21.")
+    # https://github.com/shadow/shadow/issues/3267
+    message(WARNING "Go version '${GO_VERSION}' is not supported. If running Shadow's extra tests, the Go-based extra tests will fail. See issue #3267.")
+endif()
+
 include(CheckSymbolExists)
 list(APPEND CMAKE_REQUIRED_INCLUDES ${GLIB_INCLUDE_DIRS})
 list(APPEND CMAKE_REQUIRED_LIBRARIES ${GLIB_LIBRARIES})


CMakeLists.txt
```diff
@@ -77,6 +77,13 @@ include(FindPkgConfig)
 # Keep consistent with Cargo files that pull in glib: `git grep glib -- '*/Cargo.toml'`
 pkg_check_modules(GLIB REQUIRED IMPORTED_TARGET glib-2.0>=2.58)
 
+## show a warning if Go is installed and is version 1.21.x
+execute_process(COMMAND bash -c "go version | { read _ _ v _; echo \${v#go}; }" OUTPUT_VARIABLE GO_VERSION OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_QUIET)
+if(GO_VERSION MATCHES "^1.21.")
+    # https://github.com/shadow/shadow/issues/3267
+    MESSAGE(WARNING "Go version '${GO_VERSION}' is not supported. Some extra tests will likely fail. See issue #3267.")
```
robgjansen, line 84: Since this is being checked even when `--extra` is not specified, perhaps we want to change the warning to indicate that tests will only fail if trying to run the extra tests?

Something like:

```diff
- Some extra tests will likely fail.
+ If running the extra tests, the go-based extra tests will likely fail.
```

robgjansen (approved):

CMakeLists.txt
stevenengler: Sounds good, changed it to "If running Shadow's extra tests, the Go-based extra tests will fail."

PR #3309, opened 2024-03-14
Add ability to add extra rust tests

stevenengler:
This will hopefully allow us to add rust tests that are only built when `--extra` is passed to `./setup build --test`.

New test bins can be built by adding `required-features = ["extra_tests"]` to the Cargo.toml. New tests can be run by adding a CMake test with `set_property(TARGET test-name PROPERTY EXCLUDE_FROM_ALL true)` and `add_dependencies(extra_tests test-name)`.


robgjansen (approved):
It might be worth documenting how we ought to be adding new extra tests after this is merged, so we can point at it in future PRs if needed.

stevenengler:
> It might be worth documenting how we ought to be adding new extra tests after this is merged, so we can point at it in future PRs if needed.

Yeah, I'd like to wait until we have one extra test using this method first, since I haven't actually written a test that uses this so I'm not 100% sure it will work how it's documented.

PR #3314, opened 2024-03-22
Move the `kill` syscall handler to rust

stevenengler:
I changed two `ENOSYS` return values to `ENOTSUP`. I've found it to be confusing when a shadow syscall handler returns `ENOSYS`, so I think `ENOTSUP` is more clear.

### DIFF
--- a/src/main/host/syscall/handler/signal.rs
+++ b/src/main/host/syscall/handler/signal.rs
@@ -1,8 +1,11 @@
+use linux_api::errno::Errno;
+use linux_api::signal::{siginfo_t, Signal};
 use shadow_shim_helper_rs::syscall_types::ForeignPtr;
 use syscall_logger::log_syscall;
 
 use crate::cshadow as c;
-use crate::host::syscall::handler::{SyscallContext, SyscallHandler};
+use crate::host::process::Process;
+use crate::host::syscall::handler::{SyscallContext, SyscallHandler, ThreadContext};
 use crate::host::syscall::types::SyscallError;
 
 impl SyscallHandler {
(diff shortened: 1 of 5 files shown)


robgjansen (approved):
Always happy to see a net reduction in code :)

src/main/host/syscall/handler/signal.rs
```diff
@@ -1,20 +1,82 @@
...
+            //
+            // Currently unimplemented, and unlikely to be needed in the context of a shadow
+            // simulation.
+            log::warn!("kill with pid=-1 unimplemented");
+            return Err(Errno::ENOTSUP.into());
+        } else if pid == 0 {
+            // kill(2): If pid equals 0, then sig is sent to every process in the process group of
+            // the calling process.
+            //
+            // Currently every emulated process is in its own process group.
```
sporksmith, line 33: Oops, this comment is wrong. I just filed https://github.com/shadow/shadow/issues/3315; could you update the comment and link to that?

src/main/host/syscall/handler/signal.rs
```diff
@@ -1,20 +1,82 @@
...
+            ctx.objs.process.id()
+        } else if pid < -1 {
+            // kill(2): If pid is less than -1, then sig is sent to every process in the process
+            // group whose ID is -pid.
+            //
+            // Currently every emulated process is in its own process group, where pgid=pid.
```
sporksmith, line 39: Same here

src/main/host/syscall/handler/signal.rs
```diff
@@ -1,20 +1,82 @@
...
+        };
+
+        if signal.is_realtime() {
+            log::warn!("Unimplemented signal {signal:?}");
+            return Err(Errno::ENOTSUP.into());
+        }
+
+        let sender_pid = objs.process.id().into();
+        let siginfo = unsafe { siginfo_t::peel(siginfo_t::new_for_kill(signal, sender_pid, 0)) };
+        let siginfo = unsafe { siginfo_t::wrap_ref_assume_initd(&siginfo) };
```
stevenengler, line 76: I realized that this can be simplified to just the following, so I've also made this change:

```rust
let siginfo = siginfo_t::new_for_kill(signal, sender_pid, 0);
```

src/main/host/syscall/handler/signal.rs
stevenengler: Added.

src/main/host/syscall/handler/signal.rs
stevenengler: Added.

PR #3320, opened 2024-04-01
Support lseek syscall on pipes

stevenengler:
`lseek()` on a pipe should return `ESPIPE`. I just added a C test in the existing `file.c` since it was simplest and keeps the `lseek` tests in one place.

### DIFF
--- a/src/main/host/syscall/handler/file.rs
+++ b/src/main/host/syscall/handler/file.rs
@@ -1,11 +1,14 @@
+use linux_api::errno::Errno;
 use linux_api::posix_types::kernel_mode_t;
 use shadow_shim_helper_rs::syscall_types::ForeignPtr;
 use syscall_logger::log_syscall;
 
 use crate::cshadow;
+use crate::host::descriptor::CompatFile;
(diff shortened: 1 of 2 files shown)


src/main/host/syscall/handler/file.rs
```diff
@@ -104,9 +107,35 @@ impl SyscallHandler {
...
+                    return Self::legacy_syscall(cshadow::syscallhandler_lseek, ctx)
+                        .map(Into::into);
+                }
+            }
+        };
+
+        match file.inner_file() {
+            File::Pipe(_) => Err(Errno::ESPIPE.into()),
+            _ => {
+                warn_once_then_debug!("lseek() is not implemented for non-pipes");
```
sporksmith, line 135: This is slightly inaccurate since it's also implemented for legacy file descriptor types. Maybe just "not implemented for this type"?

sporksmith (approved):

src/main/host/syscall/handler/file.rs
stevenengler: Updated to that string.

PR #3321, opened 2024-04-01
Implement alarm syscall

sporksmith:
Possibly needed by python's subprocess module and/or tor's bandwidth scanner. See #3306

### DIFF
--- a/src/main/host/syscall/handler/time.rs
+++ b/src/main/host/syscall/handler/time.rs
@@ -99,6 +99,55 @@ impl SyscallHandler {
         Ok(0.into())
     }
 
+    #[log_syscall(/* rv */ std::ffi::c_uint, /* seconds */ std::ffi::c_uint)]
+    pub fn alarm(
+        ctx: &mut SyscallContext,
+        seconds: std::ffi::c_uint,
+    ) -> Result<std::ffi::c_uint, SyscallError> {
+        // We return the remaining time of any previously armed timer, or zero if there is none.
+        let prev_remaining = ctx
+            .objs
+            .process
+            .realtime_timer_borrow()
+            .remaining_time()
+            .unwrap_or(SimulationTime::ZERO);
+        let prev_remaining = std::time::Duration::from(prev_remaining);
+        // Linux rounds to the *nearest* second, but for sub-second values always rounds *up*.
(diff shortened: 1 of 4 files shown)


src/main/host/syscall/handler/time.rs
```diff
@@ -99,6 +99,40 @@ impl SyscallHandler {
...
+    #[log_syscall(/* rv */ std::ffi::c_uint, /* seconds */ std::ffi::c_uint)]
+    pub fn alarm(ctx: &mut SyscallContext, seconds: std::ffi::c_uint) -> Result<u32, SyscallError> {
+        // We return the remaining time of any previously armed timer, or zero if there is none.
+        let prev_remaining = ctx
+            .objs
+            .process
+            .realtime_timer_borrow()
+            .remaining_time()
+            .unwrap_or(SimulationTime::ZERO);
+        let prev_remaining_secs = prev_remaining.as_secs();
```
stevenengler, line 111: The kernel seems to round to the nearest second whereas this truncates the sub-second value. The kernel also seems to never return 0 if there is a timer (ex: 1 ns results in a return value of 1 sec).

```c
  /*
	 * We can't return 0 if we have an alarm pending ...  And we'd
	 * better return too much than too little anyway
	 */
	if ((!it_old.it_value.tv_sec && it_old.it_value.tv_nsec) ||
	      it_old.it_value.tv_nsec >= (NSEC_PER_SEC / 2))
		it_old.it_value.tv_sec++;

```

src/main/host/syscall/handler/time.rs
```diff
@@ -99,6 +99,40 @@ impl SyscallHandler {
         Ok(0.into())
     }
 
+    #[log_syscall(/* rv */ std::ffi::c_uint, /* seconds */ std::ffi::c_uint)]
+    pub fn alarm(ctx: &mut SyscallContext, seconds: std::ffi::c_uint) -> Result<u32, SyscallError> {
```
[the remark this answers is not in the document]
sporksmith: Done

src/main/host/syscall/handler/time.rs
sporksmith: Oof, good catch! Done

PR #3326, opened 2024-04-25
Add support for Ubuntu 24.04 and Fedora 40, and drop Fedora 38

stevenengler:
Fedora 38 becomes EOL in two weeks (14 May 2024). We already dropped CI testing for this version.

Ubuntu 24.04 and Fedora 40 were just released. I also tested Shadow in an Ubuntu 24.04 VM and the standard tests passed, so the updated kernel (6.8.0-31-generic) should be fine.

The required CI tests will need to be updated before merging.


robgjansen (approved):
Looks good!

robgjansen:
I just added Fedora 40 and Ubuntu 24.04 as required CI checks. (We had already previously removed Fedora 38, so no change there.)

PR #3333, opened 2024-05-15
Set Cargo edition/publish in workspace, and change non-main crates to 0.0.0

stevenengler:
This means we only need to update the edition and publish options in a single place (the workspace Cargo.toml file). It's a little unfortunate that we need to specify `foo.workspace = true` in all crates, but it's not too bad.

Documentation for this feature: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table

Previously most of our libraries were version 0.1.0, and now they are changed to 0.0.0 to indicate that we don't actually update their version numbers. This shouldn't have any user-facing changes.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -17,5 +17,3 @@ values =
 [bumpversion:file:CMakeLists.txt]
 
 [bumpversion:file:src/main/Cargo.toml]
-
-[bumpversion:file:src/test/Cargo.toml]
--- a/src/Cargo.lock
+++ b/src/Cargo.lock
@@ -516,7 +516,7 @@ checksum = "0ce7134b9999ecaf8bcd65542e436736ef32ddca1b3e06094cb6ec5755203b80"
 
 [[package]]
 name = "formatting-nostd"
-version = "0.1.0"
+version = "0.0.0"
 dependencies = [
  "cc",
  "libc",
@@ -564,7 +564,7 @@ checksum = "d2fabcfbdc87f4758337ca535fb41a6d701b65693ce38287d856d1674551ec9b"
 
 [[package]]
 name = "gml-parser"
-version = "0.1.0"
+version = "0.0.0"
 dependencies = [
  "nom",
 ]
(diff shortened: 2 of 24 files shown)


robgjansen (approved):
I'm fine with the change.

It's a little weird to be bumping version numbers in crates to stay consistent with the Shadow version even when no code is changing in those crates. which I think is the intention here. But I don't think we want to start tracking separate version numbers for each crate, because that could get crazy.

Keeping the version number for each crate consistent with Shadow's version number is probably better than just keeping the crate versions at 0.1.0 indefinitely. At least that way it would be possible to reference different versions of the crate - even if many of them are identical, some of them won't be.

robgjansen:
Hmm, perhaps dropping the version number is actually a clearer signal that we are not actually tracking versions for the crates. But ultimately I don't have a strong opinion.

I wonder if @sporksmith has a preference?

sporksmith:
Sorry for the slow response -

Bumping all of the crates from 0.1.0 to 3.1.0 to match Shadow means making them all "stable" (post 1.0.0) in terms of semver. If we wanted to continue following semver, it'd imply that a semver-breaking change to an "internal" crate's API would require a major version bump for that crate. And that major version bump would be for the whole workspace if we're locking them all to the same version.

I guess we could just not follow semver for those crates. There's kind of no real consequences since we're not publishing them, but it might be confusing.

> Another alternative is to remove the version from all libraries, which gives them an implicit version of 0.0.0.

I like this idea. It makes it explicit that these are really not intended for consumption outside of shadow.

I'm also fine with our current strategy of leaving everything at 0.1.0. Is there any particular issue that came up?

stevenengler:
> I'm also fine with our current strategy of leaving everything at 0.1.0. Is there any particular issue that came up?

No particular issue, it just seemed weird to have them all permanently at 0.1.0, but I agree that having them match Shadow might be weird too. Either way they don't really follow semver. I think using 0.0.0 makes the most sense though, and it seems like everyone is okay with that approach.

I've changed it so that all libraries and the 'shadow-tests' crate use version 0.0.0. Only the edition/publish options are shared between all crates.

PR #3338, opened 2024-06-05
syscall handler mod: fix double-warn race condition

sporksmith:
Detect when another thread has raced to warn about a missing syscall number before us, ensuring we really do only warn once about a particular missing syscall number.

This doesn't seem like a substantial problem in practice, but looking at this code again it seems silly not to do this slightly better handling.

### DIFF
--- a/src/main/host/syscall/handler/mod.rs
+++ b/src/main/host/syscall/handler/mod.rs
@@ -581,6 +581,7 @@ impl SyscallHandler {
                 // only show a warning the first time we encounter this unsupported syscall
                 static WARNED_SET: RwLock<Option<HashSet<SyscallNum>>> = RwLock::new(None);
 
+                // First check with a (cheap) read-lock.
                 let has_already_warned = WARNED_SET
                     .read()
                     .unwrap()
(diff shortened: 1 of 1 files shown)


robgjansen (approved):

src/main/host/syscall/handler/mod.rs
```diff
@@ -581,23 +581,23 @@ impl SyscallHandler {
...
-                if !has_already_warned {
-                    // `insert()` returns `false` if the syscall num was already in the set.
-                    // This can happen if another thread added the syscall after we released
-                    // the read-lock, above, and took the write lock, below.
-                    WARNED_SET
+                // If it looks like we haven't already warned, add this to the
+                // already-warned set. Also detect the (rare) case that another
+                // thread already warned after we released the read-lock above.
+                let has_already_warned = has_already_warned
+                    || WARNED_SET
```
stevenengler, line 596: This seems backwards to me? `insert` should return true when it's added for the first time, which would mean `has_already_warned` should be false.

sporksmith: Doh, right. Will follow up with another fix.

Whenever I look at this  code I'm tempted to just use a Mutex instead of a RwLock since it *shouldn't* be on the hot path, but always end up deciding "oh well, this one last tweak should fix it once and for all anyway"

stevenengler: Yeah it originally used a mutex, but then was changed to RwLock during review. As far as I remember there wasn't a strong argument for either one.

I did refactor this into a separate module as part of https://github.com/shadow/shadow/pull/3332, but I'm not sure if/when I'll merge that.

PR #3339, opened 2024-06-05
syscall handler: fix inverted "warn once" logic

sporksmith:
Fixes a bug introduced in 0412e0af74f59790515a2cc6b7653005a98a4a63.


sporksmith:
I tried looking at the test output this time to validate the behavior, but am not seeing this line logged at all (at DEBUG nor WARN). I'll see if I can run with the extra tests...

stevenengler (approved):

PR #3348, opened 2024-06-08
Small changes for rust 1.79

stevenengler:
Rust 1.79 will be released in a few days, and allows us to make a few small improvements to Shadow's code. This also updates the CI to use rust 1.79 (CI expected to fail until 1.79 is released).


stevenengler:
While I think this would be a small improvement, it's not worth requiring a more recent rust version for.

PR #3349, opened 2024-06-08
Add `OnceSet` for unhandled syscall warnings

stevenengler:
This pulls the `OnceSet` commit out of #3332. It moves the locking correctness logic to its own type. This also adds changes for the following review comments from #3332:

- https://github.com/shadow/shadow/pull/3332#discussion_r1604012045
    > Nit: Maybe append
    >
    > ```rust
    > assert!(!set.insert("BAR".to_string()));
    > ```
    >
    > to make sure inserting still works after attempting to insert a duplicate?
- https://github.com/shadow/shadow/pull/3332#discussion_r1604020948
    > I think this is still replicating the `warn_once_then_debug` behavior though, right? If so, I would add that detail to the comment. And maybe a back reference to here from `warn_once_then_debug` so that if that function is ever changed we know to update this code too.

### DIFF
--- a/src/main/utility/once_set.rs
+++ b/src/main/utility/once_set.rs
@@ -1,0 +1,60 @@
+use std::collections::HashSet;
+use std::sync::RwLock;
+
+/// A [`HashSet`] that only allows insertions and uses interior mutablity. This allows it to be used
+/// in a global static.
+#[derive(Debug, Default)]
+pub struct OnceSet<T>(RwLock<Option<HashSet<T>>>);
+
+impl<T> OnceSet<T>
+where
+    T: std::cmp::Eq + std::hash::Hash,
+{
+    pub const fn new() -> Self {
+        Self(RwLock::new(None))
+    }
+
+    /// Insert `val` into the set. Returns `false` if `val` had previously been added to the set;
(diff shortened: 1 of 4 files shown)


src/main/utility/once_set.rs
```diff
@@ -0,0 +1,61 @@
...
+{
+    pub const fn new() -> Self {
+        Self(RwLock::new(None))
+    }
+
+    /// Insert `val` into the set. Returns `false` if `val` had previously been added to the set;
+    /// otherwise returns `true`.
+    pub fn insert(&self, val: T) -> bool {
+        // First check with a (cheap) read-lock.
+        let already_added = self
```
sporksmith, line 21: Optional nit - could get rid of the already_added` variable, simplifying this function a little bit.

```
  if self./*...*/.unwrap_or(false) {
    return false;
  }
  
  self.0./*...*/.insert(val)
}
```

sporksmith (approved):
Nice! With this modularized maybe we can apply this to other unhandled parameters such as socket options...

src/main/utility/once_set.rs
stevenengler: Good idea, added!

PR #3350, opened 2024-06-09
Small doc/comment improvements

stevenengler:
Some misc documentation and comment changes.

### DIFF
--- a/src/lib/scheduler/src/sync/thread_parking.rs
+++ b/src/lib/scheduler/src/sync/thread_parking.rs
@@ -109,20 +109,23 @@ impl ThreadUnparker {
 
     /// Unpark the assigned thread.
     pub fn unpark(&self) {
-        // TODO: Rust does not guarantee any synchronization between the thread that parks and the
-        // thread that unparks:
+        // NOTE: Rust now does guarantee some synchronization between the thread that parks and the
+        // thread that unparks, so the change to `ready_flag` should be seen by the parked thread:
         //
-        // https://doc.rust-lang.org/nightly/std/thread/fn.park.html#park-and-unpark
+        // https://doc.rust-lang.org/std/thread/fn.park.html#memory-ordering
         //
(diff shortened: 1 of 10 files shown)


src/lib/scheduler/src/sync/thread_parking.rs
```diff
@@ -119,10 +119,12 @@ impl ThreadUnparker {
         // > inefficient, implementation to make both park and unpark return immediately without doing
         // > anything.
         //
-        // There is no guarentee that the change to `ready_flag` will be visible after the parked
+        // There is no guarantee that the change to `ready_flag` will be visible after the parked
         // thread resumes, which may lead to a deadlock. In practice, rust currently does use an
         // atomic with release-acquire ordering, but this might be relaxed in the future.
         // https://github.com/rust-lang/rust/blob/21b246587c2687935bd6004ffa5dcc4f4dd6600d/library/std/src/sys_common/thread_parker/futex.rs#L21-L27
+        //
+        // TODO: Stop using thread parking/unparking in the scheduler and remove this code.
```
robgjansen, line 127: Maybe add a few more details/hints or link to an issue on Github? For example, does removing parking/unparking require a redesign, or is it relatively simple?

robgjansen (approved):

src/lib/scheduler/src/sync/thread_parking.rs
stevenengler: I rechecked the documentation, and they've now decided to give thread parking/unparking actual memory guarantees.

Previously the docs said:

> Notice that being unblocked does not imply any synchronization with someone that unparked this thread, [...]

Now the docs say:

> Calls to park synchronize-with calls to unpark, meaning that memory operations performed before a call to unpark are made visible to the thread that consumes the token and returns from park. [...]

I've replaced all of the text including the TODO with a comment that says this does correctly synchronize the memory.

(We could now probably replace the release/acquire operations on the `ready_flag` with relaxed operations, but I won't worry about that.)

Issue #3345, opened 2024-06-06
Migrate away from the nix crate


sporksmith:
Since we started using the nix crate, we've also:

* Started using the `rustix` crate for code that runs in the shim, since that allows us to avoid depending on libc, which in turn avoids some subtle bugs.
* Created our own `linux_api` crate for directly dealing with linux kernel APIs.

While there's nothing wrong with the `nix` crate per se, I don't think we have any usage of it that we couldn't replace with `rustix`, `linux_api`, or occasionally our own small wrappers around `libc`. Migrating away from it would reduce complexity and remove some maintenance burden (e.g. when it makes breaking changes to its APIs, which has been... not infrequent).

stevenengler:
I'm good with moving away from nix, but I think there are two downsides of using rustic/linux_api over nix/libc worth mentioning:

1. We need to keep in mind that the documentation (man pages) for the posix libc syscall wrappers do not always match with the Linux syscall of the same name, so calling `rustix::foo()` may behave differently from `libc:foo()` or `nix::foo()`.
2. Miri has special handling for libc syscall wrappers such as `libc::futex`. If we want to run custom synchronization primitives in Miri, they must use libc or nix and not rustix or linux_api.

So I agree that we'll sometimes still want to write rust wrappers around the libc functions.

PR #3351, opened 2024-06-10
Less nix

sporksmith:
Progress on https://github.com/shadow/shadow/issues/3345


sporksmith:
Obsoletes #3346 . This is the results of rearranging and squashing a bit to hopefully make the individual commits a bit more sensible.

src/main/host/syscall/handler/mman.rs
```diff
@@ -95,13 +95,22 @@ impl SyscallHandler {
...
-        if prot as u32 as u64 != prot {
-            warn_once_then_trace!("Ignoring truncated prot flags from mprotect: {prot}");
-        }
-
-        let prot = prot as i32;
+        let Some(prot) = ProtFlags::from_bits(prot) else {
+            for flag in ProtFlags::from_bits_retain(prot)
+                .difference(ProtFlags::all())
+                .iter()
+            {
```
stevenengler, line 102: I don't think the iter is needed here. The value that `iter()` is called on will only be unknown bit values, and `iter()` returns all unknown bit values together as a single value.

> Each yielded flags value will correspond to a defined named flag. Any unknown bits will be yielded together as a final flags value.

Since this iterator will only ever return a single value of all the unknown bits, I think it would be simpler just to log `ProtFlags::from_bits_retain(prot).difference(ProtFlags::all())` directly.

src/main/host/syscall/handler/mman.rs
```diff
@@ -136,20 +145,37 @@ impl SyscallHandler {
...
-        let flags = match MapFlags::from_bits(flags) {
-            Some(x) => x,
-            None => {
-                warn_once_then_debug!("Invalid mmap flags: {flags}");
-                MapFlags::from_bits_truncate(flags)
+        let Some(prot) = ProtFlags::from_bits(prot) else {
+            for flag in ProtFlags::from_bits_retain(prot)
+                .difference(ProtFlags::all())
+                .iter()
+            {
```
stevenengler, line 152: Same as above.

src/main/utility/childpid_watcher.rs
```diff
@@ -244,6 +248,8 @@ impl ChildPidWatcher {
         let mut inner = self.inner.lock().unwrap();
         let pidfd = rustix::process::pidfd_open(pid.into(), PidfdFlags::empty())
             .unwrap_or_else(|e| panic!("pidfd_open failed for {pid:?}: {e:?}"));
+        let flags = rustix::io::fcntl_getfd(&pidfd).unwrap();
+        rustix::io::fcntl_setfd(&pidfd, flags | FdFlags::CLOEXEC).unwrap();
```
stevenengler, line 252: I don't think you need to set the cloexec flag here, linux should do it automatically for `pidfd_open`:

> The  pidfd_open() system call creates a file descriptor that refers to the process whose PID is specified in pid.  The file descriptor is returned as the function result; the close-on-exec flag is set on the file descriptor.

src/main/utility/childpid_watcher.rs
```diff
@@ -249,11 +242,15 @@ impl ChildPidWatcher {
     /// an unrelated process with a recycled `pid`.
     pub fn register_pid(&self, pid: Pid) {
         let mut inner = self.inner.lock().unwrap();
-        let pidfd =
-            pidfd_open(pid).unwrap_or_else(|e| panic!("pidfd_open failed for {pid:?}: {e:?}"));
-
-        let event = EpollEvent::new(EpollFlags::EPOLLIN, pid.as_raw().try_into().unwrap());
-        self.epoll.add(&pidfd, event).unwrap();
+        let pidfd = rustix::process::pidfd_open(pid.into(), PidfdFlags::empty())
+            .unwrap_or_else(|e| panic!("pidfd_open failed for {pid:?}: {e:?}"));
```
stevenengler, line 250: I think it would be good to make the pidfd non-blocking. I have had bad experiences from epoll saying something is ready to read/write, and then the read/write syscall blocks. By adding it for the pidfd, we should catch any errors when the eventual waitid is run (if it were to block with a blocking pidfd, we'd never know there was an error since the thread would just stall, but if it's a non-blocking pidfd and waitid returns `EAGAIN` instead, we know there's a bug).

stevenengler (approved):
I added more optional comments than I meant to, so feel free to ignore them :)

src/main/host/syscall/handler/mman.rs
```diff
@@ -95,13 +95,22 @@ impl SyscallHandler {
...
+                log_once_per_value_at_level!(
+                    flag,
+                    ProtFlags,
+                    log::Level::Warn,
+                    log::Level::Debug,
+                    "Unrecognized prot flag: 0x{:x}",
```
stevenengler, line 108: I think this can be "Unrecognized prot flag: {:#x}" instead?

src/lib/linux-api/src/errno.rs
```diff
@@ -390,6 +390,58 @@ impl Errno {
         assert!(rv.0 as u32 == val);
         rv
     }
+
+    /// Get libc's errno (generally global for the current thread).
+    #[cfg(feature = "libc")]
+    pub fn from_libc_errno() -> Self {
+        // SAFETY: We check for NULL, but otherwise assume that libc gives
+        // us a safely accessible pointer.
+        let raw = *unsafe { libc::__errno_location().as_ref().unwrap() };
```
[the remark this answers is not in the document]
sporksmith: A *well behaved* signal handler is supposed to always restore the previous value of `errno` (according to `signal-safety(7)`), but yeah some extra resilience here wouldn't hurt. Done.

src/main/host/managed_thread.rs
```diff
@@ -137,35 +141,34 @@ impl ManagedThread {
                 // The process died before initializing the shim.
                 //
                 // Reap the dead process and return an error.
-                let status = nix::sys::wait::waitpid(native_pid, None).unwrap();
-                match status {
-                    nix::sys::wait::WaitStatus::Exited(pid, 127) if pid == native_pid => {
```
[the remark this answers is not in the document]
sporksmith: rustix's `waitpid` doesn't return the `pid`. This seems like an oversight TBH, but also really shouldn't be needed for this case.

src/main/host/managed_thread.rs
```diff
@@ -487,7 +490,7 @@ impl ManagedThread {
         // Alternatively we could use `set_tid_address` or `set_robust_list` to
         // be notified on a futex. Those are a bit underdocumented and fragile,
         // though. In practice this shouldn't have to loop significantly.
-        trace!("Waiting for native thread {native_pid}.{native_tid} to exit");
+        trace!("Waiting for native thread {native_pid:?}.{native_tid:?} to exit");
```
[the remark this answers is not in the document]
sporksmith: I mildly prefer not; the debug formatting is nice for these sort of log messages to make the type of the integer unambiguous. Maybe it'd be nice for cases like below where we care about formatting as exactly an integer, but in those cases I don't mind forcing the caller to do the explicit integer conversion.

src/main/host/memory_manager/memory_mapper.rs
```diff
@@ -26,15 +27,21 @@ use crate::utility::interval_map::{Interval, IntervalMap, Mutation};
...
-const HEAP_PROT: i32 = libc::PROT_READ | libc::PROT_WRITE;
-const STACK_PROT: i32 = libc::PROT_READ | libc::PROT_WRITE;
+// TODO: replace with a `const` if there's ever a const implementation of `|`
+fn heap_prot() -> ProtFlags {
+    ProtFlags::PROT_READ | ProtFlags::PROT_WRITE
+}
+// TODO: replace with a `const` if there's ever a const implementation of `|`
+fn stack_prot() -> ProtFlags {
+    ProtFlags::PROT_READ | ProtFlags::PROT_WRITE
+}
```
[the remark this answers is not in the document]
sporksmith: Nice; thanks

src/main/host/memory_manager/mod.rs
```diff
@@ -791,7 +792,7 @@ mod export {
     #[no_mangle]
     pub unsafe extern "C-unwind" fn memorymanager_new(pid: libc::pid_t) -> *mut MemoryManager {
         Box::into_raw(Box::new(unsafe {
-            MemoryManager::new(nix::unistd::Pid::from_raw(pid))
+            MemoryManager::new(Pid::from_raw(pid).unwrap())
         }))
     }
```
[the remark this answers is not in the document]
sporksmith: Done

src/main/host/syscall/handler/mman.rs
sporksmith: Done. Breaking it into the individual bits would be preferable, but I'll leave that yak alone for now.

src/main/host/syscall/handler/mman.rs
sporksmith: done

src/main/host/syscall/handler/mman.rs
sporksmith: done

src/main/utility/childpid_watcher.rs
sporksmith: Good catch; replaced this with a `debug_assert` to validate this property.

src/main/utility/macros.rs
```diff
@@ -43,6 +43,28 @@ macro_rules! log_once_at_level {
...
+
+            let level = if LOGGED_SET.insert($value) {
+                $lvl_once
+            } else {
+                $lvl_remaining
+            };
+            // NOTE: Some parts of shadow duplicate the "(LOG_ONCE)" string in their own log
+            // messages (for example `SyscallHandler::run_handler`). If we change this string
+            // here, we should change it in other places as well.
+            log::log!(level, "(LOG_ONCE) {}", format_args!($str $($x)*))
```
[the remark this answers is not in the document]
sporksmith: Done

src/main/utility/macros.rs
```diff
@@ -43,6 +43,28 @@ macro_rules! log_once_at_level {
     };
 }
 
+#[allow(unused_macros)]
+#[macro_export]
+macro_rules! log_once_per_value_at_level {
```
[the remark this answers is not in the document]
sporksmith: Done

sporksmith (commented):
Thanks for the careful review!

src/main/utility/childpid_watcher.rs
sporksmith: Seems like a good idea; done

PR #3361, opened 2024-07-02
Support fstat on pipes

### LINKED ISSUE #3319: Support fstat on pipes
Shadow's `fstat` syscall handler is in C, which only supports `RegularFile`s. This means that `fstat` in shadow currently returns `EBADF` for pipes. Shadow should properly handle pipe fds in the `fstat` syscall handler. This is used by Python when the subprocess module executes programs.

```bash
strace python3 -c 'import subprocess; subprocess.run(["ls", "-l", "/dev/null"], capture_output=True)'
```

```strace
pipe2([3, 4], O_CLOEXEC)                = 0
pipe2([5, 6], O_CLOEXEC)                = 0
fstat(3, {st_mode=S_IFIFO|0600, st_size=0, ...}) = 0
```

To implement:

1. Fill in the correct syscall arguments in `SyscallHandler::fstat` in `src/main/host/syscall/handler/file.rs`.
2. Figure out what the `struct stat` equivalent is in the linux kernel, and create bindings for it in linux-api library.
3. Figure out what values to return for each field in `struct stat`.
4. Add `fstat` syscall tests with pipes.

**Note:** The `SYS_fstat` syscall calls the kernel's `sys_newfstat` on x86-64, not `sys_fstat`: [`arch/x86/entry/syscalls/syscall_64.tbl#L16`](https://github.com/torvalds/linux/blob/2c71fdf02a95b3dd425b42f28fd47fb2b1d22702/arch/x86/entry/syscalls/syscall_64.tbl#L16), which should be [`fs/stat.c#L475`](https://github.com/torvalds/linux/blob/2c71fdf02a95b3dd425b42f28fd47fb2b1d22702/fs/stat.c#L475).

stevenengler:
This adds support for the fstat syscall on pipe fds. Not all of the fields of `struct stat` are set properly (for example the access time field, inode field, etc), but it is enough for python's `subprocess.run()` to work correctly. Shadow logs a warning in any case.

Linux installs two different `struct stat` structures, one in "asm/stat.h" and the other in "asm-generic/stat.h". I think "asm/stat.h" is the right one.

Closes #3319.

```yaml
processes:
- path: python3
  args: -c 'import subprocess; print(subprocess.run(["ls", "-l", "/dev/null"], capture_output=True).stdout)'
  start_time: 1
```

```text
b'crw-rw-rw-. 1 nobody nobody 1, 3 Jun 23  2024 /dev/null\n'
```


sporksmith (approved):
Cool!

PR #3378, opened 2024-07-31
Replace the 'syscall-logger' proc macro crate with a macro_rules macro

stevenengler:
I think the new macro_rules macro is simpler than the old proc macro, but still kind of weird. I think it's useful to have the logging definition directly before the syscall handler implementation, so I think some weirdness is unavoidable. This new version doesn't rewrite the syscall handler function and should hopefully be more robust.

One notable change with this new version is that it reads plugin memory *after* the syscall handler runs, rather than before. While reading plugin memory to show sycall arguments before the syscall handler runs seems like the more logical choice, I think reading plugin memory after the syscall handler runs is more useful and is what the strace program does. This makes it easier to see the value of out-parameters.

I don't have strong feelings about this change, but I do think it's more maintainable than the old version.


sporksmith (approved):
Nice! I agree both that this seems more maintainable and that logging the values after the handler has executed seems like an improvement.

robgjansen (commented):
I think the new `log_syscall!( ... )` format with one arg per line improves readability too :)

stevenengler:
I also checked the performance just to make sure:

run_time

PR #3379, opened 2024-08-02
Various improvements to syscall handlers

stevenengler:
Most of these changes involve changing argument/return types of syscall handlers.

- Made `SyscallHandler::legacy_syscall` nicer to call
- Fixed argument types for ioctl syscall handler
- Fixed argument types for fcntl syscall handler
- Improved the fcntl syscall handler
- Simplified mmap syscall handler error handling
- Improved getsockname/getpeername socket func return types
- Improved listen socket func return type
- Return `DescriptorHandle` directly from syscall handlers
- Changed syscall handlers to return `()` rather than 0


sporksmith (approved):
Nice!

PR #3382, opened 2024-08-05
Update kernel sources in linux-api

stevenengler:
This PR:
 - updates the bindgen version (no change to bindings)
 - updates linux-api to use new kernel definitions (hopefully I updated everything that needs to be updated)
 - updates the rseq syscall handler to work with different sized rseq structs

We don't know if the calling application is using a smaller or larger rseq struct than what we have in linux-api. The rseq struct is also no longer `Copy` or `Pod`, so we cannot use it directly with the memory manager. I added a `field_ptr` macro to get a pointer to a field in a slice of bytes. I also wrote a macro version that returns references rather than pointers, but it's quite a bit more complicated and doesn't work with unaligned data, so I think the pointer version is best.


sporksmith (approved):
I think if `struct rseq` were made to be still `Copy` and `Pod` (which may require somehow ignoring the last field?), we could still use it directly (after verifying  in the syscall handler that the length is at least the length of our definition).

I suppose bindgen dropped `Copy` as a result of the unsized array member at the end of the struct definition, though? And IIRC we currently require `Copy` for `Pod`...?

I suppose we could manually define our own `rseq` definition in the linux-api crate, without the unsized array member, and have tests to validate that it matches otherwise.

I think the way you have it is ok too, though. It's a little unwieldy to work with in the handler, but we don't need to do a lot with this one.

stevenengler:
> I think if `struct rseq` were made to be still `Copy` and `Pod` (which may require somehow ignoring the last field?), we could still use it directly (after verifying  in the syscall handler that the length is at least the length of our definition).

The length will be smaller than the length of our definition on any currently supported platform. They all still use the shorter rseq struct definition, so we couldn't cast it as our definition. We could maybe zero fill our definition, then copy the user provided data into our definition (for example if our struct is 20 bytes and theirs is 10, then zero-fill the 20 byte struct and copy in their 10 bytes). 

> I suppose bindgen dropped `Copy` as a result of the unsized array member at the end of the struct definition, though? And IIRC we currently require `Copy` for `Pod`...?

Yeah the Linux struct added the unsized array member I think to indicate that the struct is dynamically sized (there might be more data in the struct, but outside the range of `0..sizeof(struct rseq)`) so bindgen won't add `Copy` to those structs. And `Pod` requires `Copy`.

https://github.com/rust-lang/rust-bindgen/issues/1431

> I suppose we could manually define our own `rseq` definition in the linux-api crate, without the unsized array member, and have tests to validate that it matches otherwise.
> 
> I think the way you have it is ok too, though. It's a little unwieldy to work with in the handler, but we don't need to do a lot with this one. 

I feel like using the pointers is simpler than copying the struct definition, testing that it's correct, and maintaining it as the kernel headers change. At least for now. But if we need to add more to the rseq syscall handler in the future, then maybe yeah we should write our own definition that ignores the FAM.

sporksmith:
Oh right, I was thinking of the case that userspace was using a larger definition than ours, but I didn't think through the reverse.

Yeah I think your approach makes the most sense :)

stevenengler:
@sporksmith I realized that there was an issue in the rseq syscall handler (the pointers violated stacked borrow rules, which would have raised an error with miri if there were a test for the syscall handler). I added another commit which fixes this but isn't super nice and meant to get another review from you, but forgot that I had enabled auto-merge.

Anyways I think the code is correct now and has an extra test, but it's not very nice. You can see the change in the "Fix UB in last commit" commit. If you have any comments let me know and I will make a new PR to fix them. I'll also see about making a new PR with the more complicated macro, which should isolate all of the unsafe code to the macro so that we hopefully wouldn't need to worry about UB in the syscall handlers since it's not easy to test them with miri.

PR #3385, opened 2024-08-06
Replace 'field_ptr' macro with 'field_project'

stevenengler:
Instead of returning `*mut T` like in `field_ptr`, `field_project` returns a `&mut MaybeUninit<T>`. The macro can return mutable references to multiple fields at once, which means the caller doesn't need to worry about writing unsafe code or following stacked borrows rules; the compiler will ensure the macro is being used correctly.

A downside is that the macro needs to uphold a lot of rust's safety guarantees, such as making sure no fields are overlapping and are within bounds.

Example:

```rust
let mut foo: libc::nlmsghdr = shadow_pod::zeroed();
let foo_bytes = unsafe { shadow_pod::as_u8_slice_mut(&mut foo) };

let (foo_nlmsg_type, foo_nlmsg_flags) =
    field_project!(foo_bytes, libc::nlmsghdr, (nlmsg_type, nlmsg_flags)).unwrap();

// these are both mutable references to different fields within the foo_bytes slice
foo_nlmsg_type.write(10);
foo_nlmsg_flags.write(20);

assert_eq!(foo.nlmsg_type, 10);
assert_eq!(foo.nlmsg_flags, 20);
```

@sporksmith Do you think this approach is an improvement? Or do you have ideas for entirely different approaches? (Sorry about the bad diff formatting, I'm not sure how I can improve that. It might be easier to look at the code directly rather than the diff.)

### DIFF
--- a/src/main/host/syscall/handler/sched.rs
+++ b/src/main/host/syscall/handler/sched.rs
@@ -161,60 +161,16 @@ impl SyscallHandler {
         //
         // For now we just update to reflect that the thread is running on CPU 0.
 
-        Self::rseq_internal(rseq_bytes)?;
+        let Some((cpu_id, cpu_id_start)) = field_project!(rseq_bytes, rseq, (cpu_id, cpu_id_start))
+        else {
+            return Err(Errno::EINVAL);
+        };
+
+        cpu_id.write(CURRENT_CPU);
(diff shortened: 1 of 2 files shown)


sporksmith (approved):
The macro is a bit gnarly, but I think the stacked-borrows-compatible and generally nicer and safer interface is worth it.

Issue #3390, opened 2024-08-14
Shadow panics on trying to close tcp socket for exec


datacompboy:
**Describe the issue**
thread 'shadow-worker' panicked at ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/atomic_refcell-0.1.13/src/lib.rs:129:23:
already mutably borrowed

More below

**To Reproduce**
I can't trigger it reproducibly :(

**Operating System (please complete the following information):**
Ubuntu 22.04.4 LTS
5.15.153.1-microsoft-standard-WSL2

**Shadow (please complete the following information):**
Shadow built from 217887cd2fc506b550f10a98786d79e921fb1449 (chdir)
Reproduces with "-g" and without.
Can't share repro setup, unfortunately :(

**Additional context**

The full panic backtrace is below.

The panic is triggered at frame 20 `inetsocket_hasDataToSend` at  src/main/host/descriptor/socket/inet/mod.rs:704:9
when it tries to `socket.borrow().has_data_to_send()`.

The socket comes it seems at frame 60, which, apparently is call for the close on already borrowed ref, no?
```
  60:     0x559ada81733b - shadow_rs::host::descriptor::socket::inet::legacy_tcp::LegacyTcpSocket::close::h108a311beebe05a6
                               at ~/shadow/src/main/host/descriptor/socket/inet/legacy_tcp.rs:229:9
  61:     0x559ada81733b - shadow_rs::host::descriptor::socket::inet::InetSocketRefMut::close::hba7fbbe2b03ffff7
                               at ~/shadow/src/main/utility/enum_passthrough.rs:27:38
```

and the chain of calls is started from frame 71, as part of lambda inside of `Thread::update_for_exec`.

At this point it's beyond my understanding how that supposed to work and what went wrong, but all I know it fails on trying to exec() one binary from another binary. 

*Sigh*.

Any ideas where to look / how to unwind this?

```
thread 'shadow-worker' panicked at ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/atomic_refcell-0.1.13/src/lib.rs:129:23:
already mutably borrowed

stack backtrace:
   0:     0x559ada2d6ba5 - std::backtrace_rs::backtrace::libunwind::trace::h58eed11393533053
                               at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/../../backtrace/src/backtrace/libunwind.rs:116:5
[...]

PR #3391, opened 2024-08-16
Add missing `with_global_cb_queue` wrappers

stevenengler:
This is likely causing #3390. We should have been calling `with_global_cb_queue` in `Thread::update_for_exec` to initialize the global thread-local callback queue so that callbacks made from C won't occur while we already have the rust socket borrowed. I also added `with_global_cb_queue` in a few other places where it was missing.

### DIFF
--- a/src/main/host/syscall/handler/unistd.rs
+++ b/src/main/host/syscall/handler/unistd.rs
@@ -109,7 +109,11 @@ impl SyscallHandler {
         if let Some(replaced_desc) = replaced_desc {
             // from 'man 2 dup2': "If newfd was open, any errors that would have been reported at
             // close(2) time are lost"
-            CallbackQueue::queue_and_run(|cb_queue| replaced_desc.close(ctx.objs.host, cb_queue));
+            crate::utility::legacy_callback_queue::with_global_cb_queue(|| {
+                CallbackQueue::queue_and_run(|cb_queue| {
+                    replaced_desc.close(ctx.objs.host, cb_queue)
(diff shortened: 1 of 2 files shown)


src/main/host/syscall/handler/unistd.rs
```diff
@@ -109,7 +109,11 @@ impl SyscallHandler {
         if let Some(replaced_desc) = replaced_desc {
             // from 'man 2 dup2': "If newfd was open, any errors that would have been reported at
             // close(2) time are lost"
-            CallbackQueue::queue_and_run(|cb_queue| replaced_desc.close(ctx.objs.host, cb_queue));
+            crate::utility::legacy_callback_queue::with_global_cb_queue(|| {
```
sporksmith, line 112: I wonder if we can make it harder to forget to do this. Would it make sense for `queue_and_run` to set up and tear down the legacy callback queue?

sporksmith (approved):

src/main/host/syscall/handler/unistd.rs
stevenengler: I thought about that originally, but at the time I wanted to keep them separate so that it would be easy to rip out the legacy code when we didn't have C file objects anymore, and so that this global stuff would be explicit and not a hidden side effect when using a `CallbackQueue`. But it might make sense now to put it in `CallbackQueue` anyways.

I'll try it and if it works out okay then I'll make a separate PR for that.

Issue #3395, opened 2024-08-19
"Alternate stack already in use." -- any ideas how to debug?


datacompboy:
**Describe the issue**
I am trying to get large system under the shadow to shake the tree and see where fruits fall. I've wrapped the parts of system under https://bitbucket.org/sivann/restarter.git to get things auto-restarted and some also auto-killed. That working well, until recently I've got to the condition, when shadow stuck.

Digging into stuck state shown shadow waiting for zombie, and the zombie has the "Alternate stack already in use." panic in shimlog; the syscall: tgkill(self, self, 15).

**To Reproduce**
I can't reliable reproduce it under reduced setup (yet).  :(

**Operating System (please complete the following information):**
 - OS:	Ubuntu 22.04.4 LTS
 - Kernel: 5.15.153.1-microsoft-standard-WSL2
 - Shadow: Some local patches over freshly pulled origin/main daa5b470969226322a50e110f3f02b6e9e7d1517
 - The failing one is -- some golang binary under restarter mentioned above

**Additional context**

The panic itself:

```
thread '<unnamed>' panicked at lib/shim/src/signals.rs:195:17:
Alternate stack already in use.
stack backtrace:
   0:     0x7ffff7f58c95 - std::backtrace_rs::backtrace::libunwind::trace::h1a07e5dba0da0cd2
                               at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/std/src/../../backtrace/src/backtrace/libunwind.rs:105:5
   1:     0x7ffff7f58c95 - std::backtrace_rs::backtrace::trace_unsynchronized::h61b9b8394328c0bc
                               at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x7ffff7f58c95 - std::sys_common::backtrace::_print_fmt::h1c5e18b460934cff
                               at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/std/src/sys_common/backtrace.rs:68:5
   3:     0x7ffff7f58c95 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h1e1a1972118942ad
                               at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/std/src/sys_common/backtrace.rs:44:22
   4:     0x7ffff7f7566b - core::fmt::rt::Argument::fmt::h07af2b4071d536cd
[...]

PR #3396, opened 2024-08-19
Allow signal handlers to switch to in-use stack

sporksmith:
This *can* be safe in some circumstances, and golang appears to do it in its default SIGTERM handler.

Fixes https://github.com/shadow/shadow/issues/3395

### DIFF
--- a/src/test/golang/CMakeLists.txt
+++ b/src/test/golang/CMakeLists.txt
@@ -94,3 +94,10 @@ add_shadow_tests(
       # This test can take a bit longer in debug builds
       TIMEOUT 40
       LABELS golang)
+
+add_golang_test_exe(BASENAME test_sigterm)
+add_shadow_tests(
+    BASENAME sigterm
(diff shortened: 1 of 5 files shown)


src/test/golang/CMakeLists.txt
```diff
@@ -93,4 +93,11 @@ add_shadow_tests(
...
-      LABELS golang)
\ No newline at end of file
+      LABELS golang)
+
+add_golang_test_exe(BASENAME test_sigterm)
+add_shadow_tests(
+    BASENAME sigterm
+    CONFIGURATIONS extra
+    PROPERTIES
+      LABELS golang)
```
stevenengler, line 103: Should there be is this missing a test?

stevenengler (approved):

src/test/golang/CMakeLists.txt
sporksmith: Sure was!

PR #3401, opened 2024-09-05
verify_plugin_path: support scripts

### LINKED ISSUE #3400: support interpreter scripts
from execve(2):

```
       pathname must be either a binary executable, or a script starting with a line of the
       form:

           #!interpreter [optional-arg]
```

We currently don't support the latter. In particular,  `verify_plugin_path` rejects such scripts since the file is not itself a dynamically linked ELF.

Supporting this would allow us to directly execute scripts from shadow's config file, instead of having to explicitly invoke the relevant interpreter.

More importantly, it'd allow programs that attempt to *themselves* execute other scripts (such as a shell script that invokes other shell scripts) to work correctly unmodified. (I ran into this issue when attempting to run chutney under shadow; I'm not sure that itself is really a great idea, but it'd be nice to remove this blocker in any case)

I initially thought to support this we might need to rewrite the `execve` args ourselves to explicitly invoke the correct interpreter with our LD_PRELOAD etc. However, I just tried commenting out the calls to `verify_plugin_path` and that's enough to get things working correctly, so it looks like `LD_PRELOAD` is applied to the interpreter when invoking a script directly in this way.

So I think all we really need to do here is update `verify_plugin_path` to recognize such scripts and validate that the *interpreter* is a dynamically linked ELF.

sporksmith:
We can handle scripts using the "shebang line" mechanism, provided that the specified interpreter is a dynamically linked ELF (or another script whose interpreter is a dynamically linked ELF, etc).

Fixes #3400

### DIFF
--- a/src/main/utility/mod.rs
+++ b/src/main/utility/mod.rs
@@ -25,7 +25,8 @@ pub mod syscall;
 pub mod units;
 
 use std::collections::HashSet;
-use std::ffi::CString;
+use std::ffi::{CString, OsStr};
+use std::io::Read;
 use std::marker::PhantomData;
 use std::os::unix::fs::{DirBuilderExt, MetadataExt};
 use std::os::unix::prelude::OsStrExt;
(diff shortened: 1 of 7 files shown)


src/main/utility/mod.rs
```diff
@@ -362,70 +383,112 @@ impl std::fmt::Display for VerifyPluginPathError {
...
     // but be owned by a different user), but it should catch most errors
     let mask = libc::S_IXUSR | libc::S_IXGRP | libc::S_IXOTH;
     if (metadata.mode() & mask) == 0 {
+        log::debug!("{path:?} not executable");
         return Err(VerifyPluginPathError::NotExecutable);
     }
+    // Get up to PATH_MAX bytes; that should be enough to ensure we get the
+    // interpreter name where applicable.
+    let mut reader = std::io::BufReader::with_capacity(linux_api::limits::PATH_MAX, file);
+    let buf = reader.fill_buf()?;
```
stevenengler, line 413: I think `fill_buf` only makes a single `read` syscall and doesn't loop, so it won't necessarily read as many bytes as we want. It might be better to do `file.take(PATH_MAX).read_to_end()` instead?

stevenengler (approved):

src/main/utility/mod.rs
sporksmith: Thanks; yeah that's better

src/main/utility/mod.rs
```diff
@@ -362,70 +383,112 @@ impl std::fmt::Display for VerifyPluginPathError {
...
+    // Verify and strip "shebang"
+    let mut header = header.strip_prefix(b"#!")?;
+    // Skip any spaces. (Other whitespace isn't skipped AFAIK).
+    while header.first() == Some(&b' ') {
+        header = &header[1..];
+    }
+    // The path is the next contiguous set of non-space or newline characters.
+    let interp_path = match header.splitn(2, |b| b == &b' ' || b == &b'\n').next() {
+        Some(i) => i,
+        None => header,
```
sporksmith, line 467: Changing this to return `None` in this case instead of returning a *prefix* of the interpreter path.

PR #3402, opened 2024-09-06
Adding a commit to force a CI run

sporksmith:
Before submitting your pull request, please see our documentation about contributing to Shadow:

https://shadow.github.io/docs/guide/contributing.html

### DIFF
--- a/src/Cargo.lock
+++ b/src/Cargo.lock
@@ -1592,6 +1592,7 @@ dependencies = [
  "rustix",
  "signal-hook",
  "static_assertions",
+ "tempfile",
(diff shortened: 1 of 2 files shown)


stevenengler:
Is the CI failure caused by just adding tempfile?!

sporksmith:
It seems so! I'm guessing the tempfile crate does something with thread local storage that is incompatible with my... questionable thread-local-storage initialization in the `clone_test`. Ironically, it passes under shadow.

It'll probably take some digging to fix the test. I might as well look into it though or this could mysteriously bite us again...

PR #3411, opened 2024-09-12
Drop packet instead of panicking if dst IP is unknown

### LINKED ISSUE #3148: Shadow panics if a packet is sent to an unknown IP address
If a packet is sent to an IP address that is not 127.0.0.1 or an IP address in the configuration file, Shadow will panic.

```yaml
general:
  stop_time: 10s

network:
  graph:
    type: 1_gbit_switch

hosts:
  client:
    network_node_id: 0
    processes:
    - path: python3
      args:
      - -c
      - |
        import socket
        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
        s.sendto(b"hello", ("1.2.3.4", 80))
```

```text
thread 'shadow-worker' panicked at 'No host ID for dest address {dst_ip}', main/core/worker.rs:356:18
thread 'shadow-worker' panicked at 'assertion failed: self.shim_shmem_lock.borrow().is_none()', main/host/host.rs:996:9
thread '<unnamed>' panicked at 'A work thread panickedfatal runtime error: thread local panicked on drop
', main/core/scheduler/pools/unbounded.rs:193:17
```

We should probably just drop the packet instead of panicking. But we should also inform the user in some way, probably through a warn-level log message. It would probably be best to track which IP/port addresses we've already warned about so we don't flood the log with warnings. And also since this will show for some packets which should be delivered, but shadow doesn't support (such as 127.0.0.0/8, excluding 127.0.0.1).

stevenengler:
Closes #3148. Also added a regression test.

This was really easy to add now that we have `log_once_per_value_at_level!()`.

```
[WARN] [testnode:11.0.0.1] [worker.rs:354] [shadow_rs::core::worker] (LOG_ONCE) Packet has destination 13.62.123.22 which doesn't exist in the simulation. Dropping the packet.
```

### DIFF
--- a/src/main/core/worker.rs
+++ b/src/main/core/worker.rs
@@ -320,6 +320,9 @@ impl Worker {
         .unwrap();
     }
 
+    /// The packet will be dropped if the packet's destination IP is not part of the simulation (no
+    /// host has been configured for the IP).
+    ///
     /// # Safety
     ///
     /// `packet` must be valid and not accessed by another thread while this function is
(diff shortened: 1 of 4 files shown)


src/main/core/worker.rs
```diff
@@ -346,12 +349,17 @@ impl Worker {
...
 
-        let dst_host_id = Worker::with(|w| {
-            w.shared
-                .resolve_ip_to_host_id(dst_ip)
-                .expect("No host ID for dest address {dst_ip}")
-        })
-        .unwrap();
+        let Some(dst_host_id) = Worker::with(|w| w.shared.resolve_ip_to_host_id(dst_ip)).unwrap()
+        else {
+            log_once_per_value_at_level!(
```
sporksmith, line 354: Should this also set the delivery status to `cshadow::_PacketDeliveryStatusFlags_PDS_INET_DROPPED` as done for random packet drops below?

sporksmith (approved):
Nice use of python for a succinct regression test :)

src/main/core/worker.rs
stevenengler: Good catch! Added.

PR #3414, opened 2024-09-18
DNS: Allow assignment of restricted IPs

sporksmith:
The comment that was here indicated intent to allow restricted IP addresses as long as they were unique (or local), but the code rejected them.

I *think* we want to reject any non-unique + non-local IP address assignment, and don't otherwise care if it's in a restricted range.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ Instead of specifying `{path: '/usr/bin/python3', args:
 provided it has an appropriate "shebang" line (e.g. `#!/usr/bin/env python3`).
 Likewise execution of such scripts is now supported inside the simulation when
 spawning processes via `execve`.
+* Relaxed restriction on assigning restricted IPs (such as 192.168.0.*).
+Within the simulation these are treated as fully routable IPs, so are required
+to be unique as with any other IP address assignment. (#3414)
 
 PATCH changes (bugfixes):
 
--- a/src/main/routing/dns.c
+++ b/src/main/routing/dns.c
@@ -81,31 +81,6 @@ static gboolean _dns_isIPInRange(const in_addr_t netIP, const gchar* cidrStr) {
 }
 
 /* Address must be in network byte order. */
-static gboolean _dns_isRestricted(DNS* dns, in_addr_t netIP) {
-    /* http://en.wikipedia.org/wiki/Reserved_IP_addresses#Reserved_IPv4_addresses */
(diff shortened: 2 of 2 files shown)


src/main/routing/dns.c
```diff
@@ -120,10 +120,11 @@ Address* dns_register(DNS* dns, HostId id, const gchar* name, in_addr_t requeste
...
-    /* restricted is OK if this is a localhost address, otherwise it must be unique */
+    /* non-unique are OK only if this is a localhost address */
+    /* TODO: Support other localhost addresses; e.g. 127.0.0.2? */
     if (requestedIP == address_stringToIP("127.0.0.1")) {
         isLocal = TRUE;
-    } else if (_dns_isRestricted(dns, requestedIP) || !_dns_isIPUnique(dns, requestedIP)) {
+    } else if (!_dns_isIPUnique(dns, requestedIP)) {
         gchar* ipStr = address_ipToNewString(requestedIP);
         warning("Invalid IP %s (restricted: %s, unique: %s)", ipStr,
                 _dns_isRestricted(dns, requestedIP) ? "true" : "false",
```
robgjansen, line 130: I think you want to remove this call to `_dns_isRestricted()` in the warning log message because your change makes that irrelevant and inconsistent.

After that, there are no callers left to `_dns_isRestricted()` and so you probably want to delete that function implementation on lines 83-107.

robgjansen (approved):
I'm not seeing any harm in being more flexible in IP address assignment, as long as all addresses are routable in the network graph.

There are some changes needed, but I don't need to review them afterwords.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ Instead of specifying `{path: '/usr/bin/python3', args:
 provided it has an appropriate "shebang" line (e.g. `#!/usr/bin/env python3`).
 Likewise execution of such scripts is now supported inside the simulation when
 spawning processes via `execve`.
+* Fixed a bug preventing assignment of restricted IPs (such as 192.168.0.*). (#3414)
```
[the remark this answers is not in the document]
sporksmith: Ah, the code comment made it sound like it was already intended to be allowed, but maybe it was the comment that was wrong and not the code :). (Or I was misinterpreting it)

src/main/routing/dns.c
sporksmith: Done

sporksmith:
Thanks!

PR #3421, opened 2024-10-06
Fix nightly warnings

stevenengler:
Fixes `elided_named_lifetimes`, `static_mut_refs`, and `non_local_definitions` warnings.

[Apparently](https://github.com/rust-lang/rust/issues/120363) they are considering making non-local `impl` definitions a hard error in the future (so we won't be able to `allow(non_local_definitions)` anymore), which would be unfortunate since we wouldn't be able to have our syscall logger macros at the beginning of the syscall handler functions anymore. But this wouldn't take effect until 2028, so I'm happy to ignore this.

For `static_mut_refs` we can make it a little nicer using raw references when they release in rust 1.82 in ~10 days.

### DIFF
--- a/src/lib/gml-parser/src/parser.rs
+++ b/src/lib/gml-parser/src/parser.rs
@@ -42,7 +42,7 @@ fn take_verify<'a, E: GmlParseError<'a>>(
 }
 
 /// Parse a GML key.
-pub fn key<'a, E: GmlParseError<'a>>(input: &'a str) -> IResult<&str, &str, E> {
+pub fn key<'a, E: GmlParseError<'a>>(input: &'a str) -> IResult<&'a str, &'a str, E> {
     // a key starts with the a character [a-zA-Z_], and has remaining characters [a-zA-Z0-9_]
     let take_first = take_verify(1, |chr| is_alphabetic(chr as u8) || chr == '_');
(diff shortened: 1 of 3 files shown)


robgjansen (approved):
Their plan to make non-local impl defs a hard error is troubling. I know we have a couple of years but still 🤯

stevenengler:
> Their plan to make non-local impl defs a hard error is troubling. I know we have a couple of years but still 🤯 

Yeah I don't think I'm a fan of the change; the way we use it is a bit of a hack, but there aren't any nice alternatives. A good thing is that it would only be in the 2027 edition, so if we didn't update the edition number in the `Cargo.toml`, it should still theoretically build fine past that date.

Also it's only the syscall logger macros, so if we needed to we could easily drop the syscall logging from shadow.

sporksmith:
> > Their plan to make non-local impl defs a hard error is troubling. I know we have a couple of years but still 🤯
> 
> Yeah I don't think I'm a fan of the change; the way we use it is a bit of a hack, but there aren't any nice alternatives. A good thing is that it would only be in the 2027 edition, so if we didn't update the edition number in the `Cargo.toml`, it should still theoretically build fine past that date.
> 
> Also it's only the syscall logger macros, so if we needed to we could easily drop the syscall logging from shadow.

+1 that would be a bummer. I remember I was a bit skeptical about adding it in the first place, but it's been SUPER nice to have.

PR #3441, opened 2024-11-18
Support Go netlink sockets

ppopth:
When you get the interfaces in Go using `net.Interfaces()`, it uses netlink sockets to do so.

However, the messages sent by Go sometimes are too short and just ignore some fields.

For example, the RTM_GETLINK message sent by Go looks like this: [17, 0, 0, 0, 18, 0, 1, 3, 1, 0, 0, 0, 0, 0, 0, 0, 0]

You can see that `nlmsghdr` took 16 bytes already and it left one byte at the end as a payload.

In fact, the length should be at least 32 not 17 (16 bytes for `nlmsghdr` and 16 bytes for `ifinfomsg`).

This PR pads the message with zeroes until it reaches the minimum length.

I'm not sure how Linux handles it, but I think it's not worth spending time to investigate more.

Before submitting your pull request, please see our documentation about contributing to Shadow:

https://shadow.github.io/docs/guide/contributing.html


src/main/host/descriptor/socket/netlink.rs
```diff
@@ -689,8 +689,28 @@ impl InitialState {
...
-                RTM_GETADDR => self.handle_ifaddrmsg(common, &packet_buffer[..]),
+                RTM_GETLINK => {
+                    let min_len =
+                        std::mem::size_of::<nlmsghdr>() + std::mem::size_of::<ifinfomsg>();
+                    // Pad the message if it's too short and update the len field
+                    if packet_buffer.len() < min_len {
+                        packet_buffer.resize(min_len, 0);
+                        packet_buffer[memoffset::span_of!(nlmsghdr, nlmsg_len)]
+                            .copy_from_slice(&(min_len as u32).to_le_bytes()[..]);
+                    }
```
stevenengler, line 700: I think it's better to only pad the bytes if we have at least an entire `nlmsghdr`. So I think something like:

```rust
let nlmsghdr_len = std::mem::size_of::<nlmsghdr>();
let ifinfomsg_len = std::mem::size_of::<ifinfomsg>();
if (nlmsghdr_len..(nlmsghdr_len + ifinfomsg_len)).contains(packet_buffer.len()) { ... }
```

If this ends up being too restrictive, then we can change it in the future.

src/main/host/descriptor/socket/netlink.rs
```diff
@@ -689,8 +689,28 @@ impl InitialState {
...
-                RTM_GETLINK => self.handle_ifinfomsg(common, &packet_buffer[..]),
-                RTM_GETADDR => self.handle_ifaddrmsg(common, &packet_buffer[..]),
+                RTM_GETLINK => {
+                    let min_len =
+                        std::mem::size_of::<nlmsghdr>() + std::mem::size_of::<ifinfomsg>();
+                    // Pad the message if it's too short and update the len field
+                    if packet_buffer.len() < min_len {
+                        packet_buffer.resize(min_len, 0);
+                        packet_buffer[memoffset::span_of!(nlmsghdr, nlmsg_len)]
+                            .copy_from_slice(&(min_len as u32).to_le_bytes()[..]);
```
stevenengler, line 699: Should this be `to_ne_bytes`?

src/test/golang/test_netlink.go
```diff
@@ -0,0 +1,31 @@
...
+	"runtime"
+)
+
+func println(args ...interface{}) {
+	// Runtime error on go 1.17
+	// https://github.com/golang/go/issues/47873
+	if (runtime.Version() != "go1.17") {
+		fmt.Println(args...);
+	}
+}
```
stevenengler, line 15: The go test is great, but it would also be nice to have a rust test that tests the syscall specifically with the shorter buffer. I think you could re-use your `flood` test to construct a `Nlmsghdr`, shrink the `nl_len` field, and call `sendto` with the smaller buffer.

src/main/host/descriptor/socket/netlink.rs
stevenengler, line 700: Can you add a `debug!()` call here to say that we're zero-filling the buffer?

And can you also add a comment saying that we typically try not to zero-fill structs when the bytes are missing, but it should be okay here since we don't yet support most of the fields of `ifaddrmsg`?

stevenengler:
One of the issues with using third-party libraries for handling kernel data structures that we've continually run into in Shadow is that they're always designed for sending data to the kernel, whereas in Shadow we need to receive data from applications. Most rust libraries are not designed to handle edge cases like this. This is common in the networking syscalls (especially socket addresses). I think there are a few other syscalls too where we need to be able to handle fewer bytes than the size of the struct.

In general, I think we try not to add "fake zero bytes" since they may have some unintended meaning. Instead it's usually better to only use the fields if they're given (if they're within the bounds given by `len`) rather than pretending that they're zero. For example in the rseq syscall handler, the size of the `rseq` struct can change between kernel versions so we use the `field_project` macro for this:

https://github.com/shadow/shadow/blob/79e9fd6f72ee00be6921ba1bcf1278a34a261868/src/main/host/syscall/handler/sched.rs#L164-L167

(We have a test using this with `nlmsghdr`)

https://github.com/shadow/shadow/blob/79e9fd6f72ee00be6921ba1bcf1278a34a261868/src/main/utility/macros.rs#L348-L358

But I don't think neli gives us this option. Enough bytes need to be given to fill the entire [`Ifaddrmsg`](https://docs.rs/neli/latest/neli/rtnl/struct.Ifaddrmsg.html) struct. So I think zero-filling the struct is the only option here.

stevenengler (changes requested):

stevenengler:
@sporksmith Do you agree with my comments here? We've discussed this topic (fewer bytes than the size of the struct) before I think, but I don't remember what your thoughts were.

sporksmith:
> @sporksmith Do you agree with my comments here? We've discussed this topic (fewer bytes than the size of the struct) before I think, but I don't remember what your thoughts were.

Yup, what you wrote makes sense to me

src/test/netlink/test_ifaddrmsg.rs
```diff
@@ -0,0 +1,217 @@
...
+        libc::socket(
+            libc::AF_NETLINK,
+            libc::SOCK_RAW | libc::SOCK_NONBLOCK,
+            libc::NETLINK_ROUTE,
+        )
+    };
+
+    // buffer is 15 bytes long which is shorter than nlmsghdr which is 16 bytes long
+    // 22 is RTM_GETADDR
+    let buffer: Vec<u8> = vec![15, 0, 0, 0, 22, 0, 1, 3, 185, 42, 24, 254, 0, 0, 0];
```
stevenengler, line 96: Can you build this as a `nlmsghdr` struct (either using neli, libc, or linux-api) and then just truncate the buffer afterwards to <16 bytes? That will make it easier to tell that the nlmsghdr was constructed properly.

src/test/netlink/test_ifaddrmsg.rs
```diff
@@ -0,0 +1,217 @@
...
+        libc::socket(
+            libc::AF_NETLINK,
+            libc::SOCK_RAW | libc::SOCK_NONBLOCK,
+            libc::NETLINK_ROUTE,
+        )
+    };
+
+    // buffer is 17 bytes long which is shorter than nlmsghdr+ifaddrmsg which is 24 bytes long
+    // 22 is RTM_GETADDR
+    let buffer: Vec<u8> = vec![17, 0, 0, 0, 22, 0, 1, 3, 185, 42, 24, 254, 0, 0, 0, 0, 0];
```
stevenengler, line 140: Like above, can you build this as a `nlmsghdr`+`ifaddrmsg` struct and then truncate?

src/test/netlink/test_ifinfomsg.rs
```diff
@@ -0,0 +1,217 @@
...
+        libc::socket(
+            libc::AF_NETLINK,
+            libc::SOCK_RAW | libc::SOCK_NONBLOCK,
+            libc::NETLINK_ROUTE,
+        )
+    };
+
+    // buffer is 15 bytes long which is shorter than nlmsghdr which is 16 bytes long
+    // 18 is RTM_GETLINK
+    let buffer: Vec<u8> = vec![15, 0, 0, 0, 18, 0, 1, 3, 185, 42, 24, 254, 0, 0, 0];
```
stevenengler, line 96: Like above, can you build this as a `nlmsghdr` struct and then truncate?

src/test/netlink/test_ifinfomsg.rs
```diff
@@ -0,0 +1,217 @@
...
+        libc::socket(
+            libc::AF_NETLINK,
+            libc::SOCK_RAW | libc::SOCK_NONBLOCK,
+            libc::NETLINK_ROUTE,
+        )
+    };
+
+    // buffer is 17 bytes long which is shorter than nlmsghdr+ifinfomsg which is 32 bytes long
+    // 18 is RTM_GETLINK
+    let buffer: Vec<u8> = vec![17, 0, 0, 0, 18, 0, 1, 3, 185, 42, 24, 254, 0, 0, 0, 0, 0];
```
stevenengler, line 140: Like above, can you build this as a `nlmsghdr`+`ifinfomsg` struct and then truncate?

stevenengler (changes requested):
Looks good, thanks for the tests! I have one last change request in the comments, then I think it's ready. Then you can rebase on main, squash, and I'll merge it.

ppopth:
@stevenengler Can you confirm that it's looking good? and then I will rebase and squash it.

stevenengler:
> @stevenengler Can you confirm that it's looking good? and then I will rebase and squash it.

Yep looks good! Thanks for the changes. It's ready for rebase and squash.

stevenengler (approved):

PR #3442, opened 2024-11-18
Add real-world values for our uname return strings

### LINKED ISSUE #3440: Providing a standard 'Linux' OS name instead of 'shadowsys'
**Describe the issue**

Followup issue to #3435 

Java has 'shadowsys' as the `os.name` system property when running under Shadow instead of a standard `Linux` string. Some libraries with platform specific components may depend on this property and fail if the OS name is unknown.

The workaround would be supply this system property manually via JVM CLI option `-Dos.name=Linux`

TBH I'm not quite sure where this value is retrieved from (@stevenengler mentioned `uname`) 

**To Reproduce**

Run the following simple Java class under Shadow `java OsName.java` :
```java
public class OsName {
    public static void main(String[] args) {
        System.out.println(System.getProperty("os.name"));
    }
}
```

**Operating System:**
Please check the #3435 

**Shadow:**
Please check the #3435

robgjansen:
Closes #3440

### DIFF
--- a/src/main/host/syscall/handler/unistd.rs
+++ b/src/main/host/syscall/handler/unistd.rs
@@ -933,10 +933,11 @@ impl SyscallHandler {
 
         let nodename = u8_to_i8_slice(ctx.objs.host.info().name.as_bytes());
 
-        let sysname = u8_to_i8_slice(&b"shadowsys"[..]);
-        let release = u8_to_i8_slice(&b"shadowrelease"[..]);
-        let version = u8_to_i8_slice(&b"shadowversion"[..]);
-        let machine = u8_to_i8_slice(&b"shadowmachine"[..]);
+        // Currently hardcoded with values reported in Debian 12
+        let sysname = u8_to_i8_slice(&b"Linux"[..]);
+        let release = u8_to_i8_slice(&b"6.1.0-25-amd64"[..]);
+        let version = u8_to_i8_slice(&b"#1 SMP PREEMPT_DYNAMIC Debian 6.1.106-3 (2024-08-26)"[..]);
+        let machine = u8_to_i8_slice(&b"x86_64"[..]);
 
         name.sysname[..sysname.len()].copy_from_slice(sysname);
         name.nodename[..nodename.len()].copy_from_slice(nodename);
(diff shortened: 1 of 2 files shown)


src/main/host/syscall/handler/unistd.rs
```diff
@@ -933,10 +933,11 @@ impl SyscallHandler {
...
         let nodename = u8_to_i8_slice(ctx.objs.host.info().name.as_bytes());
 
-        let sysname = u8_to_i8_slice(&b"shadowsys"[..]);
-        let release = u8_to_i8_slice(&b"shadowrelease"[..]);
-        let version = u8_to_i8_slice(&b"shadowversion"[..]);
-        let machine = u8_to_i8_slice(&b"shadowmachine"[..]);
+        // Currently hardcoded with values reported in Debian 12
+        let sysname = u8_to_i8_slice(&b"Linux"[..]);
+        let release = u8_to_i8_slice(&b"6.1.0-25-amd64"[..]);
+        let version = u8_to_i8_slice(&b"#1 SMP PREEMPT_DYNAMIC Debian 6.1.106-3 (2024-08-26)"[..]);
```
sporksmith, line 939: `SMP` and `PREEMPT_DYNAMIC` seem a little unfortunate since we emulate a single-core machine with no preemption. Unless we have an example of a "real" string without those I suppose it's probably better to just leave them though.

sporksmith (approved):

src/main/host/syscall/handler/unistd.rs
robgjansen: This might be a configure option when the kernel is built. It looks like we could at least just drop the `PREEMPT_DYNAMIC` part

https://www.reddit.com/r/linuxquestions/comments/sc14th/change_kernel_version_during_build_uname_v/

and we could probably dig for more examples.

sporksmith: Yeah, IDK. Dropping `PREEMPT_DYNAMIC` seems more truthful, but also as likely to cause problems with code trying to do sniffing here than solve them since it's probably "weird".

I'm fine with either way. I think I'd lean towards just keeping what you have, and if we run into more related problems making it configurable.

Issue #1719, opened 2021-10-14
Redesign tracker in Rust and write stats to file


robgjansen:
From #1629:

Tracker log messages look like this:

```
00:00:07.393281 [thread-531] 00:00:02.000000000 [message] [peer678~11.0.2.166] [tracker.c:457] [_tracker_logNode] [shadow-heartbeat] [node] 1,2037126,2143726,0.000000,0,0.000000;0,0,0,0,0,0,0,0,0,0,0,0;0,0,0,0,0,0,0,0,0,0,0,0;1911,2037126,0,0,0,0,1911,80262,1956864,0,0,0;2011,2143726,0,0,0,0,2011,84462,2059264,0,0,0
```

This is unreadable and the code is using a C counter type. We should be using the newer rust counter type for counting things like number of packets and bytes sent/recv in the last second, and write this data to a structured stats file rather than the log file.

robgjansen:
See #1399

Issue #1450, opened 2021-06-17
Reduce number of log messages when run with default settings


stevenengler:
Shadow runs with a default log level of "info", which can lead to a large number of log messages. We should consider increasing the default log level so that we print fewer log messages by default.

The user can use the `general.heartbeat_info` option to have Shadow print periodic messages to show that Shadow is still running and not stuck. The general heartbeat messages currently log usage stats, but if we implement #1399, we can reuse these general heartbeat messages to log the simulation progress (current-time, current-simulation-time, % (based on current-sim-time/stop-time)).

Issue #1399, opened 2021-06-03
Consider storing heartbeat log information as structured data rather than log messages


stevenengler:
Shadow logs heartbeat messages containing various simulation data (for example bytes sent/received, bytes allocated/deallocated, CPU usage, retransmit header sizes, etc).

https://github.com/shadow/shadow/blob/main/docs/2-Getting-Started-Tutorial.md#gathering-statistics

All of this information is periodically written to the log file as a line of comma-delimited numbers. These log lines are [difficult to understand](https://github.com/shadow/shadow/discussions/1130), difficult to parse by external tools, and make the log file harder to read. Since some heartbeat messages are logged once per host, there can be many of these lines in larger simulations.

We should instead consider writing this simulation information to a dedicated file in `shadow.data/`, or writing it to host-specific files in `shadow.data/hosts/hostname/`. We could write this information in a dedicated data format such as [streaming JSON](https://en.wikipedia.org/wiki/JSON_streaming) to make it easier to parse, and optionally compress it on the fly.

We may still wish to keep the simulation heartbeat in the log file so that you can tell if a simulation has stalled.

sporksmith:
Another option might be to change the current log-line to be JSON. That'd make it human readable and easier to parse, though might blow up the log size (particularly in contrast to your proposal of a *compressed* auxiliary json file).

sporksmith:
> Another option might be to change the current log-line to be JSON. That'd make it human readable and easier to parse, though might blow up the log size (particularly in contrast to your proposal of a compressed auxiliary json file).

I wonder if it's worth thinking about adding an option for doing [streaming compression](https://docs.rs/flate2/1.0.20/flate2/write/struct.DeflateEncoder.html) for the *whole* log. That'd mitigate the large log-sizes in general, and make logging json records more practical.

stevenengler:
> I wonder if it's worth thinking about adding an option for doing streaming compression for the whole log. That'd mitigate the large log-sizes in general, and make logging json records more practical.

Since the log messages are to stdout, I think we could just pipe to something like gzip first and then to a file. If we want to build that in to the logger, maybe we should write log messages directly to a file rather than to stdout (for example with a `--log-file` option).

PR #3446, opened 2024-11-23
Remove tracker

### LINKED ISSUE #800: Segementation fault when logging ram usage
**Describe the issue**
When running with the option "-i ram", shadow crashes with a segmentation fault during/after shutdown.
"node" and "socket" are working.

**To Reproduce**
Follow the readme to install shadow and tgen. Run the example in "resource/examples" with 

`shadow -i ram shadow.config.xml`

Debian GNU/Linux 10 (buster)
Linux debian 4.19.0-9-amd64 1 SMP Debian 4.19.118-2 (2020-04-29) x86_64 GNU/Linux

Fresh install. Only installed packages from the readme.

 - Version: post the output of `shadow --version`  
Shadow v1.13.2-608-g035872e6 2020-05-11 (built 2020-05-12) running GLib v2.58.3 and IGraph v0.7.1
tgen v1.0.0

**Additional context**
Output:

```
[..]
00:00:22.514973 [thread-0] 00:59:59.000000000 [message] [server~11.0.0.1] [tracker.c:560] [_tracker_logRAM] [shadow-heartbeat] [ram] 1,37123,37123,321864,143,0
00:00:22.514980 [thread-0] 00:59:59.000000000 [message] [client~11.0.0.2] [tracker.c:560] [_tracker_logRAM] [shadow-heartbeat] [ram] 1,0,0,318083,87,100
00:00:22.515097 [thread-0] n/a [message] [n/a] [scheduler.c:108] [_scheduler_stopHosts] starting to shut down 2 hosts
00:00:22.515101 [thread-0] n/a [message] [server~11.0.0.1] [process.c:1301] [process_stop] terminating main thread of process 'server.tgen.1000'
00:00:22.515104 [thread-0] n/a [message] [server~11.0.0.1] [process.c:887] [_process_executeCleanup] cleaning up process 'server.tgen.1000': aborting 0 auxiliary threads and calling 1 atexit functions
00:00:22.515350 [thread-0] n/a [message] [server~11.0.0.1] [host.c:319] [host_shutdown] host 'server' has been shut down, total execution time was 10.544271 seconds
00:00:22.515473 [thread-0] n/a [message] [client~11.0.0.2] [host.c:319] [host_shutdown] host 'client' has been shut down, total execution time was 10.845080 seconds
00:00:22.515477 [thread-0] n/a [message] [n/a] [scheduler.c:110] [_scheduler_stopHosts] 2 hosts are shut down
00:00:22.515482 [thread-0] n/a [message] [n/a] [master.c:441] [master_run] log message buffering is disabled during cleanup
00:00:22.515485 [thread-0] n/a [message] [n/a] [master.c:445] [master_run] simulation finished, cleaning up now
00:00:22.515496 [thread-0] n/a [message] [n/a] [scheduler.c:234] [scheduler_shutdown] scheduler is shutting down now
00:00:22.515499 [thread-0] n/a [message] [n/a] [scheduler.c:244] [scheduler_shutdown] waiting for 0 worker threads to finish
00:00:22.515506 [thread-0] n/a [message] [n/a] [scheduler.c:315] [_scheduler_free] 0 worker threads finished
00:00:22.515523 [thread-0] n/a [message] [n/a] [slave.c:240] [slave_free] ObjectCounter: counter values: task_new=755638 task_free=755638 event_new=755638 event_free=755638 packet_new=518009 packet_free=518009 payload_new=155136 payload_free=155136 router_new=2 router_free=2 host_new=2 host_free=2 netiface_new=4 netiface_free=4 process_new=2 process_free=2 descriptor_new=105128 descriptor_free=105128 channel_new=4 channel_free=4 tcp_new=201 tcp_free=201 udp_new=0 udp_free=0 epoll_new=6 epoll_free=6 timer_new=104917 timer_free=104917 
00:00:22.515529 [thread-0] n/a [message] [n/a] [slave.c:241] [slave_free] ObjectCounter: counter diffs: task=0 event=0 packet=0 payload=0 router=0 host=0 netiface=0 process=0 descriptor=0 channel=0 tcp=0 udp=0 epoll=0 timer=0 
00:00:22.515546 [thread-0] n/a [message] [n/a] [topology.c:1280] [_topology_clearCache] path cache cleared, spent 0.000000 seconds computing 0 shortest paths with dijkstra, and 0.000000 seconds computing 0 shortest self paths
00:00:22.515568 [thread-0] n/a [message] [n/a] [master.c:130] [master_free] simulation master destroyed
00:00:22.515571 [thread-0] n/a [message] [n/a] [main.c:730] [_main_helper] Shadow v1.13.2-608-g035872e6 2020-05-11 (built 2020-05-12) simulation was shut down cleanly, returning code 0
00:00:22.515632 [thread-0] n/a [message] [n/a] [shadow_logger.c:340] [_logger_logShutdownMessage] logging system stopped at 2020-05-12 23:27:01, run time was 00:00:22
[Thread 0x7ffff460c700 (LWP 11397) exited]
** Stopping Shadow, returning code 0 (success)

Thread 1 "shadow" received signal SIGSEGV, Segmentation fault.
0x00007ffff06cbd4c in ?? ()
(gdb) bt
#0  0x00007ffff06cbd4c in ?? ()
#1  0x00007fffffffe0e0 in ?? ()
#2  0x00007ffff3646260 in ?? ()
#3  0x00007fffffffe130 in ?? ()
#4  0x00007ffff7a19718 in _nl_current_default_domain () from /lib/x86_64-linux-gnu/libc.so.6
#5  0x000000000000001b in ?? ()
#6  0x00007ffff06cbf5a in ?? ()
#7  0x0000000000000001 in ?? ()
#8  0x00007ffff7fe80fa in call_fini (file=<error reading variable: Cannot access memory at address 0xffffffffffffffc9>)
    at /home/kevin/shadow/src/external/elf-loader/vdl-fini.c:37
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
```

Backtrace when I load the python code from the readme:

```
#0  0x00007ffff06cbd4c in ?? () from /lib/x86_64-linux-gnu/libgfortran.so.5
#1  0x00007ffff06cbf5a in ?? () from /lib/x86_64-linux-gnu/libgfortran.so.5
#2  0x00007ffff7fe80fa in call_fini (file=0x7ffff0770810) at /home/kevin/shadow/src/external/elf-loader/vdl-fini.c:37
#3  0x00007ffff7fea562 in vdl_list_iterate (list=0x7ffff03e0670, iterator=0x7ffff7fe7fc9 <call_fini>)
    at /home/kevin/shadow/src/external/elf-loader/vdl-list.c:511
#4  0x00007ffff7fe820c in vdl_fini_call (files=0x7ffff03e0670) at /home/kevin/shadow/src/external/elf-loader/vdl-fini.c:82
#5  0x00007ffff7fe4b79 in stage2_finalize () at /home/kevin/shadow/src/external/elf-loader/stage2.c:420
#6  0x00007ffff7fe3cc2 in stage1_finalize () at /home/kevin/shadow/src/external/elf-loader/stage1.c:218
#7  0x00007ffff7897d8c in __run_exit_handlers (status=0, listp=0x7ffff7a19718 <__exit_funcs>, 
    run_list_atexit=run_list_atexit@entry=true, run_dtors=run_dtors@entry=true) at exit.c:108
#8  0x00007ffff7897eba in __GI_exit (status=<optimized out>) at exit.c:139
#9  0x00007ffff78820a2 in __libc_start_main (main=0x5555555b8c14 <main>, argc=5, argv=0x7fffffffe308, init=<optimized out>, 
    fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffe2f8) at ../csu/libc-start.c:342
#10 0x000055555556973a in _start ()

```

robgjansen:
The tracker is C code that is blocking converting other modules such as DNS to rust cleanly. We previously decided that we would no longer support the C tracker code, and that new statistics should be handled in rust and logged to data files rather than the log file.

Closes #800, #1399

Progress on #1450,  #1719

Stats planning: #2727


docs/log_format.md
```diff
@@ -14,32 +14,32 @@ Shadow produces simulator log messages in the following format:
 real-time [thread-id:thread-name] virtual-time [loglevel] [hostname:ip] [src-file:line-number] [function-name] MESSAGE
 ```
 
-- `real-time`:  
+- `real-time`:
```
sporksmith, line 17: I think these trailing whitespaces are being used to force a newline in the output. I'm fine with dropping that newline, but wasn't sure whether that was your intent :)

sporksmith:
> The above produces graphs like this:
https://github.com/shadow/benchmark-results/blob/master/tgen/2024-11-27-T03-28-15/plots/shadow.results.pdf

Are the other files in that directory all coming from `tornettools plot`? I don't see RAM usage or tornettool's combined pdf of all of its plots, but maybe they can be added in?

docs/log_format.md
robgjansen: Yeah, I noticed that too. I think at some point I set my VS Code to cut trailing whitespace when saving a file, which is why that happened. I was going to undo it, and then on second thought decided it didn't matter.

docs/shadow_config_spec.md
```diff
@@ -125,7 +122,7 @@ General experiment settings.
 
 #### `general.bootstrap_end_time`
 
-Default: "0 sec"  
+Default: "0 sec"
 Type: String OR Integer
```
stevenengler, line 126: I don't think we want to remove the spaces here, otherwise the rendered html will show as

```text
Default: "0 sec" Type: String OR Integer
```

instead of

```text
Default: "0 sec"
Type: String OR Integer
```

src/main/host/process.rs
```diff
@@ -467,15 +467,8 @@ impl RunnableProcess {
...
-        if let Some(mut tracker) = host.tracker_borrow_mut() {
-            unsafe {
-                cshadow::tracker_addProcessingTimeNanos(
-                    &mut *tracker,
-                    delta.as_nanos().try_into().unwrap(),
-                )
-            };
-            host.cpu_borrow_mut().add_delay(delta);
-        }
+        host.cpu_borrow_mut().add_delay(delta);
```
stevenengler, line 470: Was this a bug before, that we were only adding the delay if the tracker was enabled?

Actually is this still a bug, that we only add the cpu delay if perf timers are enabled? I don't think we want enabling performance timers to change the simulation behaviour?

I don't actually know how this cpu object is supposed to work, but it's used in `Host::execute` to reschedule events.

stevenengler (approved):
Removing the tracker sounds good to me! I agree that we'll need to remove the `shadow.results.pdf` and everything that generates it from the benchmark repo. That change will probably need to be done first so that the benchmark CI doesn't break.

This should probably also have a changelog entry even though it's experimental.

src/main/host/process.rs
robgjansen: Yes, I think this was a bug, and we probably want to add even if perf timers are not enabled. I don't think the CPU module does anything with those delays now, but still would be nice to call the `add_delay()` function in the correct spot :)

docs/shadow_config_spec.md
robgjansen: Ahh yes, thanks! I'll undo those, and make sure to curse my VS Code setting while doing it :)

robgjansen:
Removed tracker data from benchmarks
https://github.com/shadow/benchmark/commit/cb46d0a4abd2615ea58facbc47a481512f09697d

src/main/host/process.rs
robgjansen: The CPU was designed to accumulate delay and then if a delay threshold is exceeded, we would simulate a "blocked" CPU by rescheduling events into the future at a time when the CPU became "unblocked". However, I believe this CPU blocking feature got disabled when rewriting to Rust, because we didn't understand it well and the current design potentially causes nondeterministic behavior.

It is configured as part of the `HostInfo` structure but is initialized to `None` here:

https://github.com/shadow/shadow/blob/d2a8eb7302b65451ac66a17cf39549dba137e8e1/src/main/core/sim_config.rs#L249-L250

Which gets copied by the manager for each host in the simulation:

https://github.com/shadow/shadow/blob/d2a8eb7302b65451ac66a17cf39549dba137e8e1/src/main/core/manager.rs#L577-L578

And passed into `Cpu::new` on host creation:

https://github.com/shadow/shadow/blob/d2a8eb7302b65451ac66a17cf39549dba137e8e1/src/main/host/host.rs#L234-L239

So the CPU threshold is `None` and the CPU always returns that it has zero delay:

https://github.com/shadow/shadow/blob/d2a8eb7302b65451ac66a17cf39549dba137e8e1/src/main/host/cpu.rs#L81-L94

and thus we never execute the event rescheduling block in here:

https://github.com/shadow/shadow/blob/d2a8eb7302b65451ac66a17cf39549dba137e8e1/src/main/host/host.rs#L820-L848

robgjansen: Tracking native CPU performance times is performance sensitive, which is probably why `host.cpu_borrow_mut().add_delay(delta);` is configured behind the `perf_timers` feature. I'm not going to reverse that decision in the PR.

Issue #3445, opened 2024-11-23
Build error in rust-unit-tests with rustc 1.82.0


robgjansen:
There is an error building during the `rust-unit-tests` test, as follows:

```
./setup build -cgtj32
./setup test rust-unit-tests
```

For some reason `libremora.a` is not getting linked correctly, causing undefined references to `retransmit_tally_*`. Output is

```
Start testing: Nov 23 14:41 EST
----------------------------------------------------------
1/208 Testing: rust-unit-tests
1/208 Test: rust-unit-tests
Command: "/usr/bin/bash" "-c" "              SHADOW_VERSION="" RUSTFLAGS="-C force-frame-pointers=y" CFLAGS=" -std=gnu11 -O0 -ggdb -fno-omit-frame-pointer -Wreturn-type -Wswitch -DDEBUG" CXXFLAGS=" -std=c++11 -O0 -ggdb -fno-omit-frame-pointer -Wreturn-type -Wswitch -DDEBUG" CC="/usr/bin/cc" CXX="/usr/bin/c++"             cargo test                                       --manifest-path "/home/rjansen/dev/shadow/src"/Cargo.toml             --target-dir "/home/rjansen/dev/shadow/build/src/target"             --features ""             "
Directory: /home/rjansen/dev/shadow/build/src
"rust-unit-tests" start time: Nov 23 14:41 EST
Output:
```

...

```
= note: /usr/bin/ld: /home/rjansen/dev/shadow/build/src/target/debug/build/shadow-rs-8a085ae6b3bd174c/out/libshadow-c.a(134261f062bcde65-tcp.o): in function `_tcp_flush':
          /home/rjansen/dev/shadow/src/main/host/descriptor/tcp.c:1280: undefined reference to `retransmit_tally_num_lost_ranges'
          /usr/bin/ld: /home/rjansen/dev/shadow/src/main/host/descriptor/tcp.c:1284: undefined reference to `retransmit_tally_populate_lost_ranges'
          /usr/bin/ld: /home/rjansen/dev/shadow/src/main/host/descriptor/tcp.c:1298: undefined reference to `retransmit_tally_mark_retransmitted'
          /usr/bin/ld: /home/rjansen/dev/shadow/build/src/target/debug/build/shadow-rs-8a085ae6b3bd174c/out/libshadow-c.a(134261f062bcde65-tcp.o): in function `_tcp_runRetransmitTimerExpiredTask':
[...]

PR #3448, opened 2024-11-27
Bump rust to 1.82 and fix link error

sporksmith:
Fix link error with rust 1.82 (#3445), and bump our toolchain version to 1.82.


robgjansen (approved):
Nice! I also verified the fix in my local environment. 👍

PR #3451, opened 2024-12-03
Remove support for debian:10

sporksmith:
It's passed EOL: https://wiki.debian.org/LTS. Its python3 package is also pretty old (3.7), which would have required substantial changes to #3449 to support.


robgjansen (approved):
Nice to see we were able to remove a kludge! :)

PR #3455, opened 2024-12-08
Limit address scope

robgjansen:
Limit accesses to C `Address` structures in preparation for removing the C Address and DNS code.


robgjansen:
Did I introduce a dep on glib to the failing lint tests?

```
  The system library `glib-2.0` required by crate `shadow-shim-helper-rs` was not found.
  The file `glib-2.0.pc` needs to be installed and the PKG_CONFIG_PATH environment variable must contain its parent directory.
  The PKG_CONFIG_PATH environment variable is not set.

  HINT: if you have installed the library, try setting PKG_CONFIG_PATH to the directory containing `glib-2.0.pc`.
```

Not sure how I did that with the code I changed here. If anyone has ideas, lmk!

stevenengler:
> Did I introduce a dep on glib to the failing lint tests?

No, the CI image `ubuntu-latest` just switched over from ubuntu 22.04 to ubuntu 24.04. We should probably hard-code it to a specific ubuntu version rather than having the CI randomly break each time the `-latest` base image is updated.

stevenengler:
> FYI, I have two more follow on branches to (1) globalize the DNS (limit its scope to the worker), and (2) rewrite the DNS in Rust and remove the C DNS and Address code. Let me know if you prefer I push those as separate PRs now, push them all together in 1 PR, or push them one at a time after the previous is merged.

Whichever you prefer. It would probably be best to do either "push those as separate PRs now" or "push them all together in 1 PR", and then I can review them both this weekend. If you do the first option, I'll just ignore the duplicate commits from the first PR in the second PR.

PR #3458, opened 2024-12-09
Change some workflows from 'ubuntu-latest' to 'ubuntu-22.04'

stevenengler:
This leaves some workflows as 'ubuntu-latest', for example very simple workflows and workflows that run in containers.

This is needed due to the github change of 'ubuntu-latest' from 'ubuntu-22.04' to 'ubuntu-24.04', which breaks the CI.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,7 +14,7 @@ env:
 jobs:
 
   lint-python:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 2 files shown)


robgjansen (approved):
I guess eventually we'll want to update these to the more recent 24.04 and fix the outstanding errors, but pinning to 22.04 for now so we can progress for now seems smart too :)

stevenengler:
> I guess eventually we'll want to update these to the more recent 24.04 and fix the outstanding errors, but pinning to 22.04 for now so we can progress for now seems smart too :)

Yeah, I just wanted to get it working for now.

I'll make an issue for the upgrade.

PR #3462, opened 2024-12-11
Bump pinned action images to ubuntu 24.04

### LINKED ISSUE #3459: Upgrade CI from ubuntu-22.04 to ubuntu-24.04
Many of our CI workflows are using ubuntu-22.04. We should update them to use ubuntu-24.04. We might need to fix some dependencies, such as GLib.

sporksmith:
* Install libglib in workflows that run our rust build scripts
* Bump our pinned action images from ubuntu 22.04 to 24.04

Fixes #3459

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -100,6 +100,10 @@ jobs:
           # See https://github.com/shadow/shadow/issues/2166
           ref: ${{ github.event.pull_request.head.sha }}
 
+      # Our rust build scripts require libglib.
+      - name: Install system dependencies
+        run: sudo apt-get install -y libglib2.0-dev
+
       - name: Set Rust toolchain
         run: ln -s ci/rust-toolchain-stable.toml rust-toolchain.toml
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -100,6 +100,10 @@ jobs:
           # See https://github.com/shadow/shadow/issues/2166
           ref: ${{ github.event.pull_request.head.sha }}
 
+      # Our rust build scripts require libglib.
+      - name: Install system dependencies
+        run: sudo apt-get install -y libglib2.0-dev
```
stevenengler, line 105: Does clippy run build scripts? I'm guessing so if this is failing, but it seems weird to me.

stevenengler (approved):

.github/workflows/lint.yml
sporksmith: It seems so. I think it's integrated with the compiler so it sort of makes sense?

PR #3463, opened 2024-12-12
Limit dns scope

robgjansen:
Limit accesses to C DNS in preparation for removing the C Address and DNS code.

Follow-up to #3455 ; this should be rebased and merged AFTER #3455 is merged.

### DIFF
--- a/src/main/core/manager.rs
+++ b/src/main/core/manager.rs
@@ -255,30 +255,41 @@ impl<'a> Manager<'a> {
         let dns = unsafe { c::dns_new() };
         assert!(!dns.is_null());
 
-        for (i, info) in manager_config.hosts.iter().enumerate() {
-            let id = HostId::from(u32::try_from(i).unwrap());
-            let std::net::IpAddr::V4(addr) = info.ip_addr.unwrap() else {
-                unreachable!("IPv6 not supported");
-            };
+        // Assign the host id only once to guarantee it stays associated with its host.
+        let host_init: Vec<(&HostInfo, HostId)> = manager_config
+            .hosts
+            .iter()
+            .enumerate()
+            .map(|(i, info)| {
+                // Set up the host identity.
+                let id = HostId::from(u32::try_from(i).unwrap());
+                let std::net::IpAddr::V4(addr) = info.ip_addr.unwrap() else {
+                    unreachable!("IPv6 not supported");
+                };
+                let name = info.name.clone();
 
-            let hostname = info.name.clone();
-            let chostname = CString::new(&*hostname).unwrap();
-            let caddr = u32::from(addr).to_be();
+                // Register in the global DNS.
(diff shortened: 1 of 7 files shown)


src/main/core/manager.rs
```diff
@@ -255,30 +255,41 @@ impl<'a> Manager<'a> {
...
+        let host_init: Vec<(&HostInfo, HostId)> = manager_config
+            .hosts
+            .iter()
+            .enumerate()
+            .map(|(i, info)| {
+                // Set up the host identity.
+                let id = HostId::from(u32::try_from(i).unwrap());
+                let std::net::IpAddr::V4(addr) = info.ip_addr.unwrap() else {
+                    unreachable!("IPv6 not supported");
+                };
```
stevenengler, line 268: (Optional) Maybe split these up into two steps? I think it's a little nicer for the functional style methods like `map` to not have side effects, but I'm sure there are places in shadow I've done this.

```rust
let host_init: Vec<(&HostInfo, HostId)> = manager_config
    .hosts
    .iter()
    .enumerate()
    .map(|(i, info)| (info, HostId::from(u32::try_from(i).unwrap())));

for (info, id) in &host_init {
    let std::net::IpAddr::V4(addr) = info.ip_addr.unwrap() else {
        unreachable!("IPv6 not supported");
    };
    ...
    unsafe { c::dns_register(dns, id, chostname.as_ptr(), caddr) };
}
```

src/main/core/worker.rs
```diff
@@ -677,8 +677,10 @@ mod export {
     use super::*;
 
     #[no_mangle]
-    pub extern "C-unwind" fn worker_getDNS() -> *mut cshadow::DNS {
-        Worker::with_dns(std::ptr::from_ref).cast_mut()
+    pub extern "C-unwind" fn worker_getHostsFilePath() -> *mut std::ffi::c_char {
+        Worker::with_dns(|dns| unsafe {
+            cshadow::dns_getHostsFilePath(std::ptr::from_ref(dns).cast_mut())
+        })
```
stevenengler, line 683: This should have a comment about whether the returned string pointer needs to be freed, or how long it lives. (Or ignore if you plan on removing this anyways as part of the rust migration in the follow up MR.)

stevenengler (approved):

stevenengler:
Reviewed "Remove dead dns code in worker" and later.

src/main/core/worker.rs
robgjansen: Yes, this is changed in #3464, and the safety requirements are documented there too.

src/main/core/manager.rs
robgjansen: Done in e1ba7cba601d1d329c1ea3da3bf365b66f63ba12

PR #3469, opened 2024-12-22
Fix clippy warnings with rust 1.83 and update version in CI

stevenengler:


src/lib/scheduler/src/thread_per_core.rs
```diff
@@ -129,6 +129,7 @@ where
     runner: TaskRunner<'pool, 'scope>,
 }
 
+#[allow(clippy::needless_lifetimes)]
```
robgjansen, line 132: A comment about why we allow them here would be helpful. Same for the other override locations too.

robgjansen (approved):
These are great!

src/lib/scheduler/src/thread_per_core.rs
stevenengler: Added comments.

PR #3470, opened 2024-12-22
Fix clippy warnings in unit tests and add '--all-targets' to clippy CI

### LINKED ISSUE #3468: CI doesn't run clippy on unit tests
We don't run clippy on unit tests in the CI. This was probably unintentional from whenever we added clippy CI linting.

I think our clippy lints are fairly reasonable (I don't think we add any that are particularly annoying), so I don't think it should add much friction to also lint the unit tests. There are currently about 18 warnings in unit tests.

We would need to pass the `--tests` flag to `cargo clippy` (or maybe `--all-targets`).

stevenengler:
Closes #3468.

This is branched from #3469, ~so the first 4 commits should be ignored.~

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -114,7 +114,7 @@ jobs:
         run: rustup component add clippy
 
       - name: clippy
-        run: (cd src && cargo clippy -- -Dwarnings)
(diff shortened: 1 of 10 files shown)


robgjansen (approved):
Ahh clippy ❤️ :)

PR #3473, opened 2024-12-23
Add `SyscallNonDeterministicArg` strace logger wrapper

### LINKED ISSUE #3460: Determinism tests fail if ASLR can't be disabled
The determinism tests are failing for me. They have been failing for a while:

```text
$ ./setup test determinism
[...]
The following tests FAILED:
	42 - determinism1-compare-shadow (Failed)
	46 - determinism2-shadow-compare (Failed)
```

`determinism1`:

```text
$ diff /shadow/build/src/test/determinism/determinism1a-shadow.data/hosts/testnode1/test-determinism.1000.strace /shadow/build/src/test/determinism/determinism1b-shadow.data/hosts/testnode1/test-determinism.1000.strace
12c12
< 00:00:01.000000000 [tid 1000] ioctl(3, 21505, 140732830198864) = -25 (ENOTTY)
---
> 00:00:01.000000000 [tid 1000] ioctl(3, 21505, 140736331501840) = -25 (ENOTTY)
17c17
< 00:00:01.000000000 [tid 1000] ioctl(3, 21505, 140732830198864) = -25 (ENOTTY)
---
> 00:00:01.000000000 [tid 1000] ioctl(3, 21505, 140736331501840) = -25 (ENOTTY)
```

`determinism2`:

```text
$ diff /shadow/build/src/test/determinism/determinism2a-shadow.data/hosts/peer1/test-phold.1000.strace /shadow/build/src/test/determinism/determinism2b-shadow.data/hosts/peer1/test-phold.1000.strace
26c26
< 00:00:01.000000000 [tid 1000] futex(<pointer>, 129, 2147483647, <pointer>, <pointer>, 401523232) = 0
---
> 00:00:01.000000000 [tid 1000] futex(<pointer>, 129, 2147483647, <pointer>, <pointer>, 408764464) = 0
```

These look like pointers (but we don't know they're pointers because of the `ioctl` and `futex` syscalls).

Earlier I see:

```text
00:00:00.194137 [70289:shadow] n/a [WARN] [n/a] [shadow.rs:180] [shadow_rs::shadow] Could not disable address space layout randomization. This may affect determinism: ENOSYS: Function not implemented

Stack backtrace:
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.89/src/backtrace.rs:27:14
   1: shadow_rs::shadow::run_shadow
             at /shadow/src/main/shadow.rs:178:11
   2: main_runShadow
             at /shadow/src/main/shadow.rs:473:22
   3: main
             at /shadow/src/main/main.c:17:12
   4: __libc_start_call_main
   5: __libc_start_main@@GLIBC_2.34
   6: _start
```

It seems likely that ASLR isn't being disabled, so the pointers aren't deterministic in the strace.

Maybe we should make our determinism test better handle the case where ASLR can't be disabled.

stevenengler:
This makes the determinism tests pass again on my computer.

Example:

```text
00:00:01.000000000 [tid 1000] futex(<pointer>, 265, 1001, <pointer>, <pointer>, <non-deterministic>) = 0
```

Closes #3460.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ PATCH changes (bugfixes):
 reports errors in a different format on `stderr` to make the duplication easier
 to sort out in the case that `stdout` and `stderr` are merged. (#3428)
 * Sending a packet to an unknown IP address no longer causes Shadow to panic. (#3411)
+* Improved the "deterministic" strace logging mode by hiding some non-deterministic syscall arguments. (#3473)
 
 Full changelog since v3.2.0:
 
--- a/src/main/host/syscall/formatter.rs
+++ b/src/main/host/syscall/formatter.rs
@@ -81,6 +81,17 @@ impl<'a, T> SyscallVal<'a, T> {
             _phantom: PhantomData,
         }
     }
+
+    /// Cast a syscall argument or return value to another type.
+    pub fn cast<V>(&self) -> SyscallVal<'a, V> {
+        SyscallVal {
+            reg: self.reg,
+            args: self.args,
+            options: self.options,
+            mem: self.mem,
+            _phantom: PhantomData,
+        }
+    }
 }
 
 impl<'a, T> Display for SyscallVal<'a, T>
--- a/src/main/host/syscall/handler/fcntl.rs
+++ b/src/main/host/syscall/handler/fcntl.rs
@@ -5,6 +5,7 @@ use log::debug;
 use crate::cshadow;
 use crate::host::descriptor::{CompatFile, File, FileStatus};
 use crate::host::syscall::handler::{SyscallContext, SyscallHandler};
+use crate::host::syscall::type_formatting::SyscallNonDeterministicArg;
 use crate::host::syscall::types::SyscallError;
 
 impl SyscallHandler {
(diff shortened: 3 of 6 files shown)


robgjansen (approved):
Do we (eventually) want to apply this to all pointer types too, where we don't want the value of the pointer to be printed? Or is our current method for handling those sufficient?

stevenengler:
> Do we (eventually) want to apply this to all pointer types too, where we don't want the value of the pointer to be printed? Or is our current method for handling those sufficient?

Pointer types like `*const u64` should already print "\<pointer\>" rather than an address/value if the "deterministic" option is enabled, for example like the implementation of [`deref_pointer_impl`](https://github.com/shadow/shadow/blob/19edfee8a3afce77780107a85499c7b83e8c119d/src/main/host/syscall/type_formatting.rs#L115C14-L115C32). The new wrapper is for types that are normally deterministic (such as u64), but aren't for some specific syscalls like ones that take a variable number of arguments or syscalls where integers are sometimes actually pointers. I'll add a comment to explain this better.

If you mean to reuse the new `SyscallNonDeterministicArg` wrapper in the `deref_pointer_impl` macro for example, I think we could do that, but I don't think we'd gain much from it.

PR #3474, opened 2024-12-23
Disallow broadcast and multicast in DNS

### LINKED ISSUE #3465: Re-add some restricted IP ranges
In #3414 we removed the restrictions on using restricted IP ranges, but I think we should re-add the non-unicast ranges back. For example I don't think it's valid to use a multicast IP (224.0.0.0/4) or broadcast IP (255.255.255.255) as your interface IP address.

But it's unlikely that we'll support multicast or broadcast packets in Shadow any time soon, so I think restricting these IP ranges would be more to avoid confusion than something that would result in a bug.

robgjansen:
closes #3465

### DIFF
--- a/src/main/network/dns.rs
+++ b/src/main/network/dns.rs
@@ -73,7 +73,8 @@ impl DnsBuilder {
         name: String,
     ) -> Result<(), RegistrationError> {
         // Make sure we don't register reserved addresses or names.
-        if addr.is_loopback() || addr.is_unspecified() {
+        if addr.is_loopback() || addr.is_unspecified() || addr.is_broadcast() || addr.is_multicast()
+        {
             return Err(RegistrationError::InvalidAddr);
         } else if name.eq_ignore_ascii_case("localhost") {
             return Err(RegistrationError::InvalidName);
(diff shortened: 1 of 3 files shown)


robgjansen:
@stevenengler Do we want to validate the address at the config parsing step too, or is it OK just in the dns module?

stevenengler:
> @stevenengler Do we want to validate the address at the config parsing step too, or is it OK just in the dns module?

I don't have a strong opinion either way. I think it might be nice to validate at the config parsing step, but we have other address validation that doesn't happen at config parsing, such as checking that two hosts don't have the same IP address (in `assign_ips`). I think checking just at the dns module level is okay since it happens only at the beginning of the simulation. And that's what we've always done in the past anyways, so doesn't need to block this PR.

Does the error message from `register` end up being nice for the user?

robgjansen:
> Does the error message from register end up being nice for the user when this does happen?

Better now, after https://github.com/shadow/shadow/pull/3474/commits/162e4a6b121f8f9a404795a67bba4460a1a9494d :)

src/main/network/dns.rs
```diff
@@ -30,27 +30,55 @@ struct Record {
 
 #[derive(Debug, PartialEq)]
 pub enum RegistrationError {
-    InvalidAddr,
-    InvalidName,
-    AddrExists,
-    NameExists,
+    BroadcastAddrInvalid,
+    LocalhostAddrInvalid,
```
stevenengler, line 34: (Optional) Maybe `LoopbackAddrInvalid` instead? Localhost is usually a single IP address (typically 127.0.0.1 or whatever is in the hosts file), whereas loopback is the IP range 127.0.0.0/8.

src/main/network/dns.rs
```diff
@@ -30,27 +30,55 @@ struct Record {
...
+            RegistrationError::BroadcastAddrInvalid => write!(
+                f,
+                "broadcast address '{}' is invalid in DNS",
+                Ipv4Addr::BROADCAST
+            ),
+            RegistrationError::LocalhostAddrInvalid => write!(
+                f,
+                "localhost address '{}' is invalid in DNS",
+                Ipv4Addr::LOCALHOST
+            ),
```
stevenengler, line 54: nit: this may be confusing if the user used 127.0.0.10 (a loopback address which is caught by `is_loopback()` below), but the error message says the address is 127.0.0.1.

stevenengler (approved):

src/main/network/dns.rs
robgjansen: Ahh, thanks for noticing that. Fixed now! (in 69b5a5511c77f689f2f769ffdf65330a518a5bfe)

src/main/network/dns.rs
robgjansen: Fixed (in 69b5a5511c77f689f2f769ffdf65330a518a5bfe)

PR #3476, opened 2024-12-23
Generate bindings for if.h as a separate module

### LINKED ISSUE #3475: Linux headers include system headers
Some linux headers include system headers. For example in `include/linux/if.h`:

```c
#include <sys/socket.h>         /* for struct sockaddr.     */
```

This can cause bindgen to fail where clang complains with "error: redefinition of _". We can avoid including headers in standard system directories by including the `-nostdlibinc` clang argument to bindgen, but then we get "error: 'sys/socket.h' file not found" errors instead.

An example of this is that we need to create bindings for `struct ifconf`, which is defined in `include/linux/if.h`. If we include "if.h" in our bindings, we get:

```text
src/lib/linux-api/bindings-build/linux-install/include/asm/signal.h:15:23: error: typedef redefinition with different types ('unsigned long' vs '__sigset_t')
src/lib/linux-api/bindings-build/linux-install/include/linux/time.h:16:8: error: redefinition of 'timeval'
./bindings-wrapper.h:121:5: error: redefinition of enumerator 'SHUT_RD'
./bindings-wrapper.h:122:5: error: redefinition of enumerator 'SHUT_WR'
./bindings-wrapper.h:123:5: error: redefinition of enumerator 'SHUT_RDWR'
```

For some types we could just copy the definition from the linux headers into `bindings-wrapper.h`. But this doesn't work for all types. For example `struct ifconf` uses `struct ifreq` which uses `struct sockaddr` from `sys/socket.h`. I'm not sure what the best solution here is. Maybe we should write our own `sys/*` headers which we make sure don't conflict with the linux headers. But this might be a lot of work and be easy to get wrong.

sporksmith:
This is needed to work around #3475: if.h includes system headers that are incompatible with other Linux headers.  This ends up also exposing libc's sockaddr as `linux_sockaddr`.

Fixes #3475


sporksmith:
Ok, I've cleaned this up a bit. No rush; just didn't want to leave you blocked on #3475.

stevenengler (approved):

PR #3478, opened 2024-12-24
Remove coverage build code

### LINKED ISSUE #3433: Remove coverage build
The rust compiler has [removed](https://github.com/rust-lang/rust/pull/131829) the nightly `-Zprofile` option, which means that the CMake/python code we have to generate code coverage reports will no longer work.

Since it has been removed from rust, and we don't run it in the CI, I think we should remove related build code from Shadow.

stevenengler:
Closes #3433.


robgjansen (approved):
🎉

stevenengler:
I also added a changelog entry.

PR #3489, opened 2025-01-10
Fix fifo qdisc

robgjansen:
Our fifo qdisc violates the assumption that a queued item's priority does not change while the item is enqueued. This allows a socket that had only bad priority packets when it was initially pushed into the queue to move up in priority when it suddenly has new packets with better priority. This case can happen often in the legacy TCP stack because it assigns priority=0 to control packets (ACKS) without data. This means we can be more responsive to TCP controls.

This PR does two things:
1. Fixes the fifo qdisc so that it always produces a non-equal sort order over sockets. This is done by tracking the push order of the sockets and using the push order to break equality ties.
2. Tracks the socket priority that existed at push time to enable us to show the performance difference when sorting based on that priority vs the priorities that are recomputed every time the priority queue rebalances and calls the compare function on a socket.

If we only apply the first commit, the performance should be identical to nightly. If we apply both commits, the performance should be identical to #3480.

### DIFF
--- a/src/main/host/network/network_queuing_disciplines.c
+++ b/src/main/host/network/network_queuing_disciplines.c
@@ -94,6 +94,7 @@ static gint _compareSocket(const InetSocket* sa, const InetSocket* sb, FifoSocke
     bool foundb = g_hash_table_lookup_extended(fifo->items, (gpointer)sb, NULL, (gpointer*)&itemb);
     utility_debugAssert(founda && foundb);
 
+    /*
     uint64_t pa = 0;
     uint64_t pb = 0;
 
@@ -110,6 +111,7 @@ static gint _compareSocket(const InetSocket* sa, const InetSocket* sb, FifoSocke
         itema->priority = pa;
         itemb->priority = pb;
     }
+    */
 
     if (itema->priority < itemb->priority) {
         return -1;
--- a/src/main/host/network/network_queuing_disciplines.c
+++ b/src/main/host/network/network_queuing_disciplines.c
@@ -79,18 +79,50 @@ bool rrsocketqueue_find(RrSocketQueue* self, const InetSocket* socket) {
     return g_queue_find_custom(self->queue, (void*)socket, _compareInetSocket);
 }
 
-static gint _compareSocket(const InetSocket* sa, const InetSocket* sb) {
+typedef struct _FifoItem FifoItem;
+struct _FifoItem {
+    uint64_t priority;
+    uint64_t push_order;
+};
+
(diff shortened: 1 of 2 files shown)


robgjansen:
The first commit 3d73420a0693bfb464aa303aeec8686dbb61fe5d keeps the current prioritization but fixes the equality case. We can see in [the TGen benchmark](https://github.com/shadow/benchmark-results/tree/master/tgen/2025-01-10-T23-41-49) that the TGen performance is still mostly the same as the nightly before the change.

time_to_last_byte_recv_1048576_cdf

The second commit 10db06a72718decfe5661c3ea09131608c77103e changes the prioritization so that we do not modify the socket priority after it has already been pushed into the queue, meaning that the priority is set at the time it enters the queue and stays static until the socket is dequeued. We can see in [the TGen benchmark](https://github.com/shadow/benchmark-results/tree/master/tgen/2025-01-11-T00-05-58) that the TGen performance gets slightly worse because we are no longer able to move the socket ahead in priority when it needs to send control packets (ACKS) in response to incoming packets. That makes TCP slightly less responsive. This performance matches the Rust impl in #3480.

time_to_last_byte_recv_1048576_cdf

robgjansen:
@stevenengler do we want to merge one or both of these commits before we merge #3480? I think it would be good to have especially https://github.com/shadow/shadow/commit/10db06a72718decfe5661c3ea09131608c77103e in the git history since that is the minimal change that is responsible for the performance change, so I think I would lean toward merging this before #3480, even though we're basically going to remove the code in #3480 anyway.

stevenengler (approved):
Nice find! I'm glad there's a good explanation for the difference in graphs.

> This allows a socket that had only bad priority packets when it was initially pushed into the queue to move up in priority when it suddenly has new packets with better priority.

Ah I didn't realize (or forgot) that this is what the C version did. I'm not sure which behaviour we want, but I think the new one is more realistic?

> @stevenengler do we want to merge one or both of these commits before we merge #3480?

I agree, I think it's good to merge this first.

robgjansen:
> Ah I didn't realize (or forgot) that this is what the C version did. I'm not sure which behaviour we want, but I think the new one is more realistic?

I think the "old" design is "better" in the sense that it produces a result that is likely closer to a true `pfifo_fast` qdisc that linux uses. The reason is that priority band 0 packets should always be sent ahead of the others, and the "old" design made that possible, but that is no longer possible after merging this PR.  However, I think we should merge anyway, assuming that we'll get a much more accurate `pfifo_fast` design when we implement #3490.

PR #3492, opened 2025-01-13
Migrate packet to Rust

robgjansen:
Eliminates usage of the C packet and payload structs. This allows us to track application payloads using `Bytes` objects in Rust, which should make Rust code such as the Rust TCP stack and UDP payload handling more efficient.


robgjansen:
I was worried that adding an `Arc<Packet>` to every `PacketRc` was going to introduce a big performance hit compared to the unsafe inner C packet pointer that we used to use, so I ran the benchmarks:

- TGen: https://github.com/shadow/benchmark-results/tree/master/tgen/2025-01-13-T17-19-39
- Tor: https://github.com/shadow/benchmark-results/tree/master/tor/2025-01-13-T17-54-26

We do see a runtime slowdown effect due to the additional atomic operations as expected.

run_time

I'm not sure how bad of a performance hit this is. If we think it's bad, we could consider using `RootedRc`, but that's makes the code a bit less flexible and I don't remember if we wanted to phase out the use of that module.

stevenengler:
> I was worried that adding an `Arc<Packet>` to every `PacketRc` was going to introduce a big performance hit compared to the unsafe inner C packet pointer that we used to use, so I ran the benchmarks:
>
> - TGen: https://github.com/shadow/benchmark-results/tree/master/tgen/2025-01-13-T17-19-39
> - Tor: https://github.com/shadow/benchmark-results/tree/master/tor/2025-01-13-T17-54-26
>
> We do see a runtime slowdown effect due to the additional atomic operations as expected.

I'm not sure I'm convinced that the slowdown is caused by `Arc`s (or atomics in genera)l. I haven't looked at this PR closely yet, but I ran a benchmark using the C packet with [atomic reference counting](https://github.com/shadow/shadow/pull/3493), and didn't see any significant performance change.

https://github.com/shadow/benchmark-results/tree/master/tor/2025-01-14-T18-06-33

run_time

If it is the `Arc`, maybe we're cloning it much more than we were increasing the refcount of C packets. Or maybe we're doing more memory allocations, or something else.

robgjansen:
OK, thanks for checking into that! There's probably something else we could be doing better in the code then, which is good news :)

src/main/network/packet.rs
```diff
@@ -421,364 +831,763 @@ impl PacketDisplay for *const c::Packet {
...
+
+        let Data::LegacyTcp(tcp) = &packet.data else {
+            unimplemented!()
+        };
+
+        let len = usize::try_from(src_len).unwrap_or(usize::MAX);
+        let src = ForeignArrayPtr::new(src.cast::<u8>(), len);
+
+        // We want the dst buf on the heap so we don't have to copy it later.
+        let mut dst = vec![0u8; len].into_boxed_slice();
```
stevenengler, line 1217: I think it might be slightly more efficient to use `MaybeUninit` here to avoid zero-filling the vec. Something like `Box::<[u8]>::new_uninit_slice(len)`, writing bytes to the slice from the memory manager, then `dst.assume_init()` to go back to a `Box::<[u8]>`. But maybe just a comment would be good for now.

src/main/network/packet.rs
```diff
@@ -421,364 +831,763 @@ impl PacketDisplay for *const c::Packet {
...
+        // managed process.
+        if let Err(e) = mem.copy_from_ptr(&mut dst[..], src) {
+            log::warn!(
+                "Couldn't read managed process memory at {:?} into packet payload at {:?}: {:?}",
+                src,
+                dst,
+                e
+            );
+            return;
+        }
```
stevenengler, line 1235: If copying data into the payload fails, I think we should panic rather than just show a warning. Otherwise the packet data will be corrupt (not what the application wrote).

src/main/network/packet.rs
```diff
@@ -421,364 +831,763 @@ impl PacketDisplay for *const c::Packet {
...
+
+        log::trace!(
+            "We read {} bytes from the managed process into the packet's payload",
+            dst.len()
+        );
+
+        // Move the payload into the packet. Use Bytes (not BytesMut) to avoid copying the payload.
+        tcp.borrow_mut().payload.push(Bytes::from(dst));
```
stevenengler, line 1243: This function is called "set payload", but here we push the payload into an existing payload.

stevenengler (approved):
The changes look good to me! I'm not sure what's causing the performance difference though. We are doing more allocations and more packet refs/unrefs, so maybe it is one of those. I'll leave it up to you whether you want to merge with the performance as-is. If we do, I think this should also be mentioned in the changelog. I'll also take another look at the performance, but I'm not sure how best to debug other than just trying to reduce the number of allocations and packet refs/unrefs, and then seeing if that fixes it.

src/main/network/packet.rs
```diff
@@ -1,404 +1,814 @@
...
+    /// using the Rust function `PacketRc::from_raw()`, or dropped using the C function
+    /// `packet_unref()`.
+    ///
+    /// # Deprecation
+    ///
+    /// This function provides compatibility with the legacy C TCP stack and should be considered
+    /// deprecated and removed when the legacy C TCP stack is removed.
+    pub fn into_raw(self) -> *mut Packet {
+        Arc::into_raw(self.inner).cast_mut()
+    }
```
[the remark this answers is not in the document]
robgjansen: FYI, I did it this way to avoid having to change all of the C code, which mostly uses `Packet*`, into `const Packet*`. Adding a comment for now seems right, and if we want to go back later and fix all of the C instances to `const Packet*` we could do that in a separate PR.

src/main/network/packet.rs
```diff
@@ -1,404 +1,814 @@
...
+#[derive(Clone, Debug)]
+struct Metadata {
+    /// Tracks application priority so we flush packets from the interface to the wire in the order
+    /// intended by the application. This is used in the default FIFO network interface scheduling
+    /// discipline. Smaller values have greater priority.
+    // TODO: this can be removed once we support the TOS field in the `Header` struct.
+    priority: FifoPacketPriority,
+    /// Tracks the sequence of operations that happen on this packet as is transits Shadow's
+    /// network.
+    statuses: Option<AtomicRefCell<Vec<PacketStatus>>>,
```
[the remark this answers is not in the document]
robgjansen: Yes, the worker calls `let dst_packet = packetrc.new_copy_inner();` and then sends the new packet to the destination host.

This is the same general strategy that was used with the C packet too.

The C _payload_ was protected with a C mutex, but now the payload are read-only `Bytes` objects instead. We don't need a mutex around those, right?

src/main/network/packet.rs
robgjansen: Hmm, do you think zero-filling would cause the performance hit we're seeing in the benchmark?

robgjansen: Nevermind, [it didn't make a noticeable difference](https://github.com/shadow/benchmark-results/blob/master/tor/2025-02-12-T21-10-16/plots/run_time.png)

src/main/network/packet.rs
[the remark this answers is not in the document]
robgjansen: @stevenengler this does mean that the sender and receiver may have an inconsistent view of the packet status list though. So if we do not want to preserve the old C behavior, I would also be happy to just remove statuses completely. What do you think?

robgjansen:
To get a handle on the performance issue, I set up a simple experiment where a tgen client downloaded 100 MiB from a tgen server. Then I profiled performance using `perf` by running the following before and after this PR:

```
perf record --call-graph fp ~/dev/shadow/build/src/main/shadow shadow.yaml
```

Comparing the before/after diff using `perf diff` shows the following:

Screenshot 2025-02-13 at 5 01 37 PM

This highlights possible problems with `ObjectCounter`. Sure enough, it turns out I added a new `ObjectCounter` to the `PacketRc` wrapper, but `PacketRc`s were not being counted previously when it wrapped the C packet. Checking the object counts in the log shows that indeed this could be significant overhead:

```
Global allocated object counts: {PacketRc:4132234, Packet:148978,
```

That's 4.1 million new objects being counted just in this simple, small experiment. Removing the `PacketRc` object counter seems to get rid of the hotspot as `perf diff` shows:

Screenshot 2025-02-13 at 5 07 25 PM

Next I'll push that change and then run a Tor benchmark to double check.

robgjansen:
Removing the object counter from `PacketRc` has mitigated most of the performance degradation.



Full Tor benchmark results:
https://github.com/shadow/benchmark-results/tree/master/tor/2025-02-13-T22-25-31

src/main/network/packet.rs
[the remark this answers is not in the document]
stevenengler: > Yes, the worker calls `let dst_packet = packetrc.new_copy_inner();` and then sends the new packet to the destination host.

Okay, it might be good to just have a comment on `PacketRc` saying we shouldn't share it between threads (it's memory safe, but might panic).

> @stevenengler this does mean that the sender and receiver may have an inconsistent view of the packet status list though. So if we do not want to preserve the old C behavior, I would also be happy to just remove statuses completely. What do you think?

I don't have a strong preference either way. I think the packet statuses can be useful, but we don't have good tooling or documentation to make use of them. But being able to track packets within Shadow is nice, so since you've already implemented it, I think it's okay to leave in.

PR #3493, opened 2025-01-14
Use atomic refcount in packet

stevenengler:
Switches the C packet refcounts to atomics. The memory orderings try to approximately match the memory orderings used in `Arc::clone` and `Arc::drop`.

Using for a benchmark.


robgjansen:
Thank so much for this Steve! I'm closing now that we found the performance problem in #3492 and #3492 is merged now.

PR #3496, opened 2025-01-25
Fix the expired timeout issue in the futex syscall handler

magnified103:
In the current futex syscall handler:

https://github.com/shadow/shadow/blob/85d08b739741479f00100a4e5650136dc1c0acf9/src/main/host/syscall/handler/futex.c#L118-L124

When the `futex_op` is `FUTEX_WAIT_BITSET`, the timeout argument points to a `timespec` structure specifying the absolute timeout. If the provided timestamp is earlier than the current emulated time, the worker panics during event scheduling:

https://github.com/shadow/shadow/blob/85d08b739741479f00100a4e5650136dc1c0acf9/src/main/host/timer.rs#L191-L199

To fix this, I simply set `timeoutEmulatedTime` to the current emulated time in that case. I also added a unit test for this fix.

### DIFF
--- a/src/main/host/syscall/handler/futex.c
+++ b/src/main/host/syscall/handler/futex.c
@@ -117,9 +117,14 @@ static SyscallReturn _syscallhandler_futexWaitHelper(SyscallHandler* sys,
         (Trigger){.type = TRIGGER_FUTEX, .object = futex, .state = FileState_FUTEX_WAKEUP};
     SysCallCondition* cond = syscallcondition_new(trigger);
     if (timeoutSimTime != SIMTIME_INVALID) {
+        CEmulatedTime now = worker_getCurrentEmulatedTime();
         CEmulatedTime timeoutEmulatedTime = (type == TIMEOUT_RELATIVE)
-                                                ? timeoutSimTime + worker_getCurrentEmulatedTime()
+                                                ? timeoutSimTime + now
                                                 : timeoutSimTime;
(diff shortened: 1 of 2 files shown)


stevenengler (approved):
Thanks for the bug fix and test!

I wonder if we even need to schedule an event in this case, and instead should maybe just return ETIMEDOUT immediately. But either way, this PR fixes the bug, so should be good to merge.

stevenengler:
Actually thinking about this some more, I think the syscall should return right away.

There could be an issue where:

- thread A does a futex-bitset-wait with a past time
- shadow's syscall handler returns a "syscall blocked" result and schedules an event at "now" to resume the syscall
- shadow runs the next event, which happens to be on thread B
- thread B does a futex-wake on that address
- thread A resumes to handle the futex-bitset-wait event, and returns 0 instead of ETIMEDOUT

I'm going to merge anyways since the above race condition seems very rare and I think the bug fix here is more important, but I'll open an issue for it instead.

stevenengler:
> thread A resumes to handle the futex-bitset-wait event, and returns 0 instead of ETIMEDOUT

This is wrong, since it doesn't matter if there was a futex-wake or not. The syscall handler checks the timeout using `rustsyscallhandler_didListenTimeoutExpire`, and since this would return true regardless of if there was a futex-wake, the syscall handler should always return ETIMEDOUT.

So I think this PR is correct, and returning early would just be a performance optimization rather than for correctness.

PR #3504, opened 2025-01-30
Remove Debian 10 from supported platforms

stevenengler:
We dropped support in #3451.


sporksmith (approved):

stevenengler:
@sporksmith I realized I missed removing something, so I updated the commit and re-requested a review. Sorry for the noise.

PR #3506, opened 2025-02-01
Upgrade 'rand{,_chacha,_core,_xoshiro}'

stevenengler:
~~The first commit d75c397 is from #3507.~~


src/Cargo.lock
```diff
@@ -1,6 +1,6 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 3
+version = 4
```
stevenengler, line 3: Cargo automatically switched from 3 to 4 in 1.84.0, but it's supposedly backwards compatible to 1.78.0.

robgjansen (approved):

PR #3515, opened 2025-02-23
Migrate to Rust 2024 edition

stevenengler:
We don't technically need to upgrade the edition, but I think we might as well get this out of the way so that we use the modern capture rules, scopes, etc. I didn't use `cargo fix` for any of this.

There are some changes regarding capture rules, drop order, lifetime scopes, etc:
https://doc.rust-lang.org/edition-guide/rust-2024/language.html

A possible issue is that we have a lot of unsafe code and some may have relied on the old behaviour, which is now changed in rust 2025. But the updated code builds and passes all tests, so it's probably (hopefully) fine. I looked through the changes briefly, and I think it's unlikely that any broke our unsafe code.

Rustfmt now has "style editions". By upgrading rust editions, cargo automatically uses the new rustfmt style edition. This does have quite a bit of code churn. I think we could avoid this by specifying `style_edition = "2021"` in a `rustfmt.toml` file, but I decided to just do a rustfmt here using the latest (default) style edition. We can easily revert this before merging if anyone wants.

**Edit:** `lint-cargo-lock` requires https://github.com/actions/runner-images/issues/11637, but maybe it would be better to use the rust version we specify in `ci/rust-toolchain-stable.toml`. The downside would be longer a workflow run time.


sporksmith (approved):
LGTM. I think the auto-formatting changes are an improvement and worth the churn.

Issue #2066, opened 2022-04-20
Implement true preemption for managed code


sporksmith:
A thread that never makes any syscalls will stop a Shadow simulation from making progress.

We haven't encountered this in practice, but have encountered other "busy loops" that wait for another thread to do something. Our new feature for modeling latency of non-blocking syscalls helps us escape most such loops, and arguably it'd be pretty poor code that didn't have some kind of timeout, which would imply a system call to check the time.

Still, it's not too far-fetched to imagine, e.g. spawning a thread and then busy-waiting for the new thread to modify an atomic in memory. Such code probably *ought* to have a timeout or at least a `sched_yield`, but undoubtedly code exists with neither.

This is currently low priority, but let's track any real-world instances of this problem that we do find here.

sporksmith:
I think it wouldn't be terribly difficult to implement preemption using [setitimer](https://man7.org/linux/man-pages/man2/setitimer.2.html) to arrange  for an interrupt after some amount of CPU time, and having out interrupt handler move time forward and potentially reschedule the thread. This would definitely break determinism, though.

sporksmith:
> We haven't encountered this in practice, 

We have, now.

* [spin crate](https://github.com/shadow/shadow/issues/1792#issuecomment-1442987642)
* [mysql server](https://github.com/shadow/shadow/issues/1792#issuecomment-2223190467)

sporksmith:
> I think it wouldn't be terribly difficult to implement preemption using [setitimer](https://man7.org/linux/man-pages/man2/setitimer.2.html) 

A tricky bit with `setitimer` is that it sends process-targeted signals, so we'd need to be careful to block the signal in all non-running threads to ensure it gets delivered to the running (busy-looping) thread.

It looks like `timer_create` supports sending thread-targeted signals, *and* a clock that measures cpu-time of the current thread (`CLOCK_THREAD_CPUTIME_ID`). Using this API we won't have to worry about blocking and unblocking the signal. We may still want to reset the timer when rescheduling the thread, but could probably get away with not doing so. e.g. we could just track the last simulated time that the timer went off, and only actually do a preemption if it fires again and simulated time hasn't advanced.

sporksmith:
My current prototype is using `setitimer` to avoid the potential resource limits and "here be dragons" warnings  for `timer_create + SIGEV_THREAD_ID`.

[...]

PR #3517, opened 2025-02-26
linux-api: add wrappers for getitimer and setitimer

sporksmith:
Generally useful, and to be potentially used for https://github.com/shadow/shadow/issues/2066

### DIFF
--- a/src/lib/linux-api/src/time.rs
+++ b/src/lib/linux-api/src/time.rs
@@ -101,6 +101,11 @@ pub use bindings::linux_itimerval;
 pub type itimerval = linux_itimerval;
 unsafe impl shadow_pod::Pod for itimerval {}
 
+pub use bindings::linux___kernel_old_itimerval;
+#[allow(non_camel_case_types)]
+pub type kernel_old_itimerval = linux___kernel_old_itimerval;
+unsafe impl shadow_pod::Pod for kernel_old_itimerval {}
+
 /// Raw `alarm` syscall. Permits u64 arg and return value for generality with
 /// the general syscall ABI, but note that the `alarm` syscall definition itself
 /// uses u32.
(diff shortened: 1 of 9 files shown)


src/lib/linux-api/src/time.rs
```diff
@@ -118,6 +118,52 @@ pub fn alarm(secs: u32) -> Result<u32, Errno> {
...
+}
+
+/// Make a `setitimer` syscall.
+pub fn setitimer(
+    which: ITimerId,
+    new_value: &itimerval,
+    old_value: Option<&mut itimerval>,
+) -> Result<(), Errno> {
+    let old_value = old_value
+        .map(|p| p as *mut itimerval)
```
stevenengler, line 162: nit: I think `.map(std::ptr::from_mut)` would be slightly better so that there's no possibility of casting to a different type.

src/lib/linux-api/src/time.rs
```diff
@@ -118,6 +118,52 @@ pub fn alarm(secs: u32) -> Result<u32, Errno> {
...
+/// `old_value` must be safe for the kernel to write to, or NULL.
+///
+/// An invalid or inaccessible `new_value` *isn't* a safety violation, but may
+/// cause the syscall to fail e.g. with `EFAULT`.
+pub unsafe fn setitimer_raw(
+    which: i32,
+    new_value: *const itimerval,
+    old_value: *mut itimerval,
+) -> Result<(), Errno> {
+    unsafe { syscall!(linux_syscall::SYS_setitimer, which, new_value, old_value) }
```
stevenengler, line 150: In the kernel the syscall takes arguments of type `struct __kernel_old_itimerval`. Is this the same as `linux_itimerval` for x86-64, or should this function take `kernel_old_timeval` instead? (Even if they're the same, I think it would be better to have it take `kernel_old_timeval` instead just to be consistent with the kernel.)

It also takes `which` as an `int`; should we keep it as an `int` here in the wrapper too? I don't have a strong preference here though, since realistically we only target x86-64 where `int == i32`.

src/lib/linux-api/src/time.rs
```diff
@@ -118,6 +118,52 @@ pub fn alarm(secs: u32) -> Result<u32, Errno> {
     Ok(res.try_into().unwrap())
 }
 
+/// Make a `getitimer` syscall.
+///
+/// # Safety
+///
+/// `curr_value` must be safe for the kernel to write to.
+pub unsafe fn getitimer_raw(which: i32, curr_value: *mut itimerval) -> Result<(), Errno> {
+    unsafe { syscall!(linux_syscall::SYS_getitimer, which, curr_value) }
```
stevenengler, line 127: Same comment about `__kernel_old_itimerval` and `int`.

stevenengler (changes requested):

src/lib/linux-api/src/time.rs
sporksmith: Done

src/lib/linux-api/src/time.rs
sporksmith: Done

src/lib/linux-api/src/time.rs
sporksmith: Done

sporksmith (commented):
Thanks, good catches!

stevenengler (approved):

PR #3519, opened 2025-02-28
shim: improve execution context tracking

sporksmith:
* Replaces the previous boolean flag which (sort of) identified whether native syscalls were currently allowed, with an enum identifying whether the calling context is shadow code or application/managed code. This can be used to help decide whether a particular syscall ought to be native or emulated, but isn't quite 1:1 with whether syscalls are currently allowed.
* Move where we set and restore this context out to API edges where we enter/leave shadow code. This sometimes we mean to explicitly propagate the "original" execution context through multiple API layers, e.g. to keep track of whether a syscall originated from shadow code or application code, but is generally clearer and more accurate.

This is motivated primarily to help write a robust implementation of #2066, where we'll want to accurately determine in a timer signal handler whether we were running in shadow code or application code (and/or more-accurately enable/disable the timer when transitioning between the two). Even if we don't go forward with that though, these changes improve clarity and robustness IMO.

### DIFF
--- a/src/lib/shim/shim_api_c.h
+++ b/src/lib/shim/shim_api_c.h
@@ -10,15 +10,20 @@
 #include <netdb.h>
 #include <stdarg.h>
 
-/// This module defines functions that can be called by external (preloaded)
-/// libraries that are linked to the shim. Those libraries should only call
-/// functions defined here after including this header file.
+#include "lib/shim/shim_api.h"
+
+/// This module defines *C implementations* of functions that can be called by
+/// external (preloaded) libraries that are linked to the shim. Those libraries
+/// should only call functions defined here after including this header file.
+///
+/// The public interfaces to access these are defined in Rust, without prefix
+/// `shim_api_` instead of `shimc_api`.
 
 // The entry point for handling an intercepted syscall. This function remaps the
 // return value into errno upon error so that errno will be set correctly upon
 // returning control to the managed process. Be careful not to do something that
 // would overwrite errno after this function returns.
-long shimc_api_syscall(long n, ...);
+long shimc_api_syscall(ExecutionContext exe_ctx, long n, ...);
 
 // Shim implementation of `man 3 getaddrinfo`.
 int shimc_api_getaddrinfo(const char* node, const char* service, const struct addrinfo* hints,
--- a/src/lib/shim/src/lib.rs
+++ b/src/lib/shim/src/lib.rs
@@ -50,9 +50,13 @@ pub fn simtime() -> Option<SimulationTime> {
 #[derive(Debug, Copy, Clone, Eq, PartialEq, IntoPrimitive, TryFromPrimitive)]
 #[repr(u8)]
 pub enum ExecutionContext {
-    Shadow,
-    Application,
+    // Redirect through constants below so that we can access in const contexts.
+    Shadow = EXECUTION_CONTEXT_SHADOW_CONST,
+    Application = EXECUTION_CONTEXT_APPLICATION_CONST,
 }
+
+pub const EXECUTION_CONTEXT_SHADOW_CONST: u8 = 0;
+pub const EXECUTION_CONTEXT_APPLICATION_CONST: u8 = 1;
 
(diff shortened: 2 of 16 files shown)


sporksmith:
I noticed this does noticeably slow down the `go_preempt-shadow`, which exercises the related path in a pretty tight loop. It's mostly in debug builds, but only ~10% in a release build.

I suspect there won't be a measurable difference at all in more realistic simulations. Kicked off some benchmarks to verify:

* https://github.com/shadow/benchmark/actions/runs/13591816570
* https://github.com/shadow/benchmark/actions/runs/13591846492

src/lib/shim/src/lib.rs
```diff
@@ -44,18 +45,56 @@ pub fn simtime() -> Option<SimulationTime> {
     SimulationTime::from_c_simtime(unsafe { bindings::shim_sys_get_simtime_nanos() })
 }
 
-mod tls_allow_native_syscalls {
-    use super::*;
+/// Describes whether something is within the context of the shadow (shim) code,
+/// or the application/plugin code.
```
stevenengler, line 49: nit: "Describes whether something is" - all of the methods on `ExecutionContext` rely on the global `CURRENT_EXECUTION_CONTEXT`, so is it possible to be more specific in this comment since it seems like `ExecutionContext` is only used for a single thing? For example something like "Controls the execution context of the shim, which may be within the context of the ..."?

src/lib/shim/src/lib.rs
```diff
@@ -44,18 +45,56 @@ pub fn simtime() -> Option<SimulationTime> {
...
+#[derive(Debug, Copy, Clone, Eq, PartialEq, IntoPrimitive, TryFromPrimitive)]
+#[repr(u8)]
+pub enum ExecutionContext {
+    Shadow,
+    Application,
+}
+
+static CURRENT_EXECUTION_CONTEXT: ShimTlsVar<Cell<ExecutionContext>> =
+    ShimTlsVar::new(&SHIM_TLS, || Cell::new(ExecutionContext::Application));
```
stevenengler, line 62: Maybe add a comment to this saying not to access it directly, and only access it through `ExecutionContext`?

src/lib/shim/shim_api_c.h
```diff
@@ -10,15 +10,20 @@
...
-/// libraries that are linked to the shim. Those libraries should only call
-/// functions defined here after including this header file.
+#include "lib/shim/shim_api.h"
+
+/// This module defines *C implementations* of functions that can be called by
+/// external (preloaded) libraries that are linked to the shim. Those libraries
+/// should only call functions defined here after including this header file.
+///
+/// The public interfaces to access these are defined in Rust, without prefix
+/// `shim_api_` instead of `shimc_api`.
```
stevenengler, line 20: Should this be "with" rather than "without"?

src/lib/shim/src/lib.rs
sporksmith: Done.

src/lib/shim/src/lib.rs
sporksmith: Done.

src/lib/shim/shim_api_c.h
sporksmith: Oops; yes :)

sporksmith:
The Tor benchmark run-time appears to be the same https://github.com/shadow/benchmark-results/blob/master/tor/2025-03-10-T23-15-42/plots/run_time.png

I guess we don't plot wall-clock runtime for the tgen benchmark? https://github.com/shadow/benchmark-results/tree/master/tgen/2025-03-10-T22-47-59

stevenengler:
> I guess we don't plot wall-clock runtime for the tgen benchmark? https://github.com/shadow/benchmark-results/tree/master/tgen/2025-03-10-T22-47-59

We used to using the plotting scripts in the shadow repo (for example [2024-06-28-T03-12-30](https://github.com/shadow/benchmark-results/blob/master/tgen/2024-06-28-T03-12-30/plots/shadow.results.pdf)), but we removed those a few months ago.

PR #3520, opened 2025-02-28
Add preemption to escape pure-cpu busy-loops

### LINKED ISSUE #2066: Implement true preemption for managed code

sporksmith:
This new feature is off by default, and enabled via the new experimental `--native-preemption-enabled` option.

When enabled, the shim code uses a native timer to interrupt managed code that runs for longer than `--native-preemption-native-interval` (100 ms by default) without returning control to shadow, and moves simulated time forward by `--native-preemption-sim-interval` (10 ms by default).

It is intended to be used to escape CPU-only busy loops.

Fixes #2066

### DIFF
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -82,15 +82,22 @@ where work arrives quickly. However since Shadow normally doesn't advance time
 when making non-blocking syscalls or allow other threads to run, such a loop can
 run indefinitely, deadlocking the whole simulation.
 
-When feasible, it's usually good practice to modify such loops to have a bound
-on the number of iterations instead of or in addition to a bound on wallclock
-time.
+### Ideal solution: modify the managed program
 
-For cases where modifying the loop is infeasible, Shadow provides the option
-`--model-unblocked-syscall-latency`. When this option is enabled, Shadow moves
-time forward a small amount on *every* syscall (and VDSO function call), and
-switches to another thread if one becomes runnable in the meantime (e.g. because
-network data arrived when the clock moved forward, unblocking it).
+Even outside of Shadow, it's usually good practice for such loops to have a
+bound on the number of iterations instead of or in addition to a bound on
+wallclock time. When feasible, modifying the relevant loops to do this, and
+better yet upstreaming that modification, is typically the ideal solution.
+
+### Workaround: have Shadow model unblocked syscall latency
+
+For cases where modifying the loop is infeasible, and the busy loop has a bound
+on wallclock time or contains some other syscall (such as `sched_yield`), Shadow
+provides the option `--model-unblocked-syscall-latency`. When this option is
+enabled, Shadow moves time forward a small amount on *every* syscall (and VDSO
+function call, and time-check via `rdtsc` instruction), and switches to another
+thread if one becomes runnable in the meantime (e.g. because network data
+arrived when the clock moved forward, unblocking it).
 
 This feature should only be used when it's needed to get around such loops. Some
 limitations:
@@ -119,10 +126,54 @@ the impact on simulation results.
 is more accurate, since syscalls on real systems *do* take non-zero time, but it
 makes the time model more complex to understand and reason about.
 
-* It still doesn't account for time spent by the CPU executing code,
-which also means that a busy-loop that makes no syscalls at all can still lead
-to deadlock. Fortunately such busy loops are rare and are generally agreed upon
-to be bugs, since they'd also potentially monopolize a CPU indefinitely when run
-natively.
+### Workaround: have Shadow preempt CPU-only busy-loops{#cpu-busy-loops}
+
+In cases where enabling `--model-unblocked-syscall-latency` doesn't get the
+simulation out of the busy loop, it may be because the busy-loop makes no syscalls or
+time-checks at all, and instead is waiting indefinitely for another thread to
+modify memory (e.g. to flip a "ready" flag). In Shadow's default mode of
+operation, it will never regain control from such loops, and hence can't move
+the simulation forward.
+
+Such loops can be escaped by enabling the experimental option
+`--native-preemption-enabled`. In this mode of operation, Shadow uses a native
+Linux timer to preempt the thread in such situations, moving simulated time
+forward and allowing other threads in the process to run.
+
+Drawbacks:
+
+* Loss of simulation determinism. In different runs of the simulation, the native timer
+  may fire at different times. In cases where it is truly interrupting a
+  CPU-only-busy-loop that would otherwise run indefinitely, this is unlikely to
+  ultimately change the results of the simulation, but *could*, e.g. if the busy
+  loop counts how many times it iterates and that count is used later. Worse,
+  the native timer may fire during long CPU-only operations that *would* finish
+  on their own (e.g. copying, encrypting, or decrypting large chunks of data in
+  memory), firing at different points in some runs of the simulation, or not at
+  all in some runs of the simulation.
+
+* It causes the simulation to run slower.
+
+  * Even when the timer never fires, there is substantial additional overhead
+    for the additional bookkeeping and management of the native timers.
+
+  * As with `--model-unblocked-syscall-latency`, when this feature actually does
+    cause a thread to be rescheduled, there is some performance overhead for that
+    rescheduling.
+
+* It's not meant as an accurate model of CPU-time spent; for that, see
+  [#2060](https://github.com/shadow/shadow/issues/2060). e.g. CPU-time consumed
+  between syscalls still takes 0 simulation time, if not enough time passes for
+  the native timer to fire.
+
+* Nonetheless it *does* affect simulation results (in simulations where the
+  timer actually triggers a preemption). Arguably this makes the simulation
+  somewhat more accurate, since it at least causes some CPU-heavy sections of
+  code that make no syscalls to take some non-zero amount of simulated time
+  instead of zero simulated time, but it also makes the passing of simulated
+  time more complex to understand (and potentially non-deterministic, as noted
+  above).
+
+### Further discussion
 
 For more about this topic, see [#1792](https://github.com/shadow/shadow/issues/1792).
--- a/src/main/core/configuration.rs
+++ b/src/main/core/configuration.rs
@@ -174,6 +174,37 @@ impl ConfigOptions {
         SimulationTime::from_nanos(nanos)
     }
 
+    pub fn native_preemption_enabled(&self) -> bool {
+        self.experimental.native_preemption_enabled.unwrap()
+    }
+
+    pub fn native_preemption_native_interval(
+        &self,
+    ) -> anyhow::Result<linux_api::time::kernel_old_timeval> {
+        let t = self.experimental.native_preemption_native_interval.unwrap();
+        let t = core::time::Duration::from(t);
+        // TODO: Would be a little nicer to surface this error when we parse the
+        // config. I think ideally we'd update the type such that some bounds
+        // can be enforced at parse time.
+        if t < core::time::Duration::from_micros(1) {
+            return Err(anyhow::anyhow!(
+                "native_preemption_native_interval must be >= 1 microsecond. Got {t:?}."
+            ));
+        }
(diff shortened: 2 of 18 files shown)


sporksmith:
Tor benchmark: https://github.com/shadow/benchmark/actions/runs/13932935805

stevenengler:
> Tor benchmark: https://github.com/shadow/benchmark/actions/runs/13932935805. (With this feature disabled as per default, to ensure it doesn't introduce unacceptable overhead when not actually used)

I also [started](https://github.com/shadow/benchmark/actions/runs/13933155484) a tor benchmark with preemption enabled and with 3 repetitions. Not that it will affect whether we merge this or anything, I'm just curious what the results look like. If it looks like it will take a long time, then we can cancel it.

sporksmith:
> Tor benchmark: https://github.com/shadow/benchmark/actions/runs/13932935805. (With this feature disabled as per default, to ensure it doesn't introduce unacceptable overhead when not actually used)

Results:

https://github.com/shadow/benchmark-results/blob/master/tor/2025-03-18-T20-32-33/plots/run_time.png

Looks like a small but noticeable performance hit :(. [7318 s](https://github.com/shadow/benchmark-results/blob/master/tor/2025-03-18-T20-32-33/tornet-0.15-1/tornet.plot.data/resource_usage.json#L1215C16-L1215C27)  vs [7219 s](https://github.com/shadow/benchmark-results/blob/master/tor/2025-03-18-T04-14-04/tornet-0.15-1/tornet.plot.data/resource_usage.json#L1213), so about 1.3%.

My prime suspect is the syscall handler, where I'm no longer skipping some code when `model-unblocked-syscall-latency` is disabled <https://github.com/shadow/shadow/pull/3520/files#diff-dbfa63cb768b43e86e5072a4389a1ea55c4224eb25185587495f636863c7a227L271>. I did this since most of that logic also applies to when this new feature is enabled. I suppose I could check for both features to potentially skip the whole block, but that's a little unwieldy and I may have to do a little plumbing to be able to check there. Or I could just require `model-unblocked-syscall-latency` is enabled for native preemption to work, which you'd probably want anyways in practice.

I'll do some more experimentation tomorrow to try to confirm if this is the culprit. (I suppose I should also run multiple trials, but I think this is a big enough regression to probably not just be "unlucky")

src/main/host/syscall/handler/mod.rs
```diff
@@ -268,16 +268,14 @@ impl SyscallHandler {
...
-        {
+        if ctx.process.is_running() && !matches!(rv, Err(SyscallError::Blocked(_))) {
             let max_unapplied_cpu_latency = ctx.host.shim_shmem().max_unapplied_cpu_latency;
 
             // increment unblocked syscall latency, but only for non-shadow-syscalls, since the
             // latter are part of Shadow's internal plumbing; they shouldn't necessarily "consume"
             // time
-            if !is_shadow_syscall(syscall) {
+            if ctx.host.shim_shmem().model_unblocked_syscall_latency && !is_shadow_syscall(syscall)
+            {
```
stevenengler: The interaction of the new `native_preemption_sim_interval` with the old `max_unapplied_cpu_latency` is not clear to me from the docs.

My understanding is that what the timer goes off, the shim increases `unapplied_cpu_latency` by `native_preemption_sim_interval` and does a yield syscall. Then here in shadow we only update the time `if unapplied_cpu_latency > max_unapplied_cpu_latency`. Otherwise we do nothing.

But the docs don't mention any relationship between `native_preemption_sim_interval` and `max_unapplied_cpu_latency`.

stevenengler (commented):
I didn't finish reviewing it yet (I haven't looked at most of the shim changes), so just leaving the comments I have so far.

> I'll do some more experimentation tomorrow to try to confirm if this is the culprit. (I suppose I should also run multiple trials, but I think this is a big enough regression to probably not just be "unlucky")

If the benchmark I started is still running when you want to start another, please cancel mine, I don't want it to block anything.

> Looks like a small but noticeable performance hit :(. 7318 s vs 7219 s, so about 1.3%.
>
> My prime suspect is the syscall handler, where I'm no longer skipping some code when `model-unblocked-syscall-latency` is disabled

IMO I don't think the difference is big enough to prevent merging. But if it turns out that the performance regression can be eliminated by skipping that code path, I think it would be nice to have.

sporksmith:
Haven't been able to reproduce the performance regression locally, possibly just because my system is too noisy. I made some tweaks to the hot path in the shim and am trying again with 3 trials: https://github.com/shadow/benchmark/actions/runs/13936903437

sporksmith:
After that last change and running 3 trials, the performance difference looks to be in the noise. <https://github.com/shadow/benchmark-results/blob/master/tor/2025-03-19-T06-28-41/plots/run_time.png>

Performance with the feature enabled is definitely slower (as expected), though not horrendous <https://github.com/shadow/benchmark-results/blob/master/tor/2025-03-18-T22-45-17/plots/run_time.png>. I think the only way to improve it substantially would be to use the per-thread `timer_create` timers *without* continuously destroying and recreating them when we switch threads (and thus instead potentially running into the system-wide resource limit). I don't plan to spend more time on it right now, though.

sporksmith:
Also worth noting that as-expected having preemption enabled doesn't appear to have changed any of the results *inside* the simulation <https://github.com/shadow/benchmark-results/tree/master/tor/2025-03-18-T22-45-17>.

I expect there were probably no preemptions. When I've run the test suite locally (including the tor-minimal test) with the feature globally enabled, the only test besides the new cpu-busy-loop test that triggers it is the mmap test when initializing large chunks of memory, and that only in debug builds.

src/main/core/configuration.rs
```diff
@@ -174,6 +174,25 @@ impl ConfigOptions {
...
+        self.experimental.native_preemption_enabled.unwrap()
+    }
+
+    pub fn native_preemption_native_interval(&self) -> linux_api::time::kernel_old_timeval {
+        let t = self.experimental.native_preemption_native_interval.unwrap();
+        let t = core::time::Duration::from(t);
+        linux_api::time::kernel_old_timeval {
+            tv_sec: t.as_secs().try_into().unwrap(),
+            tv_usec: t.subsec_micros().into(),
+        }
```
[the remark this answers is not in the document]
sporksmith: Done. FWIW I think a zero value would "just" result in never actually arming the timer, and hence behave similarly to native preemption not being enabled at all, but yeah it'd be confusing.

src/main/host/syscall/handler/mod.rs
sporksmith: Right. Added a note to the config spec.

FWIW I'd expect that normally `native_preemption_sim_interval` would be > `max_unapplied_cpu_latency`, making this moot. If it *is* less, than the unapplied cpu latency will still accumulate over multiple native timer intervals until either it exceeds `max_unapplied_cpu_latency` or control is returned to shadow and shadow schedules a different thread.

sporksmith: (That reminds me I noticed that I couldn't find where unapplied cpu latency is discarded when switching threads, despite that being noted in a couple places in the docs. That's semi-orthogonal to this PR, though; it mostly means that more preemptions may happen than ideal/expected when unblocked syscall latency is enabled, which mostly means slightly worse performance)

sporksmith: > (That reminds me I noticed that I couldn't find where unapplied cpu latency is discarded when switching threads, despite that being noted in a couple places in the docs. That's semi-orthogonal to this PR, though; it mostly means that more preemptions may happen than ideal/expected when unblocked syscall latency is enabled, which mostly means slightly worse performance)

Ah, I was mistaken. It is reset in `Process::continue`

src/main/host/syscall/handler/mod.rs
```diff
@@ -268,48 +268,41 @@ impl SyscallHandler {
...
-                        .unapplied_cpu_latency = SimulationTime::ZERO;
+                    log::trace!(
+                        "Reached max-unapplied-cpu-latency, but not max runahead; Incrementing time"
+                    );
                     Worker::set_current_time(new_time);
                 } else {
-                    log::trace!("Reached unblocked syscall limit; Yielding");
+                    log::trace!(
+                        "Reached max-unapplied-cpu-latency, and max runahead; Rescheduling"
+                    );
```
stevenengler, line 305: These two cases seem like they behave differently. If `new_time <= max_time`, then we update the current time for all threads of all processes on this host. But if `new_time > max_time`, we block the current thread and other threads and processes continue running at the current time. I wonder if we should set the current time to `max_time` here?

Maybe this is outside the scope of this MR. Is it worth opening an issue for?

stevenengler (approved):
Looks good, I think this will be a nice feature to have.

stevenengler:
> After that last change and running 3 trials, the performance difference looks to be in the noise. https://github.com/shadow/benchmark-results/blob/master/tor/2025-03-19-T06-28-41/plots/run_time.png

(Just pasting the image into a comment so we have it even if we need to prune benchmark results from the repo in the future.)

run_time

docs/limitations.md
```diff
@@ -119,10 +126,56 @@ the impact on simulation results.
...
+
+* It may cause the simulation to run slower.
+
+  * As with `--model-unblocked-syscall-latency`, when this feature actually does
+    cause a thread to be rescheduled, there is some performance overhead for that
+    rescheduling.
+
+  * Even when the timer never fires, there is some additional overhead for the
+    additional bookkeeping and management of the native timers. This overhead
+    *usually* won't be very noticeable other than in managed code that makes
```
sporksmith, line 163: Note to self to update this: this is definitely not true; the overhead is substantial.

sporksmith: I think when I wrote it I was thinking about the overhead of the feature when not enabled. When it is enabled it introduces about 4 syscalls to every managed-code-syscall (or other transition from managed code to shim code and back), so unsurprisingly has noticeable overhead.

robgjansen: Can we add a note about the substantial overhead to the CLI help string (if we didn't already) so that if a user sees the help menu and then enables the switch we'll have tempered their expectations WRT performance?

src/main/host/syscall/handler/mod.rs
sporksmith: Added comments to try to clarify -

when `new_time < Worker::max_event_runahead_time`, we can safely bump the worker's time to `new_time`.

When it's not (because another scheduled event or the end of round occurs before `new_time`), we can't safely bump the worker's time to `new_time`, so we block and schedule ourselves to run at `new_time` instead.

i.e. 

> If new_time <= max_time, then we update the current time for all threads of all processes on this host.

Correct, but this condition only holds if no other threads or processes on the host would actually run before `new_time`.

src/test/cli/help-long-expected
```diff
@@ -93,6 +93,23 @@ Experimental (Unstable and may change or be removed at any time, regardless of S
...
+          When true, and when managed code runs for an extended time without returning control to
+          shadow (e.g. by making a syscall), shadow preempts the managed code and moves simulated
+          time forward. This usually shouldn't be needed, but can be used to escape "pure-CPU busy
+          loops". [default: false]
+
+      --native-preemption-native-interval <seconds>
+          When `native_preemption_enabled` is true, amount of native CPU-time to wait before
+          preempting managed code that hasn't returned control to shadow.
+          
+          Only supports microsecond granularity, and values below 1 microsecond are rejected.
```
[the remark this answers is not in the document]
sporksmith: Done. Yup, that appears to fix it.

Issue #3526, opened 2025-03-06
Shadow panics when sending a packet in CI test


cohosh:
**Describe the issue**

Shadow is crashing since we updated the version we used in our Gitlab CI integration tests. 

```
** Starting Shadow 3.2.0
thread 'shadow-worker' panicked at main/core/worker.rs:357:86:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/panicking.rs:692:5
   1: core::panicking::panic_fmt
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:75:14
   2: core::panicking::panic
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:145:5
   3: core::option::unwrap_failed
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/option.rs:2015:5
   4: shadow_rs::core::worker::Worker::send_packet
   5: shadow_rs::core::worker::Worker::with
   6: <shadow_rs::network::router::Router as shadow_rs::network::PacketDevice>::push
   7: shadow_rs::network::relay::Relay::run_forward_task
   8: shadow_rs::host::host::Host::execute
   9: shadow_rs::core::worker::Worker::with
  10: scheduler::thread_per_core::HostIter<HostType>::for_each
  11: shadow_rs::core::manager::Manager::run::{{closure}}::{{closure}}
  12: scheduler::thread_per_core::SchedulerScope<HostType>::run_with_data::{{closure}}
  13: scheduler::pools::unbounded::work_loop
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
thread 'shadow-worker' panicked at main/host/host.rs:971:9:
assertion failed: self.shim_shmem_lock.borrow().is_none()
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/panicking.rs:692:5
   1: core::panicking::panic_fmt
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:75:14
   2: core::panicking::panic
[...]

PR #3531, opened 2025-03-12
Fix the issue where loopback-bound sockets attempt to connect to a public IP

magnified103:
In the current sys_connect handler, sockets bound to localhost are not handled correctly.

https://github.com/shadow/shadow/blob/3e478a9cd1b3277650a6d957a22ddb8de87e6e34/src/main/host/descriptor/socket/inet/legacy_tcp.rs#L760-L765

Shadow does not check cases where it connects to an external address using a socket bound to 127.0.0.1, which later causes crashes when the worker fails to resolve the localhost address to a host.

https://github.com/shadow/shadow/blob/3e478a9cd1b3277650a6d957a22ddb8de87e6e34/src/main/core/worker.rs#L356-L358

The same issue also occurs with UDP (and possibly in tcp.rs).

This PR brings a fix to this issue, although I think we'd want to mimic a routing table and perform these handling there.


magnified103:
The current fix was based on two-month-old commits and passed all tests at that time. I'm still investigating why the tests are failing now.

stevenengler:
> The current fix was based on two-month-old commits and passed all tests at that time. I'm still investigating why the tests are failing now.

In the past month, we've migrated Shadow to the rust 2024 edition, the rustfmt 2024 edition, and are building with rust 1.85 in the CI now. There have also been some CI fixes due to a breaking change in rustup that may affect older code.

stevenengler:
Also thanks for the PR! One additional thing we'll need is some tests in `src/test/socket/` to ensure it follows Linux behaviour and so that we don't accidentally break it in the future. Probably in `src/test/socket/send_recv/test_send_recv.rs`. We have some documentation here at https://shadow.github.io/docs/guide/writing_tests.html#system-call-tests.

(Also see #3526.)

magnified103:
This PR is ready for review.

I believe the new TCP implementation also has this issue. I attempted to fix it but couldn't write a test for the fix, as adding `connect-new-tcp` to the CMake config already causes existing tests to fail.

Additionally, I found a minor issue: calling connect() to a public IP on a socket bound to 127.0.0.1 returns EINVAL, but I believe it should return EISCONN.

src/test/socket/connect/test_connect.rs
```diff
@@ -1144,6 +1149,60 @@ fn test_server_close_during_blocking_connect(
...
+
+    let other_ip: std::net::Ipv4Addr = if test_utils::running_in_shadow() {
+        // this IP is the IP for the host 'othernode' in the shadow config file
+        "26.153.52.74".parse().unwrap()
+    } else {
+        // if running outside of shadow, we use a local network address here so that the tests
+        // running outside of shadow would only be trying to connect to a server on a local
+        // network rather than some random server on the internet
+        "192.168.1.100".parse().unwrap()
+    };
```
stevenengler, line 1186: Thanks for making this a local address when run outside of shadow :) Can we use "192.168.1.100" for both inside and outside of Shadow? I know it's a little weird to give a simulated host the local network address "192.168.1.100", but this is just for a test, and is something we support now as of #3414.

src/test/socket/send_recv/test_send_recv.rs
```diff
@@ -2441,6 +2452,69 @@ fn test_unix_dgram_multiple_senders() -> Result<(), String> {
...
+
+    let other_ip: std::net::Ipv4Addr = if test_utils::running_in_shadow() {
+        // this IP is the IP for the host 'othernode' in the shadow config file
+        "26.153.52.74".parse().unwrap()
+    } else {
+        // if running outside of shadow, we use a local network address here so that the tests
+        // running outside of shadow would only be trying to connect to a server on a local
+        // network rather than some random server on the internet
+        "192.168.1.100".parse().unwrap()
+    };
```
stevenengler, line 2492: Same comment here about the address.

stevenengler (changes requested):
Thanks for the fixes and for the tests. This looks great! I just have one small change request, and then it's ready to merge.

stevenengler:
> I believe the new TCP implementation also has this issue. I attempted to fix it but couldn't write a test for the fix, as adding `connect-new-tcp` to the CMake config already causes existing tests to fail.

Don't worry about the new TCP implementation. It still needs lots of work, and once we eventually enable the tests, they should fail and we'll remember to fix this there too.

> Additionally, I found a minor issue: calling connect() to a public IP on a socket bound to 127.0.0.1 returns EINVAL, but I believe it should return EISCONN. The EISCONN check is currently in the C code and needs to be moved out.
> 
> https://github.com/shadow/shadow/blob/3e478a9cd1b3277650a6d957a22ddb8de87e6e34/src/main/host/descriptor/tcp.c#L1713-L1716
> 
> Do you want me to include the fix for that issue within this PR?

If there's a bug here, feel free to fix this either in this MR, or a new MR.

But I'm not clear on what the issue is here. Doesn't the new `test_loopback_bound_connect` test you added check this case? The test checks that this returns `EINVAL` for both TCP and UDP, both in shadow and outside shadow?

src/main/host/descriptor/socket/inet/legacy_tcp.rs
stevenengler, line 1: And also, can you add a changelog entry to `CHANGELOG.md` under the "PATCH" section. I think this is an important bug fix.

stevenengler:
I'm going to merge the PR as-is, and I'll do a follow up with the small changes. I hope you don't mind. I think this PR is good and I'd like to get this in to shadow.

magnified103:
> I'm going to merge the PR as-is, and I'll do a follow up with the small changes. I hope you don't mind. I think this PR is good and I'd like to get this in to shadow.

Hi, sorry, I was a bit busy last week. I'll make the necessary changes today, and it'll be good.

stevenengler:
> > I'm going to merge the PR as-is, and I'll do a follow up with the small changes. I hope you don't mind. I think this PR is good and I'd like to get this in to shadow.
> 
> Hi, sorry, I was a bit busy last week. I'll make the necessary changes today, and it'll be good.

Okay no problem, I don't mean to rush you. I just figured the changes I mentioned would be easy for me to add. But if you want to do them then that's great too, whenever you get time.

src/test/socket/send_recv/test_send_recv.rs
magnified103: Changed to `192.168.1.100`

src/test/socket/connect/test_connect.rs
magnified103: I've changed it to `192.168.1.100`.

src/main/host/descriptor/socket/inet/legacy_tcp.rs
magnified103: I've added an entry to CHANGELOG.

magnified103:
I've added another fix and test for loopback-bound listening sockets: sys_connect should return `EISCONN` regardless of whether the connecting address is loopback or not.

Can you review the changes I just made?

stevenengler (approved):
> I've added another fix and test for loopback-bound listening sockets: sys_connect should return `EISCONN` regardless of whether the connecting address is loopback or not.
> 
> Can you review the changes I just made?

These look good, thanks!

PR #3535, opened 2025-03-24
Drop support for Ubuntu 20.04

stevenengler:
The CI workflow says Ubuntu 20.04 EOL is April 2025, so just queuing up an MR so that we can drop support in April.

We also had some code for installing 'clang-12' in the container scripts. I don't think we need this anymore, so I also dropped that code.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -20,8 +20,6 @@ jobs:
     strategy:
       matrix:
         container:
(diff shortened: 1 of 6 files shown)


robgjansen:
FYI, it looks like the end of standard support for Ubuntu 20.04 LTS has been pushed back to May 31st, 2025.
https://wiki.ubuntu.com/Releases

robgjansen (approved):
Woo! I went ahead and removed 20.04 from the required CI action list too.

Issue #3538, opened 2025-03-25
nondeterminism in tor simulations with --model-unblocked-syscall-latency=true


sporksmith:
The tor minimal test is quite close to fully deterministic normally, but enabling `--model-unblocked-syscall-latency=true` causes the strace logs to diverge, seemingly  usually starting with the return time of a blocked `epoll_wait` call.

# Repro steps

First enable both `--model-unblocked-syscall-latency` and deterministic strace logging for the tor test:

```patch
diff --git a/src/test/tor/minimal/CMakeLists.txt b/src/test/tor/minimal/CMakeLists.txt
index 246363d21..c73ae4836 100644
--- a/src/test/tor/minimal/CMakeLists.txt
+++ b/src/test/tor/minimal/CMakeLists.txt
@@ -15,7 +15,8 @@ add_shadow_tests(BASENAME tor-minimal
                  ARGS
                    --use-cpu-pinning true
                    --parallelism 2
-                   --strace-logging-mode off
+                   --strace-logging-mode deterministic
+                   --model-unblocked-syscall-latency=true
                    # Disable to support fork
                    --use-memory-manager false
                    --template-directory "shadow.data.template"
```

Build and run the test as usual. Locally I've also hacked up the tests cmake file to omit the golang tests, since building them requires root (an unshaven yak for another day)

```shell
# build
./setup build --debug --test --extra
# run the test
./setup test --extra tor-minimal
# save the results
mv build/src/test/tor/minimal/tor-minimal-shadow.data/ build/src/test/tor/minimal/tor-minimal-shadow.data.0/
# run the test again
./setup test --extra tor-minimal
# move again
mv build/src/test/tor/minimal/tor-minimal-shadow.data/ build/src/test/tor/minimal/tor-minimal-shadow.data.1/
# diff any relay strace file
$ diff --side-by-side --suppress-common-lines build/src/test/tor/minimal/tor-minimal-shadow.data.[01]/hosts/relay1/tor.1000.strace | head -n10
00:15:04.206800750 [tid 1000] epoll_wait(3, <pointer>, 32, 55 |	00:15:04.206794100 [tid 1000] epoll_wait(3, <pointer>, 32, 55
000000904206800750 [tid 1000] clock_gettime(...) = 0	      |	000000904206794100 [tid 1000] clock_gettime(...) = 0
000000904206801760 [tid 1000] time(...) = 946685704	      |	000000904206795110 [tid 1000] time(...) = 946685704
[...]

Issue #2693, opened 2023-01-24
Non-determinism in golang runtime (etcd example, golang tests, obfs4proxy, ...)


stevenengler:
The syscall order isn't always consistent in the etcd example. For example, see the following two runs:

```diff
diff --git a/shadow.data/hosts/server1/server1.etcd.1000.strace b/shadow.data.2/hosts/server1/server1.etcd.1000.strace
index 3ac9e59f..17eb6044 100644
--- a/shadow.data/hosts/server1/server1.etcd.1000.strace
+++ b/shadow.data.2/hosts/server1/server1.etcd.1000.strace
@@ -207,10 +207,13 @@
 00:00:00.000000000 [tid 1003] gettid(...) = 1003
 00:00:00.000000000 [tid 1003] futex(...) = 1
 00:00:00.000000000 [tid 1000] futex(...) = 0
+000000000000000000 [tid 1000] clock_gettime(...) = 0
 00:00:00.000000000 [tid 1000] futex(...) = 1
 00:00:00.000000000 [tid 1003] futex(...) = 0
+000000000000000000 [tid 1003] clock_gettime(...) = 0
 00:00:00.000000000 [tid 1003] futex(...) = 1
 00:00:00.000000000 [tid 1000] futex(...) = 0
+000000000000000000 [tid 1000] clock_gettime(...) = 0
 00:00:00.000000000 [tid 1000] clock_gettime(...) = 0
 00:00:00.000000000 [tid 1000] rt_sigprocmask(...) = 0
 00:00:00.000000000 [tid 1000] mmap(0x0, 8392704, PROT_NONE, MAP_PRIVATE | MAP_ANON | MAP_ANONYMOUS | MAP_STACK, -1, 0) = 0x7fffceedd000
@@ -221,15 +224,15 @@
 00:00:00.000000000 [tid 1000] rt_sigprocmask(...) = 0
 00:00:00.000000000 [tid 1000] rt_sigprocmask(...) = 0
 000000000000000000 [tid 1000] clock_gettime(...) = 0
-000000000000000000 [tid 1000] clock_gettime(...) = 0
-000000000000000000 [tid 1000] clock_gettime(...) = 0
 00:00:00.000000000 [tid 1000] fcntl(0, 3) = 32768
 00:00:00.000000000 [tid 1000] fcntl(1, 3) = 32769
 00:00:00.000000000 [tid 1000] fcntl(2, 3) = 32769
[...]

PR #3542, opened 2025-03-27
Make AT_RANDOM in the linux auxiliary vector deterministic

### LINKED ISSUE #3539: auxval AT_RANDOM isn't deterministic
Given the following test program:

```
#include <sys/auxv.h>
#include <stdio.h>

int main() {
    unsigned long *x = (void*)getauxval(AT_RANDOM);
    printf("*getauxval(AT_RANDOM) %lx\n", *x);
}
```

shadow currently produces non-deterministic results:

```
$ shadow-exec ./a.out 
*getauxval(AT_RANDOM) 280741edbe5ef200
$ shadow-exec ./a.out 
*getauxval(AT_RANDOM) f2cf2d8a17f3400c
```

i.e. this is initialized with some random value before shadow gets control. Maybe by the dynamic linker/loader? Or maybe even by the linux kernel itself?

I'm not aware of any real-world programs that use this value, though it would be a subtle source of nondeterminism to detect if and when we do bump into it. (I was hoping maybe it was the source of nondeterminism in #2693 , but it doesn't appear to be)

sporksmith:
Fixes #3539, and progress (maybe completion?) on #2693 and #3538.

### DIFF
--- a/src/lib/shadow-shim-helper-rs/src/shim_event.rs
+++ b/src/lib/shadow-shim-helper-rs/src/shim_event.rs
@@ -11,7 +11,7 @@ pub struct ShimEventStartRes {
     /// in its auxiliary vector.
     // TODO: Consider replacing this with a seed for a shim-side PRNG, particularly
     // when implementing the getrandom vdso (#3362).
-    pub aux_at_random: [u8; 16],
+    pub auxvec_random: [u8; 16],
 }
 
 #[derive(Copy, Clone, Debug, VirtualAddressSpaceIndependent)]
--- a/src/lib/shadow-shim-helper-rs/src/shim_event.rs
+++ b/src/lib/shadow-shim-helper-rs/src/shim_event.rs
@@ -5,6 +5,17 @@ use crate::syscall_types::{ForeignPtr, SyscallArgs, SyscallReg, UntypedForeignPt
 
 #[derive(Copy, Clone, Debug, VirtualAddressSpaceIndependent)]
 #[repr(C)]
+/// Data for [`ShimEventToShim::StartRes`]
+pub struct ShimEventStartRes {
+    /// Pseudorandom data to be used to overwrite the data linux provides
+    /// in its auxiliary vector.
+    // TODO: Consider replacing this with a seed for a shim-side PRNG, particularly
+    // when implementing the getrandom vdso (#3362).
+    pub aux_at_random: [u8; 16],
+}
+
+#[derive(Copy, Clone, Debug, VirtualAddressSpaceIndependent)]
+#[repr(C)]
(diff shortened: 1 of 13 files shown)


src/lib/shadow-shim-helper-rs/src/shim_event.rs
```diff
@@ -3,6 +3,17 @@ use vasi::VirtualAddressSpaceIndependent;
 
 use crate::syscall_types::{ForeignPtr, SyscallArgs, SyscallReg, UntypedForeignPtr};
 
+#[derive(Copy, Clone, Debug, VirtualAddressSpaceIndependent)]
+#[repr(C)]
+/// Data for [`ShimEventToShim::Syscall`] and [`ShimEventToShadow::Syscall`]
```
stevenengler, line 8: I see this used in `ShimEventToShim::StartRes`, but not `ShimEventToShim::Syscall` or `ShimEventToShadow::Syscall`?

src/lib/shadow-shim-helper-rs/src/shim_event.rs
```diff
@@ -3,6 +3,17 @@ use vasi::VirtualAddressSpaceIndependent;
...
+pub struct ShimEventStartRes {
```
stevenengler, line 9: What does "Res" stand for? I'd usually guess "result", but I'm not sure that makes sense in this context.

src/lib/shim/src/lib.rs
```diff
@@ -544,7 +544,15 @@ fn wait_for_start_event(get_initial_working_dir: bool) {
...
         ipc.from_shadow().receive().unwrap()
     });
-    assert!(matches!(res, ShimEventToShim::StartRes));
+    let ShimEventToShim::StartRes(res) = res else {
+        panic!("Unexpected response: {res:?}");
+    };
+    if is_first_thread {
+        // SAFETY: We're ensuring serial execution in this process, and no other
+        // Rust code in this library should have tried accessing the auxiliary
+        // vector yet, so no references should exist.
```
stevenengler, line 554: Maybe note that we assume/hope plugin code hasn't run yet and accessed or made a constant reference to the `AT_RANDOM` data? And that the loader may have already used this value, but that's fine?

src/lib/shim/src/reinit_auxvec_random.rs
```diff
@@ -1,3 +1,5 @@
+use core::{ffi::c_ulong, mem::size_of, ptr};
```
stevenengler, line 1: `size_of` is part of the core prelude, so you shouldn't need to include it here. (I'm surprised this doesn't give a clippy warning.)

stevenengler (approved):

Issue #3546, opened 2025-03-31
`ioctl-linux` test fails


magnified103:
Commit https://github.com/shadow/shadow/commit/24343d0ef1da38e2e9fe6684bdffe3cf32c1153d . Shadow assumes that UDP packets traveling over local interfaces will immediately reach the destination socket; however, this might not always be the case.

In my local testing, the behavior seems inconsistent. Most of the time, the packet won't travel without active polling (sys_read) from the receiver (as checked using SIOCGSTAMP), though occasionally it does with low probability.
```
"ioctl-linux" start time: Mar 31 11:56 +07
Output:
----------------------------------------------------------
Testing test_tty <domain=2,sock_type=1>... ✓
Testing test_tty <domain=2,sock_type=2>... ✓
Testing test_fionread <init_method=Inet, sock_type=1>... ✓
Testing test_siocgstamp <init_method=Inet, sock_type=1>... ✓
Testing test_fionread <init_method=Inet, sock_type=2>... ✓
Testing test_siocgstamp <init_method=Inet, sock_type=2>... ✗ (Time difference was too large)
Error: "Time difference was too large"
<end of output>
Test time =   0.14 sec
```

**Operating System:**
 - OS and version:```Ubuntu 24.04.1 LTS```
 - Kernel version:
```
Linux foo 6.8.0-56-generic #58-Ubuntu SMP PREEMPT_DYNAMIC Fri Feb 14 15:33:28 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
```

**Shadow:**
 - Version and build information:
```
Shadow 3.2.0 — v3.2.0-409-g24343d0ef 2025-03-31--04:23:31
GLib 2.80.0
Built on 2025-03-31--05:03:08
Built from git branch main
Shadow was built with PROFILE=release, OPT_LEVEL=3, DEBUG=false, RUSTFLAGS="-C force-frame-pointers=y", CFLAGS="-std=gnu11 -O3 -ggdb -fno-omit-frame-pointer -Wreturn-type -Wswitch -DNDEBUG"
```

stevenengler:
> In my local testing, the behavior seems inconsistent. Most of the time, the packet won't travel without active polling (sys_read) from the receiver (as checked using SIOCGSTAMP), though occasionally it does with low probability.

Can you expand on this a bit, I'm not sure I follow. How big are the delays that you see? Are they large (around the 50ms sleep duration)?

magnified103:
[...]

Issue #3207, opened 2023-10-12
Consider removing the `--summarize` rust test option


stevenengler:
Many of Shadow's rust tests have a `--summarize` option which will [continue to run tests](https://github.com/shadow/shadow/blob/8470133417c3ca29b5aad0036659b30fa495e79f/src/test/test_utils.rs#L68-L95) even after a test fails. The goal was to provide an overview of what tests were passing/failing, but I don't think we've ever really used this. It also makes writing tests a little more painful, since any tests using this feature need to return a result rather than panic. Using panicking failures in tests is nicer than results, since then you can use `assert!` and `panic!`. Returning a result also means that we need to write cleanup code for tests that fail, since other tests may still run after the failed test.

It would simplify the tests if we removed this option and use panicking tests instead.

PR #3547, opened 2025-03-31
Add better error logging to `test_siocgstamp` socket test

stevenengler:
Just outputs some extra info if the test fails.

Related: #3546

### DIFF
--- a/src/test/socket/ioctl/test_ioctl.rs
+++ b/src/test/socket/ioctl/test_ioctl.rs
@@ -303,7 +303,11 @@ fn test_siocgstamp(init_method: SocketInitMethod, sock_type: libc::c_int) -> Res
 
                 // since it was sent over localhost, the difference between the send time and
                 // receive time should be very small
-                test_utils::result_assert(difference < threshold, "Time difference was too large")?;
+                test_utils::result_assert_lt(
+                    difference,
+                    threshold,
+                    "Time difference was too large",
+                )?;
             }
             Some(e) => test_utils::result_assert_eq(
                 ioctl_siocgstamp(fd_peer),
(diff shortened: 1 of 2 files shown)


src/test/socket/ioctl/test_ioctl.rs
```diff
@@ -303,7 +303,12 @@ fn test_siocgstamp(init_method: SocketInitMethod, sock_type: libc::c_int) -> Res
 
                 // since it was sent over localhost, the difference between the send time and
                 // receive time should be very small
-                test_utils::result_assert(difference < threshold, "Time difference was too large")?;
+                assert!(
```
sporksmith, line 306: don't we want to return an error instead of panicking?

stevenengler: Originally we did because we wanted the `--summarize` option, which required us to return an error instead of panicking. But in #3207 I think we want to remove this option, which means we can switch to panics instead of returning errors.

But for consistency, I added new `result_assert_lt` and `result_assert_gt` utils, and changed this back to returning a result.

sporksmith: Ah ok. Yeah I'm in favor of simplifying things as per #3207, but makes sense to be  consistent for now.

I'd have settled for just sticking a (mildly wasteful) `format!` call into the `result_assert` call we had before, but the new helpers look good too :)

sporksmith (approved):

src/test/socket/ioctl/test_ioctl.rs
stevenengler: I didn't even think of the `format!` macro... Yeah that would have been simpler

PR #3554, opened 2025-04-05
Fix the test for getauxval when the type is not found

magnified103:
`AT_MINSIGSTKSZ` is not available for amd64 until Linux 5.14 ([source](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=1c33bb0507508af24fd754dd7123bd8e997fab2f)), so this test will fail on earlier kernel versions.

https://github.com/shadow/shadow/blob/c7976835f8cb183de76ede64155d0a3b4009eb68/src/lib/shim/src/reinit_auxvec_random.rs#L99-L104

```
test tls::test::test_multithread_mutate_mixed_alignments ... ok
test tls::test::test_multithread_mutate_small_alignment ... ok

failures:

---- reinit_auxvec_random::test::test_getauxvec stdout ----

thread 'reinit_auxvec_random::test::test_getauxvec' panicked at lib/shim/src/reinit_auxvec_random.rs:101:57:
called `Option::unwrap()` on a `None` value


failures:
    reinit_auxvec_random::test::test_getauxvec
```

This PR simply sets 0 as the default value when the type is not found (see man `getauxval(3)`), and adds additional tests to cover all auxiliary vector types.

### DIFF
--- a/src/lib/shim/src/reinit_auxvec_random.rs
+++ b/src/lib/shim/src/reinit_auxvec_random.rs
@@ -97,9 +97,18 @@ mod test {
         assert_eq!(super::getauxval(AuxVecTag::AT_RANDOM).unwrap(), unsafe {
             libc::getauxval(libc::AT_RANDOM)
         });
-        assert_eq!(
-            super::getauxval(AuxVecTag::AT_MINSIGSTKSZ).unwrap(),
-            unsafe { libc::getauxval(libc::AT_MINSIGSTKSZ) }
-        );
+        for value in 0..100 {
+            if value == libc::AT_HWCAP || value == libc::AT_HWCAP2 {
+                // libc handles AT_HWCAP and AT_HWCAP2 differently
+                continue;
+            }
+            if let Ok(tag) = AuxVecTag::try_from(value) {
+                assert_eq!(
+                    super::getauxval(tag).unwrap_or(0),
+                    unsafe { libc::getauxval(value) },
+                    "value mismatch for type {tag:?}",
+                );
+            }
+        }
     }
 }
--- a/src/lib/shim/src/reinit_auxvec_random.rs
+++ b/src/lib/shim/src/reinit_auxvec_random.rs
@@ -94,21 +94,43 @@ mod test {
     #[cfg(not(miri))]
     fn test_getauxvec() {
         // Test consistency with libc
-        assert_eq!(super::getauxval(AuxVecTag::AT_RANDOM).unwrap(), unsafe {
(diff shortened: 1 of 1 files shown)


src/lib/shim/src/reinit_auxvec_random.rs
```diff
@@ -97,9 +97,18 @@ mod test {
         assert_eq!(super::getauxval(AuxVecTag::AT_RANDOM).unwrap(), unsafe {
             libc::getauxval(libc::AT_RANDOM)
         });
-        assert_eq!(
-            super::getauxval(AuxVecTag::AT_MINSIGSTKSZ).unwrap(),
-            unsafe { libc::getauxval(libc::AT_MINSIGSTKSZ) }
-        );
+        for value in 0..100 {
```
sporksmith, line 100: It's a little hard when looking at this to know that this covers the whole range. Minimally let's add a comment about how this range is picked, though even better to verify in code that this is at least covering some non-zero chunk of real values. e.g.:

```
// Cover all tag values currently defined in AuxVecTag, skipping over "holes".
let tags: Vec<_> = (u64::from(AuxVecTag::AT_NULL)..=u64::from(AuxVecTag::AT_MINSIGSTKSZ))
    .filter_map(|x| AuxVecTag::try_from(x).ok())
    .collect();
assert_ne!(tags.len(), 0);
for tag in tags {
```

sporksmith (changes requested):
Thanks, very nice! Just a small request for clarity

src/lib/shim/src/reinit_auxvec_random.rs
sporksmith: Actually I was thinking about this some more, and part of the reason I prefer using the symbolic names directly instead of integers is that there's really no guarantee that the libc constant values (`libc::AT_*`) and kernel constant values (`AuxVecTax::AT_*`) use the same integer values at all. The most obvious implementation in libc would use the same values, but once in a while they do differ for these kinds of things, and when they do it can be very confusing.

So I'd actually lean towards dropping the loop here and just "spot checking" a couple individual symbolic names as it was before; but just picking ones that are more widely available.

Or, if you'd really like to make this exhaustive, make an explicit mapping instead of doing the int conversion trick. e.g. `for (linux_tag, libc_tag) in [(AuxVecTag::AT_RANDOM, libc::AT_RANDOM), ...]` ...

sporksmith: I went ahead and pushed my suggested changes so we can get this in. Thank you!

PR #3559, opened 2025-04-09
Make shadow-shim-helper (mostly) no-std

sporksmith:
It currently still ends up pulling in `std` via the `tcp` crate (and maybe other places I missed); but this is some progress and more importantly will help prevent accidental "back-sliding" of using more `std` in this crate.

Progress on https://github.com/shadow/shadow/issues/2919


stevenengler:
> It currently still ends up pulling in `std` via the `tcp` crate (and maybe other deps I missed)

For the tcp crate, it's only used for a `impl tcp::util::time::Duration for SimulationTime` and `impl tcp::util::time::Instant for EmulatedTime`. Maybe we can add a `std` feature flag to `shadow-shim-helper` and put the tcp crate behind that? (I thought we had a `std` feature in `shadow-shim-helper`, but I must be misremembering since I don't see it.)

sporksmith:
> For the tcp crate, it's only used for a impl tcp::util::time::Duration for SimulationTime and impl tcp::util::time::Instant for EmulatedTime. Maybe we can add a std feature flag to shadow-shim-helper and put the tcp crate behind that? (I thought we had a std feature in shadow-shim-helper, but I must be misremembering since I don't see it.)

Maybe! I had actually started picking at this because I was thinking about moving `tcp::util::time::Instant` to some no-std helper crate, for use in `vasi-sync`'s `scchannel` to set an absolute deadline, and was surprised to find that `shadow-shim-helper-rs` wasn't already `no_std`.  If it's fairly easy to make `tcp` `no_std` maybe we don't even have to move it (though having `vasi-sync` depend on `tcp` would be a little weird).

Anyway, this is semi skunk-works (towards getting shadow to notice when a managed process seems unresponsive) and just seemed like a reasonable checkpoint to get merged.

src/lib/shadow-shim-helper-rs/src/rootedcell/mod.rs
```diff
@@ -53,7 +53,13 @@ impl Tag {
         // requiring different versions
         // https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility.
         static TAG_PREFIX: OnceCell<TagGlobalId> = OnceCell::new();
-        let prefix = *TAG_PREFIX.get_or_init(rand::random);
+        let prefix = *TAG_PREFIX.get_or_init(|| {
+            let mut buf = [0u8; core::mem::size_of::<TagLocalId>()];
```
stevenengler, line 57: I think `size_of` is part of the prelude, so the full path shouldn't be needed?

stevenengler (approved):
> Maybe we can add a `std` feature flag to `shadow-shim-helper`

Looks like you already did that. Should have reviewed the PR first :)

> If it's fairly easy to make `tcp` `no_std`

The `tcp` lib makes heave use of things that require allocators (`BinaryHeap`, `VecDeque`, `LinkedList`, `bytes::Bytes`, etc). I think most of these are in `alloc`, and `bytes::Bytes` is no_std with a default `std` feature enabled. So it may be possible to make the tcp code no_std+alloc.

> Anyway, this is semi skunk-works (towards getting shadow to notice when a managed process seems unresponsive) and just seemed like a reasonable checkpoint to get merged.

SGTM!

Issue #3558, opened 2025-04-08
Netcat under bash does not exit properly (blocks instead of exiting)


stevenengler:
When running netcat under bash in shadow, the process never exits and appears to get stuck on a `poll` syscall.

```yaml
general:
  stop_time: 30s
experimental:
  strace_logging_mode: standard
network:
  graph:
    type: gml
    inline: |
      graph [
        directed 0
        node [
          id 0
          host_bandwidth_down "1 Gbit"
          host_bandwidth_up "1 Gbit"
        ]
        edge [
          source 0
          target 0
          latency "50 ms"
        ]
      ]
hosts:
  server:
    network_node_id: 0
    ip_addr: 1.2.3.4
    processes:
    - path: nc
      args: -l -k 8080
      expected_final_state: running
  client:
    network_node_id: 0
    processes:
    - path: bash
      args:
        - -c
        - echo "foo" | nc -q 0 1.2.3.4 8080 
      start_time: 2s
      expected_final_state: { exited: 0 }
```

```text
$ shadow shadow.yaml > shadow.log
** Starting Shadow 3.2.0
Error: process 'client.bash.1000' exited with status StoppedByShadow; expected end state was exited: 0 but was running
Error: Failed to run the simulation
Error: 
Error: Caused by:
Error:     1 managed processes in unexpected final state
** Shadow did not complete successfully: Failed to run the simulation
**   1 managed processes in unexpected final state
** See the log for details
```

```text
$ cat shadow.data/hosts/server/nc.openbsd.1000.stdout
foo
$ tail shadow.data/hosts/client/bash.1000.strace 
00:00:02.100000000 [tid 1002] pselect6(4, 0x0, 0x7ffc9b303730, 0x0, 0x0 <null>, 0x0) = 1
00:00:02.100000000 [tid 1002] getsockopt(3, 1, 4, 0x7ffc9b3036ec, 4 (0x7ffc9b3036f8)) = 0
00:00:02.100000000 [tid 1002] fcntl(3, 4, 2050) = 0
00:00:02.100000000 [tid 1002] poll(0x7ffc9b2fb7b0, 4, -1) = 1
00:00:02.100000000 [tid 1002] read(0, 0x7ffc9b2fb7e0, 16384) = 4
00:00:02.100000000 [tid 1002] poll(0x7ffc9b2fb7b0, 4, -1) = 2
00:00:02.100000000 [tid 1002] read(0, 0x7ffc9b2fb7e4, 16380) = 0
[...]

Issue #3100, opened 2023-08-09
FIN is sent before earlier data


stevenengler:
If an application does a large write and then a `shutdown(SHUT_WR)`, the FIN packet seems to be sent before the data (and even before acknowledging the peer's SYN). For example with the client:

```python3
s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.connect(('server', 8080))
s.send(b'1'*1000)
s.shutdown(socket.SHUT_WR)
assert b'' == s.recv(1)
s.close()
```

Shadow gives the pcap:

![1691601064_grim](https://github.com/shadow/shadow/assets/3708797/f0df27c3-5c29-4bce-a2f6-d8cbcdcaec3d)

We can see that the 1000 bytes are sent after the FIN packet, which leads to bugs.

@robgjansen This might be related to the tgen stream issue, where we can close the socket while there is still data in the send buffer, and it seems like the FIN might be sent out-of-order.

<details><summary>Server code</summary>
<p>

```python3
import asyncio
import socket
import sys

async def handle_client(client, addr):
    loop = asyncio.get_event_loop()

    while True:
        buf = await loop.sock_recv(client, 1024)
        if len(buf) == 0:
            break

    print("Connection closed:", addr)
    client.close()

async def run_server(port):
    server = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    server.bind(('0.0.0.0', port))
    server.listen(100)
    server.setblocking(False)

    loop = asyncio.get_event_loop()

    while True:
        client, addr = await loop.sock_accept(server)
        print("New connection:", addr)
        loop.create_task(handle_client(client, addr))

asyncio.run(run_server(int(sys.argv[1])))
```

</p>
</details>

stevenengler:
Also worth noting that the sequence numbers seem to be ordered correctly (the data has SEQ=2 and the FIN has SEQ=3), even though they're sent out-of-order.

I'm not sure exactly what's happening within shadow's TCP code. It *seems* like the server is dropping the FIN, receiving the data, and then the client is never retransmitting the FIN.

stevenengler:
> @robgjansen This might be related to the tgen stream issue, where we can close the socket while there is still data in the send buffer, and it seems like the FIN might be sent out-of-order.

I think I reproduced this:

```yaml
general:
  stop_time: 60 seconds
host_option_defaults:
  pcap_enabled: true
network:
  graph:
    type: gml
    inline: |
      graph [
        directed 0
[...]

PR #3562, opened 2025-04-20
Fix packet order of FIN from `SHUT_WR`

stevenengler:
Shadow assigns a priority to packets, and packets with a lower priority will be sent before other packets. Previously, shadow only assigned a priority to non-empty packets (here meaning containing a payload), and a priority of 0 to other packets. But this meant that empty packets like FIN (with a priority of 0) could be sent before other packets, even if the FIN had a higher sequence number. Shadow seems to have discarded these out-of-order FIN packets. See #3100 for an example.

This MR changes the legacy TCP code to always assign a non-zero priority if the packets has a sequence number, so that they won't be sent out of order. This (**edit:** partially; see https://github.com/shadow/shadow/issues/3100#issuecomment-2823009618) fixes (**edit:** \<text to prevent github from closing the linked issue\>) #3100, but there are still a lot of other shutdown-related bugs (see all the disabled tests in `src/test/socket/shutdown/test_shutdown.rs`), so `shutdown()` still won't work 100% as expected.

Related: #3558 (I think this MR fixes this shutdown issue as well, but it looks like there are still some other issues.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ to sort out in the case that `stdout` and `stderr` are merged. (#3428)
   as the obfs4proxy pluggable transport (#3538). (#3542)
 * Fixed the behavior of sockets bound to the loopback interface: Shadow no longer panics in some cases where `connect` and `sendmsg` syscalls are used with a non-loopback address. (#3531)
 * Fixed the behaviour of an implicit bind during a `sendmsg` syscall for UDP sockets. (#3545)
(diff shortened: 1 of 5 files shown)


robgjansen (approved):
Good find!

If only we had a nice Rust TCP stack so we could throw this janky C stack in the trash :D

stevenengler:
> If only we had a nice Rust TCP stack so we could throw this janky C stack in the trash :D

Every time I look at the rust tcp code, I remember how much work still needs to be done. Someday hopefully :)

PR #3570, opened 2025-04-24
Add another regression test for issue 3100

### LINKED ISSUE #3100: FIN is sent before earlier data

stevenengler:
Closes #3100. This is the second example simulation in that issue.


src/test/regression/3100/regression-3100-b.yaml
```diff
@@ -0,0 +1,62 @@
+general:
+  stop_time: 60 seconds
```
sporksmith, line 2: Maybe explicitly disable `model-unblocked-syscall-latency` in case the default switches in the future, since having it enabled can mask this issue by inserting a delay between the send and the close.

sporksmith (approved):

src/test/regression/3100/regression-3100-b.yaml
stevenengler: Ah good point. I'll add this to both regression tests (3100-a and 3100-b).

stevenengler:
Added changes to set `model-unblocked-syscall-latency: false`, and two more small changes to update the changelog and to remove the pcaps from the test.

I think these changes should be uncontroversial, but I'll wait a little before merging just in case.

stevenengler:
Noticed a typo in a comment, so did a [fixup/rebase](https://github.com/shadow/shadow/compare/e7cddc3650876030216aad811949ed2c92743af9..b362422bb5239a159d2397fb330579b87c7b7d23) to fix it.

sporksmith (approved):

PR #3578, opened 2025-05-02
Fix faccessat and add faccessat2

sporksmith:
Our `faccessat` handler incorrectly took an extra flags parameter, which was garbage data. This led to a heisenbug in chutney shell scripts, since sometimes the corresponding register would happen to have 0 and work, but sometimes it would have garbage data and cause the syscall to return an error.

This renames the old handler to `faccessat2`, which *does* take the flags parameter, and adds `faccessat`, which doesn't.

### DIFF
--- a/src/main/host/syscall/handler/fileat.c
+++ b/src/main/host/syscall/handler/fileat.c
@@ -297,7 +297,7 @@ SyscallReturn syscallhandler_utimensat(SyscallHandler* sys, const SyscallArgs* a
         regularfile_utimensat(dir_desc, pathname, times, flags, plugin_cwd));
 }
 
-SyscallReturn syscallhandler_faccessat(SyscallHandler* sys, const SyscallArgs* args) {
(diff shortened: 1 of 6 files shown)


stevenengler (approved):
Nice catch!

This looks similar to #3384. I wonder if we should audit all of these to make sure we don't have other instances of these 2/not-2 syscall handlers.

Also maybe double check if you haven't that we don't need to make any changes to the libc wrappers in the shim? For example do we need to add `faccessat2` somewhere in `src/lib/preload-libc/gen_syscall_wrappers_c.py`?

sporksmith:
> Maybe double check if you haven't that we don't need to make any changes to the libc wrappers in the shim? For example do we need to add faccessat2 somewhere in src/lib/preload-libc/gen_syscall_wrappers_c.py?

Yup, those are already excluded. 

https://github.com/shadow/shadow/blob/b6cfd08df5751c8876e08ae35c49015bd0bdd183/src/lib/preload-libc/gen_syscall_wrappers_c.py#L62

PR #3590, opened 2025-05-15
shadow-exec: add ability to pass through arbitrary shadow args

sporksmith:
I had started to add extra arguments to allow setting the simulation seed etc., but we can save some plumbing by just having a generic pass-through.


shadowtools/src/shadowtools/shadow_exec.py
```diff
@@ -50,6 +51,7 @@ def _main(
     stderr: TextIO = sys.stderr,
     shadow_bin: Path = Path("shadow"),
     model_unblocked_syscall_latency: bool = True,
+    shadow_args: List[str] = [],
```
stevenengler, line 53: I think this would be better as `= None` and then `if shadow_args is None: shadow_args = []`. Especially since we append to this global list below (`shadow_args.append(...)`). I know `_main` should only be called once per program, but I think it's best to avoid this footgun anyways.

shadowtools/src/shadowtools/shadow_exec.py
```diff
@@ -111,11 +113,20 @@ def _main(
     config_path.write_text(yaml.safe_dump(config))
 
     data_dir = tmpdir.joinpath("shadow.data")
-    shadow_args = [
-        f"--data-directory={data_dir}",
-        "--progress=false",
-        str(config_path),
-    ]
+    if any((re.match(r"^--data-directory[=$]|^-d", s) for s in shadow_args)):
```
stevenengler, line 113: I'm not sure that this regex is working as intended. For example the following returns `None`.

```python3
re.match(r"^--data-directory[=$]|^-d", "--data-directory")
```

shadowtools/src/shadowtools/shadow_exec.py
```diff
@@ -111,11 +113,20 @@ def _main(
...
+        # below.
+        print(
+            f"ERROR: Overriding shadow's --data-directory currently unsupported.",
+            file=stderr,
+        )
+        sys.exit(1)
+    else:
+        shadow_args.append(f"--data-directory={data_dir}")
+    if not any((re.match(r"^--progress[=$]", s) for s in shadow_args)):
+        shadow_args.append(f"--progress=false")
```
stevenengler, line 125: Does it make sense to put `progress` in the generated config file instead so that we don't need extra handling here?

stevenengler (approved):

shadowtools/src/shadowtools/shadow_exec.py
sporksmith: (since mutating a list that the caller passes is also potentially surprising)

PR #3591, opened 2025-05-17
Fix warnings for rust 1.87 and upgrade CI rust version

stevenengler:
Fixes clippy warnings in rust 1.87, and upgrades the rust version in CI.

When running clippy with rust 1.88-beta, we get "'shadow-rs' (lib test) generated 152 warnings", so we have that to look forward to in 6 weeks :)

### DIFF
--- a/src/test/socket/socket/test_socket.rs
+++ b/src/test/socket/socket/test_socket.rs
@@ -12,8 +12,8 @@ enum Cond<'a, T> {
 impl<T: std::cmp::Eq> Cond<'_, T> {
     fn matches(&self, compare: T) -> bool {
         match self {
-            Cond::Only(vals) => vals.iter().any(|val| *val == compare),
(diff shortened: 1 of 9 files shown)


src/test/socket/socket/test_socket.rs
```diff
@@ -12,8 +12,8 @@ enum Cond<'a, T> {
 impl<T: std::cmp::Eq> Cond<'_, T> {
     fn matches(&self, compare: T) -> bool {
         match self {
-            Cond::Only(vals) => vals.iter().any(|val| *val == compare),
-            Cond::Not(vals) => vals.iter().all(|val| *val != compare),
+            Cond::Only(vals) => vals.contains(&compare),
+            Cond::Not(vals) => !vals.contains(&compare),
```
stevenengler, line 16: This `Not` case was not actually part of the clippy warning, but I changed it for consistency.

sporksmith (approved):

PR #3593, opened 2025-05-19
Finish migration of signal syscall handlers to rust

stevenengler:
There's no rush on this, I just did this while procrastinating on other things.

This also removes the `Debug` trait bound from the memory manager, since I don't see why we need it, and it was preventing us from using it with `sigaction`.

@sporksmith It could probably use an extra close look to make sure everything was translated properly, but I think it's alright and the tests pass.


src/main/host/memory_manager/memory_copier.rs
sporksmith, line 1: I suspect the Debug bound was to support debug-printing values inside the implementation of the memory manager. But yeah, I'm fine with removing it.

src/main/host/syscall/handler/signal.rs
```diff
@@ -371,11 +370,89 @@ impl SyscallHandler {
...
+                    new_ss.flags_retain(),
+                );
+                // Unrecognized flag.
+                return Err(Errno::EINVAL);
+            }
+
+            *unsafe { thread_protected.sigaltstack_mut() } = new_ss;
+        }
+
+        // TODO: should we move this before we modify the alt stack, so that we don't return EFAULT
```
sporksmith: Makes sense to me; ideally with a test for this behavior and to confirm consistency with linux behavior, but not strictly necessary.

sporksmith (approved):
Awesome!

src/main/host/memory_manager/memory_copier.rs
stevenengler: Another option is to add a empty `Debug` impl for `sigaction`, and leave the `Debug` bounds on the memory manager. Do you have a preference here?

sporksmith: I think removing the bound from the memory manager makes sense, but no strong preference either way

src/main/host/syscall/handler/signal.rs
stevenengler: Thanks for bringing up adding a test. It turns out that my TODO (and following change) was wrong, so I've reverted the change that resolves this TODO. (GitHub doesn't show the original change anymore for some reason.)

I will add a test for the correct behaviour in a follow up MR.

PR #3609, opened 2025-06-24
More strace logging

sporksmith:


src/main/host/syscall/type_formatting.rs
```diff
@@ -278,15 +278,11 @@ fn fmt_string(
...
     len: Option<usize>,
-    options: FmtOptions,
+    _options: FmtOptions,
     mem: &MemoryManager,
 ) -> std::fmt::Result {
     const DISPLAY_LEN: usize = 40;
 
-    if options == FmtOptions::Deterministic {
-        return write!(f, "<pointer>");
-    }
```
robgjansen, line 288: This change seems wrong to me. Shouldn't it be up to the caller to decide if the string might be deterministic or not, and if it is non-deterministic then it should set `FmtOptions::Deterministic`?

I could imagine cases where a string _could_ be non-deterministic, e.g., a file path in the managed process from the `/proc` filesystem would maybe have different PIDs on multiple runs.

It seems like getting rid of this code, but just for formatting strings, introduces an inconsistency that might introduce future problems. What am I missing?

robgjansen (approved):

src/main/host/syscall/type_formatting.rs
sporksmith: This is only called for formatting `SyscallStringArg`. I'll push this change up to the caller though to make it easier to later reuse for another type, like `NondeterministicSyscallStringArg` or `Nondeterministic<SyscallStringArg>`

sporksmith: Actually this turned out to be a little awkward, since `option` represents what the global formatting options are, not whether the input string is deterministic. So to keep this code and push it up to the caller in a clear way I think we'd need to keep `options` but also add something like `string_is_deterministic: bool`.

That seems like a lot of unnecessary faff for now since we'd currently always be passing `string_is_deterministic=true`, until if/when we added a different syscall string type.

I pushed a fixup to remove the `options` parameter instead of taking one and ignoring it, to at least make it a little clearer what's happening.

Issue #3541, opened 2025-03-26
Opening /proc/self/* and /proc/<pid>/* opens the wrong files (especially /proc/self/maps)


sporksmith:
Noticed while working on #2693 - the golang runtime opens and reads `/proc/self/maps`, but because we open the file from shadow and don't otherwise remap the path name, this opens *shadow's* `maps` file. In addition to possibly breaking whatever the runtime is trying to do with those maps, the result is non-deterministic since ASLR isn't disabled for the shadow process itself (and changes dynamically).

I think in our file path handling we want to replace `/proc/self/*` (and `/proc/<virtual-pid>/*`) with `/proc/<native-pid>/*`

stevenengler:
Ah, nice catch! I wonder if rust programs that open `/proc/self/maps` in the std lib runtime initialization do this before or after the shim is initialized?

> I think in our file path handling we want to replace `/proc/self/*` (and `/proc/<virtual-pid>/*`) with `/proc/<native-pid>/*`

I think this can work in some cases, but not generally. For example all of the entries in `/proc/self/fd/*`, `/proc/self/fdinfo/*`, `/proc/self/task/*`, or `/proc/self/net/*` will be wrong. I think `/proc/self/cwd` will also be wrong since we only change the virtual cwd?

We could maybe add some small number of edge cases into the existing C code to fix a few cases, but if we wanted to fix this in a more comprehensive way, I think we'd need some way to specify various rules about how to redirect file/directory paths, or to open virtual files/directories for some paths. Which effectively is a virtual overlay filesystem. I think this gets into:

- https://github.com/shadow/shadow/discussions/2734
- https://github.com/shadow/shadow/issues/1772

This is also listed as "Virtual filesystem (to enforce determinism on FS operations)" in https://github.com/shadow/shadow/discussions/2995.

sporksmith:
> Ah, nice catch! I wonder if rust programs that open /proc/self/maps in the std lib runtime initialization (like arti) do this before or after the shim is initialized?

I haven't seen other programs opening `/proc/self/maps`; just the golang runtime.

> I think this can work in some cases, but not generally for all of `/proc/self`

Yeah, I'm planning to try a hacky cludge for `/proc/self/maps` to see if it actually improves the situation for the golang. If it doesn't then it may not be worth prioritizing at all.

[...]

PR #3613, opened 2025-06-25
Add some /proc handling

### LINKED ISSUE #3536: Warn when plugin opens a file in `/proc`
I think it might be useful to log a warning if a plugin opens a file in `/proc`, since most files (all?) will not be intercepted by Shadow. These files have a high likelihood of leading to non-determinism, or at least won't be correct inside the simulation. For example `/proc/sys/net/ipv4/tcp_rmem`, `/proc/meminfo`, `/proc/self` ,`/proc/uptime`, etc.

There are two issues though:

1. This might be too noisy. We might need to evaluate some common programs to make sure they don't access too many things in `/proc` and flood the logs.
2. The `open()` syscall handler (and related syscall handlers) are in C. This means that we can't use things like `log_once_per_value_at_level` to limit the number of log messages.

sporksmith:
Remap /proc/self/* paths to point to the correct native path. Notably this fixes behavior for reading `/proc/self/maps`, which the rust runtime does, and for reading `/proc/self/exe` which asan does.

Log a warning when opening proc files that we haven't specifically tested, thought about, or otherwise vetted.

Progress on #3541
Fixes #3536
Progress on #2790

### DIFF
--- a/src/main/host/descriptor/regular_file.c
+++ b/src/main/host/descriptor/regular_file.c
@@ -7,6 +7,7 @@
 
 #include <errno.h>
 #include <fcntl.h>
+#include <linux/limits.h>
 #include <poll.h>
 #include <stdbool.h>
 #include <stdio.h>
@@ -290,6 +291,8 @@ int regularfile_openat(RegularFile* file, RegularFile* dir, const char* pathname
      * an absolute path to compare for special files. */
     char* abspath = _regularfile_getAbsolutePath(dir, pathname, workingDir);
 
+    const char* proc_prefix = "/proc/";
+
     /* Handle special files. */
     if (utility_isRandomPath(abspath)) {
         file->type = FILE_TYPE_RANDOM;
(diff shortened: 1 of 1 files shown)


sporksmith:
I had tried adding a determinism test that checksums the contents of `/proc/self/maps`, but it doesn't quite pass due to the shared memory regions that the shim loads getting different names and inode numbers on each run. e.g.:

```diff
< 7ffff71df000-7ffff79df000 rw-s 00000000 00:1e 126629                     /dev/shm/shadow_shmemfile_701540.832816749-1656503
---
> 7ffff71df000-7ffff79df000 rw-s 00000000 00:1e 126630                     /dev/shm/shadow_shmemfile_701540.931295999-1656567
```

It might be nice to still write a test that "checksums all of the contents of /proc/maps except for shadow_shmemfile lines" or better yet add a more specialized `/proc/maps` handler in shadow that filters out these lines altogether, but for now I'd like to just get this 90% solution that I've been sitting on merged :)

robgjansen (approved):
Thanks looks great!

Issue #3610, opened 2025-06-24
Nondeterminism in chutney+tor+arti


sporksmith:
chutney tests involving arti hidden services appear to be non-deterministic under shadow. I haven't narrowed down yet whether the non-determinism extends to other chutney networks (e.g. not involving arti or not involving hidden services). It's more obvious in the (arti) hidden service case since those transfers are flaky, resulting in corresponding tests run under shadow also being flaky. (instead of always passing or always failing with a given simulation seed)

Locally running a slightly modified version of arti's "shadow-test-networks" CI job, extracted to this script:

```
set -xeuo pipefail

for flavor in \
      hs-v3-arti \
      ; do
  shadow-exec \
    --preserve=always \
    --shadow-args="--model-unblocked-syscall-latency=true --max-unapplied-cpu-latency=10ms --strace-logging-mode=deterministic" \
    -- \
  bash -c \
  "
    set -x
    set -euo pipefail
    export CHUTNEY_ARTI=\"$CHUTNEY_ARTI\"
    # Too noisy
    export CHUTNEY_DEBUG=0
    # Tell the "verify" test to connect to hidden services with *every*
    # client. We want this to e.g. validate cross-compatibility between tor
    # and arti clients and servers.
    export CHUTNEY_HS_MULTI_CLIENT=1
    export PATH=\"$PATH\"
    tools/test-network.sh \
    --chutney-path \"$CHUTNEY_PATH\" \
    --net-dir \"$CHUTNEY_DATA_DIR\" \
    --sandbox 0 \
    --no-controlsocket \
    --no-ipv6 \
    --tor \"$(which tor)\" \
    --tor-gencert \"$(which tor-gencert)\" \
    --stop-time 0 \
    --flavor \"$flavor\" || \"$CHUTNEY_PATH\"/chutney stop \"$flavor\""
done
```

I eliminated some sources of nondeterminism by setting the output directory (`CHUTNEY_DATA_DIR`) to the same fresh empty directory between runs (and moving previous results out of the way).

I still get substantial diffs using:
`git diff --no-index --text --word-diff -- ~/tmp/net[01]/shadow-exec-*/shadow.data/hosts/host/bash.1000.strace`

Going roughly in order:

There are a few of these `sched_getaffinity` calls using different PIDs (in `arti`). I think this is the native PID that the dynamic linker gets hold of before we get control. The result is the same in both cases `ESRCH`, so it doesn't seem likely to itself cause further divergence.

```diff
[...]

PR #3614, opened 2025-06-26
DescriptorTable store descriptors in deterministic order

sporksmith:
Previously we were storing them in the `HashMap`'s arbitrary order, and them in that arbitrary order `DescriptorTable::remove_all` and `DescriptorTable::remove_range`. Callers of both then close the descriptors and push callbacks in that arbitrary order..

Changing the data structure to `BTreeMap` ensures we process the descriptors in a deterministic (sorted) order. Alternatively we could just sort the descriptors returned by `DescriptorTable::remove_range` and `DescriptorTable::remov_all`, but using a sorted data structure in the first place will make it harder for future subtle bugs of this nature to sneak back in.

This seems to fix some of the nondeterministic scheduling (presumably due to callbacks being called in different order) and nondeterministic file descriptor assignments observed in #3610. Unfortunately not all of it though, and it's difficult to be sure whether it's actually making a difference due to the, err, nondeterministic nondeterminism.

### DIFF
--- a/src/main/host/descriptor/descriptor_table.rs
+++ b/src/main/host/descriptor/descriptor_table.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeSet, HashMap};
+use std::collections::{BTreeMap, BTreeSet};
 
 use log::*;
 use shadow_shim_helper_rs::explicit_drop::ExplicitDrop;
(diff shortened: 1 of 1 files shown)


sporksmith:
This also affects `DescriptorTable::remove_all` and its callers; updated description.

sporksmith:
> Maybe we should search for other internal uses of HashMap, since it seems to commonly cause determinism problems for us.

Yeah, I did that locally to some other ones that looked suspicious, but it didn't seem to help the particular determinism issues I'm running into. Maybe I still ought to clean up and merge them though to eliminate  them from future scrutiny/trouble.

PR #3617, opened 2025-06-26
Handle /proc/sys/kernel/random/uuid

### LINKED ISSUE #3188: Intercept /proc/sys/kernel/random/uuid
From `proc(2)`:

```
       /proc/sys/kernel/random/uuid (since Linux 2.4)
              Each read from this read-only file returns a randomly generated 128-bit UUID, as a string in the standard UUID format.
```

I noticed the tor unit tests read this file; not sure whether tor reads it in production.

sporksmith:
Fixes #3188

### DIFF
--- a/src/main/host/descriptor/regular_file.c
+++ b/src/main/host/descriptor/regular_file.c
@@ -29,10 +29,21 @@
 #include "main/host/descriptor/descriptor.h"
 #include "main/host/syscall/kernel_types.h"
 #include "main/utility/utility.h"
+#include "regular_file.h"
 
 #define OSFILE_INVALID -1
 
 const int SHADOW_FLAG_MASK = O_CLOEXEC;
+
+/* Callback to (re)-generate contents of a `FILE_TYPE_IN_MEMORY` file.
+ *
+ * Should set `contents` to a malloc'd pointer containing the new contents, and
+ * `contentsLen` to its length.
+ *
+ * Future uses might need more parameters or other flexibility, but let's not
+ * add complexity we don't need yet.
+ */
+typedef void (*_generateInMemoryFileContentCbType)(char** contents, size_t* contentLen);
 
 struct _RegularFile {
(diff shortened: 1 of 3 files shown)


src/main/host/descriptor/regular_file.c
```diff
@@ -60,7 +72,15 @@ struct _RegularFile {
             int flagsAtOpen;
             /* The permission mode the file was opened with. */
             mode_t modeAtOpen;
-            /* The path of the file when it was opened. */
+            /* Callback to (re)-generate contents. */
+            _generateInMemoryFileContentCbType generate_contents_cb;
+            /* Whether contents should be re-generated on an lseek operation.
+             * Typical for some? all? proc files)
```
robgjansen, line 78: Unmatched closing `)`

src/main/host/descriptor/regular_file.c
```diff
@@ -60,7 +72,15 @@ struct _RegularFile {
...
+             *
+             * TODO: Actually implement lseek for inMemoryFile. Currently
+             * just returns an error.
+             */
+            bool regen_after_lseek;
```
robgjansen, line 83: This variable is currently not read. Just wanted to flag in case you think it's better to add this bit alongside an lseek implementation instead of now.

I'm OK with keeping it in now too, because you've already worked it out and so we wouldn't have to remember how to track it later.

robgjansen (approved):

src/main/host/descriptor/regular_file.c
sporksmith: Yeah I had originally planned to implement the regeneration, before realizing that lseek isn't implemented at all for these files. It would be terribly difficult, but doing it + adding the appropriate tests was a little more scope creep than I'd bargained for. But yeah, I figured might as well still leave in the bits that will make implementing the regeneration trivial once lseek is implemented.

PR #3628, opened 2025-07-13
Migrate futex syscall handlers to rust

stevenengler:
The rust code should be pretty similar to the C code. The actual `Futex` object isn't migrated yet.


src/main/host/syscall/handler/futex.rs
```diff
@@ -19,14 +24,195 @@ impl SyscallHandler {
...
+        uaddr: ForeignPtr<u32>,
+        op: std::ffi::c_int,
+        val: u32,
+        utime: ForeignPtr<linux_api::time::kernel_timespec>,
         _uaddr2: ForeignPtr<u32>,
-        _val3: u32,
+        val3: u32,
     ) -> Result<std::ffi::c_int, SyscallError> {
-        Self::legacy_syscall(c::syscallhandler_futex, ctx)
+        // TODO: currently only supports uaddr from the same virtual address space (i.e., threads)
```
sporksmith, line 34: nit - s/threads/process/. (I see this was wrong in the source comment too)

sporksmith (approved):
Nice!

PR #3632, opened 2025-07-20
Fix warnings for rust 1.88 and upgrade CI rust version

stevenengler:
Fixes clippy warnings and errors in rust 1.88, and upgrades the rust version in CI.


stevenengler:
> I'm not sure about [1f98f17](https://github.com/shadow/shadow/commit/1f98f176b45e61d8d344e08eb706b7d5974ad936), so if you want feedback about that specific change you might want to ask @sporksmith.

I removed that commit from this MR and replaced it with a new commit that does just an `#[allow(...)]`. I've moved the original commit to a new MR: shadow#3636.

PR #3636, opened 2025-07-24
Change `MemoryMapper::get_mut` to take a `&mut self`

stevenengler:
This clippy error was added in rust 1.88. In shadow#3632 I added an `#[allow(...)]`, but I think it would be better using a `&mut self` for now.

See https://github.com/shadow/shadow/pull/3632#issuecomment-3111528848 for context.

```text
error: mutable borrow from immutable input(s)
    --> main/host/memory_manager/memory_mapper.rs:1079:77
     |
1079 |     pub unsafe fn get_mut<T: Pod>(&self, src: ForeignArrayPtr<T>) -> Option<&mut [T]> {
     |                                                                             ^^^^^^^^
     |
note: immutable borrow here
    --> main/host/memory_manager/memory_mapper.rs:1079:35
     |
1079 |     pub unsafe fn get_mut<T: Pod>(&self, src: ForeignArrayPtr<T>) -> Option<&mut [T]> {
     |                                   ^^^^^
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#mut_from_ref
     = note: `#[deny(clippy::mut_from_ref)]` on by default
```

-------

We previously used the mutable reference in a higher-layer type (`MemoryManager::mapped_mut`) to ensure there would be only one mutable reference to plugin memory. This PR changes `MemoryMapper::get_mut` to also take a mutable reference to ensure we don't accidentally introduce a bug where we allow multiple mutable references.

### DIFF
--- a/src/main/host/memory_manager/memory_mapper.rs
+++ b/src/main/host/memory_manager/memory_mapper.rs
@@ -1072,8 +1072,11 @@ impl MemoryMapper {
         Some(unsafe { std::slice::from_raw_parts(notnull_debug(ptr), src.len()) })
     }
 
-    #[allow(clippy::mut_from_ref)]
-    pub unsafe fn get_mut<T: Pod>(&self, src: ForeignArrayPtr<T>) -> Option<&mut [T]> {
+    // This takes `&mut self` even though it doesn't *need* a mutable reference. But it helps to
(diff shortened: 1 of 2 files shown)


sporksmith (approved):
Yup, makes sense

PR #3637, opened 2025-07-24
Upgrade schemars to 1.0.4

stevenengler:
Closes shadow#3625.

There have been some changes to schemars here. The "cli" tests make sure that there are no changes to the final cli help rendering.

One slightly painful change is that schemars has changed how they format the description for a field. So if a field is:

```rust
/// Hello
/// world.
foo: String,
```

It will now have a description of "Hello\nworld." instead of "Hello world" like in the previous schemars version. This causes clap to also format it as two lines. So we need to perform some rudimentary formatting to replace the new lines with spaces. This isn't ideal, but is close to what schemars was doing originally. If we find that this is too limiting, we can add some markdown parser into shadow in the future.

For my own reference, this is where clap-derive does it's parsing/formatting: https://github.com/clap-rs/clap/blob/4c039309b614ee4523c67a243afc38af11860de9/clap_derive/src/utils/doc_comments.rs#L9

### DIFF
--- a/src/Cargo.lock
+++ b/src/Cargo.lock
@@ -1876,6 +1876,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "ref-cast"
+version = "1.0.24"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


robgjansen (approved):
Thanks for fixing this!

PR #3639, opened 2025-07-26
Upgrade neli to 0.7.1

stevenengler:
Closes shadow#3635.

Part of shadow#3555.


robgjansen (approved):
Oof, hopefully neli stabilizes so new updates don't require quite as much tweaking on our end.

PR #3643, opened 2025-08-19
TCP: Un-parent child sockets once they've been accepted

### LINKED ISSUE #3563: Parent TCP sockets aren't disassociated from the network interface until all child sockets are closed
In Shadow's C TCP code, a parent socket is responsible for forwarding packets to its children. For example, if a listening socket `accept()`s a new child socket, any incoming packets for the child socket are received by the parent listening socket, and then forwarded to the child socket.

https://github.com/shadow/shadow/blob/aa0445f42d4b5413b27b27d267d0a7f3ec29bed1/src/main/host/descriptor/tcp.c#L2063-L2072

This is problematic when the listening socket is closed before all of its children, or when the child sockets are `close()`d but need to spend some time in one of the closing states like LAST_ACK before being completely closed. When this happens, the listening socket cannot be disassociated from the network interface, which means that even if the listening socket is closed, another socket cannot bind to that same IP/port.

https://github.com/shadow/shadow/blob/aa0445f42d4b5413b27b27d267d0a7f3ec29bed1/src/main/host/descriptor/tcp.c#L709-L737

Related: #2590, #1781

sporksmith:
Once a child socket has been `accept`ed (associated with a file descriptor and moved off of the listening socket's queue), we now remove the parent-child relationship and associate the local:remote address pair directly with the corresponding network interface, instead of continuing to route packets to the listening socket and having the listening socket route them to the child.

This allows a listening socket to be unbound and cleaned up when it's closed instead of keeping it around to route packets to the child sockets, which allows us to reuse the address it had been bound to.

Fixes #3563.

I noticed and filed #3644 while working on this but verified that it was present before this PR as well.

### DIFF
--- a/src/main/host/descriptor/socket/inet/legacy_tcp.rs
+++ b/src/main/host/descriptor/socket/inet/legacy_tcp.rs
@@ -883,8 +883,8 @@ impl LegacyTcpSocket {
 
     pub fn accept(
         &mut self,
-        _net_ns: &NetworkNamespace,
-        _rng: impl rand::Rng,
+        net_ns: &NetworkNamespace,
+        rng: impl rand::Rng,
         _cb_queue: &mut CallbackQueue,
     ) -> Result<OpenFile, SyscallError> {
         let is_valid_listener = unsafe { c::tcp_isValidListener(self.as_legacy_tcp()) } == 1;
@@ -946,8 +946,12 @@ impl LegacyTcpSocket {
             );
         };
 
-        // sanity check: make sure new socket peer address matches address returned from
-        // tcp_acceptServerPeer() above
+        // Associate the new socket with the local:peer address pair. In
+        // previous versions of shadow, the new socket was never registered this
+        // way. Instead the packets continued to be routed to the parent
+        // listening-socket, which was responsible for routing them to this new
+        // "child" socket.  But that led to bugs such as
+        // https://github.com/shadow/shadow/issues/3563.
         {
             let File::Socket(Socket::Inet(InetSocket::LegacyTcp(new_socket))) =
                 open_file.inner_file()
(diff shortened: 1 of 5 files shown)


sporksmith:
I think we might lose data if the client sends data and then closes the connection before the server accepts the connection. This may or may not be the case before this PR.

TODO: figure out whether this happens and whether it happened before; fix it or file/update an issue.

sporksmith:
> I think we might lose data if the client sends data and then closes the connection before the server accepts the connection. This may or may not be the case before this PR.
> 
> TODO: figure out whether this happens and whether it happened before; fix it or file/update an issue.

Hmm, maybe this is why the tgen tests are failing

sporksmith:
> I think we might lose data if the client sends data and then closes the connection before the server accepts the connection. This may or may not be the case before this PR.

> TODO: figure out whether this happens and whether it happened before; fix it or file/update an issue.

Added a test for this. It's handled  correctly before and after.

> Hmm, maybe this is why the tgen tests are failing

It was something else - destroying the TCP server state while while references to the child sockets can still exist elsewhere and try to access it. Fixed.

src/main/host/descriptor/socket/inet/legacy_tcp.rs
```diff
@@ -946,28 +946,66 @@ impl LegacyTcpSocket {
...
+                InetSocket::LegacyTcp(Arc::clone(new_socket)),
+                local_addr,
+                peer_addr,
+                /* Allow the parent/listening socket to be bound to the same address,
+                 * with a missing/generic peer. */
+                /* check_generic_peer= */
+                false,
+                net_ns,
+                rng,
+            )?;
```
robgjansen, line 1004: Something is unclear for me here. Is the local address that is associated with the new socket here using the same ip/port as the listening socket? I think before it was, and I think it still should be, but now I can't tell if we've changed that behavior from the diff.

The reason we had the awkward listening parent / child relationship was to make sure the server port does not change after the connection is accepted. So if a client connects to a server at `1.1.1.1:443`, then from the client's perspective all packets coming from the server on that connection should come from `1.1.1.1:443`, even if the server uses a new (child) socket to handle it after calling `accept()`. I just want to make sure this diff is not changing that, e.g., changing the port `443` to some other port.

robgjansen (commented):
This looks reasonable to me other than my question about port numbers, but I think we might want @stevenengler to take a look too because of the importance of TCP correctness for us.

src/main/host/descriptor/socket/inet/legacy_tcp.rs
sporksmith: Added some assertions to `LegacyTcpSocket::accept` to validate expectations about the local address of the child socket.

`_tcpchild_new` was already setting the child local address to be that of the parent, *but* that included setting the address to unspecified if the listening socket address was unspecified (i.e. bound to all interfaces).

I'm looking at adding a test for this but it doesn't fit neatly into `test_accept.rs` since the bind helper functions always bind to `INADDR_LOOPBACK`. It probably makes sense to extend them to also test ~everything when binding to `INADDR_ANY` instead of `INADDR_LOOPBACK`. I'll poke at that a bit more, but I might file an issue instead of climbing down that rabbit hole right now.

sporksmith: > but that included setting the address to unspecified if the listening socket address was unspecified (i.e. bound to all interfaces).

Forgot to say I fixed this in the just-pushed squash commit. The child socket now gets bound to the address on which it received the connection

robgjansen: Yeah that seems right to me (it should keep the 4-tuple consistent). Thanks!

src/main/host/descriptor/tcp.c
```diff
@@ -305,6 +305,10 @@ static TCPChild* _tcpchild_new(TCP* tcp, TCP* parent, int handle, in_addr_t peer
     in_port_t parentPort;
     legacysocket_getSocketName(&(parent->super), &parentAddress, &parentPort);
     legacysocket_setSocketName(&(tcp->super), parentAddress, parentPort);
+    /* to handle the case that the parent address is unspecified (i.e. bound to
+     * all interfaces), the child socket should instead be bound to the address
+     * on which it received the connection. */
+    tcp->super.boundAddress = parent->server->lastIP;
```
stevenengler, line 311: I think this

src/test/socket/bind/test_bind.rs
```diff
@@ -177,9 +178,112 @@ fn get_tests() -> Vec<test_utils::ShadowTest<(), String>> {
...
+    let domain = libc::AF_INET;
+    let sock_type = libc::SOCK_STREAM;
+    let socket_flag = 0;
+    let listen_fd = unsafe { libc::socket(domain, sock_type | socket_flag, 0) };
+    assert_with_errno!(listen_fd >= 0);
+
+    // We set SO_REUSEADDR to avoid having to wait for TIME_WAIT before allowing
+    // the address to be reused. This currently isn't strictly needed under
+    // shadow, since the `close` syscall immediately fully closes the listening
+    // socket without waiting TIME_WAIT (possibly a bug?).
```
stevenengler, line 214: I think that the listening socket on Linux doesn't ever enter the TIME_WAIT state. This doesn't follow the wording of RFC 9293, but \*shrug emoji\*. (See https://github.com/shadow/shadow/blob/main/src/lib/tcp/notes/listener-close-with-pending-conn.md)

My understanding on Linux is that what can happen is that any previously accepted sockets can go to TIME_WAIT, which prevents new listening sockets from binding to the same address if the socket wasn't originally bound with `SO_REUSEADDR`. So it's not the (closed) listening socket itself that prevents new sockets from reusing the same address, but previously accepted sockets that are now in the TIME_WAIT state.

PR #3653, opened 2025-09-07
Fix rust/clippy warnings in rust 1.89.0

stevenengler:
Another rust release, another painful collection of warnings to fix. Might be easier to review with `git diff -w` (edit: it looks like this is available in the github view as well).

This also upgrades the bindgen version used in linux-api bindings.

**Edit:** This was originally an MR to fix the warnings and update our CI to use rust 1.89.0. But we can't update to 1.89.0 due to #3654. So this MR now just fixes the warnings without updating the CI.


stevenengler:
~This is blocked on #3654.~

I changed this MR to not update the rust version in the CI, so we're no longer blocked.

Issue #3544, opened 2025-03-30
`clone_leader_exits_early-linux` segfaults


stevenengler:
I noticed this in a debian sid container. Since debian sid is experimental, it's possible this is a bug in one of its packages rather than shadow. But there might also be a shadow issue here.

**Edit:** This segfaults on newer stable distros too like Fedora 42.

I haven't looked into this issue at all, but posting some output below.

```text
$ ./setup test clone_leader_exits_early-linux
2025-03-30 22:02:21,409 [INFO] calling ['ctest', '-j12', '--timeout', '20', '-R', 'clone_leader_exits_early-linux']
Test project /tmp/shadow/build
    Start 13: clone_leader_exits_early-linux
1/1 Test #13: clone_leader_exits_early-linux ...***Exception: SegFault  0.10 sec

0% tests passed, 1 tests failed out of 1

Label Time Summary:
linux    =   0.10 sec*proc (1 test)

Total Test time (real) =   0.11 sec

The following tests FAILED:
	13 - clone_leader_exits_early-linux (SEGFAULT)         linux
Errors while running CTest
Output from these tests are in: /tmp/shadow/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
2025-03-30 22:02:21,538 [INFO] ctest returned 8
```

```text
$ ./build/src/test/clone/test_clone_leader_exits_early
TAP version 14
# random seed: R02Sc7ffc3f1c71ac349b5ced31b4e17c410
1..1
# Start of clone tests
ok 1 /clone/clone_child_exits_after_leader
# End of clone tests
Segmentation fault (core dumped)
```

```text
(gdb) run
Starting program: /tmp/shadow/build/src/test/clone/test_clone_leader_exits_early 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
TAP version 14
# random seed: R02Sa0a5a062b0100f100332559512760f42
1..1
# Start of clone tests
[New Thread 0x5555555580d0 (LWP 31199)]

Thread 2 "test_clone_lead" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x5555555580d0 (LWP 31199)]
[...]

PR #3663, opened 2025-09-22
Remove Fedora 40, add 42, and add Debian 13

stevenengler:
This updates our supported versions of Fedora and Debian.

This also makes a few small changes for things that were needed for Debian 10, which was dropped a while ago. There's also a comment about using `posix_spawn_file_actions_addchdir_np` when we don't need to support Debian 10, but I didn't try to change this.

I also updated the tor version (was just a minor version bump), and disabled the `clone_leader_exits_early-linux` test which segfaults on newer distros (#3544).


stevenengler:
On Debian 13 we get:

```text
FAIL: test_bash_script_sleep (tests.test_shadow_exec.TestShadowExecCLI.test_bash_script_sleep)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/__w/shadow/shadow/shadowtools/tests/test_shadow_exec.py", line 65, in test_bash_script_sleep
    self.assertEqual(
    ~~~~~~~~~~~~~~~~^
        res,
        ^^^^
        "2000-01-01T00:00:00,000000000+00:00\n2000-01-01T00:00:01,001000000+00:00\n",
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
AssertionError: '2000-01-01T00:00:00,000000000+00:00\n2000-01-01T00:00:01,001000001+00:00\n' != '2000-01-01T00:00:00,000000000+00:00\n2000-01-01T00:00:01,001000000+00:00\n'
  2000-01-01T00:00:00,000000000+00:00
- 2000-01-01T00:00:01,001000001+00:00
?                             ^
+ 2000-01-01T00:00:01,001000000+00:00
?                             ^
```

It looks like it advances 1 ns extra on Debian 13. We disable `model_unblocked_syscall_latency`, so not sure what might be causing this. Maybe the bash version in Debian 13 does something slightly different that takes a 1 ns delay, or maybe there's some floating point conversion difference with "1.001".

sporksmith:
> It looks like it advances 1 ns extra on Debian 13.

Weird. Maybe change the test to get the time back in an easier to parse format (e.g. unix epoch time: `date +%s.%N`)?. That might make the problem go away already if it's a formatting issue. If not then we could parse what we get back and add some tolerance. Not ideal, but probably not worth chasing down a 1ns conversion discrepancy somewhere in the chain here. (LMK if you'd rather I take a stab at this)

stevenengler:
Fedora 42:

```text
$ strace sleep 1.001 
[...]
clock_nanosleep(CLOCK_REALTIME, 0, {tv_sec=1, tv_nsec=1000000}, 0x7ffcb3f95a90) = 0
$ strace sleep 1
[...]
clock_nanosleep(CLOCK_REALTIME, 0, {tv_sec=1, tv_nsec=0}, 0x7ffcb9cd0df0) = 0
$ sleep --version
sleep (GNU coreutils) 9.6
```

Debian 13:

```text
$ strace sleep 1.001 
[...]
clock_nanosleep(CLOCK_REALTIME, 0, {tv_sec=1, tv_nsec=1000001}, 0x7ffc4e3a4840) = 0
$ strace sleep 1
[...]
clock_nanosleep(CLOCK_REALTIME, 0, {tv_sec=1, tv_nsec=1}, 0x7fff6e4d85d0) = 0
$ sleep --version
sleep (GNU coreutils) 9.7
Packaged by Debian (9.7-3)
```

😞

This is probably due to https://github.com/coreutils/coreutils/commit/cb7c210d30c5a6a4a879bda96d61926d77f16d96, which is the only change recent enough to seem like it would have this effect.

We can instead use `coproc read -t 1.001; wait` instead of `sleep 1.001` to get a more reliable sleep.

stevenengler:
@robgjansen If/once this is approved, you'll need to update the CI checks to add:

- tor (tor-0.4.8.18)
- shadow (fedora:42, gcc, release)
- shadow (debian:13-slim, gcc, release)

and remove:

- tor (tor-0.4.8.13)
- shadow (fedora:40, gcc, release)

shadowtools/tests/test_shadow_exec.py
```diff
@@ -58,7 +58,10 @@ def test_bash_script_sleep(self) -> None:
                 "--",
                 "bash",
                 "-c",
-                "date -Ins; sleep 1.001; date -Ins",
+                # we would use 'sleep' here, but /bin/sleep in newer coreutils seems to add an extra
+                # nanosecond to the specified time:
+                # https://github.com/shadow/shadow/pull/3663#issuecomment-3341062224
+                "date -Ins; coproc read -t 1.001; wait; date -Ins",
```
sporksmith, line 64: TIL about coproc; clever! It'd be nice if we could keep this less esoteric, but nothing comes to mind that wouldn't add a fair bit of complexity. I'm good with this solution.

sporksmith:
> This is probably due to https://github.com/coreutils/coreutils/commit/cb7c210d30c5a6a4a879bda96d61926d77f16d96, which is the only change recent enough to seem like it would have this effect.

Nice find!

robgjansen:
@stevenengler I have updated the CI checks as requested.

shadowtools/tests/test_shadow_exec.py
stevenengler: Yeah, I'm happy if someone changes it later, but I'm not sure of a nicer way.

PR #3666, opened 2025-09-25
Fix missing syscall clobbers

### LINKED ISSUE #3654: Shadow fails when built with rust 1.89.0 in release mode
If you build Shadow with rust 1.89.0 in release mode, most of the tests fail.

Since Shadow works fine with rust 1.89.0 in debug mode, it's probably an issue due to optimizations. I think it's likely that we have UB, or there's a miscompilation bug somewhere.

Shadow runs fine when built with rust 1.88.0 in both debug and release mode.

(@sporksmith I had previously mistakenly thought that it was a bug in newer distros, not newer rust versions.)

Example:

```yaml
general:
  stop_time: 10 seconds
network:
  graph:
    type: 1_gbit_switch
hosts:
  testnode:
    network_node_id: 0
    processes:
    - path: python3
      args: -u -c "print('foo')"
```

```text
Error: process 'testnode.python3.13.1000' exited with status Signaled(Signal(11)); expected end state was exited: 0 but was signaled: SIGSEGV
Error: Failed to run the simulation
Error: 
Error: Caused by:
Error:     1 managed processes in unexpected final state
```

sporksmith:
Fixes #3654

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,7 @@ https://gitlab.torproject.org/tpo/core/arti/-/issues/1972).
 * Fixed a bug where a listening TCP socket would be cleaned up only after all its child sockets were closed. (#3643)
 * Fixed a bug where files could have their `close()` delayed until the application next made a blocking syscall. (#3652)
 * Fixed incorrect handling of most file -at syscalls when called with an empty path string. (#3661)
+* Fixed missing register clobbers that resulted in breakage in rust 1.89. (#3666, fixing #3654)
 
 Full changelog since v3.2.0:
 


sporksmith:
Locally tested a release build using rust 1.89.0, first reproducing #3654 and then verifying that adding the missing clobbers in the shim's `native_syscall` function fixed the issue, and that it was still fixed after the other changes as well.

stevenengler (approved):
Nice find!! I can confirm that the tests pass for me when using this branch built with `RUSTUP_TOOLCHAIN=1.89 ./setup build --test`.

PR #3682, opened 2025-10-15
Emulate resource limits

### LINKED ISSUE #3681: sshd (openssh) fails due to native pass-through of setrlimit
I'm working on getting ssh working under shadow (https://github.com/shadow/shadow/discussions/3671). The current snag I've hit with opensshd is that sshd forks a process to handle the incoming connection, and then tries to set `RLIMIT_FSIZE` to 0. This results in the kernel immediately sending the process a `SIGXFSZ`, killing it. Presumably this is because of shadow's shim code writing to file(s) (maybe the strace log file).

strace:

```
...
00:00:10.003000000 [tid 1003] setrlimit(RLIMIT_FSIZE, linux_rlimit { rlim_cur: 0, rlim_max: 0 } (0x7fffffffe400)) = <native>
00:00:10.003000000 [tid 1002] poll(0x7fffffffe390, 2, -1) = 2
00:00:10.003000000 [tid 1002] read(9, 0x5555556ee770, 4) = 0
00:00:10.003000000 [tid 1002] getpid() = 1002
00:00:10.003000000 [tid 1002] write(6, "debug1: monitor_read_log: child log fd c"..., 47) = 47
00:00:10.003000000 [tid 1002] close(9) = 0
00:00:10.003000000 [tid 1002] poll(0x7fffffffe390, 1, -1) = 1
00:00:10.003000000 [tid 1002] getpid() = 1002
00:00:10.003000000 [tid 1002] write(6, "debug3: mm_request_receive: entering\r\n", 38) = 38
00:00:10.003000000 [tid 1002] read(8, 0x7fffffffe334, 4) = 0
00:00:10.003000000 [tid 1002] getpid() = 1002
00:00:10.003000000 [tid 1002] write(6, "debug1: do_cleanup\r\n", 20) = 20
00:00:10.003000000 [tid 1002] getpid() = 1002
00:00:10.003000000 [tid 1002] write(6, "debug1: Killing privsep child 1003\r\n", 36) = 36
00:00:10.003000000 [tid 1002] kill(1003, 9) = 0
00:00:10.003000000 [tid 1002] getpid() = 1002
00:00:10.003000000 [tid 1002] write(6, "debug1: audit_event: unhandled event 12"..., 41) = 41
00:00:10.003000000 [tid 1002] exit_group(255) = <native>
00:00:10.003000000 [tid 1001] ppoll(0x5555556eb570, 101, 0x0 <null>, 0x7fffffffe650, 8) = 1
00:00:10.003000000 [tid 1001] rt_sigprocmask(2, 0x7fffffffe650, 0x0, 8) = 0
00:00:10.003000000 [tid 1001] wait4(-1, 65280 (0x7ffff7fa0c04), 1, 0x0) = 1002
00:00:10.003000000 [tid 1001] wait4(-1, 65280 (0x7ffff7fa0c04), 1, 0x0) = -10 (ECHILD)
00:00:10.003000000 [tid 1001] read(6, 0x7fffffffe4f7, 1) = 0
00:00:10.003000000 [tid 1001] close(6) = 0
00:00:10.003000000 [tid 1001] rt_sigprocmask(0, 0x7fffffffe5d0, 0x7fffffffe650, 8) = 0
00:00:10.003000000 [tid 1001] ppoll(0x5555556eb570, 101, 0x0 <null>, 0x7fffffffe650, 8) = <blocked>
```

sshd log file:

```
...
Connection from 11.0.0.1 port 55625 on 11.0.0.2 port 22
debug1: Local version string SSH-2.0-OpenSSH_8.9p1 Ubuntu-3ubuntu0.13
debug1: Remote protocol version 2.0, remote software version OpenSSH_8.9p1 Ubuntu-3ubuntu0.13
debug1: compat_banner: match: OpenSSH_8.9p1 Ubuntu-3ubuntu0.13 pat OpenSSH* compat 0x04000000
debug2: fd 4 setting O_NONBLOCK
debug3: ssh_sandbox_init: preparing seccomp filter sandbox
debug2: Network child is on pid 1003
debug3: preauth child monitor started
debug1: monitor_read_log: child log fd closed
debug3: mm_request_receive: entering
debug1: do_cleanup
debug1: Killing privsep child 1003
debug1: audit_event: unhandled event 12
```

(I'm a little concerned about the "preparing seccomp filter sandbox" too, but I haven't bumped into an associated failure yet)

shadow output:
```
...
00:00:00.116029 [2840179:shadow-worker] 00:00:10.003000000 [DEBUG] [testnode:11.0.0.2] [process.rs:1260] [shadow_rs::host::pro
cess] Process testnode.sshd.1003 exited while running thread 1003
00:00:00.116029 [2840179:shadow-worker] 00:00:10.003000000 [DEBUG] [testnode:11.0.0.2] [process.rs:1472] [shadow_rs::host::pro
cess] process 'testnode.sshd.1003' has completed or is otherwise no longer running
00:00:00.116051 [2840179:shadow-worker] 00:00:10.003000000 [DEBUG] [testnode:11.0.0.2] [process.rs:1557] [shadow_rs::host::pro
cess] process 'testnode.sshd.1003' exited with status Signaled(Signal(25))
...
```

Investigating solutions, including:

* return failure from `setrlimit`. Easy to implement and provides self-consistent behavior (in case of `getrlimit` etc), but ssh or other software might not like the failure.
* return success for `setrlimit` without actually doing anything. would probably mostly work, but might cause some confusing misbehavior if the simulation includes intentionally reaching those limits and handling them somehow (via signal handler) or just using `getrlimit` and seeing something other than what was set.
* return success and store the set limit, returning it for `getrlimit`. This wouldn't be too much work or overhead, and would at least provide a bit more consistent behavior.
* full-on emulation - actually track resource usage, separating out the "native" usage from the shim's usage, and emulating the kernel's behavior when the limit is reached. This is almost certainly too much work to be worth it.
* some combination of above and the current behavior of passing-through-natively, on a resource-by-resource basis

sporksmith:
Primarily this is to prevent shadow from lowering the native resource limits, since the shim may require resources that the process alone doesn't. Fixes #3681

Also:

* Add tests for prlimit64, getrlimit, and setrlimit
* Add support for prlimit64 with non-zero PIDs (getting or setting limits for a process other than one's self)
* Update generate libc syscall wrappers, and stop wrapping getrlimit and setrlimit since glibc actually implements these using prlimit64.
* Expose more resource-related constants and functionality in linux_api

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@ MINOR changes (backwards-compatible):
 
 PATCH changes (bugfixes):
 
-*
+* Avoid lowering native resource limits of managed processes beyond what is needed
+for shadow's `LD_PRELOAD`d shim to function. (#3682 fixing #3681)
 
 Full changelog since v3.3.0:
 


stevenengler (commented):
I didn't finish reviewing yet, but leaving the few comments I have. Will review the rest on Wednesday.

src/lib/linux-api/src/resource.rs
```diff
@@ -37,3 +42,55 @@ pub enum Resource {
...
     RLIMIT_RTTIME = bindings::LINUX_RLIMIT_RTTIME,
 }
+
+/// Call the `prlimit64` syscall
+///
+/// # Safety
+///
+/// Causing limits to be exceeded by lowering them can result in various kernel
+/// behaviors (signalling the process, syscalls failing, ...) that may break
+/// other safety assumptions.
```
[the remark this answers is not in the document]
sporksmith: Hmm, yeah maybe making this `unsafe` is overly conservative. I guess I was thinking along the lines of "bad, surprising, potentially-difficult-to-debug things may happen", but:

* The limits that cause kernel to send signals when exceeded use signals that kill the process by default. This is potentially bad but not unsafe in the "undefined behavior" sense.
* There's potential unsafety around blocking or handling such signals, but the relevant signal-handler installation functions etc are already unsafe.
* Some syscalls may also start to fail when limits are exceeded, and programs might not handle that well, but in general it's up to the code that's making syscalls to handle errors in a way that doesn't compromise capital-s Safety.

For now I left it `unsafe` but updated the safety comments

src/lib/linux-api/src/resource.rs
```diff
@@ -20,6 +20,13 @@ unsafe impl Pod for rlimit {}
 pub type rlimit64 = crate::bindings::linux_rlimit64;
 unsafe impl Pod for rlimit64 {}
 
+// `include/linux/resource.h` defines this as (~0ULL); for some reason
+// bindgen is generating LINUX_RLIM64_INFINITY as an i32, though.
+pub const RLIM64_INFINITY: u64 = !0;
+// `include/asm-generic/resource.h` defines this as (~0UL); for some reason
+// bindgen is generating LINUX_RLIM_INFINITY as an i32, though.
+pub const RLIM_INFINITY: u64 = !0;
```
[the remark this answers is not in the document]
sporksmith: Good idea, done. I actually changed it to use `as` coercion from the source bindings (which [sign-extends](https://doc.rust-lang.org/reference/expressions/operator-expr.html#r-expr.as.numeric.int-extension)), and added  the assertions to validate that the results are as-expected.

src/test/resource/test_resource.rs
```diff
@@ -0,0 +1,377 @@
...
+            ),
+            Err(Errno::EINVAL)
+        );
+
+        // reducing rlim_max below rlim_cur
+        let cur = LibcGetRLimit
+            .getrlimit(linux_api::resource::Resource::RLIMIT_NOFILE.into())
+            .unwrap();
+        assert!(cur.rlim_cur > 0);
+        assert!(cur.rlim_cur < i64::MAX as u64);
```
stevenengler: Maybe skip if the current is >= `i64::MAX`? I wonder if any systems set this as unlimited?

src/test/resource/test_resource.rs
```diff
@@ -366,6 +366,24 @@ fn test_prlimit_child() {
...
+    let initial = LibcGetRLimit
+        .getrlimit(linux_api::resource::Resource::RLIMIT_FSIZE.into())
+        .unwrap();
+    // Set file-write limit to 0.
+    // This should be ok as long as we don't try writing to any files while this is set.
+    LibcSetRLimit.setrlimit(linux_api::resource::Resource::RLIMIT_FSIZE.into(), &linux_api::resource::rlimit{ rlim_cur: 0, rlim_max: initial.rlim_max }).unwrap();
+
+    // Do some syscall that should be handled shim-side, causing the shim
+    // to try writing to the strace log file.
+    linux_api::time::clock_gettime(linux_api::time::ClockId::CLOCK_REALTIME).unwrap();
```
stevenengler: nit: maybe enable strace logging explicitly in the yaml, with a comment linking here?

src/main/host/process.rs
```diff
@@ -674,6 +679,79 @@ impl RunnableProcess {
...
+        let idx = usize::try_from(u32::from(resource)).unwrap();
+        let cur = &self.common.rlimits[idx];
+        if let Some(old_rlim) = old_rlim {
+            *old_rlim = *cur;
+        }
+        let Some(new_rlim) = new_rlim else {
+            // Nothing else to do.
+            return Ok(());
+        };
+        if new_rlim.rlim_cur > new_rlim.rlim_max {
```
stevenengler: nit: I think `new_rlim.rlim_cur > std::cmp::min(new_rlim.rlim_max, cur.rlim_max)` would be nicer. I think this is checked transitively through the `new_rlim.rlim_max > cur.rlim_max` below, but I think it would be nice to have this explicitly.

src/main/host/process.rs
```diff
@@ -1074,6 +1152,28 @@ impl Process {
                 .unwrap();
         }
 
+        // Initialize emulate rlimits to their native values.
+        // TODO: For determinism, we may want to use fixed limits for some or all of these.
+        // Some applications like Tor will change behavior depending on these limits.
+        // Ultimately they'd need to be compatible with the native limits though.
+        let mut rlimits: [linux_api::resource::rlimit64;
+            linux_api::resource::RLIM_NLIMITS as usize] =
```
stevenengler: I think that this size `linux_api::resource::RLIM_NLIMITS as usize` could just be an underscore, rather than specifying it twice?

src/main/host/process.rs
```diff
@@ -1074,6 +1152,28 @@ impl Process {
...
+            [shadow_pod::zeroed(); linux_api::resource::RLIM_NLIMITS as usize];
+        for r in 0..linux_api::resource::RLIM_NLIMITS {
+            let r = linux_api::resource::Resource::try_from(r).unwrap();
```
stevenengler: This seems a little fragile if linux increases `RLIM_NLIMITS` without adding a constant for each new value (or if there is some value that isn't exposed on x86-64). But it works currently, and if linux ever does cause this to break, it will fail loudly since it's in a common code path, so I think it's okay.

stevenengler (approved):
I have just some small comments. But it looks good!

src/test/resource/test_resource.rs
```diff
@@ -0,0 +1,377 @@
...
+    } else {
+        rlim - 1
+    }
+}
+
+trait GetRLimit: std::fmt::Debug {
+    fn getrlimit(
+        &self,
+        resource: u32,
+    ) -> Result<linux_api::resource::rlimit, linux_api::errno::Errno>;
```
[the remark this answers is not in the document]
sporksmith: done

src/test/resource/test_resource.rs
```diff
@@ -0,0 +1,377 @@
...
+                let new_val = linux_api::resource::rlimit {
+                    rlim_cur: initial.rlim_cur + 1,
+                    rlim_max: initial.rlim_max,
+                };
+                g.setrlimit(resource.into(), &new_val).unwrap();
+                assert_eq!(LibcGetRLimit.getrlimit(resource.into()).unwrap(), new_val);
+
+                g.setrlimit(resource.into(), &initial).unwrap();
+            } else {
+                println!("  not incrementing; already at max");
```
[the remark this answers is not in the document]
sporksmith: We test this case in `test_setrlimits_errors`

sporksmith (commented):
Thanks!

PR #3695, opened 2025-11-05
Adding experimental per-edge bandwidth limiting to Shadow

vaibhav-mattoo:
- Most Tor path-selection and load balancing papers assume circuit throughput is limited entirely by relay capacities, approximated by **BW_circuit = min(BW_guard, BW_middle, BW_exit)**. This omits constraints from the underlying overlay links between relays.

- This PR adds experimental per-edge bandwidth limiting in Shadow, allowing for simulations where circuit performance can be influenced by link capacities as well as node capacities, which aids future work on link-informed path selection and overlay-aware routing strategies for Tor. This PR does not affect any previous shadow setups and these features are only activated if edge bandwidth constraints are defined and the flag to enable the feature is used. 
- Changes are:
1) New experimental flags:
--edge-bandwidth-limiting-enabled (default false): toggles the feature.
--edge-bandwidth-algorithm (default "token-bucket").
--edge-bandwidth-burst-ratio (default 1.5).
and --help and its tests are updated with these flags grouped under the experimental section.
2) The graph attribute gml parsing logic now supports edge_bandwidth_down/edge_bandwidth_up for directional bandwidth limits (in src/main/network/graph/mod.rs), these are totally optional per edge.
src/main/core/sim_config.rs then normalizes units to base SI, constructs the (src_node, dst_node) -> bytes/sec map, and instantiates per-edge token buckets.
3) src/main/core/worker.rs iterates over the per hop node sequence (adjacent pairs u->v) for each packet and enforces the per-edge token bucket on each directed hop, accumulating any extra simulated delay across the path.

-  All other changes are docs updates and testcases for this feature (which comprise a major portion of the new lines of code here, which is why this seems pretty large).
1) New integration test  edge_bw-shadow to test this feature. This is client-server transfer through an edge with constrained bandwidth and we verify that transfer duration exceeds a minimal threshold to show throttling (along with log messages).
2) CLI help compare test cli-long-compare updaetd to account for this.
3) small port binding change to bindc-linux

### DIFF
--- a/src/main/core/worker.rs
+++ b/src/main/core/worker.rs
@@ -369,7 +369,7 @@ impl Worker {
 
         let mut delay = Worker::with(|w| w.shared.latency(src_ip, dst_ip).unwrap()).unwrap();
 
-        // Optional experimental per-edge bandwidth limiting: if enabled and not in bootstrap,
+        // experimental per-edge bandwidth limiting: if enabled and not in bootstrap,
         // enforce bandwidth per hop along the selected route. For hops without a bucket, assume
         // no limit. We sum the required conforming delays across hops.
         if Worker::with(|w| w.shared.edge_bw_enabled).unwrap() && !is_bootstrapping {
@@ -388,7 +388,6 @@ impl Worker {
                 })
                 .unwrap();
                 if let Some(nodes) = nodes_vec {
-                    // Approximate per-hop packet size to engage the token bucket.
                     // Use a conservative MTU-sized decrement so large application payloads
                     // that are segmented at lower layers still experience rate limiting.
                     let needed_bytes = if payload_size == 0 {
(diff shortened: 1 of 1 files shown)


sporksmith:
From a quick look this mostly seems plausible, though I'm not sure how one would go about finding the real-world edge-link capacities, or how plausible it is that they would become the bottleneck. Are you saying this is intended to help model the case where tor relays communciate with each-other over *some other* overlay network?

This one probably needs review from @robgjansen , at least for the high-level questions of whether this is a sensible extension to the network model

nikitaborisov:
We're working on some measurements to see if there are cases where link-level, rather than node-level, congestion dominates the performance of links between two Tor nodes. We wanted to extend Shadow with link-level constraints so that we can study how such link-level congestion would manifest itself.

We are a little ways from having a realistic model of link-level congestion, so it's not a priority to mainline this patch, but any comments you have about our approach would be most welcome!

Thanks, 
Nikita

robgjansen:
There is a lot here and I haven't reviewed in detail yet, but sharing some initial thoughts.

I'm happy to see that the new parts are added as experimental options, are optional, and are off by default. And it is great that the tests and the documentation components were added.

Most of the logic is extended from existing files in multi-level nested code that might be difficult to follow and maintain long term. I wonder if there is some module-based design that would be more intuitive and allow us to break up the new functionality into much smaller units of code that would be easier to test using unit tests.

More generally, we have resisted adding too many Internet-level components to Shadow in order to avoid it snowballing into an Internet/routing simulator, which is not something that we would be able to do well and maintain. This patch gets us a little closer to that, though just a little. I think we would need to tease apart if this feature is generally useful or is more like a research prototype that maybe isn't worth full integration quite yet.

I wonder if this patch would maybe not be needed at all if we implemented a design something like the one discussed in #3701:

```
Sender Interface -> TokenBucket1 -> Router(CoDel) -> TokenBucket2 ->
(== Unlimited Core Link ==)
-> TokenBucket3  -> Router(CoDel) -> TokenBucket4 -> Receiver Interface
```

If you wanted to test out link-level congestion effects, you could configure token buckets 2 and 3 to have lower rate limits than 1 and 4. Can you comment on this? It would be great if we could handle two related issues with the same design, as it would simplify code design and maintenance on our side.

PR #3711, opened 2026-01-08
Update apt before installing packages

robgjansen:
Whenever using apt, we should be updating it first to pull in any package server changes made since the OS image was created.

~~(Also, I think apt is more modern than apt-get. apt-get output is stable while apt is not, but we're not relying on the output here so :shrug: I made the change.)~~

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -67,6 +67,7 @@ jobs:
           # date with target branch before merging, anyway.
           # See https://github.com/shadow/shadow/issues/2166
           ref: ${{ github.event.pull_request.head.sha }}
+      - run: sudo apt-get update
       - run: sudo apt-get install shellcheck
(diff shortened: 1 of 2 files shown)


sporksmith (approved):
I don't think apt-get is going away anytime soon, and that apt isn't intended for scripting usage in general. OTOH I'm not aware of any practical problems with using apt when we're not parsing output, other than the annoying warnings

sporksmith:
The warning `WARNING: apt does not have a stable CLI interface. Use with caution in scripts.` implies that it may be a bit more likely than apt-get for given arguments to stop working or have different behavior in the future, though again IDK how likely this is to be a problem in practice

robgjansen:
OK, I'll switch back to `apt-get` if for no other reason than to avoid the annoying warnings :) Thanks!

PR #3742, opened 2026-04-19
resolver: map ::1 to loopback in getaddrinfo

starius:
I'm trying to run `bitcoind` (regtest) in shadow. This PR fixes one of the issues preventing it from running there.

Bitcoin Core's [LookupHost("::1", false)](https://github.com/bitcoin/bitcoin/blob/v31.0/src/httpserver.cpp#L152) goes through WrappedGetAddrInfo() and [calls getaddrinfo("::1")](https://github.com/bitcoin/bitcoin/blob/v31.0/src/netbase.cpp#L69). Shadow used to return `gaierror(-2, "Name or service not known")` for this, now it return the loopback.

Added a regression test: new subtest in `test_numeric_host`.

### DIFF
--- a/src/lib/shim/shim_api_addrinfo.c
+++ b/src/lib/shim/shim_api_addrinfo.c
@@ -164,6 +164,17 @@ static void _getaddrinfo_appendv6(struct addrinfo** head, struct addrinfo** tail
     }
 }
 
+// Returns the IPv4-mapped IPv6 address for the IPv4 address `addr`.
+static struct in6_addr _getaddrinfo_make_v4mapped_addr(uint32_t addr) {
+    static const unsigned char prefix[] = {
+        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xff, 0xff,
+    };
+    struct in6_addr mapped_addr = {0};
+    memcpy(mapped_addr.s6_addr, prefix, sizeof(prefix));
+    memcpy(&mapped_addr.s6_addr[sizeof(prefix)], &addr, sizeof(addr));
+    return mapped_addr;
+}
+
 // Looks for matching IPv4 addresses in /etc/hosts and them to the list
 // specified by `head` and `tail`.
 static void _getaddrinfo_add_matching_hosts_ipv4(struct addrinfo** head, struct addrinfo** tail,
@@ -403,27 +414,25 @@ int shimc_api_getaddrinfo(const char* node, const char* service, const struct ad
     }
 
     // "`node` specifies either a numerical network address..."
+    // Parse both families when needed so we can distinguish a wrong-family
+    // numeric address (EAI_ADDRFAMILY) from a nonnumeric string
+    // (EAI_NONAME with AI_NUMERICHOST).
+    const bool check_ipv6_numeric = add_ipv6 || hints->ai_family == AF_INET;
+    const bool check_ipv4_numeric = add_ipv4 || hints->ai_family == AF_INET6;
+
     struct in6_addr addr6;
-    const bool parsed_ipv6 = inet_pton(AF_INET6, node, &addr6) == 1;
+    const bool parsed_ipv6 = check_ipv6_numeric && inet_pton(AF_INET6, node, &addr6) == 1;
     if (parsed_ipv6 && add_ipv6) {
         _getaddrinfo_appendv6(res, &tail, add_tcp, add_udp, add_raw, &addr6, port);
     }
     uint32_t addr;
-    const bool parsed_ipv4 = inet_pton(AF_INET, node, &addr) == 1;
+    const bool parsed_ipv4 = check_ipv4_numeric && inet_pton(AF_INET, node, &addr) == 1;
     if (parsed_ipv4) {
         if (add_ipv4) {
             _getaddrinfo_appendv4(res, &tail, add_tcp, add_udp, add_raw, addr, port);
         } else if (hints->ai_family == AF_INET6 && (hints->ai_flags & AI_V4MAPPED)) {
-            struct in6_addr mapped_addr = {
-                .s6_addr =
-                    {
-                        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
-                        0x00, 0x00, 0xff, 0xff,
-                    },
-            };
-            memcpy(&mapped_addr.s6_addr[12], &addr, sizeof(addr));
-            _getaddrinfo_appendv6(
-                res, &tail, add_tcp, add_udp, add_raw, &mapped_addr, port);
+            struct in6_addr mapped_addr = _getaddrinfo_make_v4mapped_addr(addr);
+            _getaddrinfo_appendv6(res, &tail, add_tcp, add_udp, add_raw, &mapped_addr, port);
         }
     }
     // If we successfully parsed as a numeric address, there's no need to
@@ -432,6 +441,8 @@ int shimc_api_getaddrinfo(const char* node, const char* service, const struct ad
         return 0;
     }
     if (parsed_ipv4 || parsed_ipv6) {
+        // We recognized `node` as a numeric address, but it doesn't belong to
+        // an address family that this lookup is allowed to return.
         return EAI_ADDRFAMILY;
     }
     // "If  hints.ai_flags  contains the  AI_NUMERICHOST  flag,  then  node
--- a/src/lib/shim/shim_api_addrinfo.c
+++ b/src/lib/shim/shim_api_addrinfo.c
@@ -88,8 +88,8 @@ static int _getaddrinfo_service(in_port_t* port, const char* service,
 // Creates an `addrinfo` pointing to `addr`, and adds it to the linked list
 // specified by `head` and `tail`. An empty list can be passed in by setting
 // `*head` and `*tail` to NULL.
-static void _getaddrinfo_append(struct addrinfo** head, struct addrinfo** tail, int socktype,
-                                struct sockaddr* addr, socklen_t addrlen) {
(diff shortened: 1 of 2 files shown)


src/lib/shim/shim_api_addrinfo.c
```diff
@@ -382,6 +382,14 @@ int shimc_api_getaddrinfo(const char* node, const char* service, const struct ad
         // TODO: try parsing as IPv6
     }
     if (add_ipv4) {
+        // Shadow doesn't support IPv6 sockets, but some applications still
+        // probe the IPv6 loopback address during startup. Treat "::1" as the
+        // IPv4 loopback so they keep working in Shadow's IPv4-only model.
+        if (strcmp(node, "::1") == 0) {
+            _getaddrinfo_appendv4(
+                res, &tail, add_tcp, add_udp, add_raw, htonl(INADDR_LOOPBACK), port);
+        }
```
stevenengler: Does Linux ever return IPv4 addresses for an IPv6 address string? This seems like it could be confusing, but I'm not familiar enough with things like IPv4-mapped addresses to know whether this is a reasonable thing to return.

sporksmith: +1 this doesn't seem like a reasonable general solution. Maybe the bitcoin software you're trying to run can be configured and/or modified to be better-behaved in an ipv4-only environment?

starius: Thanks, this was a good catch. The previous implementation was too ad hoc: mapping the literal IPv6 loopback string `::1` to an IPv4 loopback result does not match Linux `getaddrinfo` behavior.

I reworked the patch to make the numeric-address path behave like Linux instead of special-casing `::1` as IPv4:

- numeric IPv6 literals such as `::1` now produce `AF_INET6` results
- `AF_INET` + `::1` now returns `EAI_ADDRFAMILY`
- `AI_V4MAPPED` handling for IPv4 literals in `AF_INET6` requests is preserved

I also updated the resolver test to check the Linux-compatible behavior directly, including `sockaddr_in6` comparison.

I re-ran a real Bitcoin Core 31.0 `bitcoind` binary under the rebuilt Shadow. With this change it no longer fails immediately at the earlier startup point triggered by `LookupHost("::1", false)`. It now gets past that stage and continues into later initialization, where it runs into separate Shadow limitations unrelated to this resolver change.

src/lib/shim/shim_api_addrinfo.c
```diff
@@ -378,20 +406,37 @@ int shimc_api_getaddrinfo(const char* node, const char* service, const struct ad
...
             _getaddrinfo_appendv4(res, &tail, add_tcp, add_udp, add_raw, addr, port);
+        } else if (hints->ai_family == AF_INET6 && (hints->ai_flags & AI_V4MAPPED)) {
+            struct in6_addr mapped_addr = {
+                .s6_addr =
+                    {
+                        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+                        0x00, 0x00, 0xff, 0xff,
+                    },
+            };
+            memcpy(&mapped_addr.s6_addr[12], &addr, sizeof(addr));
```
sporksmith: I *think* I understand what's going on here, but it could use some comments, particularly for the "12". Better yet if we can get rid of the magic "12" altogether - maybe something like:

```
const char v4map_prefix[] = { ... };
memcpy(&mapped_attr.s6_addr[0], v4map_prefix, sizeof(v4map_prefix);
memcpy(&mapped_attr.s6_addr[sizeof(v4map_prefix), &addr, sizeof(addr);
```

src/lib/shim/shim_api_addrinfo.c
```diff
@@ -378,20 +406,37 @@ int shimc_api_getaddrinfo(const char* node, const char* service, const struct ad
...
 
     // "`node` specifies either a numerical network address..."
-    if (add_ipv6) {
-        // TODO: try parsing as IPv6
-    }
-    if (add_ipv4) {
-        uint32_t addr;
-        if (inet_pton(AF_INET, node, &addr) == 1) {
+    struct in6_addr addr6;
+    const bool parsed_ipv6 = inet_pton(AF_INET6, node, &addr6) == 1;
```
sporksmith, line 424: Nit - we should only attempt this parse at all if `add_ipv6` is true, right? (though it probably doesn't really matter much)

src/lib/shim/shim_api_addrinfo.c
```diff
@@ -378,20 +406,37 @@ int shimc_api_getaddrinfo(const char* node, const char* service, const struct ad
...
+                res, &tail, add_tcp, add_udp, add_raw, &mapped_addr, port);
         }
     }
     // If we successfully parsed as a numeric address, there's no need to
     // continue on to doing name-based lookups.
     if (*res != NULL) {
         return 0;
     }
+    if (parsed_ipv4 || parsed_ipv6) {
+        return EAI_ADDRFAMILY;
```
sporksmith, line 447: Why return an error here?

sporksmith (changes requested):
I think you're on a reasonable track here; just a couple small comments

src/lib/shim/shim_api_addrinfo.c
starius: Good point. I removed the magic offset in `resolver: clarify numeric family mismatch handling`.

The IPv4-mapped construction now goes through a small helper with a named `::ffff:0:0/96` prefix and uses `sizeof(prefix)` for the copy, so the `AI_V4MAPPED` path is less ad hoc and easier to follow.

src/lib/shim/shim_api_addrinfo.c
starius: The extra parse is intentional.

Even when `add_ipv6` is false, we still need to recognize that `node` is a numeric IPv6 literal so we can distinguish "wrong-family numeric address" from "not a numeric address at all". In particular, `AF_INET + "::1"` should return `EAI_ADDRFAMILY`, not fall through to the nonnumeric path.

I tried to make that explicit in `resolver: clarify numeric family mismatch handling` by separating the "should we parse this family at all?" conditions from the "can we append a result for this family?" conditions, and by adding a comment on that distinction.

src/lib/shim/shim_api_addrinfo.c
starius: Because by that point we have already recognized `node` as a numeric address, but the requested family does not allow us to return it.

In that case Linux returns `EAI_ADDRFAMILY` rather than continuing on to hostname lookup.

I added a short comment there in `resolver: clarify numeric family mismatch handling`, and `resolver: test and preserve AI_V4MAPPED results` now covers the relevant cases directly.

starius:
@sporksmith 

Thanks! I addressed these in `resolver: clarify numeric family mismatch handling` and `resolver: test and preserve AI_V4MAPPED results`.

The cleanup removes the magic offset in the IPv4-mapped IPv6 construction, makes the numeric parse conditions explicit, and documents why the wrong-family numeric case returns `EAI_ADDRFAMILY`.

I also added resolver coverage for the Linux-compatible cases around this path:
- `AF_INET + "::1"` -> `EAI_ADDRFAMILY`
- `AF_INET6 + "127.0.0.1"` without `AI_V4MAPPED` -> `EAI_ADDRFAMILY`
- `AF_INET6 + "127.0.0.1"` with `AI_V4MAPPED` -> mapped IPv6 result, including `ai_flags == AI_V4MAPPED`

After that I reran the resolver tests and a Bitcoin smoke test. `bitcoind` still gets past the original `LookupHost("::1", false)` startup failure site.

src/lib/shim/shim_api_addrinfo.c
```diff
@@ -403,27 +414,25 @@ int shimc_api_getaddrinfo(const char* node, const char* service, const struct ad
     }
 
     // "`node` specifies either a numerical network address..."
+    // Parse both families when needed so we can distinguish a wrong-family
+    // numeric address (EAI_ADDRFAMILY) from a nonnumeric string
+    // (EAI_NONAME with AI_NUMERICHOST).
+    const bool check_ipv6_numeric = add_ipv6 || hints->ai_family == AF_INET;
+    const bool check_ipv4_numeric = add_ipv4 || hints->ai_family == AF_INET6;
```
sporksmith: I find the added indirection of these `check_ipvx_numeric` vars harder to follow than before. I think the commit adding these also changes the behavior, rather than just being a refactor as in the commit description? I don't understand the `hints` check that this adds, especially since `add_ipvx` already take the hints into account, and `AF_INET` and `AF_INET6` look swapped...?

sporksmith (changes requested):

src/lib/shim/shim_api_addrinfo.c
starius: Thanks, I agree.

The `check_ipv4_numeric` / `check_ipv6_numeric` indirection was trying to preserve the "parsed numeric literal, but wrong family" -> `EAI_ADDRFAMILY` behavior, but it made the control flow harder to follow. It also means that commit `resolver: clarify numeric family mismatch handling` is no longer just a cleanup.

I addressed that in commit `resolver: restore direct numeric literal parsing`. That commit removes those guard vars and goes back to direct `inet_pton` parsing for both families, which keeps this part of the code much closer to the earlier version.

sporksmith (approved):

sporksmith:
I'm going to see if GH will let me rebase and force-push this branch so that I don't need to re-review the rebase result ...

sporksmith:
> I'm going to see if GH will let me rebase and force-push this branch so that I don't need to re-review the rebase result ...

Looks like yes - just waiting to hear whether @stevenengler wants another look before merging ...

sporksmith:
> Looks like yes - just waiting to hear whether @stevenengler wants another look before merging ...

Confirmed out-of-band.

Thanks for the contribution! Merging ...

PR #3743, opened 2026-04-20
host: implement minimal SysV shared memory support

starius:
Add a host-scoped SysV shared memory namespace backed by memfd, and wire in handlers for shmget, shmat, shmdt, and shmctl with IPC_STAT and IPC_RMID.

Route attachments through the memory manager so shared mappings are tracked correctly, clean them up on exit and exec, and carry attachment bookkeeping across fork.

Allow segment owners to look up and attach their own segments even when the permission bits are 0000, matching Linux behavior.

Add focused native and Shadow tests covering single-process use, multi-process sharing on one host, owner access to mode-0000 segments, and isolation across hosts.

This makes shared-memory create, attach, stat, and remove flows work inside Shadow and gets PostgreSQL past the previous SysV shared-memory blocker into its bootstrap path.


sporksmith:
Could you force-push your rebased and reworked version that uses the linux-api bindings?

starius:
I rebased and used linux-api for shm/ipc keys, flags, commands, and shmid64_ds.

Also fixed a bug in my own test `src/test/sysv_shm/test_sysv_shm.c`: native Linux allows root to attach to a mode-0000 segment, and Shadow does too. Non-root still expects EACCES.

Here is [the diff of PR patches](https://gist.github.com/starius/8c2c4f3bd26e75048ac70a248de3cb1b) if that makes sense.

starius:
Fixed the test. A root user could actually get both outcomes depending on capabilities.

src/main/host/syscall/handler/sysv_shm.rs
```diff
@@ -0,0 +1,94 @@
...
+
+use crate::host::syscall::handler::{SyscallContext, SyscallHandler};
+use crate::host::syscall::types::ForeignArrayPtr;
+
+impl SyscallHandler {
+    log_syscall!(
+        shmget,
+        /* rv */ std::ffi::c_int,
+        /* key */ kernel_key_t,
+        /* size */ libc::size_t,
```
sporksmith: Throughout - `linux_api::posix_types::kernel_size_t`

PR #3744, opened 2026-04-20
sched: emulate scheduler query and update syscalls

starius:
Add Shadow-side handling for `sched_getparam`, `sched_getscheduler`, `sched_setparam`, and `sched_setscheduler`.

Track scheduling policy and priority per simulated thread so callers can query and update scheduler state without falling back to ENOSYS.

This keeps the implementation narrow to the scheduler modes and priority rules needed by common userspace code and adds a regression test that covers the new syscall surface from both the main thread and a spawned thread.

This is another change needed to run Bitcoin Core under shadow. In the Bitcoin Core, a newly created thread calls `sched_getparam` and `sched_getscheduler`.


src/main/host/syscall/handler/sched.rs
```diff
@@ -9,12 +9,93 @@ use shadow_shim_helper_rs::syscall_types::ForeignPtr;
 use crate::host::syscall::handler::{SyscallContext, SyscallHandler};
 use crate::host::syscall::type_formatting::SyscallNonDeterministicArg;
 use crate::host::syscall::types::ForeignArrayPtr;
-use crate::host::thread::ThreadId;
+use crate::host::thread::{Thread, ThreadId};
 
 // We always report that the thread is running on CPU 0, Node 0
 const CURRENT_CPU: u32 = 0;
 
+const LINUX_SCHED_NORMAL: std::ffi::c_int = 0;
```
sporksmith: Throughout - we'll want to use the bindgen'd + wrapped constants and types from our linux-api crate. I went ahead and added the ones you need (i think) in https://github.com/shadow/shadow/pull/3749 ; let me know if any are missing. (You may want to wait until it's completed and merged before updating your PR, in case there are any other changes)

sporksmith: I just added the rseq flags to that PR as well.

> Throughout - we'll want to use the bindgen'd + wrapped constants and types from our linux-api crate.

Clarification - this of course doesn't apply to tests exercising the libc interface. It *does* apply to tests that test the syscall directly instead of going through libc, but the latter usually isn't needed.

starius: I reworked it using linux-api. Sorry for force-pushing! It was too much effort to resolve conflicts when merging linux-api, so I decided to start from the latest `main` instead.

stevenengler:
In the past, we've tended to avoid adding stubs for syscalls (see the discussion in #3280), but we never came to a conclusion about what to do with stubs. We should probably decide on that so that we don't add stubs for some people and not others.

For example this tracks the scheduler settings, but doesn't actually change anything. But maybe it doesn't make sense for us to ever implement any of the linux scheduling options, so the only option is to have these as stubs?

In any case, I think that these new syscalls should emit warnings to say that we're returning fake values and not emulating these features. For example see the implementation for `set_robust_list` where we call `warn_once_then_debug!`.

starius:
> In the past, we've tended to avoid adding stubs for syscalls (see the discussion in #3280), but we never came to a conclusion about what to do with stubs. We should probably decide on that so that we don't add stubs for some people and not others.
> 
> For example this tracks the scheduler settings, but doesn't actually change anything. But maybe it doesn't make sense for us to ever implement any of the linux scheduling options, so the only option is to have these as stubs?
> 
> In any case, I think that these new syscalls should emit warnings to say that we're returning fake values and not emulating these features. For example see the implementation for `set_robust_list` where we call `warn_once_then_debug!`.

I agree these are compatibility stubs rather than real scheduling emulation. The current implementation only tracks requested policy/priority and doesn't affect Shadow scheduling behavior. I added a `warn_once_then_debug!` warning to make that explicit at runtime, similar to `set_robust_list`.

More generally, I agree it would be good to settle a consistent policy for when we prefer stubs versus ENOSYS.

src/main/host/syscall/handler/sched.rs
```diff
@@ -4,17 +4,82 @@ use bitflags::Flags;
...
 use crate::host::syscall::handler::{SyscallContext, SyscallHandler};
 use crate::host::syscall::type_formatting::SyscallNonDeterministicArg;
 use crate::host::syscall::types::ForeignArrayPtr;
-use crate::host::thread::ThreadId;
+use crate::host::thread::{Thread, ThreadId};
 
 // We always report that the thread is running on CPU 0, Node 0
 const CURRENT_CPU: u32 = 0;
 
+fn normalize_tid(ctx: &SyscallContext, tid: kernel_pid_t) -> Result<ThreadId, Errno> {
```
sporksmith: This function name isn't very informative, though I don't have a much better idea.

Given that the behavior seems a bit specific to these syscalls, and it has one call site, I'd lean towards just inlining it. Otherwise, some function doc-comment would be helpful.

src/main/host/syscall/handler/sched.rs
```diff
@@ -4,17 +4,82 @@ use bitflags::Flags;
...
+
+    let tid = ThreadId::try_from(tid).or(Err(Errno::ESRCH))?;
+    if !ctx.objs.host.has_thread(tid) {
+        return Err(Errno::ESRCH);
+    }
+
+    Ok(tid)
+}
+
+fn with_target_thread<T>(
```
sporksmith: Needs a function doc-comment

src/main/host/syscall/handler/sched.rs
```diff
@@ -4,17 +4,82 @@ use bitflags::Flags;
...
+    tid: kernel_pid_t,
+    f: impl FnOnce(&Thread) -> T,
+) -> Result<T, Errno> {
+    let target_tid = normalize_tid(ctx, tid)?;
+
+    if target_tid == ctx.objs.thread.id() {
+        return Ok(f(ctx.objs.thread));
+    }
+
+    let Some(thread_rc) = ctx.objs.host.thread_cloned_rc(target_tid) else {
```
sporksmith, line 35: I think this will panic in debug builds, due to `RootedRc::explicit_drop` not being called. The CI should catch this once the new tests are actually being executed (see comment on test file)

src/main/host/syscall/handler/sched.rs
```diff
@@ -4,17 +4,82 @@ use bitflags::Flags;
...
+        return Err(Errno::ESRCH);
+    };
+    let thread = thread_rc.borrow(ctx.objs.host.root());
+    Ok(f(&thread))
+}
+
+fn base_policy(policy: std::ffi::c_int) -> Option<Sched> {
```
sporksmith: Needs a doc-comment

src/main/host/syscall/handler/sched.rs
```diff
@@ -4,17 +4,82 @@ use bitflags::Flags;
...
+    let policy = policy & !SCHED_RESET_ON_FORK;
+    Sched::try_from(policy).ok().filter(|policy| {
```
sporksmith: Why the `filter` here? (Probably needs an explanatory comment)

src/main/host/syscall/handler/sched.rs
```diff
@@ -4,17 +4,82 @@ use bitflags::Flags;
...
+            Sched::SCHED_NORMAL
+                | Sched::SCHED_FIFO
+                | Sched::SCHED_RR
+                | Sched::SCHED_BATCH
+                | Sched::SCHED_IDLE
+        )
+    })
+}
+
+fn validate_sched_attrs(policy: std::ffi::c_int, priority: std::ffi::c_int) -> Result<(), Errno> {
```
sporksmith: Needs a doc comment

src/main/host/syscall/handler/sched.rs
```diff
@@ -4,17 +4,82 @@ use bitflags::Flags;
...
+    match base_policy(policy) {
+        Some(Sched::SCHED_NORMAL | Sched::SCHED_BATCH | Sched::SCHED_IDLE) => {
```
sporksmith: May need some inline comments, if not covered as part of the doc comment, particularly explaining the 0, 1, and 99 "magic numbers"

src/main/host/syscall/handler/sched.rs
```diff
@@ -103,6 +168,88 @@ impl SyscallHandler {
...
+    log_syscall!(
+        sched_getparam,
+        /* rv */ i32,
+        /* pid */ kernel_pid_t,
+        /* param */ *const std::ffi::c_void,
+    );
+    pub fn sched_getparam(
+        ctx: &mut SyscallContext,
+        tid: kernel_pid_t,
+        param_ptr: ForeignPtr<libc::sched_param>,
```
sporksmith: This should be the kernel type, `linux_api::sched::sched_attr`, not the libc type.

src/main/host/syscall/handler/sched.rs
```diff
@@ -103,6 +168,88 @@ impl SyscallHandler {
...
+        tid: kernel_pid_t,
+    ) -> Result<std::ffi::c_int, Errno> {
+        with_target_thread(ctx, tid, |thread| thread.sched_policy())
+    }
+
+    log_syscall!(
+        sched_setparam,
+        /* rv */ i32,
+        /* pid */ kernel_pid_t,
+        /* param */ *const std::ffi::c_void,
```
sporksmith: I think we ideally want the proper type here, so that the  pointee value gets logged. I think you'd need to add a logging implementation for `linux_api::sched::sched_attr` to make it work, though.

Leaving it void and adding a TODO comment would be ok for now.

src/main/host/syscall/handler/sched.rs
```diff
@@ -103,6 +168,88 @@ impl SyscallHandler {
...
+    );
+    pub fn sched_setparam(
+        ctx: &mut SyscallContext,
+        tid: kernel_pid_t,
+        param_ptr: ForeignPtr<libc::sched_param>,
```
sporksmith: This should be the kernel type, `linux_api::sched::sched_attr`, not the libc type.

src/main/host/syscall/handler/sched.rs
```diff
@@ -103,6 +168,88 @@ impl SyscallHandler {
...
+        })?;
+
+        Ok(())
+    }
+
+    log_syscall!(
+        sched_setscheduler,
+        /* rv */ i32,
+        /* pid */ kernel_pid_t,
+        /* policy */ std::ffi::c_int,
```
sporksmith, line 227: The proper types would be preferable here for logging. Again I'd be ok with a TODO for now, though.

src/main/host/thread.rs
```diff
@@ -46,6 +47,8 @@ pub struct Thread {
     id: ThreadId,
     host_id: HostId,
     process_id: ProcessId,
+    sched_policy: Cell<std::ffi::c_int>,
```
sporksmith: Better to use the "lifted" types `Sched` and `sched_attr` from `linux_api` here, and convert to/from int in the syscall handlers.

src/test/sched/test_sched.rs
```diff
@@ -0,0 +1,45 @@
+/*
```
sporksmith, line 1: I don't think this test is actually getting executed. You'll need to add lines to the `CMakeLists.txt` in this directory

sporksmith, line 1: The tests here could use some comments

sporksmith, line 1: Maybe also test operating on a thread other than the current one? Optionally, maybe also a process other than the current one, though `fork` in rust is a bit tricky.

src/main/host/syscall/handler/sched.rs
```diff
@@ -80,6 +80,13 @@ fn validate_sched_attrs(policy: std::ffi::c_int, priority: std::ffi::c_int) -> R
         _ => Err(Errno::EINVAL),
     }
 }
+
+fn log_sched_stub_warning() {
+    warn_once_then_debug!(
```
sporksmith: I think we want this inlined in each place we log, s.t. it gets logged once for each relevant syscall that gets used.

I'm on the fence whether we need these at all, especially for the `get` calls. I'll say more on the previous thread on this topic.

sporksmith (changes requested):

sporksmith:
> In the past, we've tended to avoid adding stubs for syscalls (see the discussion in https://github.com/shadow/shadow/issues/3280), but we never came to a conclusion about what to do with stubs. We should probably decide on that so that we don't add stubs for some people and not others.

I think it's hard to write an end-all-be-all policy, though documenting some guidelines would probably be helpful.

> For example this tracks the scheduler settings, but doesn't actually change anything. But maybe it doesn't make sense for us to ever implement any of the linux scheduling options, so the only option is to have these as stubs?

I think this is *pretty* safe, since generally, messing with scheduler parameters mostly affects performance, not behavior, and shadow is already implementing its own scheduling and pretending to have an infinite number of infinitely fast CPUs (by default), which makes it further moot.

Reporting that a realtime policy (no preemption) is active would probably be closer to the truth of shadow's actual (default) scheduling behavior, but it's probably better to return the "expected/normal" thing by default and not dive too far down this rabbit hole.

> In any case, I think that these new syscalls should emit warnings to say that we're returning fake values and not emulating these features. For example see the implementation for set_robust_list where we call warn_once_then_debug!.

I don't feel strongly that we need warnings for these per above, but I don't mind them either. I guess it's better to conservatively have them.

src/main/host/syscall/handler/sched.rs
starius: I removed the separate `normalize_tid` helper and folded that logic into a single scheduler-specific helper, so there is no longer a generic, poorly named one-call-site function here.

See commit "sched: safely target sibling threads in scheduler syscalls".

src/main/host/syscall/handler/sched.rs
starius: I added a doc-comment to the remaining scheduler-specific helper explaining the `tid == 0` behavior and the cloned-thread borrow/drop path.

```rust
/// Run `f` on the thread targeted by the scheduler syscalls.
///
/// Linux treats a tid of 0 as the current thread for these calls. For any other tid, borrow the
/// emulated thread through a cloned `RootedRc` and explicitly drop that clone before returning.
```

See commit "sched: safely target sibling threads in scheduler syscalls".

src/main/host/syscall/handler/sched.rs
starius: Yes, that was a real issue. I fixed it by wrapping the cloned thread reference in `ExplicitDropper`, matching the existing signal-handler pattern, and I added a sibling-thread scheduler test that now exercises this path under Shadow.

```rust
let thread_rc =
    ExplicitDropper::new(thread_rc, |value| value.explicit_drop(ctx.objs.host.root()));
let thread = thread_rc.borrow(ctx.objs.host.root());
```

See commit "sched: safely target sibling threads in scheduler syscalls".

src/main/host/syscall/handler/sched.rs
starius: The old helper was replaced with `parse_sched_policy`, which now has a doc-comment describing how the raw Linux policy value is parsed and how `SCHED_RESET_ON_FORK` is handled.

See commit "sched: use linux_api types for tracked scheduler state".

src/main/host/syscall/handler/sched.rs
starius: I removed the `filter`-based helper entirely and replaced it with an explicit `match` over the supported `Sched` values, so the accepted policies are now spelled out directly in the code.

```rust
match policy {
    Sched::SCHED_NORMAL
    | Sched::SCHED_FIFO
    | Sched::SCHED_RR
    | Sched::SCHED_BATCH
    | Sched::SCHED_IDLE => Some((policy, reset_on_fork)),
    _ => None,
}
```

See commit "sched: use linux_api types for tracked scheduler state".

src/main/host/syscall/handler/sched.rs
starius: `validate_sched_attrs` now has a doc-comment that explains the priority rules Shadow is enforcing for the tracked scheduler policies.

```rust
/// Validate the priority rules for the scheduler policies Shadow tracks.
///
/// Non-realtime policies use a fixed priority of 0, while Linux accepts priorities 1 through 99
/// for `SCHED_FIFO` and `SCHED_RR`.
fn validate_sched_attrs(policy: Sched, priority: std::ffi::c_int) -> Result<(), Errno> {
    match policy {
        Sched::SCHED_NORMAL | Sched::SCHED_BATCH | Sched::SCHED_IDLE if priority == 0 => Ok(()),
        Sched::SCHED_FIFO | Sched::SCHED_RR if (1..=99).contains(&priority) => Ok(()),
        _ => Err(Errno::EINVAL),
    }
}
```

See commit "sched: use linux_api types for tracked scheduler state".

src/main/host/syscall/handler/sched.rs
starius: The validation helper now documents that non-realtime policies use priority `0`, while Linux accepts priorities `1..=99` for `SCHED_FIFO` and `SCHED_RR`, so those bounds are no longer unexplained magic numbers.

```rust
/// Non-realtime policies use a fixed priority of 0, while Linux accepts priorities 1 through 99
/// for `SCHED_FIFO` and `SCHED_RR`.
```

See commit "sched: use linux_api types for tracked scheduler state".

src/main/host/syscall/handler/sched.rs
starius: `sched_getparam` now uses `ForeignPtr<linux_api::sched::sched_attr>` instead of `ForeignPtr<libc::sched_param>`.

See commit "sched: use linux_api types for tracked scheduler state".

src/main/host/syscall/handler/sched.rs
starius: I added syscall formatting support for `linux_api::sched::sched_attr` and switched the scheduler syscall pointer logging over to the typed pointer instead of keeping it as `*const c_void`.

```rust
deref_pointer_impl!(linux_api::sched::sched_attr);
```

See commit "sched: use linux_api types for tracked scheduler state".

src/main/host/syscall/handler/sched.rs
starius: `sched_setparam` and `sched_setscheduler` now take `ForeignPtr<linux_api::sched::sched_attr>` rather than the libc type.

See commit "sched: use linux_api types for tracked scheduler state".

src/main/host/syscall/handler/sched.rs
starius: The scheduler syscall logging now uses the typed `linux_api::sched::sched_attr` pointee and the corresponding formatter support, so the logging path no longer needs the untyped placeholder.

See commit "sched: use linux_api types for tracked scheduler state".

src/main/host/thread.rs
starius: `Thread` now stores a typed `Sched` plus a tracked `linux_api::sched::sched_attr`, and the integer ABI conversions are done in the syscall handlers.

```rust
sched_policy: Cell<Sched>,
sched_attr: Cell<sched_attr>,
```

See commit "sched: use linux_api types for tracked scheduler state".

src/test/sched/test_sched.rs
starius: I added `src/test/sched/CMakeLists.txt`, registered the `sched` directory from `src/test/CMakeLists.txt`, and added the Shadow YAML so the test now runs through the normal `ctest` path.

See commit "test: register scheduler syscall test with CMake".

starius: I added short comments describing the same-thread coverage in `test_sched` so the test intent is easier to read.

See commit "test: register scheduler syscall test with CMake".

starius: I added a sibling-thread case where one thread operates on another thread's TID in the same process. I did not add the cross-process `fork` case, since your note already called that out as optional and awkward in Rust.

See commit "sched: safely target sibling threads in scheduler syscalls".

src/main/host/syscall/handler/sched.rs
starius: I removed the shared warning helper and inlined `warn_once_then_debug!` in each of `sched_getparam`, `sched_getscheduler`, `sched_setparam`, and `sched_setscheduler`, so the warning is now emitted once per relevant syscall rather than once globally.

See commit "sched: warn once per scheduler stub syscall".

starius:
@sporksmith 

Thanks for the review! I addressed the actionable issues in follow-up commits on top of the PR:

- "test: register scheduler syscall test with CMake"
- "sched: safely target sibling threads in scheduler syscalls"
- "sched: use linux_api types for tracked scheduler state"
- "sched: warn once per scheduler stub syscall"
- "rustfmt: format scheduler syscall changes"

That work wires `test_sched` into the normal CMake/ctest path, adds sibling-thread coverage, fixes the cloned-`RootedRc` explicit-drop bug that test exposed, switches the tracked scheduler state and syscall-facing types over to `linux_api`, and inlines `warn_once_then_debug!` once per relevant scheduler stub syscall.

I also validated the final series with `cargo fmt --check --all --manifest-path src/Cargo.toml` and `ctest --output-on-failure -R 'sched(_affinity)?-(linux|shadow)'`.

src/main/host/syscall/handler/sched.rs
```diff
@@ -42,36 +42,31 @@ fn with_sched_target_thread<T>(
...
+    let policy = Sched::try_from(policy & !SCHED_RESET_ON_FORK).ok()?;
+
+    // Shadow only tracks the common policies handled by `sched_getparam` and friends.
+    match policy {
+        Sched::SCHED_NORMAL
+        | Sched::SCHED_FIFO
+        | Sched::SCHED_RR
+        | Sched::SCHED_BATCH
+        | Sched::SCHED_IDLE => Some((policy, reset_on_fork)),
+        _ => None,
```
sporksmith: I'm still unclear why we shouldn't parse the other policies here.

Could you make this an exhaustive match instead of using an `_` case, and add comments for the remaining cases explaining why we return `None` for them?

sporksmith (commented):
Thanks, this is looking pretty good! One small request to clarify the policy-parsing, but after that I think I'm happy with this.

src/main/host/syscall/handler/sched.rs
starius: That makes sense. I changed `parse_sched_policy` to use an exhaustive `Sched` match instead of falling back through `_ => None`, and I added inline comments for the remaining unsupported policies.

We still return `None` for `SCHED_DEADLINE` because Shadow doesn't emulate the runtime/deadline/period semantics that deadline scheduling depends on. We also still return `None` for `SCHED_EXT` because that policy delegates scheduling to a BPF-defined scheduler, and Shadow doesn't have an equivalent execution model to represent or report it. Unknown raw integer values are still rejected earlier by `Sched::try_from(...)`, so the exhaustive match now covers the known Linux scheduler policies explicitly.

See commit "sched: make scheduler policy parsing exhaustive".

sporksmith: Thanks for the explanation! The end result of returning an error when trying to set one of these makes sense, but it's a little bit confusing for this logic to live in a function called `parse_sched_policy`. The user specifying one of these doesn't result in a parse error; the problem is that shadow doesn't support setting these particular policies.

I think this part (and maybe the whole function) could be inlined into the syscall handling function. It'd also be useful to have a separate warning log entry for these cases calling out that we're returning an error here because of unimplemented shadow functionality, not because the managed process provided bad arguments. (This one could probably be a `warn!`, though `warn_once_then_debug!` would also be ok)

starius: I agree with that distinction. I removed `parse_sched_policy` and moved the policy handling into `sched_setscheduler`, so unknown raw policy values still fail through `Sched::try_from(...)`, but `SCHED_DEADLINE` and `SCHED_EXT` are now handled as parsed-yet-unsupported policies instead of looking like parse failures.

For those two cases, `sched_setscheduler` now emits a separate `warn_once_then_debug!` message before returning `EINVAL`, so the log makes it clear that the failure is due to unimplemented Shadow scheduler functionality rather than bad arguments from the managed process.

See commit "sched: inline unsupported policy handling".

sporksmith (approved):
Thanks, looks good!

PR #3745, opened 2026-04-20
prctl: emulate parent death signals

starius:
Handle PR_SET_PDEATHSIG and PR_GET_PDEATHSIG in Shadow instead of rejecting them. Store the configured signal in emulated process state so managed prctl calls no longer depend on the host kernel's parent-death signal setting.

When a parent process exits, deliver the configured signal to each child before reparenting it to INIT. This preserves the old parent PID in the generated siginfo payload while keeping the emulation separate from the shim's native PDEATHSIG used for cleanup.

Track children by pid while delivering parent-death signals so the exit path can signal each child before reparenting it without holding temporary child process references across the traversal.

The delivery model is intentionally process-granular: Shadow tracks parentage per process, so this approximates Linux's parent-thread semantics closely enough for process supervisors without claiming exact thread-level behavior. Add regression tests that cover set/get/clear of the parent-death signal and delivery to a blocked child with the correct si_pid.

This is another patch needed for PostgreSQL to work properly in Shadow. What we saw before that patch:

  - PostgreSQL startup failed with
    `FATAL: could not request parent death signal: Invalid argument`
  - Shadow logged
    `Not allowing unimplemented prctl PR_SET_PDEATHSIG`

After adding the patch, that blocker went away and Postgres could start, assuming the other Postgres-specific fixes were also in place.

### DIFF
--- a/src/main/host/host.rs
+++ b/src/main/host/host.rs
@@ -475,8 +475,7 @@ impl Host {
             return;
         }
 
-        // Reparent children, and collect IDs of children that are dead.
-        let mut orphaned_zombie_pids: Vec<ProcessId> = self
+        let child_pids: Vec<_> = self
             .processes
             .borrow()
             .iter()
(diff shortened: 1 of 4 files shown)


sporksmith:
Very nice! This looks good to me, assuming that tests pass. (I just hit retry after hitting the flaky epoll test ...)

Does anyone else want a look before merging?

It looks like you need

sporksmith:
> Very nice! This looks good to me, assuming that tests pass. (I just hit retry after hitting the flaky epoll test ...)
> 
> Does anyone else want a look before merging?
> 
> It looks like you need

Oops, forgot to finish this - it looks like you need to run rustfmt to make the lints pass. Feel free to push another commit to fix it, but please don't rebase/squash anything yet.

starius:
Thanks for looking into it! I ran `cargo fmt` and committed it separately.

starius:
[The CI failure](https://github.com/shadow/shadow/actions/runs/24945895909/job/73230158110?pr=3745) is unrelated to the PR.

I identified the root cause and fixed it in [a separate PR](https://github.com/shadow/shadow/pull/3750).

sporksmith (approved):

PR #3748, opened 2026-04-27
linux-api: Bump to linux 7.0

sporksmith:
Also fix gen-kernel-bindings.sh to ensure we format with the rustfmt matching the current effective (e.g. pinned) cargo version.


src/lib/linux-api/gen-kernel-bindings.sh
```diff
@@ -88,7 +88,13 @@ do_bindgen() {
...
   bindgen "${bindgen_flags[@]}"
   ./rename.py < "$BUILDDIR/$output_basename" > src/"$output_basename"
-  rustfmt src/"$output_basename"
+  # don't call `rustfmt` directly; we want to go through cargo to ensure
+  # we use the version matching the pinned toolchain.
+  #
+  # Unfortunately `cargo --fmt` doesn't seem to provide a way to format
+  # a specific file, and we don't want this script to format the whole repo.
+  # Just check instead and warn the user.
+  cargo fmt --check || echo 'You will need to run "cargo fmt"'
```
stevenengler, line 96: nit: `cargo --fmt` -> `cargo fmt`

And I think the 'rustfmt' call does use the pinned toolchain? For example for a rustup install:

```bash
$ which rustfmt
~/.cargo/bin/rustfmt
$ ls -l ~/.cargo/bin/rustfmt
lrwxrwxrwx. 1 user user 29 Mar  4  2025 /home/user/.cargo/bin/rustfmt -> /home/user/.cargo/bin/rustup
```

where I think the rustup binary should use the correct toolchain. In my install I can also do for example `rustfmt +nightly main/main.rs`. Maybe the rustfmt isn't being run from the shadow directory, so rustup isn't using the toolchain file?

But either the `rustfmt` or `cargo fmt --check` is good with me.

src/lib/linux-api/src/bindings.rs
```diff
@@ -357,6 +357,8 @@ pub const LINUX_RWH_WRITE_LIFE_MEDIUM: u32 = 3;
 pub const LINUX_RWH_WRITE_LIFE_LONG: u32 = 4;
 pub const LINUX_RWH_WRITE_LIFE_EXTREME: u32 = 5;
 pub const LINUX_RWF_WRITE_LIFE_NOT_SET: u32 = 0;
+pub const LINUX_F_GETDELEG: u32 = 1039;
+pub const LINUX_F_SETDELEG: u32 = 1040;
```
stevenengler, line 363: Can we add these to `FcntlCommand`, since it's an enum and can't handle unexpected fcntl ops?

src/lib/linux-api/src/bindings.rs
```diff
@@ -381,6 +388,8 @@ pub const LINUX_AT_EACCESS: u32 = 512;
 pub const LINUX_AT_REMOVEDIR: u32 = 512;
 pub const LINUX_AT_HANDLE_FID: u32 = 512;
 pub const LINUX_AT_HANDLE_MNT_ID_UNIQUE: u32 = 1;
+pub const LINUX_AT_HANDLE_CONNECTABLE: u32 = 2;
+pub const LINUX_AT_EXECVE_CHECK: u32 = 65536;
```
stevenengler, line 394: Can you add `LINUX_AT_EXECVE_CHECK` to `ExecveAtFlags`?

src/lib/linux-api/src/bindings.rs
```diff
@@ -915,6 +936,19 @@ pub const LINUX_PR_PPC_DEXCR_CTRL_CLEAR: u32 = 4;
 pub const LINUX_PR_PPC_DEXCR_CTRL_SET_ONEXEC: u32 = 8;
 pub const LINUX_PR_PPC_DEXCR_CTRL_CLEAR_ONEXEC: u32 = 16;
 pub const LINUX_PR_PPC_DEXCR_CTRL_MASK: u32 = 31;
+pub const LINUX_PR_GET_SHADOW_STACK_STATUS: u32 = 74;
+pub const LINUX_PR_SET_SHADOW_STACK_STATUS: u32 = 75;
+pub const LINUX_PR_SHADOW_STACK_ENABLE: u32 = 1;
+pub const LINUX_PR_SHADOW_STACK_WRITE: u32 = 2;
+pub const LINUX_PR_SHADOW_STACK_PUSH: u32 = 4;
+pub const LINUX_PR_LOCK_SHADOW_STACK_STATUS: u32 = 76;
```
stevenengler, line 946: Can you add `PR_GET_SHADOW_STACK_STATUS`, `PR_SET_SHADOW_STACK_STATUS`, and `PR_LOCK_SHADOW_STACK_STATUS` to `PrctlOp`?

src/lib/linux-api/src/bindings.rs
```diff
@@ -915,6 +936,19 @@ pub const LINUX_PR_PPC_DEXCR_CTRL_CLEAR: u32 = 4;
...
+pub const LINUX_PR_TIMER_CREATE_RESTORE_IDS: u32 = 77;
+pub const LINUX_PR_TIMER_CREATE_RESTORE_IDS_OFF: u32 = 0;
+pub const LINUX_PR_TIMER_CREATE_RESTORE_IDS_ON: u32 = 1;
+pub const LINUX_PR_TIMER_CREATE_RESTORE_IDS_GET: u32 = 2;
+pub const LINUX_PR_FUTEX_HASH: u32 = 78;
```
stevenengler, line 951: And `PR_FUTEX_HASH` to `PrctlOp`?

src/lib/linux-api/src/bindings.rs
```diff
@@ -1810,6 +1851,14 @@ pub const LINUX___NR_lsm_get_self_attr: u32 = 459;
...
 pub const LINUX___NR_lsm_list_modules: u32 = 461;
 pub const LINUX___NR_mseal: u32 = 462;
+pub const LINUX___NR_setxattrat: u32 = 463;
+pub const LINUX___NR_getxattrat: u32 = 464;
+pub const LINUX___NR_listxattrat: u32 = 465;
+pub const LINUX___NR_removexattrat: u32 = 466;
+pub const LINUX___NR_open_tree_attr: u32 = 467;
+pub const LINUX___NR_file_getattr: u32 = 468;
+pub const LINUX___NR_file_setattr: u32 = 469;
+pub const LINUX___NR_listns: u32 = 470;
```
stevenengler, line 1870: And these to `SyscallNum`? And `LINUX___NR_uprobe` (I highlighted it when I added a comment, but GitHub seems to have ignored it.)

src/lib/linux-api/src/bindings.rs
```diff
@@ -247,6 +248,7 @@ pub const LINUX_EALREADY: u32 = 114;
 pub const LINUX_EINPROGRESS: u32 = 115;
 pub const LINUX_ESTALE: u32 = 116;
 pub const LINUX_EUCLEAN: u32 = 117;
+pub const LINUX_EFSCORRUPTED: u32 = 117;
```
stevenengler, line 251: Can we add these two to `Errno`? Edit: The other was `LINUX_EFSBADCRC`.

src/lib/linux-api/src/bindings.rs
```diff
@@ -949,6 +951,13 @@ pub const LINUX_PR_TIMER_CREATE_RESTORE_IDS_GET: u32 = 2;
 pub const LINUX_PR_FUTEX_HASH: u32 = 78;
 pub const LINUX_PR_FUTEX_HASH_SET_SLOTS: u32 = 1;
 pub const LINUX_PR_FUTEX_HASH_GET_SLOTS: u32 = 2;
+pub const LINUX_PR_RSEQ_SLICE_EXTENSION: u32 = 79;
+pub const LINUX_PR_RSEQ_SLICE_EXTENSION_GET: u32 = 1;
+pub const LINUX_PR_RSEQ_SLICE_EXTENSION_SET: u32 = 2;
+pub const LINUX_PR_RSEQ_SLICE_EXT_ENABLE: u32 = 1;
+pub const LINUX_PR_GET_CFI: u32 = 80;
+pub const LINUX_PR_SET_CFI: u32 = 81;
```
stevenengler, line 959: And these two to `PrctlOp`?

**Edit:** And `PR_RSEQ_SLICE_EXTENSION`.

src/lib/linux-api/src/bindings.rs
```diff
@@ -1859,6 +1868,7 @@ pub const LINUX___NR_open_tree_attr: u32 = 467;
 pub const LINUX___NR_file_getattr: u32 = 468;
 pub const LINUX___NR_file_setattr: u32 = 469;
 pub const LINUX___NR_listns: u32 = 470;
+pub const LINUX___NR_rseq_slice_yield: u32 = 471;
```
stevenengler, line 1871: And another syscall :)

stevenengler (approved):
Sorry for asking for all the additions, I didn't think there would be so many that Shadow already knows about. But I think it's worth adding these incrementally as we go.

src/lib/linux-api/gen-kernel-bindings.sh
sporksmith: Hmm, yeah I was having trouble with a mismatched cargo-fmt getting invoked somehow, but maybe that wasn't the problem I was hitting here.

I verified locally via strace that `rustfmt` ends up invoking a `rustfmt` bin from the correct toolchain.

However, the output doesn't pass `cargo fmt --check`, for me.

Using strace again, I see that `cargo fmt` ends up invoking `cargo-fmt`. Apparently it *doesn't* end up invoking the `rustfmt` bin, and apparently the two don't necessarily agree with eachother.

Tweaked this to still try running `rustfmt`, before warning if the output doesn't pass `cargo fmt --check`. Locally I get the warning and have to run `cargo fmt`, but maybe this is something specific to my environment or is a temporary bug in the rust tools.

src/lib/linux-api/src/bindings.rs
sporksmith: done

src/lib/linux-api/src/bindings.rs
sporksmith: done

src/lib/linux-api/src/bindings.rs
sporksmith: done

src/lib/linux-api/src/bindings.rs
sporksmith: done

src/lib/linux-api/src/bindings.rs
sporksmith: done

src/lib/linux-api/src/bindings.rs
sporksmith: done

src/lib/linux-api/src/bindings.rs
sporksmith: done

src/lib/linux-api/src/bindings.rs
sporksmith: done

PR #3750, opened 2026-04-27
epoll: avoid oneshot test race

starius:
Replace the sleep-based sequencing in `test_oneshot_multi_write` with explicit synchronization between the waiter thread and the writer thread.

The old test assumed the second epoll_wait had already timed out before the main thread rearmed the EPOLLONESHOT registration. Under scheduler delay that assumption can fail: the second wait may still be blocked when EPOLL_CTL_MOD rearms the interest and the following write arrives, causing the second wait to observe an event instead of timing out. This matches the CI failure in `epoll-edge-rs-linux` on the unix-dgram case.

Coordinate each phase explicitly so that the second wait has definitely finished before the rearm, and the third wait is the one that observes the post-rearm event. This removes the timing race without changing the behavior being tested.

---

This fixes the flake that happened here: https://github.com/shadow/shadow/actions/runs/24945895909/job/73230158110?pr=3745

To repro the flake reliably, use this patch:
```diff
diff --git a/src/test/epoll/test_epoll_edge.rs b/src/test/epoll/test_epoll_edge.rs
index c245a55c1..b9797e65a 100644
--- a/src/test/epoll/test_epoll_edge.rs
+++ b/src/test/epoll/test_epoll_edge.rs
@@ -219,7 +219,13 @@ fn test_oneshot_multi_write(readfd: libc::c_int, writefd: libc::c_int) -> anyhow
         let thread = std::thread::spawn(move || {
             vec![
                 do_epoll_wait(epollfd, timeout, /* do_read= */ false),
-                do_epoll_wait(epollfd, timeout, /* do_read= */ false),
+                // Control-only delay to reproduce the race where the second
+                // wait is still blocked when the main thread rearms the
+                // oneshot registration and performs the final write.
+                {
+                    std::thread::sleep(timeout);
+                    do_epoll_wait(epollfd, timeout, /* do_read= */ false)
+                },
                 do_epoll_wait(epollfd, timeout, /* do_read= */ false),
             ]
         });

```

### DIFF
--- a/src/test/epoll/test_epoll_edge.rs
+++ b/src/test/epoll/test_epoll_edge.rs
@@ -161,12 +161,16 @@ fn test_threads_multi_write(readfd: libc::c_int, writefd: libc::c_int) -> anyhow
         )?;
 
         let timeout = Duration::from_millis(100);
+        let (result_tx, result_rx) = std::sync::mpsc::channel();
 
-        let threads = [
-            std::thread::spawn(move || do_epoll_wait(epollfd, timeout, /* do_read= */ false)),
-            std::thread::spawn(move || do_epoll_wait(epollfd, timeout, /* do_read= */ false)),
-            std::thread::spawn(move || do_epoll_wait(epollfd, timeout, /* do_read= */ false)),
-        ];
+        let threads: [std::thread::JoinHandle<()>; 3] = std::array::from_fn(|_| {
+            let result_tx = result_tx.clone();
+            std::thread::spawn(move || {
+                let result = do_epoll_wait(epollfd, timeout, /* do_read= */ false);
+                result_tx.send(result).unwrap();
+            })
+        });
+        drop(result_tx);
 
         // Wait for readers to block.
         std::thread::sleep(timeout / 3);
@@ -174,11 +178,24 @@ fn test_threads_multi_write(readfd: libc::c_int, writefd: libc::c_int) -> anyhow
         // Make the read-end readable.
         unistd::write(writefd, &[0])?;
 
-        // Wait again and make the read-end readable again.
-        std::thread::sleep(timeout / 3);
+        // Wait for the first edge to be consumed, then drain the byte before
+        // generating the next edge. Otherwise the second write can happen
+        // while the fd is still readable and fail to create another wakeup.
+        let first_result = result_rx.recv().unwrap();
+        unistd::read(readfd, &mut [0])?;
+
+        // Make the read-end readable again.
         unistd::write(writefd, &[0])?;
 
-        let mut results = threads.map(|t| t.join().unwrap());
+        let mut results = [
+            first_result,
+            result_rx.recv().unwrap(),
+            result_rx.recv().unwrap(),
+        ];
+
+        for thread in threads {
+            thread.join().unwrap();
+        }
 
         // Two of the threads should have gotten an event, but we don't know which one.
         // Sort results by number of events received.
@@ -257,7 +274,7 @@ fn test_oneshot_multi_write(readfd: libc::c_int, writefd: libc::c_int) -> anyhow
 
         drop(write_requested_tx);
         thread.join().unwrap()?;
-        let results = vec![results0, results1, results2];
+        let results = [results0, results1, results2];
 
         // The first wait should have received the event
         ensure_ord!(results[0].epoll_res, ==, Ok(1));
--- a/src/test/epoll/test_epoll_edge.rs
+++ b/src/test/epoll/test_epoll_edge.rs
@@ -215,37 +215,35 @@ fn test_oneshot_multi_write(readfd: libc::c_int, writefd: libc::c_int) -> anyhow
         )?;
 
         let timeout = Duration::from_millis(100);
-        let (wait_started_tx, wait_started_rx) = std::sync::mpsc::channel();
-        let (wait_finished_tx, wait_finished_rx) = std::sync::mpsc::channel();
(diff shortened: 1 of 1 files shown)


sporksmith:
Oh excellent, I was just attempting a [hackier fix](https://github.com/shadow/shadow/pull/3751) for this. @stevenengler are you able to take a look?

sporksmith (approved):
Actually, this looks sensible to me. I'm not sure it's the *only* source of the flakes we've been seeing in this test; we might still want to increase the timeouts as in the other PR. But let's go ahead and merge this.

Could you run cargo fmt and push another commit?

sporksmith:
> I'm not sure it's the only source of the flakes we've been seeing in this test;

Oops, indeed it failed again. Looking a bit more to understand what's going on. In either case please go ahead and push a "cargo fmt" commit.

src/test/epoll/test_epoll_edge.rs
```diff
@@ -215,28 +215,41 @@ fn test_oneshot_multi_write(readfd: libc::c_int, writefd: libc::c_int) -> anyhow
...
+                wait_once(),
             ]
         });
 
-        // Wait for readers to block.
-        std::thread::sleep(timeout / 3);
+        // Coordinate each phase explicitly so that the second wait has already
+        // timed out before we rearm the oneshot registration. Using fixed
+        // sleeps here is racy under scheduler delays.
+        wait_started_rx.recv().unwrap();
```
sporksmith: Actually, there's still a race here. The reader thread could have sent on the channel but not actually started the epoll. I think this change actually makes the race tighter.

sporksmith (changes requested):
Un-approving. I don't think this works, as noted in the other comment thread.

src/test/epoll/test_epoll_edge.rs
starius: I ran `cargo fmt`.

You're right about the channel-only version. It sequenced the phases, but it didn't actually prove the waiter had entered `epoll_wait`, so there was still a start-side gap there.

I fixed that by changing the shape of the test instead of trying to infer thread state. `test_oneshot_multi_write` now keeps all three `epoll_wait` calls on one thread, and uses a helper thread only for the writes. Each phase is now ordered as: perform the write, wait for the write to complete, then run the corresponding wait locally. The rearm is still strictly after `wait2`, so `wait2` can't steal the post-rearm event anymore.

That gets rid of the remaining race for the behavior this test is checking, and it also avoids the procfs/thread-state path that doesn't work cleanly under Shadow. I verified it with 10 repeated runs.

starius:
Pushed another pack of fixes.

src/test/epoll/test_epoll_edge.rs
stevenengler: > That gets rid of the remaining race for the behavior this test is checking, and it also avoids the procfs/thread-state path that doesn't work cleanly under Shadow. I verified it with 10 repeated runs.

What doesn't work cleanly under shadow? Isn't it just the -linux tests that are failing?

**Edit:** Ah I think you're referring to @sporksmith's comment in https://github.com/shadow/shadow/pull/3751#issuecomment-4330418318.

src/test/epoll/test_epoll_edge.rs
```diff
@@ -215,27 +232,34 @@ fn test_oneshot_multi_write(readfd: libc::c_int, writefd: libc::c_int) -> anyhow
...
+        let request_write = || -> anyhow::Result<()> {
+            write_requested_tx.send(()).unwrap();
+            write_finished_rx.recv().unwrap();
+            Ok(())
+        };
 
         // Make the read-end readable.
-        unistd::write(writefd, &[0])?;
+        request_write()?;
+        let results0 = do_epoll_wait(epollfd, timeout, /* do_read= */ false);
```
stevenengler, line 258: I still don't think that this does what the original test was doing. The original test wanted to perform the write() after the epoll_wait() had started, in part so that we can test the wakeup behaviour of the syscall. Here we request the write and wait for the write to finish before we do the epoll_wait(), so now there's no wakeup at all. And in this case, there's no need for threads either since we're serializing everything.

There are two downsides to using an mpsc queue like this:
1. There's no way to notify after the epoll_wait() has started.
2. These tests are designed to test epoll. By adding an mpsc queue, we're now introducing the futex() subsystem in to these tests. That's not necessarily a bad thing if it's the only option, but when the test does fail in Shadow, it makes it more confusing as to whether it's an epoll issue or a futex issue. So I think we should only add these things if there are no other options. But this may be a controversial opinion because it's hard to avoid races without synchronization primitives.

Unless I'm missing something, using a much larger timeout seems to me like the better approach for now.

starius: I agree, my PR narrows the thing being tested.

The key distinction is:

  - We can make `test_oneshot_multi_write` deterministic about the EPOLLONESHOT invariant: "wait2 must finish before rearm, so it can't consume the post-rearm event."
  - We cannot, with mpsc, atomics, or any other normal userspace sync, prove that epoll_wait() has already entered the kernel before the write happens.

[The merged fix](https://github.com/shadow/shadow/pull/3751) seems to work under normal circumstances. It still depends on timing and can break, if CI is slow enough. I propose to close my PR for now. Maybe in the future there is a better approach for the test, which would be fully deterministic and testing the right thing and not sensitive to timing.

sporksmith: sounds good; thanks for the attempt!

Issue #3410, opened 2024-09-12
`epoll-edge-rs-linux` test fails


stevenengler:
```text
"epoll-edge-rs-linux" start time: Sep 12 18:59 UTC
Output:
----------------------------------------------------------
Testing multi-write <fds="tcp",swapped=false>... ✓
Testing oneshot-multi-write <fds="tcp",swapped=false>... ✓
Testing write-then-partial-read <fds="tcp",swapped=false>... ✓
[...]
Testing oneshot-multi-write <fds="pipe">... ✓
Testing write-then-partial-read <fds="pipe">... ✓
Testing threads-multi-write <fds="pipe">... ✓
Testing multi-write <fds="pipe-direct">... ✓
Testing oneshot-multi-write <fds="pipe-direct">... ✗ (EAGAIN: Try again)
Error: EAGAIN: Try again

Stack backtrace:
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.86/src/backtrace.rs:27:14
   1: <core::result::Result<T,F> as core::ops::try_trait::FromResidual<core::result::Result<core::convert::Infallible,E>>>::from_residual
             at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/core/src/result.rs:1989:27
   2: test_epoll_edge::test_oneshot_multi_write::{{closure}}
   3: test_utils::run_and_close_fds
             at /shared/code/shadow/fix-packet-panic/src/test/test_utils.rs:166:14
   4: test_epoll_edge::test_oneshot_multi_write
             at /shared/code/shadow/fix-packet-panic/src/test/epoll/test_epoll_edge.rs:206:5
   5: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/alloc/src/boxed.rs:2084:9
   6: test_utils::ShadowTest<T,E>::run
             at /shared/code/shadow/fix-packet-panic/src/test/test_utils.rs:56:9
   7: test_utils::run_tests
             at /shared/code/shadow/fix-packet-panic/src/test/test_utils.rs:80:15
   8: test_epoll_edge::main
             at /shared/code/shadow/fix-packet-panic/src/test/epoll/test_epoll_edge.rs:577:5
   9: core::ops::function::FnOnce::call_once
[...]

PR #3751, opened 2026-04-27
test_epoll_edge: increase timeouts

sporksmith:
We're getting unexpected timeouts in the GitHub CI. Possibly the underlying hardware is overloaded.

Let's try increasing the timeouts.

See https://github.com/shadow/shadow/issues/3410

### DIFF
--- a/src/test/epoll/epoll-edge-rs.yaml
+++ b/src/test/epoll/epoll-edge-rs.yaml
@@ -1,5 +1,5 @@
 general:
-  stop_time: 10
+  stop_time: 20
 network:
   graph:
     type: 1_gbit_switch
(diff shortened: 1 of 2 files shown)


sporksmith:
As https://github.com/shadow/shadow/pull/3750 points out, this is still fundamentally racy, but I don't think we can do better than giving more time padding, short of inspecting thread state to work out when the thread is actually blocked.

starius:
I made another attempt to fix the race in https://github.com/shadow/shadow/pull/3750
Please take a look :)

stevenengler (approved):

sporksmith:
Going ahead and merging this to at least mitigate the issue. Great if we eventually also come up with a way to more fundamentally avoid the race :)

PR #3753, opened 2026-04-30
Reorganize the edge-triggered epoll tests to try to reduce timing-related flakiness

stevenengler:
Part of #3410.

We've been having issues with these tests running slowly and exceeding their expected duration in CI. I think what exacerbates the situation is that for tests running two parallel threads, we don't have any synchronization between the two threads after each epoll wait. So the timing of the threads gradually becomes more out-of-sync over time.

This commit launches each epoll wait in a new thread so that the epoll wait in one thread and the sleep in the other thread have a common reference point.

We could have alternatively used some synchronization primitive to synchronize the threads, but I think it's better to rely on the thread start/stop synchronization instead of introducing futex operations to the test.

There's still a race between the epoll waits in one thread and the writes in the other thread. This also shortens the sleeps in all tests to hopefully reduce how often things approach the timeout.

### DIFF
--- a/src/test/epoll/test_epoll_edge.rs
+++ b/src/test/epoll/test_epoll_edge.rs
@@ -65,28 +65,37 @@ fn test_multi_write(readfd: libc::c_int, writefd: libc::c_int) -> anyhow::Result
 
         let timeout = Duration::from_millis(200);
 
-        let thread = std::thread::spawn(move || {
-            vec![
-                do_epoll_wait(epollfd, timeout, /* do_read= */ false),
-                do_epoll_wait(epollfd, timeout, /* do_read= */ false),
-                // The last one is supposed to timeout.
-                do_epoll_wait(epollfd, timeout, /* do_read= */ false),
-            ]
-        });
+        // Results from the epoll waits.
+        let mut results = Vec::new();
 
         ////// First epoll wait.
+
+        let thread =
+            std::thread::spawn(move || do_epoll_wait(epollfd, timeout, /* do_read= */ false));
 
         // Wait for the epoll wait to block and make the read-end readable.
         std::thread::sleep(timeout / 3);
         unistd::write(writefd, &[0])?;
 
+        results.push(thread.join().unwrap());
+
         ////// Second epoll wait.
+
+        let thread =
+            std::thread::spawn(move || do_epoll_wait(epollfd, timeout, /* do_read= */ false));
 
         // Wait again and make the read-end readable again.
         std::thread::sleep(timeout / 3);
         unistd::write(writefd, &[0])?;
 
-        let results = thread.join().unwrap();
+        results.push(thread.join().unwrap());
+
+        ////// Third epoll wait (should time out).
+
+        let thread =
+            std::thread::spawn(move || do_epoll_wait(epollfd, timeout, /* do_read= */ false));
+
+        results.push(thread.join().unwrap());
 
         ////// Check results.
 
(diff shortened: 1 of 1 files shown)


stevenengler:
I ran the CI tests 5 times. The first 4 passed and the 5th failed at:

```text
Output:
----------------------------------------------------------
Testing multi-write <fds="tcp",swapped=false>... ✓
Testing oneshot-multi-write <fds="tcp",swapped=false>... ✓
Testing write-then-partial-read <fds="tcp",swapped=false>... ✓
Testing threads-multi-write <fds="tcp",swapped=false>... ✓
Testing multi-write <fds="tcp",swapped=true>... ✓
Testing oneshot-multi-write <fds="tcp",swapped=true>... ✓
Testing write-then-partial-read <fds="tcp",swapped=true>... ✓
Testing threads-multi-write <fds="tcp",swapped=true>... ✓
Testing multi-write <fds="udp">... ✓
Testing oneshot-multi-write <fds="udp">... ✗ (!(Ok(0) == Ok(1)))
Error: !(Ok(0) == Ok(1))
```

specifically at `test_epoll_edge.rs:291` which is:

```rust
// The first wait should have received the event
ensure_ord!(results[0].epoll_res, ==, Ok(1)); // This line here
ensure_ord!(results[0].duration, <, timeout);
ensure_ord!(results[0].events[0], ==, epoll::EpollEvent::new(EpollFlags::EPOLLIN, 0));
```

So we start a thread (which should immediately epoll-wait), sleep for 40 ms, then send a UDP message over loopback. It seems strange that within the 200 ms timeout the epoll-wait is never notified. I wonder if this isn't really just "the CI is busy so we exceed the timeout", or if this is a similar situation to #3546 where UDP messages have an unexpectedly long delay.

**Edit:** Now a different failure:

```text
Output:
----------------------------------------------------------
Testing multi-write <fds="tcp",swapped=false>... ✓
Testing oneshot-multi-write <fds="tcp",swapped=false>... ✓
Testing write-then-partial-read <fds="tcp",swapped=false>... ✓
Testing threads-multi-write <fds="tcp",swapped=false>... ✓
Testing multi-write <fds="tcp",swapped=true>... ✓
Testing oneshot-multi-write <fds="tcp",swapped=true>... ✓
Testing write-then-partial-read <fds="tcp",swapped=true>... ✓
Testing threads-multi-write <fds="tcp",swapped=true>... ✓
Testing multi-write <fds="udp">... ✓
Testing oneshot-multi-write <fds="udp">... ✓
Testing write-then-partial-read <fds="udp">... ✓
Testing threads-multi-write <fds="udp">... ✓
Testing multi-write <fds="unix-stream",swapped=false>... ✓
Testing oneshot-multi-write <fds="unix-stream",swapped=false>... ✓
Testing write-then-partial-read <fds="unix-stream",swapped=false>... ✓
Testing threads-multi-write <fds="unix-stream",swapped=false>... ✓
Testing multi-write <fds="unix-stream",swapped=true>... ✓
Testing oneshot-multi-write <fds="unix-stream",swapped=true>... ✓
Testing write-then-partial-read <fds="unix-stream",swapped=true>... ✓
Testing threads-multi-write <fds="unix-stream",swapped=true>... ✗ (!(Ok(0) == Ok(1)))
Error: !(Ok(0) == Ok(1))
```

at `test_epoll_edge.rs:213`.

So doesn't look like a UDP issue after all. If it is that the CI is overloaded, I wonder why it's the edge epoll tests specifically when we have lower timeouts in other epoll and socket-related tests.

stevenengler:
@sporksmith Even though this doesn't seem to solve the problem, lmk whether you think these changes seem useful anyways. I think it's still worthwhile, but I don't feel too strongly about it.

stevenengler:
Closing since I changed my mind. It could be that spawning threads itself is slow, so for all I know this could be making it worse.

PR #3759, opened 2026-05-19
linux-api: add bindings for types needed for shm syscalls

sporksmith:


sporksmith:
cc @starius  - let me know if you see anything wrong or missing that you'd need for https://github.com/shadow/shadow/pull/3743 to replace the libc types

starius:
@sporksmith 

Thanks! I cherry-picked this onto #3743 locally, changed the shm implementation to use the new linux-api wrappers where they fit, and rebuilt/tested that actual code.

The pieces that work well are `ShmgetFlagsParts`/`ShmgetFlags`, `ShmatFlags`, and `ShmctlCmd`. I was able to replace the direct libc constants for `IPC_CREAT`, `IPC_EXCL`, `SHM_RDONLY`, `IPC_STAT`, and `IPC_RMID` with those, and also stop using `libc::key_t` internally.

Some things which you could do in your PR:

1. Could you do some renames before merging this PR? In particular, `ShmAtFlags` -> `ShmatFlags` and `ShmCtlCmd` -> `ShmctlCmd`, plus the test helper names from generic `shmflags`/`shmflgs` to `shmgetflags`/`shmgetflgs`. Those spellings match the syscall names (`shmat`, `shmctl`, `shmget`) and would avoid landing inconsistent public API names that my PR would then immediately need to code against or rename around.

2. Naming note from the local conversion: I used local `ShmKey = c_int` and `IPC_PRIVATE = 0` aliases instead of keeping `libc::key_t`/`libc::IPC_PRIVATE`, since this PR does not currently expose a key type or private-key constant. Small nice-to-have: a public `IPC_PRIVATE` constant and a public `kernel_key_t` alias would remove the last local workaround for the key type/private-key value. Not a blocker.

3. I also intentionally did not use the provided `linux_api::shm::shmid_ds` name in the `IPC_STAT` path because that name currently refers to the old UAPI layout; naming/exposing the needed layout as `shmid64_ds`/`ipc64_perm` would make the ABI distinction explicit and avoid accidentally using the obsolete struct. The remaining blocker for replacing all of the libc shm types is the `shmctl(IPC_STAT)` buffer. The currently exported `linux_api::shm::shmid_ds` is the old UAPI `struct shmid_ds`, not the 64-bit userspace/syscall layout that glibc's `struct shmid_ds` uses on x86_64. I checked the layouts locally: glibc `struct shmid_ds` and Linux `struct shmid64_ds` are 112 bytes with `shm_segsz` at offset 48 and `shm_nattch` at offset 88, while the exported Linux `struct shmid_ds` is 88 bytes with `shm_segsz` at offset 28 and `shm_nattch` at offset 64. If my PR writes IPC_STAT using the currently exported type, the test program reads the wrong fields.

  - Could you also bind/expose `struct shmid64_ds` and `struct ipc64_perm` from the installed asm headers? Either exposing them explicitly as `linux_api::shm::shmid64_ds`/`linux_api::ipc::ipc64_perm`, or making the Rust-facing `linux_api::shm::shmid_ds` alias the 64-bit ABI type, would let my PR replace `libc::shmid_ds` safely.

---

Side note. While validating the converted code, the Linux owner-perms test showed that my PR's test/implementation was too permissive: Linux allows `shmget(key, size, 0)` as a lookup on a `0000` segment, but denies `shmat(..., 0)` with `EACCES`. I fixed that locally as part of the conversion.

sporksmith:
> 1. Could you do some renames before merging this PR? In particular, `ShmAtFlags` -> `ShmatFlags` and `ShmCtlCmd` -> `ShmctlCmd`, plus the test helper names from generic `shmflags`/`shmflgs` to `shmgetflags`/`shmgetflgs`. Those spellings match the syscall names (`shmat`, `shmctl`, `shmget`) and would avoid landing inconsistent public API names that my PR would then immediately need to code against or rename around.

Done

> Small nice-to-have: a public IPC_PRIVATE constant and a public kernel_key_t alias would remove the last local workaround for the key type/private-key value.

Done

> Could you also bind/expose struct shmid64_ds and struct ipc64_perm from the installed asm headers? Either exposing them explicitly as linux_api::shm::shmid64_ds/linux_api::ipc::ipc64_perm, or making the Rust-facing linux_api::shm::shmid_ds alias the 64-bit ABI type, would let my PR replace libc::shmid_ds safely.

Good catch! Done. I kept the `64` names to stay consistent with the kernel code/headers, and removed the bindings for the legacy versions to help avoid confusion.

starius:
Thank you for the updates! I think the PR is in good shape for https://github.com/shadow/shadow/pull/3743

sporksmith:
@robgjansen @stevenengler friendly ping - could someone take a look? LMK if you'd like me to rebase away the fixups first

src/lib/linux-api/bindings-wrapper.h
```diff
@@ -10,6 +10,7 @@
 #include <linux/fcntl.h>
 #include <linux/futex.h>
 #include <linux/in.h>
+#include <linux/ipc.h>
```
stevenengler, line 13: Did this line do anything? The generated bindings look the same, just some stuff moved around.

src/lib/linux-api/src/shm.rs
```diff
@@ -0,0 +1,146 @@
...
+
+/// parameter to `shmctl` syscall.
+pub use bindings::linux_shmid_ds as shmid_ds;
+use num_enum::IntoPrimitive;
+use num_enum::TryFromPrimitive;
+
+bitflags::bitflags! {
+    /// Flags accepted by `shmget` syscall.
+    #[derive(Copy, Clone, Debug, Default, Eq, PartialEq)]
+    pub struct ShmgetFlags: i32 {
```
stevenengler, line 31: Maybe add a comment here that this represents the flags, but not the `shmflg` argument specifically, and that the actual argument is represented by `ShmgetFlagsParts`?

src/lib/linux-api/src/shm.rs
```diff
@@ -0,0 +1,146 @@
...
+            flags: ShmgetFlags::from_bits_retain(shmflg & !tlb_size_mask & !perms_mask),
+        }
+    }
+
+    /// Convert to a `shmflg` value, suitable for `shmget` syscall param.
+    pub fn to_shmflg(self) -> c_int {
+        self.flags.bits()
+            | (self.tlb_size_log_2 << SHM_HUGE_SHIFT) as i32
+            | self.perms.bits() as i32
+    }
```
stevenengler, line 82: I think this should return an `Option` instead, and return `None` if any of the fields are invalid?

For example with a large `tlb_size_log_2` the left shift here would panic in debug builds, or wrap around (I think) in release builds which would mess with the flags/perms.

Or `perms.bits()` could overwrite bits outside of the 9 least significant bits. Or `flags.bits()` could overwrite the perms/page sizes.

src/lib/linux-api/src/shm.rs
```diff
@@ -0,0 +1,146 @@
...
+impl ShmgetFlagsParts {
+    /// Split a `shmflg` value, as provided to the `shmget` syscall, into consituent parts.
+    pub fn from_shmflg(shmflg: c_int) -> Self {
+        let tlb_size_shift = SHM_HUGE_SHIFT;
+        let tlb_size_mask = 0b11_1111 << tlb_size_shift;
+
+        let perms_shift = 0;
+        let perms_mask = 0b1_1111_1111;
+        Self {
+            tlb_size_log_2: (shmflg & tlb_size_mask) as u32 >> tlb_size_shift,
```
stevenengler, line 64: Can you add a comment here that the u32 cast is important before the right shift so that it doesn't fill in with ones? I always get confused about shifting signed integers, so having a comment would be helpful.

src/lib/linux-api/src/ipc.rs
```diff
@@ -1,5 +1,9 @@
 use crate::bindings;
+use crate::posix_types::kernel_key_t;
 
 /// Specifies permissions for an ipc object. Used in `shmid_ds`, which is used
 /// by the `shmctl` syscall.
 pub use bindings::linux_ipc_perm as ipc_perm;
+
+// Defined in linux/ipc.h, but not handled by bindgen.
+pub const IPC_PRIVATE: kernel_key_t = 0;
```
stevenengler: Does this mean we haven't included it with bindgen, or that bindgen isn't picking it up properly?

src/lib/linux-api/bindings-wrapper.h
stevenengler: I guess maybe it was already included by a different header. So if we're using it, then adding it explicitly here makes sense to me.

src/lib/linux-api/src/ipc.rs
sporksmith: bindgen isn't picking it up properly.

The source definition is in `include/linux/ipc.h`:

`#define IPC_PRIVATE ((__kernel_key_t) 0) `

Updated the comment to clarify.

src/lib/linux-api/src/shm.rs
sporksmith: good point; done

src/lib/linux-api/src/shm.rs
sporksmith: Done

src/lib/linux-api/src/shm.rs
sporksmith: Good point. The error-checking is pretty fiddly to write though, and I don't anticipate needing this outside of testing. Added a warning, renamed, and made this test-only.

src/lib/linux-api/bindings-wrapper.h
sporksmith: Right

src/lib/linux-api/src/posix_types.rs
```diff
@@ -38,6 +38,10 @@ pub use bindings::linux___kernel_gid32_t;
 #[allow(non_camel_case_types)]
 pub type kernel_gid32_t = linux___kernel_gid32_t;
 
+pub use bindings::linux___kernel_key_t;
+#[allow(non_camel_case_types)]
+pub type kernel_key_t = linux___kernel_key_t;
```
stevenengler, line 43: I don't remember, is there a reason we export the same type with two different names (`linux___kernel_key_t` and `kernel_key_t`) in some places but not others? I don't think anything needs to change, since it's better to be consistent with the surrounding code. But I'm wondering if there's a reason I have forgotten.

stevenengler (approved):

src/lib/linux-api/src/posix_types.rs
sporksmith: We use the `linux__` types when re-exposing to C, so that they don't conflict with the libc or kernel types in other C headers. Documented here: https://github.com/shadow/shadow/blob/main/src/lib/linux-api/src/lib.rs#L10

PR #3762, opened 2026-06-02
Implement utimensat null-path handling

### LINKED ISSUE #3761: utimensat (utimensat_time64) doesn't handle NULL pathname correctly
From [utimensat(2)](https://www.man7.org/linux/man-pages/man2/utimensat.2.html):

> the Linux utimensat() system call implements a nonstandard feature: if path is NULL, then the call modifies the timestamps of the file referred to by the file descriptor dirfd (which may refer to any type of file).

Our implementation ends up returning `EFAULT` when the path is `NULL`. ~~I think this happens in our syscall machinery without ever calling the handler, since the function signature expects a (non-NULL) `SyscallStringArg`.~~ (EDIT: actually that's just the "formatting" signature. The actual signature uses simple types. We're rejecting the NULL in the C function `syscallhandler_utimensat`) 

This causes an error in the `filetime` crate version 0.2.29, blocking arti from updating to it: https://gitlab.torproject.org/tpo/core/arti/-/work_items/2562

Side-note: there's also some interposition naming confusion. We interpose `utimensat` at the libc layer, mapping it to the `utimensat` syscall. IIUC, libc on a 64-bit system should actually make the `utimensat_time64` syscall. This ends up mostly working out because we implement the `utimensat` syscall handler as if it were `utimensat_time64`, and largely re-delegate to the libc library call.

sporksmith:
Fixes #3761


src/lib/shadow-shim-helper-rs/src/syscall_types.rs
```diff
@@ -560,3 +560,13 @@ impl TryFrom<SyscallReg> for linux_api::fcntl::FcntlCommand {
...
     }
 }
+
+mod export {
+    use super::*;
+
+    /// Returns whether `ptr` is NULL.
+    #[unsafe(no_mangle)]
+    extern "C-unwind" fn untypedforeignpointer_is_null(ptr: UntypedForeignPtr) -> bool {
+        ptr.val == 0
```
stevenengler: nit: Can this be just `ptr.is_null()`?

src/main/host/descriptor/regular_file.c
```diff
@@ -1197,10 +1197,27 @@ int regularfile_utimensat(RegularFile* dir, const char* pathname, const struct t
         debug("Failed to get OS fd for 'RegularFile' in 'regularfile_utimensat'");
         return -EINVAL;
     }
+    trace("RegularFile %p utimesat os-backed file %i", dir, osFd);
```
stevenengler: nit: "utimesat" -> "utimensat"

**Edit:** I see this line was only moved. But maybe would still be good to fix.

stevenengler (approved):
LGTM!

src/lib/shadow-shim-helper-rs/src/syscall_types.rs
sporksmith: done

src/main/host/descriptor/regular_file.c
sporksmith: done

PR #3765, opened 2026-06-04
patch_vdso: no malloc

### LINKED ISSUE #3763: shim: jemalloc-using managed programs self-deadlock at startup — patch_vdso reads /proc/self/maps via fopen()/malloc(), re-entering the program's allocator
# shim: jemalloc-using managed programs self-deadlock at startup — `patch_vdso` reads `/proc/self/maps` via `fopen()`/`malloc()`, re-entering the program's allocator

## Summary

A dynamically-linked managed program that installs a global C `malloc` interposer — e.g. **jemalloc** built with `unprefixed_malloc_on_supported_platforms` (the default for Rust's `tikv-jemallocator`/`jemallocator`, and similarly tcmalloc/mimalloc) — **deadlocks during process startup under Shadow**, before `main`. The process hangs in `futex_wait` on the allocator's own init lock, makes no further syscalls, the simulation's virtual time freezes, and the run never progresses.

Root cause: the shim's `patch_vdso()` (run lazily on the first intercepted syscall) calls `_getVdsoBounds()`, which reads `/proc/self/maps` with `fopen()` + `malloc()` + `getline()`. When the managed program interposes the global `malloc`, that allocation re-enters the program's allocator **while it is mid-initialization holding a non-recursive lock** → self-deadlock.

This is the same hazard class Shadow already fixed in #2766 (shim `calloc` → `__libc_calloc` lock → `futex_wait`) and is tracking generally in #2919 ("avoid memory allocation functions… we can't be sure some library call doesn't allocate internally"). The fix is to make `_getVdsoBounds` allocation-free.

## Environment

- Shadow 3.3.0 (git `5b71013`, built 2026-03-07). The offending code (`src/lib/shim/patch_vdso.c:_getVdsoBounds`) is **byte-identical on current `main`** — last touched 2023-05-26, `fopen` logic dates to the original 2022 commit `2966855`. So this is present on `main` too.
- Linux x86-64. Managed binary dynamically linked, with jemalloc as the Rust `#[global_allocator]` and the `unprefixed_malloc_on_supported_platforms` feature (jemalloc interposes the global C `malloc`).

## Symptom

- Managed process hangs immediately at startup: 0 application output, ~3 MB RSS, single thread, `State: S (sleeping)`, `wchan: futex_wait_queue`.
- `strace_logging_mode: standard` produces a **0-byte** `.strace` — the deadlock happens *inside the shim's own first-syscall init*, before any syscall is logged.
- Shadow's manager log stops right after `Running simulation` (virtual time frozen).

## Root cause — gdb backtrace of the wedged process (bottom-up)

```
#38 call_init                       dl-init.c:74          # glibc runs ELF constructors, pre-main
#37 <glib constructor>              libglib-2.0.so
#36 g_hash_table_new_full           libglib-2.0.so
#35 g_malloc                        libglib-2.0.so
#34 _rjem_je_malloc_default         jemalloc.c:2722       # -> jemalloc (global allocator)
#30 malloc_init_hard                jemalloc.c:2129
#29 malloc_init_hard_a0_locked      jemalloc.c:1816       # *** acquires non-recursive init_lock ***
#27 malloc_conf_init_helper         jemalloc.c:1085
#26 obtain_malloc_conf              jemalloc.c:1022
#25 readlink("/etc/malloc.conf")    bits/unistd.h:68      # config read, still holding init_lock
#24 readlink                        preload-libc/syscall_wrappers.c:608
#23 syscall (n=89)                  preload-libc/libc_impls.c:50   # 89 = readlink
#22 shim_api_syscall                libshadow_shim.so
#20 shim_syscallv                   shim_syscall.c:38     # first intercepted syscall ->
#19 _shim_parent_init_preload       shim.c:125            #   runs shim one-time lazy init
#18 patch_vdso                      patch_vdso.c:260
#17 _parseElf                       patch_vdso.c:97
#16 _getVdsoBounds                  patch_vdso.c:19
#15 _IO_new_fopen("/proc/self/maps")
#14 __fopen_internal                libio/iofopen.c:65
#13 imalloc_fastpath
#12 _rjem_je_malloc_default(size=472) jemalloc.c:2722     # fopen buffer -> malloc -> RE-ENTER jemalloc
#8  malloc_init_hard                jemalloc.c:2118
#3  ___pthread_mutex_lock(init_lock+64)
#0  futex_wait(init_lock+64, expected=2)                  # *** re-take init_lock (held) -> DEADLOCK ***
```

The chain:
1. A glib global constructor (pre-`main`) calls `g_malloc` → routed to jemalloc.
2. jemalloc `malloc_init_hard_a0_locked()` **takes its non-recursive `init_lock`**, then `obtain_malloc_conf()` does `readlink("/etc/malloc.conf")` while holding it.
3. That `readlink` is the process's **first shim-intercepted syscall**, so `shim_ensure_init` → `_shim_load` → `init_process` → `_shim_parent_init_preload()` runs, calling `patch_vdso(getauxval(AT_SYSINFO_EHDR))`.
4. `patch_vdso → _parseElf → _getVdsoBounds` does `fopen("/proc/self/maps")` (+ an explicit `malloc(n)`).
5. `fopen`'s buffer allocation **re-enters jemalloc**, which calls `malloc_init_hard()` again and tries to re-acquire the already-held `init_lock` → `futex_wait` forever.

The `init_process` Rust guard (`src/lib/shim/src/lib.rs`, `STARTED_INIT` `LazyLock`) does *not* help: the re-entry deadlocks one frame below, inside **jemalloc's** lock, before control ever returns to Rust. That guard's own comment is apt:
> "TODO: This shouldn't be necessary once we've gotten rid of all calls to libc from the shim's initialization."

## Reproduction

Any dynamically-linked Linux program that sets jemalloc as the global allocator with `unprefixed_malloc_on_supported_platforms` and reaches its first intercepted syscall while jemalloc is still initializing (e.g. a glib/`g_malloc` constructor before `main`). Minimal: a Rust binary with
```rust
#[global_allocator]
static A: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc; // crate feature "unprefixed_malloc_on_supported_platforms"
```
linked dynamically, run under a basic Shadow `shadow.yaml`. It hangs at startup with the backtrace above.

sporksmith:
Ideally we'd avoid libc functions at all other than those explicitly guaranteed to be async signal-safe. This moves us a bit in the right direction, by avoiding memory allocation functions known to very much *not* be async signal safe, which has been confirmed to cause issues for users (#3763).

Fixes #3763 (unconfirmed)
Progress on #2919

### DIFF
--- a/src/lib/shim/patch_vdso.c
+++ b/src/lib/shim/patch_vdso.c
@@ -1,14 +1,65 @@
 #include <assert.h>
 #include <elf.h>
 #include <errno.h>
+#include <fcntl.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <sys/mman.h>
+#include <sys/stat.h>
 #include <sys/syscall.h>
+#include <sys/types.h>
 #include <unistd.h>
 
 #include "lib/logger/logger.h"
+
+// Read `filename` into an mmap'd buffer. `*buffer` is set to the mmap'd buffer. `*buflen` is set to
+// the size of the buffer.  `*filelen` is set to the number of bytes read from the file into the
+// buffer.
+//
+// Doesn't attempt to mmap or stat the file, making it suitable for `/proc` files.
+//
+// Avoids known-not-signal-safe libc functions (e.g. anything that allocates other than mmap).
+//
+// Caller is responsible for deallocating `*buffer`, by `munmap(*buffer, *buflen)`.
+static void _read_file_into_mmapd(const char* filename, char** buffer, size_t* buflen,
+                                  size_t* filelen) {
+    *buflen = 4096;
+    *buffer = mmap(NULL, *buflen, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, 0, 0);
+    if (*buffer == MAP_FAILED) {
+        panic("mmap: %s", strerror(errno));
+    }
+    int fd = open(filename, O_RDONLY);
+    if (fd == -1) {
+        panic("open: %s", strerror(errno));
+    }
+    *filelen = 0;
+    while (true) {
+        ssize_t remaining_capacity = *buflen - *filelen;
+        if (remaining_capacity <= 0) {
+            // We ran out of space. Reallocate a larger buffer.
+            size_t old_region_size = *buflen;
+            *buflen *= 2;
+            *buffer = mremap(*buffer, old_region_size, *buflen, MREMAP_MAYMOVE);
+            if (*buffer == MAP_FAILED) {
+                panic("mremap: %s", strerror(errno));
+            }
+            continue;
+        }
+        ssize_t this_read = read(fd, &(*buffer)[*filelen], remaining_capacity);
+        if (this_read < 0) {
+            panic("read: %s", strerror(errno));
+        }
+        if (this_read == 0) {
+            // EOF
+            break;
+        }
+        *filelen += this_read;
+    }
+    if (close(fd) == -1) {
+        panic("close: %s", strerror(errno));
+    }
+}
 
 static void _getVdsoBounds(void** start, void** end) {
     assert(start);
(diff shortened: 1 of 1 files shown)


src/lib/shim/patch_vdso.c
```diff
@@ -1,54 +1,103 @@
...
+    // First read all of /proc/self/maps into an mmap'd buffer.
+    char* maps_buffer = NULL;
+    size_t maps_buffer_size = 0;
+    size_t maps_file_size = 0;
+    _read_file_into_mmapd("/proc/self/maps", &maps_buffer, &maps_buffer_size, &maps_file_size);
+
+    const char* label = "[vdso]\n";
+    char* line_end = memmem(maps_buffer, maps_file_size, label, strlen(label));
+    if (line_end == NULL) {
+        panic("Couldn't find [vdso] line");
```
robgjansen, line 79: There are several panics added in this PR. Is there really no way to handle these cases gracefully? Or is panicking always the best option?

robgjansen (approved):
I wish there was a good way to add some basic unit tests here to increase our confidence that the corner cases are handled properly. It looks OK to me, but up to you if you want Steve to look at it a bit closer.

sporksmith:
> I wish there was a good way to add some basic unit tests here to increase our confidence that the corner cases are handled properly. It looks OK to me, but up to you if you want Steve to look at it a bit closer.

Yeah, good point.

I was tempted to rewrite this in Rust; I think it wouldn't be too hard using the [elf](https://crates.io/crates/elf) crate (which supports no-std) to do the elf parsing. I'd probably still need to do the ad-hoc /proc/self/maps parsing, but at least it'd be easier to write unit tests and maybe do some error-handling.

I'm still mildly inclined to merge as-is and defer rewriting it, but otoh the original bug reporter seems to have found a work-around for their use-case and hasn't followed-up to confirm this fix, so I'd also be fine with revisiting in a bit when I have time to do the rust rewrite.

sporksmith:
> I'd probably still need to do the ad-hoc /proc/self/maps parsing

From a cursory search, there is the [procinfo](https://crates.io/crates/procinfo) crate, but it looks like it uses libc and doesn't have a no-std option.

src/lib/shim/patch_vdso.c
sporksmith: I suppose we could allow the simulation to continue without patching the vdso, and just logging a warning. The libc preloads are usually sufficient, but e.g. golang can end up using the vdso directly and circumventing shadow

sporksmith: though I'd be mildly inclined to just panic until if/when we find it to be unreliable, since warnings tend to be easy to miss, especially in the shim log. (we could arrange for it to get logged in the "main" log, but it'd take a bit more plumbing)

robgjansen:
Yeah, if no one needs it ATM, then I think a Rust version a bit later is better than a C version now. OTOH, a C version now is probably better than a Rust version never. Happy to defer to your final judgement :)

sporksmith:
> OTOH, a C version now is probably better than a Rust version never

Yeah, that was my thinking powering through with a quick C fix :laughing: 

Let's just leave this MR open for the moment and we can revisit if it becomes more urgently wanted and/or it doesn't look I'll get to the rust version soon.

sporksmith:
> Yeah, if no one needs it ATM, then I think a Rust version a bit later is better than a C version now. OTOH, a C version now is probably better than a Rust version never. Happy to defer to your final judgement :)

Yeah, I don't think I'm going to get to the Rust version soon. I'll go ahead and merge this now before it rots.

PR #3767, opened 2026-07-02
Extend and modularize shadow-exec

sporksmith:
Primarily this is to decouple shadow-exec's functionality of creating simple ad-hoc simulations, from its functionality for running simulations while streaming stdout and stderr of a simulated process. This allows the latter to be used for more complex, multihost, simulations where there is a single client or "control" process (#3671) that has most of the useful output.

### DIFF
--- a/shadowtools/src/shadowtools/shadow_exec.py
+++ b/shadowtools/src/shadowtools/shadow_exec.py
@@ -23,6 +23,7 @@ import argparse
 import glob
 import re
 import enum
+import os
 import subprocess
 import shlex
 import shutil
@@ -105,8 +106,8 @@ def _make_controller_host(args: Iterable[str]) -> scfg.Host:
         # Change back to host working dir
         cd {shlex.quote(str(Path('.').resolve()))}
 
-        # Run specified command, merging stderr to stdout
-        exec {shlex.join(args)} 2>&1
+        # Run specified command
+        exec {shlex.join(args)}
         """)
     return scfg.Host(
         network_node_id=0,
@@ -131,8 +132,8 @@ def _run_shadow_watching_process(
     shadow_config_path: Path,
     shadow_args: Iterable[str] = (),
     stdout: BinaryIO = sys.stdout.buffer,
-    stderr: TextIO = sys.stderr,
-) -> int:
+    stderr: BinaryIO = sys.stderr.buffer,
+) -> None:
     """Run shadow, forwarding the stdout of one simulated process to `stdout`
 
     watch_host: Name of the host inside the shadow sim to watch.
@@ -149,9 +150,8 @@ def _run_shadow_watching_process(
         # It wouldn't be *terribly* hard to support this, but not today.
         # Naively allowing this override would break our stdout pass-through
         # below.
-        print(
+        raise Exception(
             f"ERROR: Overriding shadow's --data-directory currently unsupported.",
-            file=stderr,
         )
     shadow_stdout_path = dstdir.joinpath("shadow.stdout")
     shadow_stderr_path = dstdir.joinpath("shadow.stderr")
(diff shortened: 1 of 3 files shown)


sporksmith:
The errors are because some new code I wrote requires Python 3.10, but debian:11 is still using 3.9.

debian:11 is EOL at the end of August (https://wiki.debian.org/LTS), but python 3.9 is already EOL since last November (https://devguide.python.org/versions/)

I think it wouldn't be that hard to change this to just be compatible with python 3.9 for now. Otherwise I think the options are to install python3.10+ in our debian:11 containers (ew), or end our support for debian:11 slightly early.

sporksmith:
> I think it wouldn't be that hard to change this to just be compatible with python 3.9 for now.

Done

shadowtools/src/shadowtools/shadow_exec.py
```diff
@@ -126,14 +127,16 @@ def _main(
...
+    shadow_stderr_path = dstdir.joinpath("shadow.stderr")
+    host_dir = data_dir.joinpath("hosts", watch_host)
     with shadow_stdout_path.open("w") as shadow_stdout_file, shadow_stderr_path.open(
         "w"
     ) as shadow_stderr_file:
         shadow_ps = subprocess.Popen(
-            [str(shadow_bin)] + list(shadow_args) + ["--", str(config_path)],
+            [str(shadow_bin)]
+            + list(shadow_args)
+            + [f"--data-directory={data_dir}", "--", str(shadow_config_path)],
```
stevenengler, line 179: Does this fail gracefully if the user provides `--data-directory` in `shadow_args`?

shadowtools/src/shadowtools/shadow_exec.py
```diff
@@ -144,14 +147,14 @@ def _main(
 
             # Try opening the simulated process's stdout if we haven't
             # successfully done so yet.
-            if simulated_stdout_file is None:
-                try:
-                    simulated_stdout_file = open(
-                        str(data_dir.joinpath("hosts/host/bash.1000.stdout")), "rb"
+            if simulated_stdout_file is None and (
+                stdout_paths := glob.glob(f"*.{watch_pid}.stdout", root_dir=host_dir)
```
stevenengler: super nit: Maybe add a `include_hidden=True` just in case the binary name starts with a '.'?

shadowtools/src/shadowtools/shadow_exec.py
```diff
@@ -43,6 +43,32 @@ class PreserveChoice(enum.Enum):
...
+def _glob_with_root_dir(pattern: str, root_dir: Path) -> List[str]:
+    """Wrapper around glob.glob, backporting root_dir param.
+
+    glob.glob added the root_dir parameter in python 3.10. When we stop
+    supporting python 3.10 we can replace calls to this function with
+    `glob.glob(pattern, root_dir=root_dir)`.
+    """
+    prefix = str(root_dir) + os.sep
+    rooted_pattern = prefix + pattern
+    return [s.removeprefix(prefix) for s in glob.glob(rooted_pattern)]
```
stevenengler: Does this work if `host_dir` has a trailing separator, like `/path/hosts/host/`? Maybe do a `root_dir.removesuffix(os.sep) + os.sep`? Or just ignore this if it can never happen.

shadowtools/src/shadowtools/shadow_exec.py
```diff
@@ -149,9 +150,8 @@ def _run_shadow_watching_process(
         # It wouldn't be *terribly* hard to support this, but not today.
         # Naively allowing this override would break our stdout pass-through
         # below.
-        print(
+        raise Exception(
             f"ERROR: Overriding shadow's --data-directory currently unsupported.",
-            file=stderr,
```
stevenengler, line 127: nit: I think the error doesn't need the "ERROR:" prefix if it's now an exception.

shadowtools/src/shadowtools/shadow_exec.py
```diff
@@ -167,28 +167,48 @@ def _run_shadow_watching_process(
...
+                    # excessive buffering.
+                    bufsize = 1_000_000
+                    data = src.read(bufsize)
+                if data:
+                    processed_data = True
+                    while data:
+                        count = dest.write(data)
+                        data = data[count:]
+                else:
+                    # Flush when there's no (more) data.  Flushing makes output
```
stevenengler: nit: I found this confusing at first, so maybe "Flush when there's no (more) data available to read"? (I originally read it as "no more data available in 'data'", but that's wrong.)

stevenengler (approved):
LGTM!

shadowtools/src/shadowtools/shadow_exec.py
```diff
@@ -100,70 +94,130 @@ def _main(
...
+                if src is not None:
+                    # Fairly arbitrary, but impose *some* limit to avoid
+                    # excessive buffering.
+                    bufsize = 1_000_000
+                    data = src.read(bufsize)
+                if data:
+                    processed_data = True
+                    while data:
+                        count = dest.write(data)
+                        data = data[count:]
```
stevenengler, line 221: Not for this MR, but I wonder if `shutil.copyfileobj(src, dest)` would simplify this loop a bit. I *think* it would copy everything from `src` to `dest` until the EOF, then we could flush `dest` and continue on.

shadowtools/src/shadowtools/shadow_exec.py
sporksmith: I had experimentally verified that the string conversion of a Path never has a trailing suffix, even if the Path was created from a string that had a trailing suffix. I didn't see any documentation explicitly guaranteeing this though; I'll go ahead and add the `removesuffix` to make this clearer and more robust.

shadowtools/src/shadowtools/shadow_exec.py
sporksmith: Done

shadowtools/src/shadowtools/shadow_exec.py
sporksmith: We detect this and raise an exception above.

shadowtools/src/shadowtools/shadow_exec.py
sporksmith: Hmm, maybe. Added a TODO

shadowtools/src/shadowtools/shadow_exec.py
sporksmith: Changed to "Flush when there's currently no data available to read."

sporksmith (commented):
Thanks!

shadowtools/src/shadowtools/shadow_exec.py
sporksmith: good catch. unfortunately `include_hidden` wasn't added until python 3.11, but adding it to our wrapper (now a reimplementation) actually simplified it a bit.

PR #3768, opened 2026-07-05
Bump deps

### LINKED ISSUE #3528: Upgrade to rustix 1.0
There's a new rustix 1.0 with breaking changes. We should probably upgrade to it at some point.

https://github.com/bytecodealliance/rustix/releases/tag/v1.0.0

https://github.com/bytecodealliance/rustix/blob/main/CHANGES.md#changes-from-038x-to-1x

robgjansen:
Batching some dependency updates.

Closes #3727, Closes #3740, Closes #3741, Closes #3528

### DIFF
--- a/src/main/utility/childpid_watcher.rs
+++ b/src/main/utility/childpid_watcher.rs
@@ -141,10 +141,11 @@ impl ChildPidWatcher {
     fn thread_loop(inner: &Mutex<Inner>, epoll: impl AsFd) {
         let mut commands = Vec::new();
         let mut done = false;
+        let mut events = Vec::<epoll::Event>::with_capacity(10);
         while !done {
-            let mut events = epoll::EventVec::with_capacity(10);
-            match epoll::wait(epoll.as_fd(), &mut events, -1) {
-                Ok(()) => (),
+            let spare_buf = rustix::buffer::spare_capacity(&mut events);
+            match epoll::wait(epoll.as_fd(), spare_buf, None) {
+                Ok(_) => (),
                 Err(rustix::io::Errno::INTR) => {
                     // Just try again.
                     continue;
@@ -159,7 +160,7 @@ impl ChildPidWatcher {
             // caller unregisters it.
             let mut inner = inner.lock().unwrap();
 
-            for event in events.into_iter() {
+            for event in events.drain(..) {
                 if event.data.u64() == 0 {
                     // We get an event for pid=0 when there's a write to the
                     // command_notifier; Ignore that here and handle below.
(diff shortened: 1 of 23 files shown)


src/main/utility/childpid_watcher.rs
```diff
@@ -141,10 +141,11 @@ impl ChildPidWatcher {
     fn thread_loop(inner: &Mutex<Inner>, epoll: impl AsFd) {
         let mut commands = Vec::new();
         let mut done = false;
+        let mut events = Vec::<epoll::Event>::with_capacity(10);
         while !done {
-            let mut events = epoll::EventVec::with_capacity(10);
-            match epoll::wait(epoll.as_fd(), &mut events, -1) {
-                Ok(()) => (),
+            let spare_buf = rustix::buffer::spare_capacity(&mut events);
+            match epoll::wait(epoll.as_fd(), spare_buf, None) {
```
stevenengler, line 147: nit: Can we call `events.clear()` before the `spare_capacity()` call? That way if we accidentally added a `continue` or something, we know we'll never have old elements in the vec. (And it makes the code easier to read if we make it explicit that the vec is always reset.)

src/test/epoll/test_epoll_edge.rs
```diff
@@ -22,7 +22,9 @@ struct WaiterResult {
 
 fn do_epoll_wait(epoll_fd: i32, timeout: Duration, do_read: bool) -> WaiterResult {
     let mut events = Vec::new();
-    events.resize(10, epoll::EpollEvent::empty());
+    // All tests assume that each thread will only receive a single event. Let's
+    // enforce that by limiting the event set to 1.
+    events.resize(1, epoll::EpollEvent::empty());
```
stevenengler, line 27: I don't see how this would fix the races that we see in the CI. I've only locally and in the CI seen where we receive too few events (0 events), never too many. I think there's a benefit of checking if we've received too many events, in which case maybe the epoll_wait waited longer than it was supposed to. But if this is an edge case that shows up in practice and is causing flakiness, then it's probably worth making this change.

I re-ran the CI for this MR, and the "epoll-edge-rs-linux" test failed again.

stevenengler (approved):
This looks good to me, but I'm not convinced by the epoll-edge test fix, since I've never seen that particular edge case before, and the CI is still flaky. But if this is an edge case that you have seen, then merging that seems okay to me.

src/test/epoll/test_epoll_edge.rs
robgjansen: Split this flaky epoll test part into https://github.com/shadow/shadow/pull/3786

src/main/utility/childpid_watcher.rs
robgjansen: Done. (See https://github.com/shadow/shadow/pull/3768/changes/623ec664a3d5749e542c0f86237260764bb115f8)

Issue #2258, opened 2022-07-06
file locks are acquired by shadow's process instead of emulated


sporksmith:
Shadow's implementation of various file locks (`fcntl` record locks, `flock` whole-flie locks) perform the native operation on the native open file descriptor, from Shadow's process.

This is probably ok in cases where the locks are never under contention, but will be incorrect otherwise. e.g. if two different managed processes attempt to acquire a write lock on the same region of the same file, Linux will see it as the Shadow process trying to acquire the lock twice. From the man page, I *think* the second attempt will succeed, which will lead to incorrect and confusing behavior:

> A  single process can hold only one type of lock on a file region; if a new lock is applied to an already-locked region, then the existing lock is converted to the new lock type.

Lock metadata as returned by e.g. `F_GETLK` also includes the pid of the process holding the lock, which will be the native PID of shadow's process instead of the emulated PID of the corresponding managed process.

There are three kinds of file locks that have this problem:

* Two kinds of record-locks operated on via `fcntl` syscalls, which *do* interact with eachother.
  * `fcntl` "process-associated" locks (posix), created with the `fcntl` operations `F_SETLK` and `F_SETLKW`
  * `fcntl` "open file description" locks, created with the `fcntl` operatiosn `F_OFD_SETLK` and `F_OFD_SETLKW`
* Whole-file locks created via the `flock` syscall. These do *not* interact with `fcntl` locks on Linux since version 2.0, according to the `flock` man page.

sporksmith:
I wondered if we could use `F_GETLK` to check whether a lock is already held over the requested range, but AFAICT there's no way to distinguish between "the lock isn't held" and "the lock is held by the current process".

```
#include <unistd.h>
#include <fcntl.h>
#include <stdio.h>
#include <stdlib.h>

int main() {
  int fd = open("test", O_CREAT|O_TRUNC|O_RDWR, S_IRUSR|S_IWUSR);
  if (fd < 0) {
    perror("open");
    abort();
  }
  if (write(fd, "test", 4) != 4) {
    perror("write");
    abort();
  }
  struct flock flock = {
    .l_type = F_WRLCK,
    .l_whence=SEEK_SET,
    .l_start=0,
    .l_len=4,
  };

  if (fcntl(fd, F_SETLK, &flock) != 0) {
    perror("fcntl");
    abort();
  }

  // Getting the lock again succeeds
  if (fcntl(fd, F_SETLK, &flock) != 0) {
    perror("fcntl");
    abort();
[...]

PR #3772, opened 2026-07-29
Add tests for posix record locks

sporksmith:
Most of these don't pass for shadow, yet.
    
Progress on <https://github.com/shadow/shadow/issues/2258>


sporksmith:
We might end up wanting more tests here, but this felt like a good checkpoint for review and to start actually implementing them in shadow.

src/test/test_utils.rs
```diff
@@ -642,3 +644,201 @@ where
...
+            }
+            total_written += usize::try_from(nwritten).unwrap();
+        }
+        Ok(())
+    }
+}
+
+/// Wraps a readable `OwnedFd` to read values of type `T`.
+// TODO: It'd be nice to wrap a generic `Read` object instead, but this is
+// tricky to do in a sound way when T may have undefined padding bytes.
```
stevenengler: I think the `Read` case would be straightforward, since the `read()` will always give initialized bytes? But if the idea is to use `OwnedFd` just to match `TypedWriter`, then that would make sense to me. So maybe just update the comment here?

src/test/test_utils.rs
```diff
@@ -642,3 +644,201 @@ where
...
+        Self {
+            fd,
+            _t: PhantomData,
+        }
+    }
+
+    /// Receive a `T`, or `None` if EOF is reached.
+    ///
+    /// If EOF is reached in the middle of a value, an error with
+    /// ErrorKind::UnexpectedEof is returned.
```
stevenengler: I find this comment a bit confusing. Maybe "or \`None\` if EOF is reached during the first read"?

src/test/test_utils.rs
```diff
@@ -642,3 +644,201 @@ where
...
+    pub fn recv(&mut self) -> Result<Option<T>, std::io::Error> {
```
stevenengler: Maybe clarify somewhere that the reader/writer wrapper types are meant to only be used with blocking files? I think non-blocking files would give some unexpected results.

stevenengler (commented):
I've reviewed everything except `test_file_locks.rs`. I figured I may as well leave the few small comments I have so far.

src/test/test_utils.rs
sporksmith: I should have just updated the comment as requested, but fell down a bit of a rabbit-hole instead.

Yes, the `Read` case is *relatively* straightforward, but the implementation was fiddly enough that I didn't feel confident in it.

I went ahead and implemented it for `Read`, along with a miri unit test to validate that it's sound.

Since this implementation is a bit more solid, I moved the core implementation into the `shadow_pod` crate as an extension trait.

I thought I'd be able to do the same for the write case, at least for `std::os::fd::AsRawFd`, if not `std::io::Write`, but found that miri doesn't allow passing unitialized bytes to `libc::write`. On further reflection it's hard to say whether it's allowable for *any* arbitrary `libc::write`, so I made it directly use the linux syscall instead.

src/test/test_utils.rs
sporksmith: done

src/test/test_utils.rs
sporksmith: done

src/lib/pod/src/lib.rs
```diff
@@ -134,6 +134,102 @@ pub fn from_array<const N: usize, T: Pod>(x: &[u8; N]) -> T {
...
+    R: std::io::Read,
+{
+    fn read_pod<T: Pod>(&mut self) -> std::io::Result<Option<T>> {
+        let mut dst = std::mem::MaybeUninit::<T>::uninit();
+        let dst_bytes = dst.as_mut_ptr().cast::<u8>();
+        // SAFETY: this byte range is writable by construction.
+        unsafe { std::ptr::write_bytes(dst_bytes, 0u8, std::mem::size_of::<T>()) };
+        // SAFETY: now that this byte range is initialized, it can be fully deref'd, read, and written.
+        let dst_slice =
+            unsafe { std::slice::from_raw_parts_mut(dst_bytes, std::mem::size_of::<T>()) };
```
stevenengler: (Optional) Maybe a bit simpler as the following to avoid dealing with buffer lengths?

```rust
let mut dst = MaybeUninit::<T>::uninit();
let dst_slice = unsafe { as_u8_slice_mut(&mut dst) };
dst_slice.fill(MaybeUninit::new(0));
let dst_slice = unsafe { dst_slice.assume_init_mut() };
```

src/lib/pod/src/lib.rs
```diff
@@ -324,3 +420,90 @@ unsafe impl Pod for libc::utmpx {}
...
+    /// it, and we want to ensure that our functions are sound in the presence
+    /// of such a reader).
+    struct NosyReader<R>(R);
+    impl<R> std::io::Read for NosyReader<R>
+    where
+        R: std::io::Read,
+    {
+        fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+            // The "nosy" access to the input buffer bytes, which will be
+            // unsound if they're undefined (in which case miri should raise an
```
stevenengler: nit: undefined -> uninitialized?

src/lib/pod/Cargo.toml
```diff
@@ -7,3 +7,7 @@ publish.workspace = true
 
 [dependencies]
 libc = { version = "0.2.176", default-features = false }
+linux-syscall = "1.2.0"
+
+[features]
+std = []
```
stevenengler: nit: since linux-syscall is only used with std, maybe use `optional = false`, and `std = ["dep:linux-syscall"]`?

src/test/file_locks/test_file_locks.rs
```diff
@@ -0,0 +1,428 @@
...
+        l_type: libc::F_RDLCK.try_into().unwrap(),
+        l_whence: libc::SEEK_SET.try_into().unwrap(),
+        l_start: 0,
+        l_len: 100,
+        l_pid: 0,
+    };
+
+    // Taking a read lock succeeds. (Doesn't matter that the file is empty; non-existent ranges can be locked).
+    // flock struct should be unmodified.
+    ensure_ord!(fcntl_lock(file.as_raw_fd(), cmd, &rd_flock)?, ==, rd_flock);
```
stevenengler, line 42: The comments in this functions seem a little confusing, since cmd can be `[libc::F_SETLK, libc::F_SETLKW, libc::F_GETLK]`, so a caller could provide `libc::F_GETLK` as `cmd`, and then I think we're not taking a read lock anymore? Maybe document `test_uncontested_pid_locks()` as only being usable with `libc::F_SETLK` or `libc::F_SETLKW`?

stevenengler (approved):
I'm not familiar with fcntl locks, but this all looks reasonable to me. The code for testing the behaviour of the child processes looks nice. Implementing this looks like it will be fun :)

src/test/test_utils.rs
stevenengler: I think the new version looks nice!

src/lib/pod/src/lib.rs
sporksmith: nice; done

src/lib/pod/src/lib.rs
sporksmith: done

src/lib/pod/Cargo.toml
sporksmith: done

src/test/file_locks/test_file_locks.rs
sporksmith: good point. documented and added an assertion.

sporksmith (commented):
thanks for the review!

PR #3773, opened 2026-08-04
Add more fcntl posix lock tests

sporksmith:
Progress on #2258

### DIFF
--- a/src/test/file_locks/test_file_locks.rs
+++ b/src/test/file_locks/test_file_locks.rs
@@ -182,6 +182,68 @@ fn test_contested_pid_locks() -> anyhow::Result<()> {
     Ok(())
 }
 
+fn test_contested_zero_len_pid_locks() -> anyhow::Result<()> {
+    let file = tempfile::tempfile().unwrap();
+    let mut child = ForkedChild::new(handle_lock_request)?;
+
+    // fcntl(2):
+    // "Specifying 0 for l_len has the special meaning: > lock all bytes
+    // starting at the location specified by l_whence and l_start through to the
+    // end of file, no matter how large the file grows."
+    let zerolen_flock = libc::flock {
+        l_type: libc::F_WRLCK.try_into().unwrap(),
+        l_whence: libc::SEEK_SET.try_into().unwrap(),
+        l_start: 100,
+        l_len: 0,
+        l_pid: 0,
+    };
+    let res = child
+        .send_recv(&SerializedLockRequest {
+            fd: file.as_raw_fd(),
+            cmd: libc::F_SETLK,
+            flock: zerolen_flock,
+        })
+        .unwrap();
+    ensure_ord!(res.rv, ==, 0);
+    ensure_ord!(res.flock, ==, zerolen_flock);
+
+    // first 100 bytes should still be available to lock
+    let begin_flock = libc::flock {
+        l_type: libc::F_WRLCK.try_into().unwrap(),
+        l_whence: libc::SEEK_SET.try_into().unwrap(),
+        l_start: 0,
+        l_len: 100,
+        l_pid: 0,
+    };
+    ensure_ord!(fcntl_lock(file.as_raw_fd(), libc::F_GETLK, &begin_flock)?.l_type, ==, i16::try_from(libc::F_UNLCK).unwrap());
+
+    // next byte should conflict with child's lock
+    ensure_ord!(fcntl_lock(file.as_raw_fd(), libc::F_GETLK, &libc::flock{
+        l_type: libc::F_WRLCK.try_into().unwrap(),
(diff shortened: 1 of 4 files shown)


src/test/file_locks/test_file_locks.rs
```diff
@@ -297,6 +297,93 @@ fn test_coalesce_and_split_pid_locks() -> anyhow::Result<()> {
...
+        file.as_raw_fd(),
+        libc::F_GETLK,
+        &libc::flock {
+            l_type: libc::F_WRLCK.try_into()?,
+            l_whence: libc::SEEK_SET.try_into()?,
+            l_start: child1_flock.l_start,
+            l_len: 1,
+            l_pid: 0,
+        },
+    )?, ==, libc::flock{
```
robgjansen: I'd prefer to unroll the `fcntl_lock` and `libc::flock` calls into their own stack variables first, and then keep the `ensure_ord!()` calls to just one line. That will force you to name the results of the lock calls (and hopefully you'll choose nice, self documenting names), and make it clearer what the ensures are verifying.

(Also applies to the other `ensure_ord!` on line 369 below).

src/test/file_locks/test_file_locks.rs
```diff
@@ -297,6 +297,93 @@ fn test_coalesce_and_split_pid_locks() -> anyhow::Result<()> {
...
+        l_pid: child1.pid(),
+        ..child1_flock
```
robgjansen: I'm embarrassed this is the first time I am encountering the Rust struct update syntax `..child1_flock`. Is it overly verbose to add a small comment?

```
    ..child1_flock // Copy remaining fields
```

(There is a second instance below too.)

src/test/file_locks/test_file_locks.rs
```diff
@@ -384,6 +384,45 @@ fn test_query_overlapping_pid_locks() -> anyhow::Result<()> {
...
+
+    // Lock a range
+    let wr_flock = libc::flock {
+        l_type: libc::F_WRLCK.try_into()?,
+        l_whence: libc::SEEK_SET.try_into()?,
+        l_start: 0,
+        l_len: 100,
+        l_pid: 0,
+    };
+    ensure_ord!(fcntl_lock(file.as_raw_fd(), libc::F_SETLK, &wr_flock)?, ==,wr_flock);
```
robgjansen: Again I generally prefer separating the function calls from the assertions, to keep the assertions smaller and quicker to comprehend, but maybe this is just a style preference so I'll stop commenting about it after this.

src/test/file_locks/test_file_locks.rs
```diff
@@ -124,18 +161,16 @@ fn test_contested_pid_locks() -> anyhow::Result<()> {
...
+    ensure_ord!(child
         .send_recv(&SerializedLockRequest {
             fd: file.as_raw_fd(),
-            cmd: libc::F_SETLK,
+            cmd: cmd.into(),
             flock: rd_flock,
         })
-        .unwrap();
-    ensure_ord!(res.rv, ==, 0);
-    ensure_ord!(res.flock, ==, rd_flock);
```
robgjansen: Ha! You've un-done my preferred style by compressing everything into a single line :)

I will say, there is value in the two assertions rather than using ? and a single assertion - I think it more precisely gets us to the exact line that failed rather than having to manually parse a backtrace for it.

The extra lines of code don't bother me at all, but I have much less cognitive load because I don't have to figure out which comma is part of the function call and which is a separator for the assertion arguments.

robgjansen (approved):
I don't have much experience with file locks, but these all look reasonable. My comments are mostly about style, feel free to ignore at your discretion.

src/test/file_locks/test_file_locks.rs
sporksmith: Done

src/test/file_locks/test_file_locks.rs
sporksmith: Un-inlined the function calls. One thing I don't like about temp variables in such cases is that it's not always clear to the reader whether this is a name they need to remember beyond this couple lines, but I mitigated that with block scopes.

src/test/file_locks/test_file_locks.rs
sporksmith: It would be too verbose to comment anywhere this syntax is used, IMO. I think this is a pretty common pattern in testing, especially in this file 

I have been in your position though and I wish there was a better way to find out what piece of syntax does when you first come across it, since it's not always easy to construct a search query or otherwise look it up 🤷

src/test/file_locks/test_file_locks.rs
sporksmith: Done. (I actually got rid of this particular struct equality check since this is just setting up what we're trying to test, not the focus of what we're trying to test in this function)

PR #3770, opened 2026-07-11
Version maintenance

robgjansen:
Upgraded dependencies and toolchains.

Closes #3760

PR #3774, opened 2026-08-06
emulate posix record locks

sporksmith:
Implements most of posix record locks (#2258).

`F_SETLKW` doesn't yet implement blocking. Instead, if the lock can't be taken immediately, it returns `EACCES` as for `F_SETLK` and logs a warning. It's probably not that much more work to add it, but I'll leave it for a future PR since this is already quite big.

This should cover the functionality used by `sqlite`.

### DIFF
--- a/src/main/host/descriptor/mod.rs
+++ b/src/main/host/descriptor/mod.rs
@@ -2,23 +2,23 @@
 
 use std::sync::Arc;
 
-use atomic_refcell::AtomicRefCell;
-use linux_api::errno::Errno;
-use linux_api::fcntl::{DescriptorFlags, OFlag};
-use linux_api::ioctls::IoctlRequest;
-use shadow_shim_helper_rs::syscall_types::ForeignPtr;
-
-use crate::core::worker;
 use crate::cshadow as c;
 use crate::host::descriptor::listener::{StateListenHandle, StateListenerFilter};
 use crate::host::descriptor::socket::{Socket, SocketRef, SocketRefMut};
+use crate::host::fcntl_lock_table::FileId;
 use crate::host::host::Host;
 use crate::host::memory_manager::MemoryManager;
+use crate::host::process::ProcessId;
 use crate::host::syscall::io::IoVec;
 use crate::host::syscall::types::{SyscallError, SyscallResult};
 use crate::utility::callback_queue::CallbackQueue;
 use crate::utility::{HostTreePointer, IsSend, IsSync, ObjectCounter};
+use atomic_refcell::AtomicRefCell;
+use linux_api::errno::Errno;
+use linux_api::fcntl::{DescriptorFlags, OFlag};
+use linux_api::ioctls::IoctlRequest;
 use shadow_shim_helper_rs::explicit_drop::ExplicitDrop;
+use shadow_shim_helper_rs::syscall_types::ForeignPtr;
 
 pub mod descriptor_table;
 pub mod epoll;
@@ -44,6 +44,19 @@ bitflags::bitflags! {
         const DIRECT = OFlag::O_DIRECT.bits();
         const NOATIME = OFlag::O_NOATIME.bits();
     }
+}
+
+/// For functions that involve closing descriptors; specifies for which
+/// `ProcessId` posix record locks ought to be freed, if any.
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub enum DropPosixRecordLocks {
+    // Indicates to *not* drop posix record locks as part of closing a
+    // descriptor.  This should only be used for "non-user-visible" closes; e.g.
+    // we end up internally cloning and dropping descriptors as part of our fork
+    // and exec implementations.
+    False,
+    // Includes associated ProcessId for which to drop locks.
+    ForPid(ProcessId),
 }
 
 impl FileStatus {
@@ -517,12 +530,37 @@ impl Descriptor {
         self.file.take().unwrap()
     }
 
-    /// Close the descriptor. The `host` option is a legacy option for legacy file.
+    fn drop_posix_record_locks(&self, host: &Host, pid: ProcessId) {
+        let owner = super::fcntl_lock_table::LockOwner::Process(pid);
+        let stat = match self.file().stat() {
+            Ok(s) => s,
+            Err(e) => {
+                log::debug!("Couldn't stat file to drop record locks: {e:?}");
+                return;
+            }
+        };
+        let fid = FileId::from(&stat);
+        host.fcntl_lock_table_borrow_mut().remove_owner(fid, &owner);
+    }
+
+    /// Close the descriptor.
+    // Caller needs to tell us which PID to drop posix record locks for, if any.
+    // We can't store the "owning" PID in the `Descriptor`, nor the
+    // `DescriptorTable`, since both can be shared across multiple processes
+    // (e.g. after `clone` with `CLONE_FILES` and without `CLONE_THREAD`).
+    //
+    // We use the Host reference both for closing legacy files, and for
+    // accessing the record-lock table.
     pub fn close(
         self,
         host: &Host,
+        drop_posix_record_locks: DropPosixRecordLocks,
         cb_queue: &mut CallbackQueue,
     ) -> Option<Result<(), SyscallError>> {
+        match drop_posix_record_locks {
+            DropPosixRecordLocks::False => (),
+            DropPosixRecordLocks::ForPid(pid) => self.drop_posix_record_locks(host, pid),
+        }
         self.into_file().close(host, cb_queue)
     }
 
@@ -592,12 +630,12 @@ impl Drop for Descriptor {
 }
 
 impl ExplicitDrop for Descriptor {
-    type ExplicitDropParam<'p> = (&'p Host, &'p mut CallbackQueue);
+    type ExplicitDropParam<'p> = (&'p Host, DropPosixRecordLocks, &'p mut CallbackQueue);
     type ExplicitDropResult = Option<Result<(), SyscallError>>;
 
     fn explicit_drop<'p>(self, param: Self::ExplicitDropParam<'p>) -> Self::ExplicitDropResult {
-        let (host, cb_queue) = param;
-        self.close(host, cb_queue)
+        let (host, drop_posix_record_locks, cb_queue) = param;
+        self.close(host, drop_posix_record_locks, cb_queue)
     }
 }
 
@@ -687,11 +725,16 @@ impl LegacyFileCounter {
 
     /// Should drop `self` immediately after calling this.
     fn close_helper(&mut self, host: &Host) {
+        // Always take out the `file` object, so that our `Drop` impl knows this object
+        // was closed properly.
+        let Some(file) = self.file.take() else {
+            warn_and_debug_panic!("Tried to close missing file");
+            #[allow(unreachable_code)]
+            return;
+        };
         // this isn't subject to race conditions since we should never access descriptors
         // from multiple threads at the same time
-        if Arc::<()>::strong_count(&self.open_count) == 1
-            && let Some(file) = self.file.take()
-        {
+        if Arc::<()>::strong_count(&self.open_count) == 1 {
             unsafe { c::legacyfile_close(file.ptr(), host) }
         }
     }
@@ -713,7 +756,20 @@ impl LegacyFileCounter {
 
 impl std::ops::Drop for LegacyFileCounter {
     fn drop(&mut self) {
-        worker::Worker::with_active_host(|host| self.close_helper(host)).unwrap();
+        if self.file.is_some() {
+            warn_and_debug_panic!("Dropped LegacyFileCounter without explicitly closing");
+            #[allow(unreachable_code)]
+            crate::core::worker::Worker::with_active_host(|host| self.close_helper(host)).unwrap();
+        }
+    }
+}
+
+impl ExplicitDrop for LegacyFileCounter {
+    type ExplicitDropParam<'p> = &'p Host;
+    type ExplicitDropResult = ();
+
+    fn explicit_drop<'p>(mut self, host: Self::ExplicitDropParam<'p>) -> Self::ExplicitDropResult {
+        self.close_helper(host);
     }
 }
 
--- a/src/main/host/descriptor/mod.rs
+++ b/src/main/host/descriptor/mod.rs
@@ -701,6 +701,14 @@ impl LegacyFileCounter {
     pub fn close(mut self, host: &Host) {
         self.close_helper(host);
     }
+
+    pub fn mode(&self) -> FileMode {
+        let raw_flags = unsafe { c::legacyfile_getFlags(self.ptr()) };
+        let oflags = OFlag::from_bits_retain(raw_flags);
+        let (mode, _other_flags) =
+            FileMode::from_o_flags(oflags).expect("Invalid flags for open file");
+        mode
+    }
 }
 
 impl std::ops::Drop for LegacyFileCounter {
@@ -747,6 +755,13 @@ impl CompatFile {
         match self {
             CompatFile::New(open_file) => open_file.inner_file().borrow().stat(),
             CompatFile::Legacy(legacy_file_counter) => Ok(legacy_file_counter.stat()?),
+        }
+    }
+
+    pub fn mode(&self) -> FileMode {
+        match self {
+            CompatFile::New(open_file) => open_file.inner_file().borrow().mode(),
+            CompatFile::Legacy(legacy_file_counter) => legacy_file_counter.mode(),
         }
     }
 }
--- a/src/main/host/descriptor/regular_file.c
+++ b/src/main/host/descriptor/regular_file.c
@@ -405,7 +405,12 @@ int regularfile_openat(RegularFile* file, RegularFile* dir, const char* pathname
         file->type = FILE_TYPE_REGULAR;
     }
 
-    int originalFlags = flags;
+    // store original flags in superclass, LegacyDescriptor.
+    // (Primarily so that rust code operating on a LegacyDescriptor
+    // can work out whether the file is readable and/or writable).
+    // TODO: track updates to status flags. Tentatively fixed in pending
+    // <https://github.com/shadow/shadow/pull/3782>.
+    legacyfile_setFlags(&file->super, flags);
 
     // move any flags that shadow handles from 'flags' to 'shadowFlags'
     file->shadowFlags = flags & SHADOW_FLAG_MASK;
--- a/src/main/host/fcntl_lock_table.rs
+++ b/src/main/host/fcntl_lock_table.rs
@@ -1,0 +1,404 @@
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    ops::Range,
+};
+
+use linux_api::fcntl::FlockType;
+use rangemap::RangeMap;
+
+use crate::host::process::ProcessId;
+
+/// Error returned by a lock operation.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum LockError {
+    // lock couldn't be taken, because it's held by another owner.
+    ConflictingLock,
+}
+
+/// Owner of a record lock.
+// Intentionally not `Copy`; see TODO below.
+#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub enum LockOwner {
+    /// For "process-associated" (posix) locks, as created via `fcntl` operation
+    /// `F_SETLK`.
+    // TODO: We might want to make the owner be the *descriptor table* to
+    // more-closely align with the (undocumented? unintentional?) Linux
+    // semantics for processes that shares a descriptor table. See
+    // <https://github.com/shadow/shadow/issues/3783>
+    Process(ProcessId),
+    // TODO: add enumerator for "open file description" locks, as created via
+    // `fcntl` operation `F_OFD_SETLK`. We might want this to involve some sort
+    // of pointer (e.g. Weak) to the file description. To support that, we don't
+    // make this type `Copy`.
+}
+
+/// Internal representation of a record lock.
+#[derive(Clone, Debug, Eq, PartialEq)]
+enum FcntlLock {
+    // Write-lock; one owner.
+    Write(LockOwner),
+    // Read-lock; multiple owners.
+    // Important to use a sorted set here, so that an "arbitrary" conflicting
+    // lock returned via a.g. `F_GETLK` is deterministic.
+    Read(BTreeSet<LockOwner>),
+}
+
+impl FcntlLock {
+    fn new(requester: &LockOwner, flock_type: FlockType) -> Option<FcntlLock> {
+        match flock_type {
+            FlockType::F_UNLCK => None,
+            FlockType::F_RDLCK => Some(FcntlLock::Read(BTreeSet::from_iter([requester.clone()]))),
+            FlockType::F_WRLCK => Some(FcntlLock::Write(requester.clone())),
+        }
+    }
+
+    fn access(&self) -> FlockType {
+        match self {
+            FcntlLock::Write(_) => FlockType::F_WRLCK,
+            FcntlLock::Read(_) => FlockType::F_RDLCK,
+        }
+    }
+
+    fn owners_contains(&self, o: &LockOwner) -> bool {
+        match self {
+            FcntlLock::Write(write_owner) => write_owner == o,
+            FcntlLock::Read(read_owners) => read_owners.contains(o),
+        }
+    }
+
+    /// Return the result of updating this lock with the given request, if possible.
+    /// Otherwise returns an error that describes one of the conflicting locks.
+    ///
+    /// Use `requested_type=FlockType::F_UNLCK` to unlock.
+    ///
+    /// Returns:
+    /// * `Ok(None)` when the result is "no lock"; i.e. when
+    ///   `requested_type` is `F_UNLCK`, and `requester` is the only owner of
+    ///   `self`.
+    /// * `Ok(Some(x))`, where x is the result of a successful update.
+    /// * `Err(x)`, where x is one of the current owners of a conflicting lock.
+    pub fn with_request_applied(
+        &self,
+        requester: &LockOwner,
+        requested_type: FlockType,
+    ) -> Result<Option<Self>, LockOwner> {
+        match self {
+            FcntlLock::Write(current_owner) => {
+                if current_owner == requester {
+                    // requester is already the exclusive owner.
+                    // Give them whatever they want.
+                    Ok(Self::new(requester, requested_type))
+                } else if requested_type == FlockType::F_UNLCK {
+                    // requester is releasing, but isn't an owner of this lock.
+                    // Return this lock unchanged.
+                    Ok(Some(self.clone()))
+                } else {
+                    // Conflict with current owner.
+                    Err(current_owner.clone())
+                }
+            }
+            FcntlLock::Read(current_owners) => {
+                match requested_type {
+                    FlockType::F_RDLCK => {
+                        // Add requester to the set of read-lock owners
+                        let mut s = current_owners.clone();
+                        s.insert(requester.clone());
+                        Ok(Some(FcntlLock::Read(s)))
+                    }
+                    FlockType::F_WRLCK => {
+                        match current_owners.iter().find(|o| o != &requester) {
+                            Some(conflicting_owner) => {
+                                // requesting a write lock, but a different owner
+                                // is holding a read lock.
+                                Err(conflicting_owner.clone())
+                            }
+                            None => {
+                                // requesting a write lock, and requester is the sole holder
+                                // of a read lock. Upgrade it to a write lock.
+                                Ok(FcntlLock::new(requester, FlockType::F_WRLCK))
+                            }
+                        }
+                    }
+                    FlockType::F_UNLCK => {
+                        // Requesting to unlock, and there's a read-lock.
+                        // Remove requester from set of read-lock owners.
+                        let new_owners = BTreeSet::from_iter(
+                            current_owners.iter().filter(|x| x != &requester).cloned(),
+                        );
+                        if new_owners.is_empty() {
+                            // No owners left -> unlocked.
+                            Ok(None)
+                        } else {
+                            Ok(Some(FcntlLock::Read(new_owners)))
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+
+/// Stable identifier for a file, which we use to look up record locks.
+/// Remains valid as long as the file has links in the file system and/or there
+/// are open file descriptors to the file.
+//
+// We implement this using device+inode, as returned by `fstat`.
+//
+// According to
+// [posix](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/sys_stat.h.html):
+//
+//   A file identity is uniquely determined by the combination of st_dev and
+//   st_ino. At any given time in a system, distinct files shall have distinct
+//   file identities; hard links to the same file shall have the same file
+//   identity. Over time, these file identities can be reused for different files.
+//   For example, the st_ino value can be reused after the last link to a file is
+//   unlinked and the space occupied by the file has been freed, and the st_dev
+//   value associated with a file system can be reused if that file system is
+//   detached ("unmounted") and another is attached ("mounted").
+//
+// While the posix definition doesn't seem to clearly specify whether an inode
+// number can be reused when there are no more links in the file system, but the
+// file is still open, my understanding is that at least in Linux, files aren't
+// destroyed and their inode numbers made available for reuse until it is no
+// longer open by anyone. e.g.
+// [unlink(2)](https://man7.org/linux/man-pages/man2/unlink.2.html):
+//
+//   If the name was the last link to a file but any processes still
+//   have the file open, the file will remain in existence until the
+//   last file descriptor referring to it is closed.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct FileId {
+    device: u64,
+    inode: u64,
+}
+
+impl From<&linux_api::stat::stat> for FileId {
+    fn from(value: &linux_api::stat::stat) -> Self {
+        FileId {
+            device: value.lst_dev,
+            inode: value.lst_ino,
+        }
+    }
+}
+
+impl From<&libc::stat> for FileId {
+    fn from(value: &libc::stat) -> Self {
+        FileId {
+            device: value.st_dev,
+            inode: value.st_ino,
+        }
+    }
+}
+
+#[derive(Debug)]
+struct FcntlLockTableForOneFile {
+    locks: RangeMap<usize, FcntlLock>,
+    // Will need something here to track sleepers
+}
+
+impl FcntlLockTableForOneFile {
+    fn new() -> Self {
+        Self {
+            locks: Default::default(),
+        }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.locks.is_empty()
+    }
+
+    /// Apply the requested lock to the given range. Similar semantics as
+    /// `fcntl` operations like `F_SETLK`.
+    ///
+    /// Returns an error if an incompatible lock exists.
+    pub fn set_lock(
+        &mut self,
+        requested_range: Range<usize>,
+        requester: &LockOwner,
+        requested_type: FlockType,
+    ) -> Result<(), LockError> {
+        // Process overlapping locks, collecting what the updated locks will look like.
+        let mut updated_results = Vec::<(Range<usize>, FcntlLock)>::new();
+        for (lock_range, lock) in self.locks.overlapping(&requested_range) {
+            match lock.with_request_applied(requester, requested_type) {
+                Ok(Some(updated_lock)) => {
+                    // Record the updated lock.
+                    let start = std::cmp::max(lock_range.start, requested_range.start);
+                    let end = std::cmp::min(lock_range.end, requested_range.end);
+                    updated_results.push((start..end, updated_lock));
+                }
+                Ok(None) => {
+                    // Result is no-lock. No need to record anything; we'll clear this below.
+                }
+                Err(_) => {
+                    // Conflicting lock; can't update.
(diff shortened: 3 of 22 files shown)


src/main/host/descriptor/regular_file.c
```diff
@@ -405,7 +405,12 @@ int regularfile_openat(RegularFile* file, RegularFile* dir, const char* pathname
         file->type = FILE_TYPE_REGULAR;
     }
 
-    int originalFlags = flags;
+    // store original flags in superclass, LegacyDescriptor.
+    // (Primarily so that rust code operating on a LegacyDescriptor
+    // can work out whether the file is readable and/or writable)
+    // TODO: maybe it'd be better to push `flags` down into the subclasses,
+    // and have a callback for fetching them from the LegacyDescriptor?
+    file->super.flags = flags;
```
opara-tor: I don't think the `super.flags` is meant for all of these flags? For example `flags` contains file creation flags like `O_RDWR`, but those aren't stored in the file flags (the flags returned by `fcntl()` with `F_GETFL`). So we'd return the wrong flags from `F_GETFL`.

In the Linux `struct file`, it has a `mode` field which stores the access mode (in shadow's rust code it's `FileMode`). Maybe `LegacyDescriptor` should have a `mode` field instead?

opara-tor: Add a comment here that we don't keep `super.flags` in sync with the actual file flags? For example we don't update these flags when we `fcntl()` with `F_SETFL`.

And this includes some flags that don't make sense for a file like `O_CLOEXEC`. Maybe remove that flag and use `legacyfile_setFlags()` here instead of accessing the super flags directly?

src/main/host/fcntl_lock_table.rs
```diff
@@ -0,0 +1,423 @@
...
+
+/// Owner of a record lock.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub enum LockOwner {
+    // For "process-associated" (posix) locks, as created via `fcntl` operation
+    // `F_SETLK`.
+    Process(ProcessId),
+    // TODO: add enumerator for "open file description" locks, as created via
+    // `fcntl` operation `F_OFD_SETLK`. We'll need some appropriate identifier
+    // to do so.
```
opara-tor: I'm a little confused by this. The fcntl locks are per-process, but file locks are per-host. So won't this need separate tables for both types of locks (one at host level and many at process levels)?

In that case, shouldn't the lock owner be a generic type on the lock table rather than an enum? Since there wouldn't be two types of owners in the same lock table.

src/main/host/fcntl_lock_table.rs
```diff
@@ -0,0 +1,423 @@
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+enum FcntlLock {
+    // Unlocked.
+    None,
+    // Write-lock; one owner.
+    Write(LockOwner),
+    // Read-lock; multiple owners.
+    // Important to use a sorted set here, so that an "arbitrary" conflicting
+    // lock returned via a.g. `F_GETLK` is deterministic.
+    Read(BTreeSet<LockOwner>),
```
opara-tor, line 43: Why can we have a record lock of "unlocked", in which case it's not really a lock anymore? This could use a comment.

src/main/host/fcntl_lock_table.rs
```diff
@@ -0,0 +1,423 @@
...
+            FcntlLock::Write(write_owner) => write_owner == &o,
+            FcntlLock::Read(read_owners) => read_owners.contains(&o),
+        }
+    }
+
+    /// If the requested lock type is compatible with this existing lock type, return the
+    /// result of merging the two. Otherwise returns an error that describes one of the
+    /// conflicting locks.
+    ///
+    /// Use `requested_access=None` to unlock.
```
opara-tor: What does "requested_access" mean here? Is this meant to be the `requested_type` argument?

src/main/host/fcntl_lock_table.rs
```diff
@@ -0,0 +1,423 @@
...
+    fn new() -> Self {
+        let mut locks = RangeMap::new();
+        // All unlocked.
+        locks.insert(usize::MIN..usize::MAX, FcntlLock::None);
+        Self { locks }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.locks.is_empty()
+    }
```
opara-tor, line 208: This seems like a footgun since a map of only `FcntlLock::None` (so no locks) will be considered non-empty.

src/main/host/fcntl_lock_table.rs
```diff
@@ -0,0 +1,423 @@
...
+            .or_insert_with(FcntlLockTableForOneFile::new);
+        let res = file_locks.set_lock(requested_range.clone(), requested_owner, requested_access);
+        if res.is_err() {
+            log::debug!(
+                "failed to lock {file_id:?}.{requested_range:?}.{requested_owner:?}.{requested_access:?}"
+            );
+        }
+        if file_locks.is_empty() {
+            self.locks.remove(&file_id);
+        }
```
opara-tor, line 358: I don't think this handles a table of `FcntlLock::None`?

src/main/host/fcntl_lock_table.rs
```diff
@@ -0,0 +1,423 @@
...
+            }
+            FcntlLock::Write(current_owner) => {
+                if current_owner == &requester {
+                    // requester is already the exclusive owner.
+                    // Give them whatever they want.
+                    Ok(Self::new(requester, requested_type))
+                } else if requested_type == FlockType::F_UNLCK {
+                    // requester is releasing, but isn't an owner of this lock.
+                    // Return this lock unchanged.
+                    Ok(self.clone())
```
opara-tor: This is confusing to me, maybe because of the name "merge".

For example we have a write lock and a read/write/un -lock request. So "write lock merged with write lock = write lock", "write lock merged with read lock = read lock", and "write lock merged with unlock = unlock". But these don't seem intuitive to me.

Also we have:

1. Write lock merged with unlock = unlock
2. Unlock merged with write lock = write lock

but I think of a "merge" as generally being commutative.

Maybe a name like "apply_change(request)" would be better?

src/main/host/fcntl_lock_table.rs
opara-tor: I feel like we'd want the "open file description" to be a `Weak` reference to a file in the future, rather than just a numeric identifier. And this would mean that `LockOwner` can't be `Copy`. I don't know if it's worth making this non-`Copy` now if you agree.

src/main/host/fcntl_lock_table.rs
```diff
@@ -0,0 +1,423 @@
...
+struct FcntlLockTableForOneFile {
+    locks: RangeMap<usize, FcntlLock>,
+    // Will need something here to track sleepers
+}
+
+impl FcntlLockTableForOneFile {
+    fn new() -> Self {
+        let mut locks = RangeMap::new();
+        // All unlocked.
+        locks.insert(usize::MIN..usize::MAX, FcntlLock::None);
```
opara-tor: Why do we require an explicit entry for unlocked regions? This seems more complicated and like it would possibly use more memory since each unlocked region also needs to be tracked.

src/main/host/fcntl_lock_table.rs
```diff
@@ -0,0 +1,423 @@
...
+                break;
+            };
+            if !(can_coalesce(prev_lock)) {
+                break;
+            }
+            conflicting_range.start = prev_range.start
+        }
+
+        // Coalesce forwards
+        while let Some(x) = conflicting_range.end.checked_add(1) {
```
opara-tor: I think the `checked_add(1)` here isn't right? Since the range is half-open, `end` is not included in the range. So if there was an adjacent 1 byte gap, we'd skip over it.

src/main/host/fcntl_lock_table.rs
```diff
@@ -0,0 +1,423 @@
...
+            let Some((next_range, next_lock)) = self.locks.get_key_value(&x) else {
+                break;
+            };
+            if !(can_coalesce(next_lock)) {
+                break;
+            }
+            conflicting_range.end = next_range.end
```
opara-tor, line 318: nit: debug assert before this that `conflicting_range.end == next_range.start`?

src/main/host/fcntl_lock_table.rs
```diff
@@ -0,0 +1,423 @@
...
+        file_id: FileId,
+        requested_range: Range<usize>,
+        requested_owner: LockOwner,
+        requested_access: FlockType,
+    ) -> Option<(LockOwner, Range<usize>, FlockType)> {
+        let file_locks = self.locks.get(&file_id)?;
+        let start = requested_range.start;
+        let end = requested_range.end;
+        let res = file_locks.get_coalesced_conflicting_lock(
+            requested_range,
```
opara-tor, line 375: nit: I think this would be simpler with just `requested_range.clone()` and removing `start` and `end`?

src/main/host/fcntl_lock_table.rs
```diff
@@ -0,0 +1,423 @@
...
+        res
+    }
+
+    /// Drop all of the given owner's locks on the given file.
+    pub fn remove_owner(&mut self, file_id: FileId, requested_owner: LockOwner) {
+        let Some(file_locks) = self.locks.get_mut(&file_id) else {
+            return;
+        };
+        file_locks.remove_owner(requested_owner);
+    }
```
opara-tor, line 397: I think this should do some sort of cleanup if the map is now empty?

Or alternatively there could be some helper method like `with_table_mut(&mut self, file_id, f: FnOnce(&mut FcntlLockTableForOneFile))` that always checks if there are any non-none entries after running the closure, so that we don't forget.

src/main/host/host.rs
```diff
@@ -993,7 +1008,6 @@ impl Drop for Host {
         assert!(self.shim_shmem_lock.borrow().is_none());
     }
 }
-
```
opara-tor, line 996: nit: this looks accidental?

src/main/host/syscall/handler/fcntl.rs
```diff
@@ -1,13 +1,112 @@
...
 use crate::host::syscall::handler::{SyscallContext, SyscallHandler};
 use crate::host::syscall::type_formatting::SyscallNonDeterministicArg;
 use crate::host::syscall::types::SyscallError;
 
+#[derive(Debug, Copy, Clone)]
+struct CommonFlockParams {
+    file_id: fcntl_lock_table::FileId,
+    start: usize,
+    end: usize,
+    owner: fcntl_lock_table::LockOwner,
```
opara-tor: nit: We already import `FileId` above so don't need the full path here, and we use `LockOwner` in a couple places, so maybe it could be included with `FileId`?

src/main/host/syscall/handler/fcntl.rs
```diff
@@ -1,13 +1,112 @@
...
+    let (start, end) = if flock.l_len == 0 {
+        // fcntl(2): Specifying 0 for l_len has the special meaning: lock all
+        // bytes starting at the location specified by l_whence and l_start
+        // through to the end of file, no matter how large the file grows.
+        let start = usize::try_from(signed_start).map_err(|_| Errno::EINVAL)?;
+        (start, FLOCK_LENGTH_0_END)
+    } else if flock.l_len < 0 {
+        // fcntl(2): if l_len is negative, the interval described by lock
+        // covers bytes l_start+l_len  up  to  and  including  l_start-1.
+        let start = usize::try_from(signed_start + flock.l_len).map_err(|_| Errno::EINVAL)?;
```
opara-tor: nit: i think `signed_start + flock.l_len` should use a checked add? Maybe it ends up being fine, but it's hard to tell without thinking about the edge case.

src/main/host/syscall/handler/fcntl.rs
```diff
@@ -1,13 +1,112 @@
...
+use crate::host::process::Process;
 use crate::host::syscall::handler::{SyscallContext, SyscallHandler};
 use crate::host::syscall::type_formatting::SyscallNonDeterministicArg;
 use crate::host::syscall::types::SyscallError;
 
+#[derive(Debug, Copy, Clone)]
+struct CommonFlockParams {
+    file_id: fcntl_lock_table::FileId,
+    start: usize,
+    end: usize,
```
opara-tor: Can we use a `Range` or `RangeInclusive` here to make it clearer whether its half-open or not?

src/main/host/syscall/handler/fcntl.rs
```diff
@@ -1,13 +1,112 @@
...
+        owner: requested_owner,
+        access: requested_access,
+    })
+}
+
+// We internally represent flock ranges using an exclusive range of `usize`s.
+// Internally, Linux returns an overflow error when the calculated end is
+// greater than i64::MAX + 1.  (Perhaps suggesting that it internally uses an
+// *inclusive* range of i64s).
+const FLOCK_MAX_RANGE_END: usize = (i64::MAX as usize) + 1;
```
opara-tor: Can you add a comment here that the `FLOCK_MAX_RANGE_END` is the max value we allow, to make it clear that this is the end of an inclusive range?

But maybe this isn't true because we allow locks with range `FLOCK_LENGTH_0_END` which is larger?

src/main/host/syscall/handler/fcntl.rs
```diff
@@ -1,13 +1,112 @@
...
+    };
+    let Some(signed_start) = origin.checked_add(flock.l_start) else {
+        return Err(Errno::EINVAL.into());
+    };
+    let (start, end) = if flock.l_len == 0 {
+        // fcntl(2): Specifying 0 for l_len has the special meaning: lock all
+        // bytes starting at the location specified by l_whence and l_start
+        // through to the end of file, no matter how large the file grows.
+        let start = usize::try_from(signed_start).map_err(|_| Errno::EINVAL)?;
+        (start, FLOCK_LENGTH_0_END)
```
opara-tor: Does this cause unexpected results if a fcntl returns a lock (as in `F_GETLK`), since the returned range will be greater than `FLOCK_MAX_RANGE_END`?

src/main/host/syscall/handler/fcntl.rs
```diff
@@ -1,13 +1,112 @@
...
+    let whence = FlockWhence::try_from(flock.l_whence).map_err(|_| Errno::EINVAL)?;
+    let origin = match whence {
+        FlockWhence::SEEK_SET => 0,
+        FlockWhence::SEEK_CUR => file.lseek(0, LSeekWhence::SEEK_CUR)?,
+        FlockWhence::SEEK_END => file.stat()?.lst_size,
+    };
+    let Some(signed_start) = origin.checked_add(flock.l_start) else {
+        return Err(Errno::EINVAL.into());
+    };
+    let (start, end) = if flock.l_len == 0 {
```
opara-tor: Add a comment here about whether `end` is inclusive or not?

src/main/host/syscall/handler/fcntl.rs
```diff
@@ -1,13 +1,112 @@
...
+        access: requested_access,
+    })
+}
+
+// We internally represent flock ranges using an exclusive range of `usize`s.
+// Internally, Linux returns an overflow error when the calculated end is
+// greater than i64::MAX + 1.  (Perhaps suggesting that it internally uses an
+// *inclusive* range of i64s).
+const FLOCK_MAX_RANGE_END: usize = (i64::MAX as usize) + 1;
+/// flock(2): Specifying 0 for l_len has the special meaning: lock all bytes
```
opara-tor: I think this should say `fcntl(2)`?

src/main/host/fcntl_lock_table.rs
```diff
@@ -0,0 +1,423 @@
...
+    /// Apply the requested lock to the given range. Similar semantics as
+    /// `fcntl` operations like `F_SETLK`.
+    ///
+    /// Returns an error if an incompatible lock exists.
+    pub fn set_lock(
+        &mut self,
+        requested_range: Range<usize>,
+        requester: LockOwner,
+        requested_type: FlockType,
+    ) -> Result<(), ()> {
```
opara-tor: I think it would be nice to have an actual error type here. Even just `struct IncompatibleLockExistsError;` or something.

src/main/host/syscall/handler/fcntl.rs
```diff
@@ -39,11 +138,80 @@ impl SyscallHandler {
...
+                    .set_lock(
+                        params.file_id,
+                        params.start..params.end,
+                        params.owner,
+                        params.access,
+                    )
+                    .map_err(|()| Errno::EACCES);
+                log::trace!("setlk[w] {params:?} -> {res:?}");
+                if cmd == FcntlCommand::F_SETLKW
+                    && [Err(Errno::EACCES), Err(Errno::EAGAIN)].contains(&res)
```
opara-tor: Why do we check `EAGAIN` here? I think nothing can set that? I think there should be a comment here.

src/main/host/syscall/handler/fcntl.rs
```diff
@@ -39,11 +138,80 @@ impl SyscallHandler {
...
+                    params.owner,
+                    params.access,
+                ) {
+                    Some((conflict_owner, conflict_range, conflict_access)) => {
+                        let length = if conflict_range.end == FLOCK_LENGTH_0_END {
+                            // "round-trip" length=0 to mean "through EOF". See `FLOCK_LENGTH_0_END`.
+                            0
+                        } else {
+                            i64::try_from(conflict_range.len()).unwrap_or_else(|_err| {
+                                panic!("Current lock range {conflict_range:?} end is out of range")
```
opara-tor: I think this can panic for a lock with range `0..(i64::MAX + 1)`? Maybe add a test?

src/main/host/descriptor/mod.rs
```diff
@@ -517,13 +530,22 @@ impl Descriptor {
         self.file.take().unwrap()
     }
 
-    /// Close the descriptor. The `host` option is a legacy option for legacy file.
+    /// Close the descriptor.
+    // Caller needs to tell us which PID to drop posix record locks for, if any.
+    // We can't store the "owning" PID in the `Descriptor`, nor the
+    // `DescriptorTable`, since both can be shared across multiple processes
+    // (e.g. after `fork` without `execve`).
```
opara-tor: Doesn't `fork()` duplicate the descriptor table? So neither the `Descriptors` or `DescriptorTable` are shared?

Maybe you meant "e.g. after \`clone()\` with \`CLONE_FILES\`"? But I feel like the current code will break if the program were to use `CLONE_FILES`, so I'm not sure this is right either.

src/main/host/descriptor/mod.rs
```diff
@@ -729,12 +793,13 @@ impl CompatFile {
...
         self,
         host: &Host,
+        drop_posix_record_locks: DropPosixRecordLocks,
         cb_queue: &mut CallbackQueue,
     ) -> Option<Result<(), SyscallError>> {
         match self {
             Self::New(file) => file.close(cb_queue),
             Self::Legacy(file) => {
-                file.close(host);
+                file.close(host, drop_posix_record_locks);
```
opara-tor: I feel like releasing the locks should be an operation on the `Descriptor` and not on the `CompatFile`/`LegacyFileCounter`, since posix file locks are released when the descriptor is closed and not the file. The `CompatFile` and `LegacyFileCounter` are mostly just a temporary thing until the legacy files go away. And files are stored in other places than the descriptor table. For example `OpenFile` is stored in `_SysCallCondition` (it should be storing `CompatFile` but it's not because of a bug in the legacy file code), which has nothing to do with posix locks.

And also don't we want this for all file types, not just legacy files?

src/main/host/fcntl_lock_table.rs
```diff
@@ -0,0 +1,423 @@
...
+use rangemap::RangeMap;
+
+use crate::host::process::ProcessId;
+
+/// Owner of a record lock.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub enum LockOwner {
+    // For "process-associated" (posix) locks, as created via `fcntl` operation
+    // `F_SETLK`.
+    Process(ProcessId),
```
opara-tor: The documentation describes this as process-associated, but then it seems broken when used with `CLONE_FILES` (and no `CLONE_THREAD`)? For example if you clone() a process with `CLONE_FILES`, you'll have two processes sharing the same descriptor table. If one process creates a lock for an fd and the other thread closes that same fd, the lock won't be freed because the pid will be different.

Maybe add a TODO to investigate this? Or if you have already, then maybe add some comments to describe how this is handled?

I wonder if it's not really process-associated but "descriptor table"-associated, so that the locks are tied to the descriptor table and not the pid.

src/main/host/syscall/handler/unistd.rs
```diff
@@ -16,9 +16,9 @@ use crate::core::work::task::TaskRef;
 use crate::core::worker::Worker;
 use crate::cshadow as c;
 use crate::host::descriptor::descriptor_table::DescriptorHandle;
-use crate::host::descriptor::pipe;
 use crate::host::descriptor::shared_buf::SharedBuf;
 use crate::host::descriptor::{CompatFile, Descriptor, File, FileMode, FileStatus, OpenFile};
+use crate::host::descriptor::{DropPosixRecordLocks, pipe};
```
opara-tor: nit: Combine these two lines?

opara-tor (changes requested):

src/main/host/descriptor/mod.rs
sporksmith: > Doesn't fork() duplicate the descriptor table? So neither the Descriptors or DescriptorTable are shared?

Oops, you're right

> Maybe you meant "e.g. after `clone()` with `CLONE_FILES`"?

Right

> But I feel like the current code will break if the program were to use CLONE_FILES, so I'm not sure this is right either.

Maybe! Our handler currently implements it, and we have some testing for it in `test_clone.rs`, but it is a pretty niche use-case (outside of also using `CLONE_THREAD`).

Updated comment. "e.g. after `clone` with `CLONE_FILES` and without `CLONE_THREAD`"

src/main/host/descriptor/mod.rs
sporksmith: Oops, yes. I'd originally implemented this more directly for `RegularFile`s, before realizing that they actually work on other file types too and implementing generic `fstat` and `lseek`.

Moved the lock-drop up into `Descriptor`.

src/main/host/descriptor/regular_file.c
sporksmith: I ended up overhauling the flag handling here and putting it in a separate PR #3782.

src/main/host/syscall/handler/fcntl.rs
sporksmith: Done.

Yeah, I had it as a range initially, and then split it because `Range<T>` isn't `Copy`, which leads to some awkward `clone` calls. I suppose the clarity is worth it and the `clone`s should be trivially cheap, though.

(I wonder why `Range<T>` isn't `Copy` even when `T` is; maybe `Copy` is special and can't be implemented conditionally this way?)

src/main/host/syscall/handler/fcntl.rs
sporksmith: done

src/main/host/syscall/handler/fcntl.rs
sporksmith: renamed `end`  to `open_end`

src/main/host/syscall/handler/fcntl.rs
sporksmith: done

src/main/host/syscall/handler/fcntl.rs
sporksmith: done

src/main/host/syscall/handler/fcntl.rs
sporksmith: Renamed to `FLOCK_MAX_VISIBLE_RANGE_OPEN_END` and tried to clarify the comment

src/main/host/syscall/handler/fcntl.rs
sporksmith: Ah, I added this check later, and reflexively checked for both because F_SETLK is allowed to return either. But yeah, it's a bit silly when we can see just above that it can only be EACCES.

I rearranged this a bit, adding a proper error type from the lock table and doing an exhaustive match on it here *before* mapping it to an errno.

src/main/host/syscall/handler/fcntl.rs
sporksmith: good catch!

Initially I was thinking this is impossible, since the maximum length of a lock that can be set is i64::MAX, but a *coalesced lock* can have one beyond that.

Added a test; it turns out that Linux reports length 0 in this case (same as if the user had set a lock with length=0).

src/main/host/syscall/handler/unistd.rs
sporksmith: done

src/main/host/syscall/handler/fcntl.rs
sporksmith: No; hopefully this is clearer now from the updated names and comments around FLOCK_LENGTH_0_END

src/main/host/fcntl_lock_table.rs
sporksmith: My expectation based on the man pages was that the broken behavior you describe would be the Linux behavior, as it matches what's in the man pages afaict.

I wrote a test though, and it indeed it appears that the descriptor-table *is* effectively the owner. a CLONE_FILES child process seems to own the same set of posix range locks that its parent does. It'd probably be nice to get the "can take a lock or not" behavior for this case right, though maybe it's worth asking kernel folks if this is intended behavior that's unlikely to change, or just some unspecified case and their implementation happens to work this way.

There's another layer of complexity trying to determine what the `l_pid` labels ought to be and which lock regions get coalesced in `F_GETLK` queries. That's probably getting further into unspecified implementation details we shouldn't try that much to worry about though; it's hard to imagine the particular behavior of those actually changing the behavior of managed processes.

sporksmith (commented):
Still working on this, but thought I'd push some of the updates

src/main/host/fcntl_lock_table.rs
sporksmith: Hmm, yeah that probably makes sense; otherwise we'd need some Host table of identifier -> file description object. Went ahead and made it non-Copy for now.

src/main/host/fcntl_lock_table.rs
sporksmith: Removed the None case, replacing this with an Option<FcntlLock> in some places.

I had added None here to correspond to FlockType::F_UNLCK. Having a 1:1 correspondence between these and the FlockTypes made some things a little nicer, but I think overall it probably is more confusing. (and I'd forgotten to update some comments and fix table GC when I'd initially added this state)

src/main/host/fcntl_lock_table.rs
sporksmith: Outdated comment; fixed.

src/main/host/fcntl_lock_table.rs
sporksmith: Renamed to `with_request_applied`

src/main/host/fcntl_lock_table.rs
sporksmith: `FcntlLock::None` removed.

src/main/host/fcntl_lock_table.rs
sporksmith: Yeah, `FcntlLock::None` removed.

src/main/host/fcntl_lock_table.rs
sporksmith: good catch!

src/main/host/fcntl_lock_table.rs
sporksmith: done

src/main/host/fcntl_lock_table.rs
sporksmith: `FcntlLock::None` removed

src/main/host/fcntl_lock_table.rs
sporksmith: done

src/main/host/fcntl_lock_table.rs
sporksmith: Rewrote to reuse `set_lock` with `F_UNLCK`, which already does the cleanup. Realized we don't need the more specialized implementation.

src/main/host/fcntl_lock_table.rs
sporksmith: I think I'm inclined to keep the implementation roughly as-is for now and file an issue for correctly handling processes with shared descriptor tables. Wdyt?

sporksmith: P.S. I checked the kernel tests to see if any covered this case, since that would probably better indicate intent. I didn't find any tests for posix range lock; only some very light tests for open file description locks :shrug:  https://github.com/torvalds/linux/blob/master/tools/testing/selftests/filelock/ofdlocks.c

src/main/host/syscall/handler/fcntl.rs
```diff
@@ -12,11 +14,10 @@ use crate::host::syscall::handler::{SyscallContext, SyscallHandler};
 use crate::host::syscall::type_formatting::SyscallNonDeterministicArg;
 use crate::host::syscall::types::SyscallError;
 
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Clone)]
 struct CommonFlockParams {
     file_id: fcntl_lock_table::FileId,
-    start: usize,
-    end: usize,
+    range: Range<usize>,
```
opara-tor: Optionally use `std::range::Range` which is `Copy` instead of `std::ops::Range`.

But this might be waiting on #3770 for a newer rust version.

src/main/host/fcntl_lock_table.rs
```diff
@@ -8,6 +8,13 @@ use rangemap::RangeMap;
 
 use crate::host::process::ProcessId;
 
+/// Error returned by a lock operaion.
```
opara-tor: nit: operaion -> operation

src/test/file_locks/test_file_locks.rs
```diff
@@ -283,6 +283,68 @@ fn test_pid_lock_overflow(cmd: FcntlPosixSetlkUncontestedCommand) -> anyhow::Res
...
+                    l_pid: 0,
+                },
+            })?
+            .to_result()?
+    };
+
+    ensure_ord!(conflict_flock, ==, libc::flock {
+                    l_type: libc::F_WRLCK.try_into().unwrap(),
+                    l_whence: libc::SEEK_SET.try_into().unwrap(),
+                    l_start: 0,
```
opara-tor: super nit: indentation

src/main/host/syscall/handler/fcntl.rs
```diff
@@ -184,17 +184,15 @@ impl SyscallHandler {
...
-                        };
                         let start = i64::try_from(conflict_range.start).unwrap_or_else(|_err| {
                             panic!("Current lock range {conflict_range:?} start is out of range")
                         });
+                        // length > i64::MAX is represented as 0.
+                        // Primarily this happens when the lock is *set* with
+                        // length 0 (see FLOCK_LENGTH_0_OPEN_END), but Linux
+                        // also uses 0 to represent the length of a coalesced
+                        // lock whose length doesn't fit in an i64.
+                        let length = i64::try_from(conflict_range.len()).unwrap_or(0);
```
opara-tor, line 199: Are we sure that it always uses 0 for >i64::MAX, or could it be overflowing and wrapping instead? So for example if the lock was `0..(i64::MAX+2)`, does it still return 0 and not 1?

src/test/test_utils.rs
```diff
@@ -806,16 +823,56 @@ where
...
-            std::cmp::Ordering::Greater => fork_res,
-            std::cmp::Ordering::Less => return Err(std::io::Error::last_os_error()),
+            Ok(CloneResult::CallerIsParent(pid)) => pid,
+            Err(e) => return Err(std::io::Error::from_raw_os_error(e.into())),
         };
+        if clone_flags.contains(CloneFlags::CLONE_FILES) {
+            // We share a descriptor table with the child. *do not* close the child's
+            // ends of the pipes.
+            let _ = cmd_reader.reader.into_raw_fd();
+            let _ = res_writer.writer.into_raw_fd();
```
opara-tor, line 833: I think we should rethink the API here, but I don't have any suggestions at the moment. I think in theory launching another process with `CLONE_FILES` is very similar to launching a thread, and `std::thread::spawn()` doesn't have this shared fd issue. So I feel that there's something nicer we could do here.

I don't want to delay this PR over this, but maybe open an issue for follow-up work?

src/main/host/fcntl_lock_table.rs
```diff
@@ -415,10 +376,13 @@ impl FcntlLockTable {
...
     pub fn remove_owner(&mut self, file_id: FileId, requested_owner: LockOwner) {
-        let Some(file_locks) = self.locks.get_mut(&file_id) else {
-            return;
-        };
-        file_locks.remove_owner(requested_owner);
+        self.set_lock(
+            file_id,
+            usize::MIN..usize::MAX,
+            requested_owner,
+            FlockType::F_UNLCK,
```
opara-tor, line 394: Add a comment here that unlocking entries with a different owner will be skipped and not result in an error?

src/main/host/syscall/handler/fcntl.rs
opara-tor: Yeah the non-Copy of `Range` has been something they've wanted to fix for a while, it's mainly because it implements `Iterator` rather than `IntoIterator`. There's a new `std::range::Range` type that is `Copy` as of some newer rust version.

src/main/host/fcntl_lock_table.rs
opara-tor: (A `LockError` was added later.)

src/main/host/fcntl_lock_table.rs
opara-tor: > I think I'm inclined to keep the implementation roughly as-is for now and file an issue for correctly handling processes with shared descriptor tables. Wdyt?

Thanks for looking into it! Yeah I agree with filing an issue for now, the current code handles the typical cases correctly.

> P.S. I checked the kernel tests to see if any covered this case, since that would probably better indicate intent. I didn't find any tests for posix range lock; only some very light tests for open file description locks 🤷 https://github.com/torvalds/linux/blob/master/tools/testing/selftests/filelock/ofdlocks.c

Yeah my impression is that the Linux people aren't big on syscall tests, so I'm not surprised :) Shadow now has more posix file lock tests than the kernel :)

src/main/host/descriptor/regular_file.c
opara-tor: Thanks, I'll take a look at that PR.

opara-tor (approved):
Thanks for the changes! Have a few last small comments.

src/main/host/host.rs
sporksmith: fixed

src/test/file_locks/test_file_locks.rs
sporksmith: fixed

src/test/test_utils.rs
sporksmith: We'd similarly want to avoid closing these descriptors if we were launching a thread. We only want to close them when the child has its own copy of the descriptors (in its own copy of the descriptor table).

Your comment could also apply to the Safety comment in `new_with_clone_files` though. As I was thinking about how to make that comment clearer I thought for a moment that we might be able to make it safe by requiring that the closure is `Send` and `Sync`, but I realized that doesn't help either. Updated the comment to:

```
    /// Create a child process that shares its descriptor table with the parent
    /// with the CLONE_FILES flag.
    ///
    /// # Safety
    ///
    /// TLDR: avoid manipulating shared file descriptors using Rust code that
    /// has safety requirements around the state of those descriptors.
    ///
    /// The spawned child process shares a descriptor table with its parent,
    /// similarly as if it were a thread. Unlike a thread, though, the child
    /// process does *not* share a virtual address space, meaning that even if
    /// we used `Send` and `Sync` requirements here, tools like
    /// `std::sync::Mutex` that might otherwise make such sharing safe won't
    /// work as expected.
    ///
    /// For example, `std::fs::File` is an "owned file descriptor", with a
    /// safety requirement that it is open for the lifetime of the object.
    /// <https://doc.rust-lang.org/stable/std/os/fd/trait.FromRawFd.html#tymethod.from_raw_fd>.
    /// Sharing such an object between the parent and the child, and dropping
    /// the object in one of them (closing the file), would violate the safety
    /// property in the other. Even if we required that the function object is
    /// `Send` and `Sync`, the type system would allow calling code to share an
    /// `Arc<Mutex<Option<File>>>`, but at runtime calling
    /// `lock().unwrap().take()` on one of the objects would close the
    /// descriptor for both processes without actually doing any in-memory
    /// synchronization, resulting in a safety violation.
```

src/main/host/fcntl_lock_table.rs
sporksmith: done

src/main/host/fcntl_lock_table.rs
sporksmith: Filed https://github.com/shadow/shadow/issues/3783 and xrefd it here

src/main/host/fcntl_lock_table.rs
sporksmith: done

src/main/host/syscall/handler/fcntl.rs
sporksmith: `test_pid_lock_overflow` verifies that the last byte that can be locked is `i64::MAX+1`

sporksmith (commented):
Thanks for all the careful review! I'll rebase and merge once you are granted approval power (hopefully it works retroactively 😆 )

src/main/host/syscall/handler/fcntl.rs
sporksmith: Ah I see. Added a TODO to switch once we're using 1.96.

src/main/host/syscall/handler/fcntl.rs
sporksmith: added a todo

stevenengler (approved):

PR #3775, opened 2026-08-09
Expand doc comment for `pod::to_u8_slice()`

stevenengler:
I was confused about why we required `Pod`, since any struct should be made up of maybe initialized bytes. But I think we use `Pod` just for its "no interior mutability" guarantee, since we're casting to an immutable slice.

### DIFF
--- a/src/lib/pod/src/lib.rs
+++ b/src/lib/pod/src/lib.rs
@@ -37,6 +37,10 @@ pub unsafe trait Pod: Copy + 'static {}
 /// Convert to a slice of raw bytes.
 ///
 /// Some bytes may be uninitialized if T has padding.
+///
+/// Requiring `Pod` for this function is overly strict,
+/// but we need some way to prevent providing types that have interior mutability.
+/// So `Pod` works well enough for us here, and in similar functions below.
 pub fn to_u8_slice<T>(slice: &[T]) -> &[MaybeUninit<u8>]
 where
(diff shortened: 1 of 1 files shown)


src/lib/pod/src/lib.rs
```diff
@@ -37,6 +37,10 @@ pub unsafe trait Pod: Copy + 'static {}
 /// Convert to a slice of raw bytes.
 ///
 /// Some bytes may be uninitialized if T has padding.
+///
+/// Requiring `Pod` for this function is overly strict,
```
sporksmith, line 41: I think this is a helpful comment, but maybe we can be a little more precise. How about:

> This function is safe because `Pod` guarantees there is no interior mutability, and therefore we can't accidentally create an immutable reference to mutable bytes. The safety here doesn't actually depend on Pod's property that any bit pattern is valid.

(I suppose we could consider splitting out `NoInteriorMutability` or something as a separate trait, but probably not worth the bother for our purposes)

"similar functions" -> I think maybe just `as_u8_slice`? In particular, functions that provide mutable u8 "views" need the "core" Pod guarantee that any bit pattern is valid. Maybe instead of referencing "other functions" here, `to_u8_slice` (and any other relevant functions I'm missing) can have something like "This is safe because `Pod` ensures no interior mutability; see also `to_u8_slice`"

Issue #3694, opened 2025-11-04
strace logging: support for RV-dependent formatting for output-pointer-args


sporksmith:
Our strace logging logs the syscall parameters after the syscall has been executed, allowing it to log the output value for out-pointers.

Currently though, the formatting code for syscall parameters doesn't get to see the return value of the syscall. e.g. this means that when formatting the  output buffer of a `read` syscall, the formatter knows what the max size of the buffer is, but not how much was actually written.

I *think* to improve this we'd need to do something like:

* Add the syscall return value to `SyscallVal` and populate it in `host::syscall::formatter::write_syscall`
* Add types like:
  * `OutputSyscallBuffer` that works like `SyscallBuffer` but assumes that the rv of the syscall tells how many bytes are valid.
  * Similarly, `OutputSyscallString` that works like `SyscallString` but assumes that the rv of the syscall tells how many bytes are valid. (Needed because some, if not all, syscalls with output strings don't write a NUL byte)
  * `SyscallOutputOnlyArg<T>`, analogous to `SyscallNonDeterministicArg<T>`, which suppresses logging of `T` if the syscall didn't succeed. (e.g. for output-only fixed-size structs)

PR #3776, opened 2026-08-09
Some syscall logger improvements, and new "long" option

stevenengler:
1. This adds a `SyscallArrayArg` for displaying arrays of values, which is used for `epoll_wait()` (and related epoll calls).

    Before:

    ```text
    00:00:00.050000000 [tid 1001] epoll_wait(5, 0x7f1cb8000ea0, 10, 100) = 1
    ```

    After:

    ```text
    00:00:00.050000000 [tid 1001] epoll_wait(5, [linux_epoll_event { events: 1, data: 0 } (0x7f5be8000ea0), ...], 10, 100) = 1
    ```

2. This also switches `shadow_hostname_to_addr_ipv4()` from using `SyscallStringArg` to `SyscallBufferArg`.

    This doesn't change the output in the typical case. But if the syscall was called without a null terminated string, the logger could try to access memory outside the given limit in `name_len`.

3. This also adds a new "long" option for `experimental.strace_logging_mode`. If set, shadow doesn't as aggressively truncate values.

    Before:

    ```text
    00:00:01.160000000 [tid 1000] write(1, "Testing test_zero_len_buf <sys_method=To"..., 79) = 79
    ```

    After (with option set):

    ```text
    00:00:01.160000000 [tid 1000] write(1, "Testing test_zero_len_buf <sys_method=ToFrom, init_method=Inet, sock_type=1>...", 79) = 79
    ```

### DIFF
--- a/src/main/host/syscall/type_formatting.rs
+++ b/src/main/host/syscall/type_formatting.rs
@@ -335,6 +335,70 @@ fn fmt_string(
     }
 }
 
+fn fmt_array<'a, T>(
+    f: &mut std::fmt::Formatter<'_>,
+    ptr: ForeignPtr<T>,
+    len: usize,
+    args: [shadow_shim_helper_rs::syscall_types::SyscallReg; 6],
+    options: FmtOptions,
+    mem: &'a MemoryManager,
+) -> std::fmt::Result
+where
+    SyscallVal<'a, *const T>: SyscallDisplay,
+{
+    // How many bytes we want to show.
+    // We try not to exceed this, but there may be some edge cases which do.
+    const DISPLAY_LEN: usize = 100;
+
+    // Separator for items.
+    const SEPARATOR: &str = ", ";
+
+    if options == FmtOptions::Deterministic {
+        return write!(f, "<pointer>");
+    }
+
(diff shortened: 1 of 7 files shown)


src/main/host/syscall/type_formatting.rs
```diff
@@ -334,6 +334,70 @@ fn fmt_string(
...
+{
+    // How many bytes we want to show.
+    // We try not to exceed this, but there may be some edge cases which do.
+    const DISPLAY_LEN: usize = 100;
+
+    // Separator for items.
+    const SEPARATOR: &str = ", ";
+
+    if options == FmtOptions::Deterministic {
+        return write!(f, "<pointer>");
```
sporksmith, line 372: Maybe a brief comment here about why we don't display more. Maybe just "see https://github.com/shadow/shadow/issues/3694" ?

sporksmith (approved):

src/main/host/syscall/type_formatting.rs
stevenengler: Added in b25ca91 for `fmt_buffer()` and `fmt_array()`.

PR #3777, opened 2026-08-10
Improve "explicit drop" API and usage

sporksmith:
Broken out of WIP #3774, which is progress on #2258

* Generalize the `ExplicitDrop` trait to have a lifetime parameter, instead of `ExplicitDrop::explicit_drop` taking a reference to `ExplicitDropParam` with an implicit lifetime. This is necessary to allow `ExplicitDropParam` to internally include lifetimes, which is used in #3774.
* Detect if `Descriptor`s and `DescriptorTable`s are dropped without explicitly closing. (This becomes more important in #3774, where we release fcntl record locks as part of the descriptor-close process)

### DIFF
--- a/src/lib/shadow-shim-helper-rs/src/explicit_drop.rs
+++ b/src/lib/shadow-shim-helper-rs/src/explicit_drop.rs
@@ -9,9 +9,9 @@
 /// dropped. One workaround is for a type to also implement `Drop`, and have
 /// that implementation validate that `ExplicitDrop::explicit_drop` was called.
 pub trait ExplicitDrop {
-    type ExplicitDropParam;
+    type ExplicitDropParam<'p>;
     type ExplicitDropResult;
-    fn explicit_drop(self, param: &Self::ExplicitDropParam) -> Self::ExplicitDropResult;
+    fn explicit_drop<'p>(self, param: Self::ExplicitDropParam<'p>) -> Self::ExplicitDropResult;
 }
 
 /// Wrapper that uses a provided function to drop the inner value.
--- a/src/main/host/descriptor/descriptor_table.rs
+++ b/src/main/host/descriptor/descriptor_table.rs
@@ -191,10 +191,13 @@ impl DescriptorTable {
 
     /// Remove and return all descriptors.
     pub fn remove_all(&mut self) -> impl Iterator<Item = Descriptor> {
-        // reset the descriptor table
-        let old_self = std::mem::take(self);
+        // reset `self` to default state.
+        let mut old_self = std::mem::take(self);
+        // take the descriptors out of the table, so that our `Drop` check
+        // doesn't think we forgot to close them.
+        let old_descriptors = std::mem::take(&mut old_self.descriptors);
         // return the old descriptors
-        old_self.descriptors.into_values()
+        old_descriptors.into_values()
     }
 
     /// Remove and return all descriptors in the range. If you want to remove all descriptors, you
@@ -252,6 +255,14 @@ impl ExplicitDrop for DescriptorTable {
                 desc.close(host, cb_queue);
             }
         });
+    }
+}
+
+impl Drop for DescriptorTable {
+    fn drop(&mut self) {
+        if !self.descriptors.is_empty() {
+            debug_panic!("Dropped DescriptorTable without closing (or taking) descriptors");
+        }
     }
 }
 
--- a/src/main/host/descriptor/descriptor_table.rs
+++ b/src/main/host/descriptor/descriptor_table.rs
@@ -236,10 +236,10 @@ impl Default for DescriptorTable {
 }
 
 impl ExplicitDrop for DescriptorTable {
-    type ExplicitDropParam = Host;
+    type ExplicitDropParam<'p> = &'p Host;
     type ExplicitDropResult = ();
 
-    fn explicit_drop(mut self, host: &Host) {
+    fn explicit_drop<'p>(mut self, host: Self::ExplicitDropParam<'p>) {
         // Drop all descriptors using a callback queue.
         //
         // Doing this explicitly instead of letting `DescriptorTable`'s `Drop`
--- a/src/main/host/descriptor/descriptor_table.rs
+++ b/src/main/host/descriptor/descriptor_table.rs
@@ -261,7 +261,9 @@ impl ExplicitDrop for DescriptorTable {
 impl Drop for DescriptorTable {
     fn drop(&mut self) {
         if !self.descriptors.is_empty() {
-            debug_panic!("Dropped DescriptorTable without closing (or taking) descriptors");
+            warn_and_debug_panic!(
+                "Dropped DescriptorTable without closing (or taking) descriptors"
+            );
         }
     }
 }
--- a/src/main/host/descriptor/mod.rs
+++ b/src/main/host/descriptor/mod.rs
@@ -17,6 +17,7 @@ use crate::host::syscall::io::IoVec;
 use crate::host::syscall::types::{SyscallError, SyscallResult};
 use crate::utility::callback_queue::CallbackQueue;
 use crate::utility::{HostTreePointer, IsSend, IsSync, ObjectCounter};
+use shadow_shim_helper_rs::explicit_drop::ExplicitDrop;
 
 pub mod descriptor_table;
 pub mod epoll;
@@ -475,7 +476,9 @@ impl std::ops::Drop for OpenFileInner {
 #[derive(Debug, Clone)]
 pub struct Descriptor {
     /// The file that this descriptor points to.
-    file: CompatFile,
+    // `Option` wrapper so that our `Drop` implementation can ensure that the
+    // file was closed or removed before drop.
+    file: Option<CompatFile>,
     /// Descriptor flags.
     flags: DescriptorFlags,
     _counter: ObjectCounter,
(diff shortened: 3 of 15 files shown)


src/lib/shadow-shim-helper-rs/src/explicit_drop.rs
```diff
@@ -8,10 +8,10 @@
 /// Unfortunately there is no built-in way to *ensure* a type is explicitly
 /// dropped. One workaround is for a type to also implement `Drop`, and have
 /// that implementation validate that `ExplicitDrop::explicit_drop` was called.
-pub trait ExplicitDrop {
-    type ExplicitDropParam;
+pub trait ExplicitDrop<'p> {
+    type ExplicitDropParam: 'p;
     type ExplicitDropResult;
-    fn explicit_drop(self, param: &Self::ExplicitDropParam) -> Self::ExplicitDropResult;
+    fn explicit_drop(self, param: Self::ExplicitDropParam) -> Self::ExplicitDropResult;
```
stevenengler: Why does it require a reference? It seems like it would be simpler with a lifetime on the associated type? Something like:

```rust
trait ExplicitDrop {
    type ExplicitDropParam<'a>;
    type ExplicitDropResult;
    fn explicit_drop<'a>(self, param: Self::ExplicitDropParam<'a>) -> Self::ExplicitDropResult;
}
```

Then it could be implemented with a reference or owned object, like:

```rust
impl ExplicitDrop for Foo {
    type ExplicitDropParam<'a> = u32;
    type ExplicitDropResult = u32;
    fn explicit_drop<'a>(self, param: Self::ExplicitDropParam<'a>) -> Self::ExplicitDropResult { ... }
}
```

or 

```rust
impl ExplicitDrop for Foo {
    type ExplicitDropParam<'a> = &'a u32;
    type ExplicitDropResult = u32;
    fn explicit_drop<'a>(self, param: Self::ExplicitDropParam<'a>) -> Self::ExplicitDropResult { ... }
}
```

src/main/host/descriptor/mod.rs
```diff
@@ -576,6 +577,24 @@ impl Descriptor {
     }
 }
 
+impl Drop for Descriptor {
+    fn drop(&mut self) {
+        if self.file.is_some() {
+            debug_panic!("Dropped descriptor without closing");
+        }
+    }
+}
```
stevenengler, line 588: I'm unclear about this. The current reason we need to manually call `close()` is because C file objects need a reference to the host in order to close properly. But rust file objects don't (in `CompatFile::close()`, the `host` is only used for the legacy file). In a future where C file objects go away, it wouldn't need to take the host anymore and we could just drop them normally.

Is the goal to add a new dependence on the host or another object for all descriptors (C and rust)? If there's a table of locks that we need to remove the descriptor from, could we just have the drop impl do that automatically, similar to how file event listeners get a `Handle` that they hold on to and when the `Handle` get dropped, it automatically unregisters the listener from the `EventSource`? So something like: a descriptor would get a handle to the lock table, and when dropped would automatically unregister itself.

It would be nice if we could move away from needing to manually clean things up, and instead could just drop things. But I don't know much about how the locks work, or maybe that would take a lot more work?

src/lib/shadow-shim-helper-rs/src/explicit_drop.rs
sporksmith: > Why does it require a reference?

I'm a little confused about this part, since this change gets rid of the baked-in reference, but:

> Then it could be implemented with a reference or owned object, like:

Oh nice, I didn't know you could introduce a generic parameter in the associated type without it being a parameter of the whole trait. I like this better; done.

src/main/host/descriptor/mod.rs
sporksmith: In the context of this PR alone, my thinking was that currently it's possible to drop a `Descriptor` without closing it, which would be a potentially non-obvious bug (at least if it's a legacy file).

I do have more reason to want an explicit close in the range lock implementation, but if you disagree that the above is reason enough, I can leave this part for the posix range lock implementation PR.

Here's the additional reasoning why I think we'll want it for the posix range lock PR:

> a descriptor would get a handle to the lock table, and when dropped would automatically unregister itself.

Some context - For posix range locks, when a process closes a descriptor, any locks that process holds to the associated file are released, regardless of whether they were taken through that particular descriptor. i.e. the lock "belongs to" the Process, not the descriptor.

Despite that, I'd still prefer to have the lock-dropping logic in the Descriptor, rather than the Process, to ensure we can't forget to do it.

But, we need to know a) which ProcessId to drop locks for, and b) whether to drop locks at all (there are some places, I think primarily in the middle of fork and exec, where we drop descriptors as an implementation detail, but they shouldn't have "user-visible" effects like dropping locks here).

I suppose the Descriptor could store something like an `Option<ProcessId>` of which pid it should drop locks for, if any, when dropped, but we'd need to be careful to keep it updated. Specifying the pid when we're doing the close/drop seems less error-prone.

If you're not convinced, I can drop the added `ExplicitDrop for Descriptor` here (or maybe just the enforcement of it), and leave it for the PR that actually implements the fnctl locks. Wdyt?

src/main/host/descriptor/descriptor_table.rs
```diff
@@ -255,6 +256,14 @@ impl ExplicitDrop for DescriptorTable {
     }
 }
 
+impl Drop for DescriptorTable {
+    fn drop(&mut self) {
+        if !self.descriptors.is_empty() {
+            debug_panic!("Dropped DescriptorTable without closing (or taking) descriptors");
```
stevenengler: nit: Maybe add a `warn_once` after this so that we catch it (non-fatally) in release builds too? (And for `Descriptor::drop`?)

src/main/host/descriptor/descriptor_table.rs
```diff
@@ -192,9 +192,10 @@ impl DescriptorTable {
     /// Remove and return all descriptors.
     pub fn remove_all(&mut self) -> impl Iterator<Item = Descriptor> {
         // reset the descriptor table
-        let old_self = std::mem::take(self);
+        let mut old_self = std::mem::take(self);
         // return the old descriptors
-        old_self.descriptors.into_values()
+        let old_descriptors = std::mem::take(&mut old_self.descriptors);
+        old_descriptors.into_values()
```
stevenengler, line 200: This could use a comment explaining why it's taking `descriptors` out.

src/main/host/descriptor/mod.rs
```diff
@@ -475,7 +476,7 @@ impl std::ops::Drop for OpenFileInner {
 #[derive(Debug, Clone)]
 pub struct Descriptor {
     /// The file that this descriptor points to.
-    file: CompatFile,
+    file: Option<CompatFile>,
```
stevenengler, line 481: Add a comment about why it's an `Option`?

src/main/host/descriptor/mod.rs
```diff
@@ -506,17 +507,17 @@ impl Descriptor {
...
 
     /// Close the descriptor. The `host` option is a legacy option for legacy file.
     pub fn close(
-        self,
+        mut self,
         host: &Host,
         cb_queue: &mut CallbackQueue,
     ) -> Option<Result<(), SyscallError>> {
-        self.file.close(host, cb_queue)
+        self.file.take().unwrap().close(host, cb_queue)
```
stevenengler: nit: `self.into_file().close()`?

src/lib/shadow-shim-helper-rs/src/explicit_drop.rs
stevenengler: > I'm a little confused about this part, since this change gets rid of the baked-in reference, but:

Sorry I read the change wrong, so I was confused.

src/main/host/descriptor/mod.rs
stevenengler: > In the context of this PR alone, my thinking was that currently it's possible to drop a `Descriptor` without closing it, which would be a potentially non-obvious bug (at least if it's a legacy file).

~Following that reasoning, I think we would want to implement it for `CompatFile` and `LegacyTcp` as well then? Since we can have file objects that aren't owned by any descriptor, and we may also drop them without closing them properly.~

~But I don't think we should do that in this MR, and could leave it for a time when we might need it.~

**Edit:** The above wasn't right, since we want to be able to hold on to file objects after we've `close()`d them. So `ExplicitDrop` wouldn't make sense for them.

> [posix range locks]

Okay thanks, I think I understand the general problem you're describing with the locks, and since descriptors can "change" processes during a fork, and we implement `Clone` on `Descriptor`.

Okay, I'll say I'm 90% convinced :)

stevenengler (approved):
Thanks for the changes/comments. Looks good to me, just a few more small comments.

src/main/host/descriptor/descriptor_table.rs
sporksmith: done

src/main/host/descriptor/mod.rs
sporksmith: done

src/main/host/descriptor/descriptor_table.rs
sporksmith: `debug_panic` actually unconditionally warns before conditionally panicking. I agree it's not obvious from the name, and checked myself when adding it here.

Renamed the macro to `warn_and_debug_panic`

src/main/host/descriptor/mod.rs
sporksmith: done

PR #3778, opened 2026-08-11
More posix (pid) range lock tests

sporksmith:
Progress on #2258

### DIFF
--- a/src/test/file_locks/test_file_locks.rs
+++ b/src/test/file_locks/test_file_locks.rs
@@ -1,4 +1,3 @@
-use anyhow::ensure;
 use linux_api::errno::Errno;
 use num_enum::{IntoPrimitive, TryFromPrimitive};
 use std::{
@@ -6,7 +5,7 @@ use std::{
     ops::Neg as _,
     os::fd::AsRawFd,
 };
-use test_utils::{ForkedChild, ShadowTest, TestEnvironment, ensure_ord, set};
+use test_utils::{ForkedChild, ShadowTest, TestEnvironment, ensure_in, ensure_ord, set};
 
 // fcntl commands that set (or clear) a range lock, and that should behave
 // identically when the requested lock is uncontested.
@@ -219,7 +218,7 @@ fn test_contested_pid_locks(cmd: FcntlPosixSetlkUncontestedCommand) -> anyhow::R
     // (We only test F_SETLK here rather than `cmd`; blocking behavior with F_SETLKW is in another test).
     {
         let res = fcntl_lock(file.as_raw_fd(), FcntlFlockCommand::F_SETLK, &wr_flock);
-        ensure!([Err(Errno::EACCES), Err(Errno::EAGAIN)].contains(&res));
+        ensure_in!(&res, [Err(Errno::EACCES), Err(Errno::EAGAIN)]);
     }
 
     // Tell the child to release its lock, which should succeed.
@@ -1012,7 +1011,7 @@ fn test_overlapping_pid_locks(cmd: FcntlPosixSetlkUncontestedCommand) -> anyhow:
                 l_pid: 0,
             },
         )?;
-        ensure!([child1.pid(), child2.pid()].contains(&out_flock.l_pid));
+        ensure_in!(&out_flock.l_pid, [child1.pid(), child2.pid()]);
     }
     // querying the last byte of the child1 lock should return that one (again
     // describing the whole range covered by that lock)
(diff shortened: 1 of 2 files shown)


src/test/file_locks/test_file_locks.rs
```diff
@@ -246,6 +249,455 @@ fn test_contested_pid_locks(cmd: FcntlPosixSetlkUncontestedCommand) -> anyhow::R
...
+            fd: file.as_raw_fd(),
+            cmd: cmd.into(),
+            flock: wr_flock,
+        })?
+        .to_result()?;
+
+    // Double-check that we see the lock.
+    {
+        let out_flock = fcntl_lock(file.as_raw_fd(), FcntlFlockCommand::F_GETLK, &wr_flock)?;
+        ensure_ord!(out_flock, ==, libc::flock {l_pid: child.pid(), ..wr_flock });
```
robgjansen, line 386: Nit: consider unnesting.

src/test/file_locks/test_file_locks.rs
```diff
@@ -246,6 +249,455 @@ fn test_contested_pid_locks(cmd: FcntlPosixSetlkUncontestedCommand) -> anyhow::R
...
+        ensure_ord!(out_flock.l_pid, ==, child.pid());
+    }
+
+    // Have child exit, which should cause the lock to be released.
+    drop(child);
+
+    // We should no longer see the lock.
+    {
+        let out_flock = fcntl_lock(file.as_raw_fd(), FcntlFlockCommand::F_GETLK, &wr_flock)?;
+        ensure_ord!(out_flock, ==, libc::flock {l_type: libc::F_UNLCK.try_into().unwrap(), ..wr_flock });
```
robgjansen, line 396: Nit: consider unnesting.

src/test/file_locks/test_file_locks.rs
```diff
@@ -246,6 +249,455 @@ fn test_contested_pid_locks(cmd: FcntlPosixSetlkUncontestedCommand) -> anyhow::R
...
+    // dropped as a result of tearing down the thread.
+    let mut child = ForkedChild::new(handle_lock_request)?;
+    let out_flock = child
+        .send_recv(&SerializedLockRequest {
+            fd,
+            cmd: FcntlFlockCommand::F_GETLK.into(),
+            flock: wr_flock,
+        })?
+        .to_result()?;
+    ensure_ord!(out_flock, ==, libc::flock {
```
robgjansen, line 431: Nit: consider unnesting.

src/test/file_locks/test_file_locks.rs
```diff
@@ -327,6 +779,46 @@ fn test_zero_len_pid_locks(cmd: FcntlPosixSetlkUncontestedCommand) -> anyhow::Re
...
+    // Use GETLK from child to inspect the lock.
+    {
+        let out_flock = child
+            .send_recv(&SerializedLockRequest {
+                fd: file.as_raw_fd(),
+                cmd: FcntlFlockCommand::F_GETLK.into(),
+                flock: negative_len_flock,
+            })?
+            .to_result()?;
+        ensure_ord!(out_flock, ==, libc::flock {
```
robgjansen, line 810: Nit: consider unnesting.

robgjansen (approved):
The tests look great! I'm surprised at the number of cases needing tests, but glad we're comprehensively adding these!

FYI, I wanted to make sure I'm not being too annoying with my requests for un-nesting the assertions, so I looked into it. You're probably already familiar, but I found out this is known as the Arrange-Act-Assert (AAA) pattern and is generally considered a best practice for test readability. Cool :)

sporksmith:
> The tests look great! I'm surprised at the number of cases needing tests, but glad we're comprehensively adding these!

Thanks! Yeah there turn out to be a lot of "fun" / interesting corner cases to test :)

> FYI, I wanted to make sure I'm not being too annoying with my requests for un-nesting the assertions, so I looked into it. You're probably already familiar, but I found out this is known as the Arrange-Act-Assert (AAA) pattern and is generally considered a best practice for test readability. Cool :)

Yeah, I think I've seen something along the AAA pattern before. I think it's an ok suggestion as a starting point, but I'm not convinced to be dogmatic about it :)

> Nit: consider unnesting.

Based on the feedback from the previous PR I kept/moved significant function calls out of being inlined into the `ensure_eq!` invocations, but IMO readability wouldn't be improved by lifting what are mostly literal structs into a `let expected = { ... }`.

e.g. to take it a step further, I think `let expected = 3; assert_ord!(actual, ==, expected)` would be worse than `assert_ord!(actual, ==, 3)`.

For a complicated enough `expected` it might be worth assigning to a temp this way, but IMO these don't quite clear the bar

PR #3779, opened 2026-08-12
Make lseek and fstat generic

sporksmith:
Broken out of #3774, which is for #2258

Experimentally fcntl record locks work for other file types than "regular files" on Linux; it's unclear whether that's useful or used in practice, but for the shadow implementation it actually works out a bit nicer anyway to implement them generically at a higher level of abstraction rather than adding more "regular file"-specific code, since that still needs to be reworked into Rust.

To facilitate that, this implements `lseek` and `fstat` at higher abstraction layers. Previous behavior is preserved for file types that don't support these operations.

This also simplifies the `lseek` and `fstat` syscall handlers, which can now call the high-level abstract versions, rather than themselves having to pick apart the file and delegate to a C syscall handler.


src/main/build.rs
```diff
@@ -263,6 +263,8 @@ fn run_bindgen(build_common: &ShadowBuildCommon) {
         .opaque_type("WorkerPool")
         .blocklist_type("HashSet_String")
         // Imported from libc crate below
+        .blocklist_type("stat")
+        // Imported from libc crate below
         .blocklist_type("siginfo_t")
```
stevenengler: We don't seem to import `siginfo_t` anymore below, so this comment doesn't seem right.

src/lib/linux-api/src/bindings.rs
```diff
@@ -589,19 +855,6 @@ pub const LINUX_SHMDT: u32 = 22;
 pub const LINUX_SHMGET: u32 = 23;
 pub const LINUX_SHMCTL: u32 = 24;
 pub const LINUX_DIPC: u32 = 25;
-pub const LINUX_NR_OPEN: u32 = 1024;
-pub const LINUX_NGROUPS_MAX: u32 = 65536;
```
stevenengler, line 593: (Off-topic) I wonder if we should `cargo fmt` this file so that things don't jump around as often. It's hard to tell what was added and what was just moved.

src/lib/linux-api/src/unistd.rs
```diff
@@ -0,0 +1,14 @@
+use crate::bindings;
+use num_enum::{IntoPrimitive, TryFromPrimitive};
+
+// Valid values for the `whence` parameter of the `lseek` syscall.
```
stevenengler: nit: `//` -> `///` (I sometimes read the cargo doc output for these things)

src/main/host/descriptor/descriptor.c
```diff
@@ -154,6 +154,29 @@ void legacyfile_close(LegacyFile* descriptor, const Host* host) {
     descriptor->funcTable->close(descriptor, host);
 }
 
+int legacyfile_fstat(LegacyFile* descriptor, struct stat* statbuf) {
+    MAGIC_ASSERT(descriptor);
+    MAGIC_ASSERT(descriptor->funcTable);
+
+    if (descriptor->funcTable->fstat == NULL) {
+        debug("fstat not implemented for LegacyFile type %d", descriptor->type);
+        return -EBADF;
```
stevenengler, line 163: I think that `EBADF` isn't the right thing to return here, since the file descriptor was valid if it made it this far. (And for `legacyfile_lseek()`). Maybe `ENOSYS` would be better? Or possibly `ESPIPE` for lseek.

src/main/host/descriptor/mod.rs
```diff
@@ -681,6 +725,30 @@ impl CompatFile {
...
+        whence: linux_api::unistd::LSeekWhence,
+    ) -> Result<linux_api::posix_types::kernel_off_t, SyscallError> {
+        match self {
+            CompatFile::New(open_file) => match open_file.inner_file() {
+                File::Pipe(_) => Err(Errno::ESPIPE.into()),
+                _ => {
+                    warn_once_then_debug!("lseek() is not implemented for this type");
+                    Err(Errno::ENOTSUP.into())
+                }
+            },
```
stevenengler: Can you pass this lseek call through to the underlying file, similar to how we do with stat? So then we won't have file-specific logic like `File::Pipe(_) => Err(Errno::ESPIPE.into())` in the generic `CompatFile`? It's a little bit of a pain to pass it through, but I think keeps the logic together better.

src/main/host/descriptor/mod.rs
```diff
@@ -635,6 +636,49 @@ impl LegacyFileCounter {
...
+            st_rdev: libc_stat.st_rdev,
+            st_size: libc_stat.st_size,
+            st_blksize: libc_stat.st_blksize,
+            st_blocks: libc_stat.st_blocks,
+            st_atime: libc_stat.st_atime as u64,
+            st_atime_nsec: libc_stat.st_atime_nsec as u64,
+            st_mtime: libc_stat.st_mtime as u64,
+            st_mtime_nsec: libc_stat.st_mtime_nsec as u64,
+            st_ctime: libc_stat.st_ctime as u64,
+            st_ctime_nsec: libc_stat.st_ctime_nsec as u64,
```
stevenengler: Is it possible to use `into()` here rather than a cast (I'm not sure what the types are)?

src/main/host/descriptor/mod.rs
```diff
@@ -635,6 +636,49 @@ impl LegacyFileCounter {
...
+            ..shadow_pod::zeroed()
```
stevenengler: Can you add a comment about what other fields are here that we're zeroing?

src/main/host/descriptor/mod.rs
```diff
@@ -635,6 +636,49 @@ impl LegacyFileCounter {
         unsafe { self.file.as_ref().unwrap().ptr() }
     }
 
+    pub fn lseek(
+        &mut self,
```
stevenengler: Why do `LegacyFileCounter::lseek()` and `CompatFile::lseek()` require `&mut`? They're both shared references, so I think can always use just `&self`?

src/main/host/descriptor/mod.rs
```diff
@@ -499,6 +499,10 @@ impl Descriptor {
         &self.file
     }
 
+    pub fn file_mut(&mut self) -> &mut CompatFile {
+        &mut self.file
+    }
+
```
stevenengler: I don't think we need this if we make the lseek methods require just `&self`.

src/test/pipe/test_pipe.rs
```diff
@@ -39,6 +39,11 @@ fn get_tests() -> Vec<test_utils::ShadowTest<(), String>> {
             test_read_write,
             set![TestEnv::Libc, TestEnv::Shadow],
         ),
+        test_utils::ShadowTest::new(
+            "test_lseek",
+            test_lseek,
+            set![TestEnv::Libc, TestEnv::Shadow],
+        ),
```
stevenengler, line 46: super nit: this test was added in the middle of the read/write tests, could it be moved up a little before `test_read_write`?

stevenengler (changes requested):
Looks nice! Just have a few comments/change requests.

src/main/build.rs
sporksmith: Hah, this comment was already here, but the diff makes it look new. I guess I should fix it anyway 😆 . Fixed.

src/test/pipe/test_pipe.rs
sporksmith: done

src/lib/linux-api/src/bindings.rs
sporksmith: This is `cargo fmt`'d, but the order tends to change when we add new  header files to parse, since some already-recursively-included files can end up getting included in a different order.

src/lib/linux-api/src/unistd.rs
sporksmith: done

src/main/host/descriptor/descriptor.c
sporksmith: Out of curiosity I checked to see what actually happens with an epoll fd; it appears to just succeed and return position 0.

```
lseek(5, 10, SEEK_CUR)                  = 0
lseek(5, 10, SEEK_CUR)                  = 0
```

Anyway, I've seen `EBADF` used in other places to mean this isn't a valid operation on the current descriptor; e.g. for `fcntl` "EBADF  cmd is F_SETLK or F_SETLKW and the file descriptor open mode doesn't match with the type of lock requested."

So, IMO `ENOSYS` and `EBADF` are equally incorrect, but `EBADF` is a little more informative.

FWIW the previous behavior for this case was to return `EINVAL`. I think that's worse than both of the above, since it is a documented error for `lseek`, but indicates a different condition. "EINVAL whence is not valid.  Or: the resulting file offset would be negative, or beyond the end of a seekable device." I suppose I could just leave that behavior in place for this PR though to make this more of a pure refactor.

LMK what you think

src/main/host/descriptor/mod.rs
sporksmith: My thinking is that this function mutates state of the file; it just happens that in the current implementation that state lives in the OS, so Rust doesn't force us to make this `&mut`. IMO it's better for this to be `&mut` mostly to avoid something holding an immutable ref from being "surprised" about the state changing out from under it. (I guess it also leaves us  free to change the implementation to keep more of the state in shadow instead of deferring as much to Linux, though I don't think we have any plans for that)

If you still disagree I can make this `&self` though, and get rid of `file_mut`

src/main/host/descriptor/mod.rs
sporksmith: In the linux definition these are signed.

I plumbed the linux stat struct definition through to the C code so that we don't need to do this conversion.

src/main/host/descriptor/mod.rs
sporksmith: Gone here, but done in `pipe.rs` which uses the same idiom.

src/main/host/descriptor/mod.rs
sporksmith: I started to do this, but recursively exploding this out to all of the file types and subtypes both is quite a lot of boiler-plate, and adds some risk of the implementations unintentionally diverging. e.g. they should probably all return the same error (except for pipes, where the error is specifically specified as EPIPE), or possibly all treat it as a no-op as appears to be the case for `epoll` fds.

src/lib/linux-api/src/bindings.rs
stevenengler: For some reason I thought cargo fmt would sort these constants, but that's wrong.

src/main/host/descriptor/mod.rs
stevenengler: > adds some risk of the implementations unintentionally diverging

But don't they diverge? Regular files return a non-error value, epoll files return 0, pipe files return EPIPE, and we don't know what other file types return.

Can we move this logic to the syscall handler instead? Then there's less of the syscall/file-specific logic spread around in different parts of the code. If it's a `CompatFile::Legacy(inner)`, we call `inner.lseek()`, if it's a `CompatFile::New(open_file)` we do the `match open_file.inner_file()`?

src/main/host/descriptor/mod.rs
stevenengler: I think that's by design though? All file objects (`CompatFile`, `OpenFile`, `InetSocket`, etc) are just handles to a file object, either through an `Arc` (in rust) or a reference counted `LegacyFile` (in C). Nothing has ownership over the underlying file object. For example `Pipe` only ever exists as an `Arc<AtomicRefCell<Pipe>>`. It's the `AtomicRefCell` where we get a mutable reference to the file, which is why other syscall handlers do something like:

```rust
match file {
    CompatFile::New(open_file) => open_file.inner_file().borrow_mut().ioctl(request, ...),
    CompatFile::Legacy(file) => // do something with the C refcounted file
}
```

So by making `CompatFile::lseek()` take a `&mut`, it seems like we have an exclusive reference, but we don't because it's just a handle.

```rust
let mut foo: CompatFile = descriptor_table.get_mut(fd);
let mut bar = foo.clone();

let foo = &mut foo;
let bar = &mut bar;

// these &mut methods seem like they have exclusive access,
// but they are both modifying the same file
foo.lseek();
bar.lseek();
```

src/main/host/descriptor/mod.rs
sporksmith: Done - pushed this down to the individual files. Your other comment that `CompatFile` is a shared reference type helped convince me that it's weird to have it here.

> But don't they diverge? Regular files return a non-error value, epoll files return 0, pipe files return EPIPE, 

Right

> and we don't know what other file types return.

Right; ideally we probably want all of these to behave the same until we work out what the correct behavior is. 

But yeah, I agree now that this isn't a good place to put implementation, and it doesn't look like we currently have another nice way of providing a "default impl".

> Can we move this logic to the syscall handler instead? 

The main motivation of this PR is to get this logic out of the syscall handler, so that it can be used in the posix range lock implementation with `whence=SEEK_CUR` (and similarly for lstat and implementing `whence=SEEK_END`)

src/main/host/descriptor/mod.rs
sporksmith: Ah, gotcha. You're right - done

src/main/Cargo.toml
```diff
@@ -80,6 +80,9 @@ system-deps = "7.0"
 # Building the C code from this crate's build script requires
 # that these bindings have been generated.
 shadow-shim-helper-rs = { path = "../lib/shadow-shim-helper-rs" }
+# Needs to be a build-dependency as well for its generated header files
+# to be present, which our C build needs.
+linux-api = { path = "../lib/linux-api", features = ["c_bindings"] }
```
stevenengler: I think this will make cargo build linux-api twice (once during the build script and once when building shadow), and the later run will overwrite the `build/src/lib/linux-api/linux-api.h` of the former? But I don't see a way around this. If you agree maybe add a comment?

src/lib/linux-api/rename.py
```diff
@@ -101,6 +101,8 @@ def prefix_field_names(fields, s):
             t.startswith('si_')
             # sigaction fields; sometimes collide with macros in libc
             or t.startswith('sa_')
+            # stat fields; collides with macros in libc
+            or t.startswith('st_')
```
stevenengler, line 105: Do you know of an example of C code that the original names cause problems for (not necessarily for stat specifically, but any struct)? I'm not arguing about including this or anything, I just don't remember why we needed these renames and am curious.

stevenengler (approved):
Thanks for the changes!

src/main/host/descriptor/mod.rs
stevenengler: FWIW in the past I tried to share common implementations as much as possible, but there's always one or two files/syscalls/etc that behave slightly differently. So that's why so much is delegated to the specific file, because having special cases throughout the code was making things more difficult to follow rather than just putting all of the implementation details for a file type in a single place (for example in 'pipe.rs').

src/lib/linux-api/rename.py
sporksmith: I ran into parse errors while compiling shadow. I *think* what was happening was once I started including `linux-api.h` in shadow C headers, which *also* include libc headers, bindgen running on *those* headers was failing to correctly parse the struct definitions in `linux-api.h` due to the libc preprocessor macros.

src/main/Cargo.toml
sporksmith: yeah, added a comment

PR #3780, opened 2026-08-14
Remove memory mapper

### LINKED ISSUE #2581: Re-evaluate --use-memory-manager

### LINKED ISSUE #3622: experimental.use_memory_manager causes sigsegv in golang programs after #3613
**Describe the issue**

After commit e1990cb4e4e7862fc76e741bc80b06a1c75fc33b from #3613, programs crash with SIGSEGV when run with `experimental.use_memory_manager`

**To Reproduce**

Files:
```
==> main.go <==
package main

import (
	"fmt"
	_ "runtime/cgo"
	"time"
)

func main() {
	time.Sleep(time.Second)
	fmt.Println("Hello world")
	time.Sleep(time.Second)
}

==> shadow.yaml <==
general:
  # Simulation duration
  stop_time: 20s

experimental:
  use_memory_manager: true

network:
  graph:
    type: gml
    inline: |
      graph [
        directed 0
        node [
          id 0
          host_bandwidth_down "1000 Mbit"
          host_bandwidth_up "1000 Mbit"
        ]
        edge [
          source 0
          target 0
          latency "1 ms"
          jitter "0 ms"
          packet_loss 0.0
        ]
      ]

hosts:
  server:
    network_node_id: 0
    ip_addr: 10.0.0.1
    processes:
      - path: ./main
```

Build binary with: `go build -linkshared ./main.go`

 - OS and version: post the output of `lsb_release -d`: `Debian GNU/Linux 12 (bookworm)`
 - Kernel version: post the output of `uname -a`: `Linux beelink-boulder-01 6.1.0-22-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.94-1 (2024-06-21) x86_64 GNU/Linux`

 - Version and build information: post the output of `shadow --show-build-info`
e.g.,
```
Shadow 3.2.0 — v3.2.0-518-ge1990cb4e 2025-06-26--16:12:42
GLib 2.74.6
Built on 2025-07-01--21:38:18
Built from git branch HEAD
Shadow was built with PROFILE=release, OPT_LEVEL=3, DEBUG=false, RUSTFLAGS="-C force-frame-pointers=y", CFLAGS="-std=gnu11 -O3 -ggdb -fno-omit-frame-pointer -Wreturn-type -Wswitch -DNDEBUG"
```
 - Which processes you are trying to run inside the Shadow simulation: Go program

**Additional context**

Relevant logs:
```
00:00:00.011003 [170019:shadow-worker] 00:00:00.000000000 [DEBUG] [server:10.0.0.1] [regular_file.c:351] [regularfile_openat] Rewriting `openat` path '/proc/self/maps' to '/proc/170020/maps'
00:00:00.011344 [170019:shadow-worker] 00:00:00.000000000 [DEBUG] [server:10.0.0.1] [process.rs:1234] [shadow_rs::host::process] process 'server.main.1000' done continuing
00:00:00.011345 [170019:shadow-worker] 00:00:00.000000000 [DEBUG] [server:10.0.0.1] [process.rs:1260] [shadow_rs::host::process] Process server.main.1000 exited while running thread 1000
00:00:00.011345 [170019:shadow-worker] 00:00:00.000000000 [DEBUG] [server:10.0.0.1] [process.rs:1472] [shadow_rs::host::process] process 'server.main.1000' has completed or is otherwise no longer running
00:00:00.011372 [170019:shadow-worker] 00:00:00.000000000 [ERROR] [server:10.0.0.1] [process.rs:1557] [shadow_rs::host::process] process 'server.main.1000' exited with status Signaled(Signal(11)); expected end state was exited: 0 but was signaled: SIGSEGV
Error: process 'server.main.1000' exited with status Signaled(Signal(11)); expected end state was exited: 0 but was signaled: SIGSEGV
00:00:00.011485 [170019:shadow-worker] 00:00:00.000000000 [DEBUG] [server:10.0.0.1] [memory_mapper.rs:326] [shadow_rs::host::memory_manager::memory_mapper] MemoryManager misses: (consider extending MemoryManager to remap regions with a high miss count)
00:00:00.011487 [170019:shadow-worker] 00:00:00.000000000 [DEBUG] [server:10.0.0.1] [memory_mapper.rs:330] [shadow_rs::host::memory_manager::memory_mapper] 	2 in Region { shadow_base: 0x0, prot: ProtFlags(PROT_READ), sharing: Private, original_path: Some(Path("/usr/lib/x86_64-linux-gnu/libc.so.6")) }
00:00:00.011487 [170019:shadow-worker] 00:00:00.000000000 [DEBUG] [server:10.0.0.1] [memory_mapper.rs:330] [shadow_rs::host::memory_manager::memory_mapper] 	1 in Region { shadow_base: 0x0, prot: ProtFlags(PROT_READ | PROT_WRITE), sharing: Private, original_path: Some(InitialStack) }
```

sporksmith:
Removes the experimental command-line option `--use-memory-manager` (experimental config option `use_memory_manager`)

Fixes #2581; previous discussion on this topic there.

Also fixes #3622, which only applies to the memory mapper.

## Background motivation

This was a complex optimization that mapped managed process memory into memory shared with shadow so that it could access it directly. At the time that it was introduced, this was a huge improvement over copying a word at a time through `ptrace` operations.

Since then, we've migrated the "copying" path from using `ptrace`, first to reading and writing `/proc/[pid]/mem`, and subsequently to use the `process_vm_readv` and `process_vm_writev` syscalls, which made it nearly as fast as the "mapped" path. 

Meanwhile, the "mapped" path has been expensive to maintain and prone to subtle bugs, and has never updated to be compatible with emulating the `exec` syscall.

## How it turned out

The first couple commits remove the memory mapper itself; IMO just doing this and nothing else wouldn't be a substantial win.

Most of the rest of the PR is deleting the MemoryManager abstractions for direct references to process memory. These abstractions were fairly complex to maintain, in particular including a sort of arena pointer system for the legacy syscalls to use.

Migrating the C syscalls from those "mapped" APIs to explicitly-copying APIs (e.g. `process_readablePointer` -> `process_readPointer`) in some cases pushes some complexity into the syscall handlers, which now have to handle allocating and freeing an intermediate buffer. I think this is an overall complexity win though, vs the previous behavior of leveraging the arena pointers needed to support the older APIs. The performance should be no worse than when using those APIs with the memory mapper disabled (which has been the default for some time now), since in those cases we were already doing the allocation and intermediate buffer copying behind the scenes. In cases where we have fixed size or small-upper-bound sized buffers, we can now get a small performance and complexity win by stack-allocating instead.


sporksmith:
tor benchmark: https://github.com/shadow/benchmark-results/tree/master/tor/2026-08-14-T23-51-35

image

Basically no change, which isn't surprising. The memory manager/mapper was already disabled in the benchmark when we switched shadow's default to having it off.

With that in mind, In the relatively hot paths like `read` and `write`, this basically just makes the dynamic allocation that was happening behind the scenes in the MemoryManager more explicit.

There are a lot of places where we got rid of dynamic allocations by reading into statically allocated buffers instead. In principle that could improve performance, but I think those aren't very hot paths, so it's unsurprising that they didn't make a noticeable impact here.

sporksmith:
tgen benchmark: https://github.com/shadow/benchmark-results/tree/master/tgen/2026-08-15-T14-51-19

I don't see a record of how long it took to simulate?

The performance inside the simulation looks unchanged, as expected, compared to the recent nightly and weekly benchmarks.

There is a substantial difference from the 3.1.0 baseline from 2023, but that's not relevant here. We should probably do an updated 3.3.0 baseline :laughing:

sporksmith:
Added both @robgjansen and @stevenengler as reviewers; I'd like to get an ack from both, though it's probably fine for only one to do a deep review

stevenengler:
> I'd like to get an ack from both, though it's probably fine for only one to do a deep review

I have no objection to removing it. (I haven't done a review yet.)

robgjansen:
Sorry, I was away touching grass for 2 weeks. I'll look more closely at this soon!

src/main/host/syscall/handler/fcntl.c
```diff
@@ -83,38 +86,35 @@ static int _syscallhandler_fcntlHelper(SyscallHandler* sys, RegularFile* file, i
...
 
         case F_SETOWN_EX: {
-            const struct f_owner_ex* foe = process_getReadablePtr(
-                rustsyscallhandler_getProcess(sys), argReg.as_ptr, sizeof(*foe));
-            result = regularfile_fcntl(file, command, (void*)foe);
+            struct f_owner_ex foe = {0};
+            result = process_readPtr(
+                rustsyscallhandler_getProcess(sys), &foe, argReg.as_ptr, sizeof(foe));
+            result = regularfile_fcntl(file, command, &foe);
             break;
```
opara-tor, line 118: I think this is missing an `if (result < 0) break`?

src/main/host/syscall/handler/fcntl.c
```diff
@@ -83,38 +86,35 @@ static int _syscallhandler_fcntlHelper(SyscallHandler* sys, RegularFile* file, i
...
         }
-#endif
 
         case F_GETOWN_EX: {
-            struct f_owner_ex* foe = process_getWriteablePtr(
-                rustsyscallhandler_getProcess(sys), argReg.as_ptr, sizeof(*foe));
-            result = regularfile_fcntl(file, command, foe);
+            struct f_owner_ex foe = {0};
+            result = regularfile_fcntl(file, command, &foe);
```
opara-tor, line 101: Just noting the behaviour change here, where we pass in a zeroed `f_owner_ex` instead of the user-supplied one. But this seems fine to me. (And for other fcntl options below.)

src/main/host/syscall/handler/file.c
```diff
@@ -116,13 +116,16 @@ SyscallReturn syscallhandler_fstat(SyscallHandler* sys, const SyscallArgs* args)
...
     /* Get some memory in which to return the result. */
-    struct stat* buf =
-        process_getWriteablePtr(rustsyscallhandler_getProcess(sys), bufPtr, sizeof(*buf));
-    if (!buf) {
-        return syscallreturn_makeDoneErrno(EFAULT);
+    struct stat buf = {0};
+
+    int res = regularfile_fstat(file_desc, &buf);
+    if (res < 0) {
+        return syscallreturn_makeDoneErrno(-res);
```
opara-tor, line 123: I'm not sure we should use a zeroed `stat` buffer here. I think Linux sets the stat fields individually, and then passes the struct off to the filesystem to set more fields. It's possible that a field won't be set (and I think I remember seeing this before). In this case I think the field should be what the user passed in and not set to zero. So I think we should do a `process_readPtr` first.

(And for `fstatfs` below.)

src/main/host/syscall/handler/file.c
```diff
@@ -116,13 +116,16 @@ SyscallReturn syscallhandler_fstat(SyscallHandler* sys, const SyscallArgs* args)
...
     }
 
-    return syscallreturn_makeDoneI64(regularfile_fstat(file_desc, buf));
+    errcode = process_writePtr(rustsyscallhandler_getProcess(sys), bufPtr, &buf, sizeof(buf));
+
+    return syscallreturn_makeDoneI64(res);
```
opara-tor, line 128: Should we return an error if `errcode` indicates a failed write?

src/main/host/syscall/handler/sched.rs
```diff
@@ -73,30 +76,28 @@ impl SyscallHandler {
...
         }
+        // We currently always write back a mask that says the thread can
+        // execute only on CPU 0.
 
-        // Shadow doesn't have users, so no need to check for permissions
-
-        if cpusetsize == 0 {
+        if cpusetsize < KERNEL_CPUSETSIZE_BYTES {
             return Err(Errno::EINVAL);
         }
```
opara-tor, line 89: I don't really understand this change. I think we want to allow sizes less than `KERNEL_CPUSETSIZE_BYTES`, just not zero? (And similarly for `sched_setaffinity()`).

src/main/host/memory_manager/mod.rs
```diff
@@ -316,7 +316,20 @@ impl MemoryManager {
...
+        let mut values = Box::<[T]>::new_uninit_slice(ptr.len());
+        let ptr = ptr.cast::<MaybeUninit<T>>().unwrap();
+        let copied = unsafe { self.memory_copier.copy_prefix_from_ptr(&mut values, ptr)? };
+
+        // Drop the still uninitd values. Is there a way to directly resize the boxed slice
+        // without having to go through Vec?
+        let mut values = Vec::from(values);
+        values.resize(copied, MaybeUninit::uninit());
+        let values = values.into_boxed_slice();
+
```
opara-tor, line 328: I think this can be done with:

```rust
let ptr = ptr.cast::<MaybeUninit<T>>().unwrap();
let mut values = Vec::with_capacity(ptr.len());
let copied = unsafe { self.memory_copier.copy_prefix_from_ptr(values.spare_capacity_mut(), ptr)? };
values.set_len(copied);
```

This would be like the original code that doesn't reallocate the buffer (as in `shrink_to_fit()` with `Vec::into_boxed_slice()`).

opara-tor (commented):
Looks a lot nicer!

robgjansen (approved):
I have reviewed and am supportive of merging this PR. I was pleasantly surprised at how much code you were able to remove and how much cleaner the replacement bits are. :)

Thank you for your hard work on this, I think it will pay dividends. And thanks Steve for doing the deep dive.

I'm marking this as approved, but assume you'll ensure Steve's comments are satisfactorily resolved before merging.

src/main/host/syscall/handler/fcntl.c
sporksmith: whoops! done

src/main/host/syscall/handler/fcntl.c
sporksmith: Yeah, I *think* these are write-only

src/main/host/syscall/handler/file.c
sporksmith: Oh interesting; fixed

src/main/host/syscall/handler/file.c
sporksmith: yes, fixed

src/main/host/syscall/handler/sched.rs
sporksmith: Oops, you're right for `sched_setaffinity`, but for `sched_getaffinity` this is supposed to be EINVAL.

> EINVAL (sched_getaffinity() and, in kernels before 2.6.9, sched_setaffinity()) cpusetsize is smaller than the size of the affinity mask used by the kernel.

fixed in sched_setaffinity

src/main/host/memory_manager/mod.rs
sporksmith: nice; done

sporksmith:
Ah, it looks like I'll need to rebase and solve the conflicts before CI will run again. @stevenengler i'll wait for your approval (or your go-ahead to rebase anyway)

src/main/host/memory_manager/mod.rs
```diff
@@ -142,20 +142,17 @@ impl MemoryManager {
...
-        values.resize(copied, MaybeUninit::uninit());
-        let values = values.into_boxed_slice();
-
-        // SAFETY: should now contain only initialized values.
-        let values = unsafe { values.assume_init() };
-
+        let mut values = Vec::with_capacity(ptr.len());
+        let copied = self
+            .memory_copier
+            .copy_prefix_from_ptr(&mut values.spare_capacity_mut(), ptr)?;
```
opara-tor: I'm confused by the `&mut values.spare_capacity_mut()` here. Doesn't `values.spare_capacity_mut()` return a `&mut [MaybeUninit]`? So I don't understand how the `&mut &mut [MaybeUninit]` works here. But if it compiles then it must be okay :)

src/main/host/memory_manager/mod.rs
```diff
@@ -142,20 +142,17 @@ impl MemoryManager {
...
+        // SAFETY: we just initialized these bytes.
+        unsafe { values.set_len(copied) };
         Ok(Vec::from(values))
```
opara-tor: nit: I think `values` is already a `Vec`?

src/main/host/syscall/handler/file.c
opara-tor, line 129: Is this missing a `return`?

opara-tor (approved):

src/main/host/memory_manager/mod.rs
sporksmith: Oops, yeah I think Rust does some auto-dereferencing. Removed the `&mut` for clarity

src/main/host/memory_manager/mod.rs
sporksmith: fixed

src/main/host/syscall/handler/file.c
sporksmith: fixed

sporksmith (commented):
Good catches; thanks!

PR #3781, opened 2026-08-18
shadow-exec: dump shadow config as json to avoid ambiguities

sporksmith:
In particular, the pyyaml encoder doesn't quote strings that it thinks unambiguously look like strings. shadow's yaml decoder sometimes disagrees here, causing it to reject output from the pyyaml encoder.

IIUC the particular here is a difference between yaml 1.1 and 1.2 spec https://github.com/yaml/pyyaml/issues/393, where pyyaml follows 1.1 <https://github.com/yaml/pyyaml/issues/486> and thinks a string only looks like a float if it has a `+` after the `e`, so feels free to emit a string like `1e6` without quoting, but shadow's decoder presumably uses yaml 1.2 and interprets such a string as a float. https://github.com/yaml/pyyaml/issues/393

Rather than continuing to try to iron those out, json is probably a better output format for a config file that is machine-generated and machine-consumed. Since json is by design a subset of yaml, this is still valid yaml. The json output should be less prone to ambiguities in shadow's yaml decoder.

### DIFF
--- a/shadowtools/src/shadowtools/shadow_exec.py
+++ b/shadowtools/src/shadowtools/shadow_exec.py
@@ -31,7 +31,7 @@ import shutil
 import sys
 import tempfile
 import textwrap
-import yaml
+import json
 
 from pathlib import Path
 from typing import TextIO, BinaryIO, Final, Optional, List, Iterable, Dict, Literal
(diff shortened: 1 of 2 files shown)


sporksmith:
A potential alternative is to use a yaml 1.2 compliant encoder:

```python
import yaml12
import yaml

print(f"yaml module:", yaml.safe_dump('1e10'))
print(f"yaml12 module:", yaml12.format_yaml('1e10'))
```

Output:
```
yaml module: 1e10
...

yaml12 module: "1e10"
```

but that introduces other potential ambiguities if something else tries to decode as yaml 1.1 (e.g. the bare string 'no' decodes to the boolean `False` in yaml 1.1). I think for an "ephemeral" config file meant to be machine written and read, json is a simpler and safer choice.

sporksmith:
Ah ha, I *thought* I'd run into this before and came up with a different workaround: https://github.com/yaml/pyyaml/issues/860#issuecomment-4596493626

So we could continue using the yaml encoder here and just force quoting with `default_style='"'`, but encoding with the json encoder produces nicer output IMO.

```
>>> d = {"myint": 3, "mybool": False, "myfloat": 3.14, "mystring": "32e4"}
>>> print(yaml.dump(d))
mybool: false
myfloat: 3.14
myint: 3
mystring: 32e4

>>> print(yaml.dump(d, default_style='"'))
"mybool": !!bool "false"
"myfloat": !!float "3.14"
"myint": !!int "3"
"mystring": "32e4"

>>> print(json.dumps(d, indent=4))
{
    "myint": 3,
    "mybool": false,
    "myfloat": 3.14,
    "mystring": "32e4"
}
```

stevenengler (approved):
LGTM! I think there are some corner cases where yaml and json differ, but we're unlikely to run into those here.

PR #3782, opened 2026-08-27
Clean up oflag handling

sporksmith:
Split from #3774

This started as wanting to be able to get the access-mode of regular files from Rust in #3774, but pulling that thread ended up becoming a wider cleanup.

Broadly, this:

* adds linux-api support for handling open flags, particularly distinguishing between "access mode", "creation", and "file status" flags
* updates legacyfile and regularfile code to be more careful about these distinctions.
* expose related functionality in CompatFile
* updates the fcntl F_SETFL and F_GETFL handling, ultimately unifying it into the rust handler

### DIFF
--- a/src/main/host/descriptor/descriptor.c
+++ b/src/main/host/descriptor/descriptor.c
@@ -279,19 +279,16 @@ void legacyfile_removeListener(LegacyFile* descriptor, StatusListener* listener)
 
 gint legacyfile_getFlags(LegacyFile* descriptor) {
     MAGIC_ASSERT(descriptor);
-    return descriptor->flags;
+    MAGIC_ASSERT(descriptor->funcTable);
+    utility_alwaysAssert(descriptor->funcTable->get_flags != NULL);
+    return descriptor->funcTable->get_flags(descriptor);
 }
 
 void legacyfile_setFlags(LegacyFile* descriptor, gint flags) {
     MAGIC_ASSERT(descriptor);
-    int creation_flags = flags & linux_file_creation_oflags();
-    if (creation_flags) {
-        warning("Caller included creation flags 0x%x, which they should have removed and handled "
-                "separately",
-                creation_flags);
-    }
-    utility_debugAssert(!creation_flags);
-    descriptor->flags = flags;
+    MAGIC_ASSERT(descriptor->funcTable);
+    utility_alwaysAssert(descriptor->funcTable->set_flags != NULL);
+    descriptor->funcTable->set_flags(descriptor, flags);
 }
 
 bool legacyfile_supportsSaRestart(LegacyFile* legacyDesc) {
--- a/src/main/host/descriptor/descriptor.c
+++ b/src/main/host/descriptor/descriptor.c
@@ -284,10 +284,13 @@ gint legacyfile_getFlags(LegacyFile* descriptor) {
 
 void legacyfile_setFlags(LegacyFile* descriptor, gint flags) {
     MAGIC_ASSERT(descriptor);
-    if (flags & O_CLOEXEC) {
-        warning("Adding CLOEXEC to legacy file when it should "
-                "have been added to the descriptor");
+    int creation_flags = flags & linux_file_creation_oflags();
+    if (creation_flags) {
+        warning("Caller included creation flags 0x%x, which they should have removed and handled "
+                "separately",
+                creation_flags);
     }
+    utility_debugAssert(!creation_flags);
     descriptor->flags = flags;
 }
 
--- a/src/main/host/descriptor/descriptor.c
+++ b/src/main/host/descriptor/descriptor.c
@@ -277,18 +277,36 @@ void legacyfile_removeListener(LegacyFile* descriptor, StatusListener* listener)
     eventsource_removeLegacyListener(descriptor->event_source, listener);
 }
 
-gint legacyfile_getFlags(LegacyFile* descriptor) {
+gint legacyfile_getAccessModeFlags(LegacyFile* descriptor) {
     MAGIC_ASSERT(descriptor);
-    MAGIC_ASSERT(descriptor->funcTable);
-    utility_alwaysAssert(descriptor->funcTable->get_flags != NULL);
-    return descriptor->funcTable->get_flags(descriptor);
+    return descriptor->accessAndCreationFlags & linux_access_mode_oflags();
 }
 
-void legacyfile_setFlags(LegacyFile* descriptor, gint flags) {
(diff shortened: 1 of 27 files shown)


src/main/host/syscall/handler/mman.rs
```diff
@@ -379,15 +379,16 @@ impl SyscallHandler {
         // /proc/<shadow-pid>/fd/<linux-fd> file, which is a symbolic link
         flags -= OFlag::O_NOFOLLOW;
 
-        let mode = unsafe { c::regularfile_getModeAtOpen(file) };
+        // Mode argument to 'open' is a no-op, since we're not creating a file.
+        let mode = 0i32;
```
opara-tor, line 364: I think this was originally a mistake (probably added by me). The comment on `_File.modeAtOpen` says "the permission mode the file was opened with", but the `mode` argument to `open()` is really "the file mode bits to be applied when a new file is created". So I think the code originally confused the "file mode" with "access mode", and probably thought it was applying the same access mode here.

**Edit:** I see you noticed this too in the following commit.

src/main/host/descriptor/descriptor.c
```diff
@@ -284,10 +284,12 @@ gint legacyfile_getFlags(LegacyFile* descriptor) {
 
 void legacyfile_setFlags(LegacyFile* descriptor, gint flags) {
     MAGIC_ASSERT(descriptor);
-    if (flags & O_CLOEXEC) {
-        warning("Adding CLOEXEC to legacy file when it should "
-                "have been added to the descriptor");
+    int creation_flags = flags & linux_file_creation_oflags();
+    if (creation_flags) {
+        warning(
+            "Setting file-creation flags 0x%x, which should already be acted on", creation_flags);
```
opara-tor: I don't think the "should already be acted on" is very clear here. I think this means something along the lines of "the function caller should have removed these flags before passing them here"?

src/main/host/descriptor/descriptor.c
```diff
@@ -279,18 +279,16 @@ void legacyfile_removeListener(LegacyFile* descriptor, StatusListener* listener)
 
 gint legacyfile_getFlags(LegacyFile* descriptor) {
     MAGIC_ASSERT(descriptor);
-    return descriptor->flags;
+    MAGIC_ASSERT(descriptor->funcTable);
+    utility_alwaysAssert(descriptor->funcTable->set_flags != NULL);
+    return descriptor->funcTable->get_flags(descriptor);
```
opara-tor: The assert checks `set_flags` but calls `get_flags`.

opara-tor (commented):
I didn't quite finish this review, so leaving whatever comments I have so far.

src/main/host/descriptor/descriptor_types.h
```diff
@@ -24,6 +24,11 @@ typedef struct _LegacyFile LegacyFile;
 typedef struct _LegacyFileFunctionTable LegacyFileFunctionTable;
 
 /* required functions */
+
+/* Set "file status" oflags */
+typedef void (*LegacyFileSetFlags)(LegacyFile* descriptor, int flags);
```
opara-tor: This says just "file status", but `legacyfile_setFlags()` says  "access" and "file status". I think these should be consistent?

src/main/host/descriptor/mod.rs
```diff
@@ -29,35 +29,7 @@ pub mod shared_buf;
...
-    }
-
-    /// Returns a tuple of the `FileStatus` and any remaining flags.
-    pub fn from_o_flags(flags: OFlag) -> (Self, OFlag) {
-        let status = Self::from_bits_truncate(flags.bits());
-        let remaining = flags.bits() & !status.bits();
-        (status, OFlag::from_bits(remaining).unwrap())
-    }
-}
+pub use linux_api::fcntl::FileStatusOFlag as FileStatus;
```
opara-tor, line 33: Can we add a TODO about removing this re-export/alias?

src/main/host/descriptor/epoll.c
```diff
@@ -113,8 +113,8 @@ struct _Epoll {
     /* A counter for sorting watches, for guaranteeing determinism when reporting events. */
     uint64_t watch_id_counter;
 
-    /* access and status flags, as returned via fcntl F_GETFL */
-    int accessAndStatusFlags;
+    /* status flags, as returned via fcntl F_GETFL */
+    int statusFlags;
```
opara-tor: I think this should specify whether the non-status but returned by `F_GETFL` flags are included here.

So maybe something like "status flags, as documented in open(2)", or "status flags and additional flags, as returned via fcntl F_GETFL".

(And for any other non-epoll files.)

src/main/host/syscall/handler/mman.rs
```diff
@@ -352,7 +352,11 @@ impl SyscallHandler {
         // get the access and status oflags (not creation flags).
         // unclear whether the status flags are actually needed, but they may effect
         // performance (e.g. O_DIRECT).
-        let oflags = OFlag::from_bits_retain(unsafe { c::regularfile_get_flags(file) });
+        let oflags = {
+            let raw_status = unsafe { c::regularfile_get_status_flags(file) };
+            let raw_access = unsafe { c::legacyfile_getAccessModeFlags(file.cast()) };
```
opara-tor: In the C code I think we usually do `&file->super` rather than a pointer cast. Can we do that here too?

opara-tor (approved):

src/main/host/descriptor/mod.rs
sporksmith: done

src/main/host/syscall/handler/mman.rs
sporksmith: Added `regularfile_getLegacyFile` and used it here.

src/main/host/syscall/handler/mman.rs
sporksmith: ack

src/main/host/descriptor/descriptor.c
sporksmith: done

src/main/host/descriptor/descriptor.c
sporksmith: done

src/main/host/descriptor/descriptor_types.h
sporksmith: fixed in a later commit

sporksmith (commented):
thanks!

src/main/host/descriptor/epoll.c
sporksmith: changed to "as documented in open(2)"

PR #3786, opened 2026-09-12
Attempt at improving the flaky epoll tests

robgjansen:
My attempt at improving the stability of the flaky epoll tests (some minimal discussion [here](https://github.com/shadow/shadow/pull/3768#discussion_r3567874827)).

Tried to as much as possible remove any dependency on sleeping a certain period of time and then assuming that a thread has already started after waking up.

Was able to reproduce a failing test case (epoll-edge-rs-linux) using `stress-ng` to eat up all CPUs while running the test. With my new code, that test now passes even when running stress-ng.

Also useful was separating our timeouts into two different ones: a long timeout (10s) which is used only in cases where we expect that the `epoll_wait` call should not reach the timeout, and a short timeout (100ms) which is used only in cases where we expect the `epoll_wait` call should exceed the timeout. (The short duration timeout keeps the test runtime lower.)

### DIFF
--- a/src/test/epoll/test_epoll_edge.rs
+++ b/src/test/epoll/test_epoll_edge.rs
@@ -1,5 +1,5 @@
 use std::io::Cursor;
-use std::time::Duration;
+use std::sync::mpsc;
 
 use neli::ToBytes;
 use neli::consts::nl::NlmF;
@@ -13,43 +13,9 @@ use nix::unistd;
 use test_utils::socket_utils::{SocketInitMethod, socket_init_helper};
 use test_utils::{ShadowTest, TestEnvironment, ensure_ord, set};
 
-#[derive(Debug)]
-struct WaiterResult {
-    duration: Duration,
-    epoll_res: nix::Result<usize>,
-    events: Vec<epoll::EpollEvent>,
-}
+use crate::util::*;
 
-fn do_epoll_wait(epoll_fd: i32, timeout: Duration, do_read: bool) -> WaiterResult {
-    let mut events = Vec::new();
-    events.resize(10, epoll::EpollEvent::empty());
-
-    let t0 = std::time::Instant::now();
-
-    let res = epoll::epoll_wait(
-        epoll_fd,
-        &mut events,
-        timeout.as_millis().try_into().unwrap(),
-    );
-
-    let t1 = std::time::Instant::now();
-
-    events.resize(res.unwrap_or(0), epoll::EpollEvent::empty());
-
-    if do_read {
-        for ev in &events {
-            let fd = ev.data() as i32;
-            // we don't care if the read is successful or not (another thread may have already read)
-            let _ = unistd::read(fd, &mut [0]);
-        }
-    }
-
-    WaiterResult {
-        duration: t1.duration_since(t0),
-        epoll_res: res,
-        events,
-    }
-}
+mod util;
 
 fn test_multi_write(readfd: libc::c_int, writefd: libc::c_int) -> anyhow::Result<()> {
     let epollfd = epoll::epoll_create()?;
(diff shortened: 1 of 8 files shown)


robgjansen:
Previous attempt #3753

robgjansen:
I suspect #3410 was actually caused by the flakiness that this PR is attempting to solve. But I'm not sure due to the discussion therein about linux kernel behavior changes. Can/should we close #3410 after merging this?

robgjansen:
@shadow/dev FYI, I was hoping to merge this before starting to handle a bunch of the version bump tasks, to hopefully reduce some of the CI flakiness during the bump process. (It seems like my recent change to the dependabot config triggered something because dependabot barraged us with PRs shortly after.) But if you think I should not block the version bumps on this PR, let me know.

src/test/epoll/util.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /// Sets the value to true and notifies the condition variable if the value
+    /// changed from false to true.
+    pub fn notify(&self) {
+        let (lock, cvar) = self.state.as_ref();
+        let mut value = lock.lock().unwrap();
+        if !*value {
+            *value = true;
+            cvar.notify_one();
```
opara-tor, line 43: I think this should be `notify_all()` instead of `notify_one()`, otherwise I think there's a race where two threads could pass `wait()` at the same time for a single `notify()`.

Or if you do want the `notify_one()` behaviour, I think this could be done by having the `wait()` reset the true back to false.

src/test/epoll/util.rs
```diff
@@ -0,0 +1,110 @@
...
+/// When we expect an event to be ready quickly, use a long timeout that
+/// shouldn't normally trigger but provides an upper bound on error.
+pub const LONG_DUR: Duration = Duration::from_secs(10);
+
+/// When we expect that the timeout will be reached, use a shorter one to
+/// prevent the test from consuming too much runtime.
+pub const SHORT_DUR: Duration = Duration::from_millis(200);
+
+#[derive(Debug, Clone)]
+pub struct Notify {
```
opara-tor, line 16: Can you add a doc comment about what this does or what it's supposed to be used for?

src/test/epoll/test_epoll_edge.rs
```diff
@@ -63,39 +29,37 @@ fn test_multi_write(readfd: libc::c_int, writefd: libc::c_int) -> anyhow::Result
...
-        // Wait for readers to block.
-        std::thread::sleep(timeout / 3);
+        let thread =
+            std::thread::spawn(move || vec![waiter1.wait(), waiter2.wait(), waiter3.wait()]);
 
-        // Make the read-end readable.
+        // Wait for the thread to start and execute the first epoll_wait,
+        // then write to the write end so the read end becomes readable.
+        block1.wait();
         unistd::write(writefd, &[0])?;
```
opara-tor, line 44: I think this has the same issue as https://github.com/shadow/shadow/pull/3750#discussion_r3150150746.

The other thread may have notified but not yet started the epoll wait. So here we may be writing before the epoll wait has started. So it may or may not be testing the wakeup behaviour.

opara-tor: The old version with `sleep()` also had a race, but in Shadow we know that sleep will always finish after the epoll wait has started. It's less clear in the new version what happens with Shadow, especially with things like the "unblocked syscall latency". I think it also depends on what order the threads run, such as in the `test_threads_multi_write()` test below.

Since we're having issues with the epoll tests anyway outside of Shadow, maybe we still want to do this change instead, and just add a disclaimer that we're hoping the "epoll_wait" happens immediately before the other thread writes? Or maybe we could do "sleep if running in shadow, otherwise use the condvar notifier"?

src/test/epoll/test_epoll_edge.rs
```diff
@@ -160,39 +119,52 @@ fn test_threads_multi_write(readfd: libc::c_int, writefd: libc::c_int) -> anyhow
...
-            std::thread::spawn(move || do_epoll_wait(epollfd, timeout, /* do_read= */ false)),
+            std::thread::spawn(move || tx1.send(waiter1.wait()).unwrap()),
+            std::thread::spawn(move || tx2.send(waiter2.wait()).unwrap()),
+            std::thread::spawn(move || tx3.send(waiter3.wait()).unwrap()),
         ];
 
-        // Wait for readers to block.
-        std::thread::sleep(timeout / 3);
+        // Wait for at least one thread to be ready.
+        block1.wait();
```
opara-tor, line 137: The original test was checking that if you have 3 threads blocked in `epoll_wait()` and you write one byte, does it only wake up a single `epoll_wait()`?

I think this should wait for all three threads to be ready?

src/test/epoll/util.rs
```diff
@@ -0,0 +1,110 @@
...
+        Self {
+            epoll_fd,
+            timeout_ms: timeout.as_millis().try_into().unwrap(),
+            notify,
+        }
+    }
+
+    pub fn wait(&self) -> WaiterResult {
+        let mut events = Vec::new();
+        events.resize(1, epoll::EpollEvent::empty());
```
opara-tor, line 72: Can we use a >1 number of events here, and then assert after the `epoll_wait()` that `res <= 1`?

If we ever get more than 1 event here, I think it means that our test is bad, or shadow has a bug, and I think it would be good to fail.

src/test/epoll/test_epoll_edge.rs
robgjansen: Yes, there is still a race. I don't think there is a facility that gives us what we want, i.e., only sets a flag/condition _after_ a syscall has entered the kernel space. Once we make the syscall (here, `epoll_wait`) we lose control. I'd love to update if there is such a facility.

The race is much tighter than before by design, which means the flakiness will be reduced but not eliminated. I chose the word "improving" in the PR title for this reason :)

I had considered adding in `sleep()` and/or `yield()` calls immediately before the waiter returns from the `Notify::wait()` function. I think that would further decrease the likelihood that the `epoll_wait` loses the race, and would also ensure the correct ordering when running in Shadow. I didn't end up adding it because the tests were passing without it, but happy to add it now.

Issue #3788, opened 2026-09-13
File lock test is flaky in Github CI


robgjansen:
I've seen this fail a few times now:

```
The following tests FAILED:
Errors while running CTest
	 66 - file-locks-linux (Failed)
2026-09-12 23:27:11,096 [INFO] ctest returned 8
```

I think it's because the Github CI is so slow that the timing being asserted in a test case is causing assertion failures.

Example: https://github.com/shadow/shadow/actions/runs/34723059269/job/103636791921#step:12:580

```
Testing block-on-pid-locks F_SETLKW... ✗ (!(106.619775ms <= 100ms))
Error: !(106.619775ms <= 100ms)

Stack backtrace:
   0: anyhow::error::<impl anyhow::Error>::msg
             at /github/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/anyhow-1.0.100/src/backtrace.rs:27:14
   1: test_file_locks::test_block_on_pid_locks
             at /__w/shadow/shadow/src/test/file_locks/test_file_locks.rs:1455:5
   2: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/alloc/src/boxed.rs:2254:9
   3: test_utils::ShadowTest<T,E>::run
             at /__w/shadow/shadow/src/test/test_utils.rs:60:9
   4: test_utils::run_tests
             at /__w/shadow/shadow/src/test/test_utils.rs:84:20
   5: test_file_locks::main
             at /__w/shadow/shadow/src/test/file_locks/test_file_locks.rs:1572:5
   6: core::ops::function::FnOnce::call_once
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/ops/function.rs:250:5
   7: std::sys::backtrace::__rust_begin_short_backtrace
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/std/src/sys/backtrace.rs:166:18
   8: std::rt::lang_start::{{closure}}
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/std/src/rt.rs:206:18
   9: <&dyn core::ops::function::Fn<(), Output = i32> + core::marker::Sync + core::panic::unwind_safe::RefUnwindSafe as core::ops::function::FnOnce<()>>::call_once
[...]

PR #3790, opened 2026-09-13
Improve file lock block-duration test

sporksmith:
* Looser upper bound on time taken, to account for busy test machines.
* Shorter "sleep" time, to shorten wall-clock test duration, and which should be tolerable with new bounds-check style

Mitigates #3788


sporksmith:
@robgjansen I *think* your comment about using explicit synchronization isn't applicable here; these tests already use explicit synchronization where applicable.

I wasn't able to repro the issue with `stress-ng --cpu 0 --cpu-method all -t 20m`. I'm still poking through the options there; I think I need to tweak things to get > 100% CPU demand so that it takes longer for the test threads to scheduled, but if you already have a suggestion of a good way of running it let me know.

sporksmith:
> @robgjansen I _think_ your comment about using explicit synchronization isn't applicable here; these tests already use explicit synchronization where applicable.

Actually I thought about it some more, and I think we could do a reasonable test here without the sleep/timing checks. We'd need to extend the `ForkedChild` and underlying pipes it uses for IPC to support non-blocking mode, and then something like:

* Parent takes lock
* Parent tells child to make a blocking SETLKW operation (and then write back to the pipe when it has completed)
* (semi-optional) Parent waits a small amount of time (maybe 100 ms). (This shouldn't contribute to flakiness, but helps ensure that we're actually testing blocking behavior and *not* just beating a race with the child)
* Parent polls the pipe to verify that child hasn't written to the pipe yet (and therefore hasn't returned from SETLKW operation)
* parent releases the lock, allow child to finish SETLKW operation and write to the pipe
* parent does a blocking read on the pipe
* (optional) verify that "not much" time passed between releasing the lock and being able to read from the pipe. (this could again be a potential source of flakiness, but we could probably be pretty loose here if we do this check at all)

I don't think I'll have time to try tackling this today though; maybe merge with the current less-flaky version and add a TODO to convert to above?

robgjansen:
I was able to trigger in my specific case just using `stress-ng --cpu 0`, but YMMV. (I am already using a machine from 2018 so maybe that helped too.)

robgjansen:
> I don't think I'll have time to try tackling this today though; maybe merge with the current less-flaky version and add a TODO to convert to above?

I think applying a quick fix is smart while we wait for the more robust approach. Perhaps you could copy/move that more robust design text over to #3788, and then keep #3788 open even after merging this.

robgjansen:
Huh a futex test failure - is that yet another time-related flakiness issue?

robgjansen (approved):

sporksmith:
> Huh a futex test failure - is that yet another time-related flakiness issue?

Yeah, added another sub-issue https://github.com/shadow/shadow/issues/3791

Issue #3627, opened 2025-07-13
Check CI workflows with zizmor


stevenengler:
There's a static analysis tool that supposedly scans for common security issues in GitHub CI workflows. I think that our workflows are probably better than most, but it might be useful to run it and see if it reports anything interesting. Or maybe just run it as its own job in the CI (but it's a rust program, so is probably slow to build, and maybe not worthwhile).

https://github.com/zizmorcore/zizmor

PR #3792, opened 2026-09-13
Label cleanup

robgjansen:
Motivated by what I perceive as too much visual noise in the issue tracker lately, I did some cleanup in our github issue labels. Hope this isn't too controversial.

- Migrated labels to the new issue field type. For example, the old tag "Priority: Low" is now set as field name "Priority" value "Low". Same for component labels.
- Migrated the "Type: " style labels to the new type field.
- Archived most old labels that were not being used. (Some, like Obsolete or Infeasible are now captured automatically by selecting the correct reason when closing an issue).

Example: #3627

I kept just a few labels that we seldom use but might be useful to tag. I figured we could create new tags organically if we find we need them, but I'm hoping we won't need it too often if we have the right metadata fields.

Sooo, github doesn't support types or fields on pull requests, which I totally absolutely 100% knew before doing any of the above migrations ;-P I think the intention is that the context should be captured on an issue, and the PR can stick to code-specific discussion, and that way we deduplicate metadata. That's OK with me! My thought was that the workflow could be that if you create a PR and find that you are wanting to add metadata that doesn't exist anymore, create an issue and add it there and then link the issue to the PR.

So now this PR removes some of our scripting that we had set up to automatically apply labels that I now consider obsolete. (Those old labels still exist but are in an "archived" state and cannot be applied to new issues.)


sporksmith (approved):
Seems fine to me. I think you're the main consumer of the labels etc when generating reports, so happy to defer to what works for you :)

opara-tor (approved):

PR #3802, opened 2026-09-14
Stop dependabot from adding labels to PRs

robgjansen:


opara-tor (approved):

robgjansen:
Hmm, setting an empty label array may not work due to a bug in some non-public Github backend code :(
https://github.com/dependabot/dependabot-core/issues/11783

PR #3810, opened 2026-09-21
Enable strace-logging of FutexOpFlags

sporksmith:


sporksmith:
I'm a little on the fence on this one. I threw it together while debugging another issue; in practice it makes the output a bit nicer but has some potentially-confusing artifacts due to that `futex_op` parameters is actually conceptually an enum or'd together with some flags. e.g. `FUTEX_WAIT` is command 0, which current gets printed as 0x0 instead of a name. Since the operations have values 0-13, some of them would probably end up getting printed as multiple "flags".

sporksmith:
It'd probably be ideal to have a bespoke type for the combined parameter value, that has methods for splitting and recombining the underlying enum + flags. But I'm not sure I want to bother with shaving that yak right now.

robgjansen:
> But I'm not sure I want to bother with shaving that yak right now.

LLM?

I'm not sure what would be more confusing, the raw int or the potential artifacts. We probably want Steve's opinion on this one.

sporksmith:
> > But I'm not sure I want to bother with shaving that yak right now.
> 
> LLM?

Ehh I've so far managed to avoid messing with them and haven't been impressed by the output I've seen :)

> I'm not sure what would be more confusing, the raw int or the potential artifacts. We probably want Steve's opinion on this one.

I've marked this as draft; I'll try to rework it using a bespoke type when I get a little time. I did something similar with the "oflag int"

sporksmith:
obsoleted by #3813

